PR #10, opened 2013-08-13
JSLint and JSHint directives should not be flagged as commented out code

david-owens-akqa:
The current version of the sonar-javascript plugin flags JSLint and JSHint directives as commented out code in some instances.

This pull request addresses these false positive in the same way that JSDoc comment blocks are currently handled.

I have updated the unit tests to cover these new checks. The directives I have included in the test are taken from our developer guidelines and match the base ruleset that we use in production.

Background

It may be desirable to include a JSLint or JSHint directive at the top of a JavaScript file.

http://www.jslint.com/lint.html#options
http://www.jshint.com/docs/config/

These allow an IDE to provide feedback during development and increase code portability.

The directives look like JavaScript code, especially when they contain keywords, and trigger the "Sections of code should not be "commented out" rule.

### DIFF
--- a/javascript-checks/src/main/java/org/sonar/javascript/checks/CommentedCodeCheck.java
+++ b/javascript-checks/src/main/java/org/sonar/javascript/checks/CommentedCodeCheck.java
@@ -65,7 +65,7 @@ public class CommentedCodeCheck extends SquidCheck<LexerlessGrammar> implements 
 
   public void visitToken(Token token) {
     for (Trivia trivia : token.getTrivia()) {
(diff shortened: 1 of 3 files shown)


weblogixx:
Same should be true for esLint I think.

lindamartin:
We a huge delay but here is the ticket: https://jira.sonarsource.com/browse/SONARJS-545

lindamartin:
Merge in 9e75f160542cfe0aebef4c9a9707c287815551ee. Thanks a lot for your contribution!

PR #11, opened 2014-06-16
Added multi module lcov analysis support

nieuwtje:
The JavaScript Sonar plugin didn't support LCOV analysis in multi module projects, when analysing a multi module project, only the root POM would be checked for a LCOV report path, while the report path was defined in a sub module (or in multiple sub modules). To support this set-up the LCOVSensor had to adjusted. Instead of the root project settings it now uses the project specific properties.

The ASTMaker class had some errors, 5 casts were removed to resolve them.


vilchik-elena:
Unfortunately we don't plan to implement this feature for the moment

PR #12, opened 2015-02-26
SONARJS-336 Replace use of getCanonicalPath() by use of SQ FileSystem

henryju:
getCanonicalPath() may transform the path (change case on Windows, resolve symlinks on Linux) in a way that will later prevent to match it against SQ InputFile. We have already done several work on this area in the platform so it is best to rely on FileSystem API to look for a file from its path (relative or absolute).


lindamartin:
Pull request merged. Thank for the fix @henryju !

PR #13, opened 2015-05-25
SONARJS-430 - Extend rule (S100) Function names should comply with a naming convention

renewolfert:
Added support for expression functions in variable assignments and expression functions in object literals.


lindamartin:
@renewolfert sorry for the late answer. There have been a lot of changes within the JavaScript plugin lately and the way the AST is done as completely changed so unfortunately we won't be able to merge this pull request. But thank you for your consideration and interest!

PR #15, opened 2015-05-28
SONARJS-368 - S1125: Raise an issue when a boolean literal is used as an operand of the ternary operator

renewolfert:
Implemented logic to support validation of conditional expressions (including the examples of the RSPEC)


vilchik-elena:
@renewolfert We discussed this ticket in our team and end up with decision to not fix it. Too many issues should be considered not to raise false positive, and still there might be false negative (as we never know for sure whether expression is boolean or not). 
The only sure case is when boolean literal is used in ternary condition (true ? expr1 : expr2). And we have separate rule for this http://jira.sonarsource.com/browse/RSPEC-2752. 
WDYT?

renewolfert:
@vilchik-elena True. It won't be easy to get a watertight solution for this.

The only change that may remain relevant is the check for parenthesised expressions in the visitExpression method.

vilchik-elena:
I closed a ticket SONARJS-368 as "won't fix" and created this one SONARJS-446 (fixed in 860a3e91a96f69253440b198f40db4f45c202b14).

PR #17, opened 2015-06-04
SONARJS-445 Fix NPE by using ProgressAstScanner from sslr-squid-bridge

pynicolas:
sslr-squid-bridge provides org.sonar.squidbridge.ProgressAstScanner:
* which does not have the same bug as org.sonar.javascript.ProgressAstScanner
* which is supposed to be re-used across plugins


lindamartin:
Thank you @pynicolas ! Merged in 71726574ae1af7be4e7e1c9d502fc94e927cbd5e

PR #18, opened 2015-06-04
SONARJS-451: Update filename building to work on Windows

pwillworth:
The building of the fileName variable was assuming a forward slash character which fails to match the actual path of the file on Windows platforms causing test results to not be imported.  Use native os file separator instead.

### DIFF
--- a/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/unittest/jstestdriver/JsTestDriverSensor.java
+++ b/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/unittest/jstestdriver/JsTestDriverSensor.java
@@ -93,7 +93,7 @@ public class JsTestDriverSensor implements Sensor {
   protected String getUnitTestFileName(String className) {
     // For JsTestDriver assume notation com.company.MyJsTest that maps to com/company/MyJsTest.js
     String fileName = getUnitTestClassName(className);
-    fileName = fileName.replace('.', File.separator.charAt(0));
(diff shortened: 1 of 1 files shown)


lindamartin:
Thanks for your pull request! Here is the associated ticket: [SONARJS-45] (http://jira.sonarsource.com/browse/SONARJS-45)1. I updated the pull request title with the ticket number.

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/unittest/jstestdriver/JsTestDriverSensor.java
```diff
@@ -93,7 +93,7 @@ protected Resource getUnitTestResource(String classKey) {
   protected String getUnitTestFileName(String className) {
     // For JsTestDriver assume notation com.company.MyJsTest that maps to com/company/MyJsTest.js
     String fileName = getUnitTestClassName(className);
-    fileName = fileName.replace('.', '/');
+    fileName = fileName.replace('.', File.separator.charAt(0));
```
lindamartin: @pwillworth why not use File#separatorChar ?

pwillworth: Oh cool, yeah that's better, I just wasn't aware of it.

lindamartin:
Merged in ab8cdfad08ff850ad0ea81e7388dfbf5131effc1

PR #33, opened 2015-07-09
Migrate S1219 from AstNode and fix SONARJS-462

lindamartin:


javascript-checks/src/main/java/org/sonar/javascript/checks/NonCaseLabelInSwitchCheck.java
```diff
@@ -42,36 +38,30 @@
 @ActivatedByDefault
 @SqaleSubCharacteristic(RulesDefinition.SubCharacteristics.READABILITY)
 @SqaleConstantRemediation("10min")
-public class NonCaseLabelInSwitchCheck extends SquidCheck<LexerlessGrammar> {
+public class NonCaseLabelInSwitchCheck extends BaseTreeVisitor {
 
   private boolean inCase;
```
benzonico: The fact that this is only a boolean is bothering me:  what about nested functions : 
```
case FOO: 
     var f = function() {
         label1: foo(); // false positive ?          
     }
```

lindamartin: @benzonico indeed, I create the ticket [SONARJS-462](http://jira.sonarsource.com/browse/SONARJS-462) for the false-positive and fixed it in fbc1178d59bf622c7c6b1f3fed61e3214a92ce44

lindamartin:
@benzonico assigned to you again to review the last commit.

benzonico:
LGTM

PR #36, opened 2015-07-09
Migrate FunctionDefinitionInsideLoop from AstNode.

lindamartin:


javascript-checks/src/main/java/org/sonar/javascript/checks/FunctionDefinitionInsideLoopCheck.java
```diff
@@ -43,51 +44,63 @@
 @ActivatedByDefault
 @SqaleSubCharacteristic(RulesDefinition.SubCharacteristics.INSTRUCTION_RELIABILITY)
 @SqaleConstantRemediation("30min")
-public class FunctionDefinitionInsideLoopCheck extends SquidCheck<LexerlessGrammar> {
+public class FunctionDefinitionInsideLoopCheck extends SubscriptionBaseVisitor {
 
-  private Stack<Integer> stack;
+  private Stack<Integer> scope = new Stack<>();
```
benzonico: You should use a deque and a linked list rather than a Stack : http://docs.oracle.com/javase/7/docs/api/java/util/Stack.html

javascript-checks/src/main/java/org/sonar/javascript/checks/FunctionDefinitionInsideLoopCheck.java
```diff
@@ -43,51 +44,63 @@
...
+      enterScope();
     }
   }
 
   @Override
-  public void leaveNode(AstNode astNode) {
-    if (CheckUtils.isIterationStatement(astNode)) {
-      stack.push(stack.pop() - 1);
+  public void leaveNode(Tree tree) {
+    if (tree instanceof IterationStatementTree) {
```
benzonico: This should probably be checked using is(Kind...) method with iterationStatementsArray rather than instance of.

lindamartin: @benzonico why so ? It looks explicit that way in my opinion, isn't it ?

benzonico: instanceof is costly in terms of perf regarding checking enum values  and an anti-pattern as per OOP.

dbolkensteyn: I'll add that you can have a class, `FooImpl` that implements both `ForStatementTree` and `WhileStatementTree`. In this case `foo instanceof ForStatementTree` will return true even if the instance `foo` actually is a `WhileStatementTree`.

lindamartin:
Merged from 49b4ebbd154f4804ca3ad552176eb8a3c1e2fba9.

PR #38, opened 2015-07-10
Migration tab character

lindamartin:


javascript-squid/src/main/java/org/sonar/plugins/javascript/api/CheckTest.java
```diff
@@ -38,9 +39,12 @@
    * Scan the given file with the given check.
    */
   public SourceFile scanFile(String fileName, JavaScriptFileScanner check) {
+    DefaultFileSystem fs = new DefaultFileSystem();
+    fs.setEncoding(Charset.defaultCharset());
```
benzonico: Given it is on tests, I would not rely on default charset but clearly specify it.

benzonico:
LGTM

javascript-squid/src/main/java/org/sonar/plugins/javascript/api/CheckTest.java
lindamartin: @benzonico fixed in 782abd9dbdd8c57ee3c9e3f63782d85b399d591b

PR #39, opened 2015-07-10
Migrate S1451 from AstNode.

lindamartin:


javascript-checks/src/main/java/org/sonar/javascript/checks/FileHeaderCheck.java
```diff
@@ -62,21 +64,23 @@ public void setCharset(Charset charset) {
   }
 
   @Override
-  public void init() {
+  public void scanFile(AstTreeVisitorContext context) {
+    super.scanFile(context);
+    // FIXME martin: should be done in a init method
```
SonarTech: MAJOR Take the required action to fix the issue indicated by this comment. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1134)

SonarTech:
SonarQube analysis reported 1 issue:
* MAJOR 1 major

javascript-checks/src/main/java/org/sonar/javascript/checks/FileHeaderCheck.java
```diff
@@ -62,21 +64,23 @@ public void setCharset(Charset charset) {
   }
 
   @Override
-  public void init() {
+  public void scanFile(AstTreeVisitorContext context) {
+    super.scanFile(context);
```
benzonico: There is actually no need to call the super method in this case. That would avoid to visit the entire when it is not needed.

lindamartin: @benzonico indeed, it was called just to set the "context" field but as in this check the context is only used in the scanFile method it can use the parameter directly. Fixed in 53ae7ca49ccbb4aaf8995d3365d580d27f511a59

PR #51, opened 2015-07-20
SONARJS-465 Migrate ConditionalComment from AstNode

pynicolas:


javascript-checks/src/main/java/org/sonar/javascript/checks/ConditionalCommentCheck.java
```diff
@@ -40,16 +42,20 @@
...
-        if (comment.startsWith("/*@cc_on") || comment.startsWith("//@cc_on")) {
-          getContext().createLineViolation(this, "Refactor your code to avoid using Internet Explorer's conditional comments.", trivia.getToken());
-        }
+  public List<Tree.Kind> nodesToVisit() {
+    return ImmutableList.of(Tree.Kind.TOKEN);
+  }
+  
+  @Override
+  public void visitNode(Tree tree) {
+    SyntaxToken token = (SyntaxToken) tree;
```
lindamartin: @pynicolas I really think we could support the method visitToken which should be cheap and will avoid this useless line.

pynicolas: I don't feel a strong need for that, but why not.

PR #53, opened 2015-07-20
Migrate TrailingWhitespace from AstNode.

lindamartin:


javascript-checks/src/main/java/org/sonar/javascript/checks/TrailingWhitespaceCheck.java
```diff
@@ -57,19 +58,30 @@ public void setCharset(Charset charset) {
...
+  public List<Tree.Kind> nodesToVisit() {
+    return Collections.emptyList();
+  }
+
+  @Override
+  public void visitFile(Tree scriptTree) {
     List<String> lines;
+
     try {
       lines = Files.readLines(getContext().getFile(), charset);
```
pynicolas: @lindamartin I would include only that line in the try block.

lindamartin: @pynicolas why so ?

pynicolas: @lindamartin I generally try to make a try block rather small: I prefer to make it obvious where the exception can come from.

lindamartin: @pynicolas agreed, updated in 457cb22a72d58b734e08e93bda4429d7bcf667e3.

lindamartin:
Merged in dcc3c48ba456b760d90673d8d28deca62c7b620e.

PR #58, opened 2015-07-27
Update README.md file

vilchik-elena:


lindamartin:
@vilchik-elena LTGM, maybe those information could be update in the same time https://github.com/SonarCommunity/sonar-javascript/blob/master/pom.xml#L60 ? WDYT ?

vilchik-elena:
@lindamartin I think there is no big sense in doing it in the same time. https://github.com/SonarCommunity/sonar-javascript/blob/master/pom.xml#L60 is about upgrading parent to 23 (if i'm not wrong). And what i've done is mostly about migrating from codehaus.

lindamartin:
@vilchik-elena sorry I was not very clear, I was thinking about updating information in "ciManagement" tag which currently point to cloubees and it has been removed in favour of Travis. It has actually nothing to do with update of parent pom.

vilchik-elena:
@lindamartin do you like PR now?

lindamartin:
@vilchik-elena I was thinking it could be part or this PR because it's about CI info, but in any case it is neither urgent nor crucial :)

lindamartin:
@vilchik-elena yes perfect! Tanks :)

PR #59, opened 2015-07-28
org.sonar.api.resources.ProjectFileSystem was deprecated.

nagaiyuki:
Just a small fix.
We should replace the deprecated classes, is it right?


lindamartin:
Yes definitely, thank you for your pull request! I will merged it.

PR #80, opened 2015-08-17
Create *TreeImpl for accessor and generator method

vilchik-elena:


javascript-checks/src/main/java/org/sonar/javascript/checks/EmptyBlockCheck.java
```diff
@@ -74,6 +76,21 @@ public void visitFunctionDeclaration(FunctionDeclarationTree tree) {
     scan(tree.body().statements());
   }
 
+  @Override
+  public void visitAccessorMethodDeclaration(AccessorMethodDeclarationTree tree) {
+    scan(tree.name());
+    scan(tree.parameters());
+    // Ignoring empty function
```
lindamartin: @vilchik-elena  why so ?

javascript-frontend/src/main/java/org/sonar/javascript/tree/impl/declaration/GeneratorMethodDeclarationTreeImpl.java
```diff
@@ -0,0 +1,106 @@
...
+    this.body = body;
+  }
+
+  public GeneratorMethodDeclarationTreeImpl completeWithStaticToken(InternalSyntaxToken staticToken) {
+    this.staticToken = staticToken;
+
+    return this;
+  }
+
+  @Nullable
```
lindamartin: @vilchik-elena I guess this one could be removed as starToken should not be nullable at this point, should it ?

javascript-frontend/src/main/java/org/sonar/plugins/javascript/api/tree/declaration/MethodDeclarationTree.java
```diff
@@ -45,4 +46,5 @@
 
   BlockTree body();
 
+  MethodDeclarationTree completeWithStaticToken(InternalSyntaxToken staticToken);
```
lindamartin: @vilchik-elena I don't think that this should appear in the interface it is a implementation detail and should not be visible on an API level.

lindamartin:
See my comments, otherwise the rest looks much better now!

javascript-checks/src/main/java/org/sonar/javascript/checks/EmptyBlockCheck.java
vilchik-elena: @lindamartin I just commented code of visitMethodDeclaration (as previously accessors and generators were processed in this method). As i understand, this rule shouldn't raise issues for empty function bodies, so in next line we `scan(tree.body().statements());` and not `scan(tree.body());`.

javascript-frontend/src/main/java/org/sonar/javascript/tree/impl/declaration/GeneratorMethodDeclarationTreeImpl.java
vilchik-elena: @lindamartin this `@Nullable` belongs to staticToken

javascript-checks/src/main/java/org/sonar/javascript/checks/EmptyBlockCheck.java
lindamartin: @vilchik-elena indeed, got it! I thought it was in BaseTreeVisitor, obviously my eyes crossed ^^

javascript-frontend/src/main/java/org/sonar/plugins/javascript/api/tree/declaration/MethodDeclarationTree.java
vilchik-elena: @lindamartin indeed. I removed this method completely :) 51dc6eaedc79c40ad653dde52bbe81b24045053a

javascript-frontend/src/main/java/org/sonar/javascript/tree/impl/declaration/GeneratorMethodDeclarationTreeImpl.java
lindamartin: @vilchik-elena indeed my bad I misread again

javascript-frontend/src/main/java/org/sonar/plugins/javascript/api/tree/declaration/MethodDeclarationTree.java
lindamartin: @vilchik-elena cool, even if it seems redundant, it looks clearer in the grammar IMHO :)

PR #86, opened 2015-08-18
SONARJS-513 NonEmptyCaseWithoutBreak: Fix FP with block

pynicolas:


javascript-checks/src/main/java/org/sonar/javascript/checks/NonEmptyCaseWithoutBreakCheck.java
```diff
@@ -51,12 +53,22 @@ public void visitSwitchStatement(SwitchStatementTree tree) {
     cases.remove(cases.size() - 1);
     for (SwitchClauseTree switchClauseTree : cases){
       List<StatementTree> statements = switchClauseTree.statements();
-      if (!statements.isEmpty()
-        && !Iterables.getLast(statements).is(Kind.BREAK_STATEMENT, Kind.RETURN_STATEMENT, Kind.THROW_STATEMENT, Kind.CONTINUE_STATEMENT)) {
+      if (!statements.isEmpty() && !endsWithJump(statements)) {
```
vilchik-elena: @pynicolas you can drop this `!statements.isEmpty()` as this condition presents in `endsWithJump ()`

javascript-checks/src/test/resources/checks/nonEmptyCaseWithoutBreak.js
```diff
@@ -20,3 +20,16 @@ switch (param) {
   case 0: // OK
     doSomethingElse();
 }
+
+switch (param) {
+  case 0: // OK
+    doSomething(); break;
+  case 1: // OK
+    { break; }
```
vilchik-elena: @pynicolas i would add couple more test-cases, like `{ continue; }`, `{ { break } }`

javascript-checks/src/main/java/org/sonar/javascript/checks/NonEmptyCaseWithoutBreakCheck.java
pynicolas: @vilchik-elena I can't drop `!statements.isEmpty()`: there's a `!` before `endsWithJump(statements)`

vilchik-elena: @pynicolas indeed, sorry

PR #88, opened 2015-08-19
SONARJS-333 FP "Parentheses" check: typeOf can require parenthesis

vilchik-elena:


javascript-checks/src/main/java/org/sonar/javascript/checks/ParenthesesCheck.java
```diff
@@ -97,4 +101,19 @@ private void checkExpression(@Nullable ExpressionTree expression) {
...
+
+      if (nestedExpr != null && !shouldBeParenthesized(nestedExpr)) {
+        String expressingString = CheckUtils.asString(nestedExpr);
+        getContext().addIssue(this, nestedExpr, String.format("The parentheses around \"%s\" are useless.", expressingString));
+      }
+    }
+  }
+
+  private static boolean shouldBeParenthesized(ExpressionTree expr) {
+    return expr instanceof BinaryExpressionTree || expr instanceof AssignmentExpressionTree;
```
pynicolas: @vilchik-elena I can think of other cases where parentheses would be required, e.g. typeof(x ? y : z)

pynicolas: @vilchik-elena could we find something better than instanceof? What about listing the acceptable kinds e.g. in CheckUtils? cc @lindamartin

vilchik-elena: @pynicolas both are fixed with feaef9052147cc25535c1efc9711bc1dc1b0ae41

PR #89, opened 2015-08-19
SONARJS-419 Add exception "ForIn" check: Loops used to clone objects should be ignored

vilchik-elena:


javascript-checks/src/main/java/org/sonar/javascript/checks/ForInCheck.java
```diff
@@ -50,11 +53,17 @@ public void visitForInStatement(ForInStatementTree tree) {
...
 
-    if (statementNode != null && !statementNode.is(Kind.IF_STATEMENT)) {
+    if (statementNode != null && !statementNode.is(Kind.IF_STATEMENT) && !isAttrCopy(statementNode)) {
       getContext().addIssue(this, tree, "Insert an if statement at the beginning of this loop to filter items.");
     }
 
     super.visitForInStatement(tree);
   }
 
+  private boolean isAttrCopy(StatementTree statement) {
```
SonarTech: MAJOR Make "isAttrCopy" a "static" method. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS2325)

SonarTech:
SonarQube analysis reported 1 issue:
* MAJOR 1 major

javascript-checks/src/main/java/org/sonar/javascript/checks/ForInCheck.java
```diff
@@ -50,11 +53,17 @@ public void visitForInStatement(ForInStatementTree tree) {
...
+    return statement.is(Kind.EXPRESSION_STATEMENT)
+        && ((ExpressionStatementTree) statement).expression().is(Kind.ASSIGNMENT)
```
pynicolas: @vilchik-elena Style: I would have created a variable for (ExpressionStatementTree) statement).expression()

javascript-checks/src/main/java/org/sonar/javascript/checks/ForInCheck.java
```diff
@@ -50,11 +53,17 @@ public void visitForInStatement(ForInStatementTree tree) {
...
+        && ((AssignmentExpressionTree) ((ExpressionStatementTree) statement).expression()).variable() instanceof MemberExpressionTree;
```
pynicolas: @vilchik-elena I would prefer to use Kind rather than instanceof

javascript-checks/src/main/java/org/sonar/javascript/checks/ForInCheck.java
```diff
@@ -50,11 +53,17 @@ public void visitForInStatement(ForInStatementTree tree) {
...
-    if (statementNode != null && !statementNode.is(Kind.IF_STATEMENT)) {
+    if (statementNode != null && !statementNode.is(Kind.IF_STATEMENT) && !isAttrCopy(statementNode)) {
       getContext().addIssue(this, tree, "Insert an if statement at the beginning of this loop to filter items.");
     }
 
     super.visitForInStatement(tree);
   }
 
+  private boolean isAttrCopy(StatementTree statement) {
+    return statement.is(Kind.EXPRESSION_STATEMENT)
```
pynicolas: @vilchik-elena It seems that we remove a lot of cases (false negatives): to check that we have a copy, shouldn't we check whether the left hand side expression uses the variable of the loop? e.g. shouldnt' we raise an issue on `for (x in obj) { a.b = x }`?

javascript-checks/src/main/java/org/sonar/javascript/checks/ForInCheck.java
vilchik-elena: @pynicolas i like better this way, because with variable i will need "if" statement to check statement kind

PR #92, opened 2015-08-20
SONARJS-453 Rule: Expression statements should not be useless

vilchik-elena:


javascript-checks/src/main/java/org/sonar/javascript/checks/StatementWithoutSideEffectsCheck.java
```diff
@@ -0,0 +1,70 @@
...
+import org.sonar.plugins.javascript.api.tree.expression.LiteralTree;
+import org.sonar.plugins.javascript.api.tree.statement.ExpressionStatementTree;
+import org.sonar.plugins.javascript.api.visitors.BaseTreeVisitor;
+import org.sonar.squidbridge.annotations.ActivatedByDefault;
+import org.sonar.squidbridge.annotations.SqaleConstantRemediation;
+import org.sonar.squidbridge.annotations.SqaleSubCharacteristic;
+
+@Rule(
+  key = "S2998",
+  name = "String literals and comparisons should have side effects",
```
pynicolas: @vilchik-elena I don't like the name of that rule: literals and comparisons should not have side effects, but they should not be useless.

javascript-checks/src/main/java/org/sonar/javascript/checks/StatementWithoutSideEffectsCheck.java
```diff
@@ -0,0 +1,70 @@
...
+
+  @Override
+  public void visitExpressionStatement(ExpressionStatementTree tree) {
+    Tree expression = tree.expression();
+
+    if (expression.is(Kind.EQUAL_TO)) {
+      getContext().addIssue(this, tree, MESSAGE);
+    }
+
+    if (expression.is(Kind.STRING_LITERAL) && !isUseStrictDirective(expression)) {
```
pynicolas: @vilchik-elena If we use isUseStrictDirective only here, why not pass it a LiteralTree instead of a Tree? That would avoid testing the kind twice.

PR #94, opened 2015-08-21
SONARJS-515 Rule: Results of operations on strings should not be ignored

pynicolas:


javascript-checks/src/main/java/org/sonar/javascript/checks/UselessStringOperationCheck.java
```diff
@@ -0,0 +1,71 @@
...
+public class UselessStringOperationCheck extends SubscriptionBaseVisitor {
+
+  @Override
+  public List<Kind> nodesToVisit() {
+    return ImmutableList.of(Kind.EXPRESSION_STATEMENT);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    Tree expression = ((ExpressionStatementTree) tree).expression();
```
lindamartin: @pynicolas maybe 1 or 2 blank line could have been used here :)

javascript-checks/src/main/java/org/sonar/javascript/checks/UselessStringOperationCheck.java
```diff
@@ -0,0 +1,71 @@
...
+    if (expression.is(Kind.CALL_EXPRESSION)) {
+      ExpressionTree callee = ((CallExpressionTree) expression).callee();
+      if (callee.is(Kind.DOT_MEMBER_EXPRESSION)) {
+        MemberExpressionTree memberExpression = (MemberExpressionTree) callee;
+        if (memberExpression.object().types().containsOnly(Type.Kind.STRING)) {
```
lindamartin: @pynicolas have you think about list all function strings and check for the property of the call expression to be one of them ? Maybe more cases could be caught ? There should not be to much risk for FP, WDYT ?

pynicolas: @lindamartin I thought about that solution and did not want to take the risk of false positives. What if someone implements the equivalent of a StringBuilder with the same method names as a string, e.g. concat? Maybe that would be acceptable.

lindamartin: @pynicolas from what I can think of:
if someone implement those functions for any kind of object:
* it might have the same kind of behaviour so is returning something. In which case the issue will be correct
* if it does not return anything but updating the current object, IMO it is a very corner case which will be a FP
* if it is an equivalent of StringBuilder, with the current implementation this kind of cases will be caught: ``myStringBuilder.toUpperCase().concat();`` where "concat()" is the last called function and not "build()" which will be kind of raising a real problem I think.

WDYT ?

lindamartin:
Merged in d9b7fd695c9e9db639fa7c2a328e58852409ff86.

PR #98, opened 2015-09-03
Retrieve plugin from target instead of local maven repository

lindamartin:


its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/Tests.java
```diff
@@ -43,22 +43,12 @@
...
-      orchestratorBuilder
-        .restoreProfileAtStartup(FileLocation.ofClasspath("/profile-javascript-custom-rules.xml"))
-        .addPlugin(FileLocation.of(TestUtils.pluginJar(CUSTOM_RULES_ARTIFACT_ID)));
-    }
-
-    orchestratorBuilder.restoreProfileAtStartup(FileLocation.ofClasspath("/empty-profile.xml"));
-    ORCHESTRATOR = orchestratorBuilder.build();
-  }
+  public static final Orchestrator ORCHESTRATOR = Orchestrator.builderEnv()
+    .addPlugin(FileLocation.of(new File(TestUtils.homeDir(), "../../sonar-javascript-plugin/target/sonar-javascript-plugin.jar")))
```
pynicolas: @lindamartin Not sure we need to use `TestUtils.homeDir()`. When launched with maven or inside an IDE, I think that the current directory should always be "its/plugin".

its/type-inference/tests/src/test/java/com/sonar/javascript/it/JavaScriptTypeInferenceTest.java
```diff
@@ -23,23 +23,20 @@
...
-
 public class JavaScriptTypeInferenceTest {
 
-  private static final String PLUGIN_KEY = "javascript";
   @ClassRule
   public static Orchestrator orchestrator = Orchestrator.builderEnv()
-    .addPlugin(PLUGIN_KEY)
-    .setMainPluginKey(PLUGIN_KEY)
-    .addPlugin(MavenLocation.create("org.sonarsource.sonar-lits-plugin", "sonar-lits-plugin", "0.5-SNAPSHOT"))
+    .addPlugin(FileLocation.of(new File(TestUtils.homeDir(), "../../sonar-javascript-plugin/target/sonar-javascript-plugin.jar")))
```
pynicolas: @lindamartin same thing here: I would suggest to change that to `.addPlugin(FileLocation.of(new File("../../sonar-javascript-plugin/target/sonar-javascript-plugin.jar")))`

pynicolas:
Looks good to me!

its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/Tests.java
lindamartin: @pynicolas without it it does not find the jar. I have to admit I have not really investigate into details...

PR #106, opened 2015-11-16
SONARJS-532 FP ConstructorFunctionsForSideEffectsCheck: "export default" declaration

vilchik-elena:


javascript-checks/src/test/resources/checks/constructorFunctionsForSideEffects.js
```diff
@@ -1,5 +1,22 @@
+import {Router} from 'backbone';
+
+class AppRouter extends Router {
```
pynicolas: @vilchik-elena do we need such a complex example?

vilchik-elena: @pynicolas indeed, we could simplify it

PR #111, opened 2015-11-19
export default statement: replace EOS with optional semicolon

vilchik-elena:


vilchik-elena:
Optional semicolon for export default statement should be treated the same way as for other statements

PR #114, opened 2015-11-23
SONARJS-553 False Negative Semicolon rule: export statements

vilchik-elena:


javascript-checks/src/test/resources/checks/semicolon.js
```diff
@@ -39,18 +39,31 @@ function f() {
...
+
+export class C { }  // OK
+export function foo() {}  // OK
+export function * foo() {} // OK
+
+export default class C { }  // OK
+export default function foo() {}  // OK
+export default function * foo() {} // OK
+export default class { }  // OK
+export default function () {}  // OK
```
pynicolas: @vilchik-elena This a function expression, not a declaration. I think we should raise an issue in such a case.

vilchik-elena: @pynicolas Actually it's function declaration according to a grammar (see http://www.ecma-international.org/ecma-262/6.0/#sec-functions-and-classes). But indeed in implementation it's presented as function expression. WDYT should we fix that to be compliant with grammar?

So semicolon here should be avoided.

pynicolas: @vilchik-elena Ok, understood. The specification seems to consider it as a function declaration in the context of `export default`.

PR #118, opened 2015-11-24
SONARJS-523 Fix parsing of "export default" and false negatives in SemiColonCheck

pynicolas:


pynicolas:
Duplicates https://github.com/SonarCommunity/sonar-javascript/pull/114

PR #119, opened 2015-11-26
SONARJS-510 Improve S930 and S2999: messages should not contain the full body of a function

vilchik-elena:


javascript-frontend/src/main/java/org/sonar/javascript/tree/symbols/type/ObjectType.java
```diff
@@ -24,7 +24,7 @@
 public class ObjectType implements Type {
 
 
-  private Callability callability;
+  protected Callability callability;
```
pynicolas: @vilchik-elena This doesn't look good. Why not remove the `callability` field from ObjectType and just keep the `callability()` method which can be overridden in subclasses?

PR #123, opened 2015-11-26
SONARJS-563 Improve issue location for OneStatementPerLineCheck

vilchik-elena:


javascript-checks/src/main/java/org/sonar/javascript/checks/OneStatementPerLineCheck.java
```diff
@@ -47,7 +48,7 @@
 @SqaleConstantRemediation("1min")
 public class OneStatementPerLineCheck extends SubscriptionBaseTreeVisitor {
 
-  private final Map<Integer, Integer> statementsPerLine = Maps.newHashMap();
+  private final Map<Integer, List<StatementTree>> statementsPerLine = Maps.newHashMap();
```
pynicolas: @vilchik-elena why not use a com.google.common.collect.ListMultimap?

vilchik-elena: @pynicolas thanks! I will refactor that

PR #135, opened 2015-11-30
Single maven reactor

pynicolas:


its/plugin/plugins/pom.xml
```diff
@@ -4,13 +4,13 @@
   <modelVersion>4.0.0</modelVersion>
 
   <parent>
-    <groupId>org.sonarsource.it.javascript</groupId>
-    <artifactId>it-javascript-plugin</artifactId>
-    <version>1.0-SNAPSHOT</version>
+    <groupId>org.sonarsource.javascript</groupId>
+    <artifactId>javascript-it-plugin</artifactId>
+    <version>2.9-SNAPSHOT</version>
```
vilchik-elena: @pynicolas do we really need hardcoded version here?

pynicolas: @vilchik-elena I think so. Is there a better way to do it?

vilchik-elena:
LGTM

PR #139, opened 2015-12-14
SONARJS-143 Rule: Parameters should be passed in the correct order

pynicolas:


javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S2234.html
```diff
@@ -0,0 +1,32 @@
+<p>When the names of parameters in a function call match the names of the function arguments, it contributes to clearer, more readable code. However, when the names match, but are passed in a different order than the function arguments, it indicates a mistake in the parameter order which will likely lead to unexpected results.</p>
```
vilchik-elena: @pynicolas IMHO function declared with **parameters** and called with **arguments**. It's vice versa here.

pynicolas: @vilchik-elena I agree. Fixed in 38a60d6705234c9124e99377b12d40deb68d20b0.

javascript-checks/src/test/resources/checks/misorderedParameterList.js
```diff
@@ -0,0 +1,28 @@
...
+  f1(p1, "p2", "x");
+  unknown(p1, p3, p2);
+  f2(p1, p3, p2);
+  (function(p1, p2) {})(p1, p2);
+  (function(p1, p2) {})(p2, p1); // Noncompliant {{Parameters to this call have the same names but not the same order as the function arguments.}}
+  var a1 = new A();
+  a1.method1(p1, p2);
+  a1.method1(p2, p1); // False negative
+  withInitializer(p2, p1);
+  arrayBindingPattern(p2, p1);
```
vilchik-elena: @pynicolas why no issue in these two cases?

javascript-checks/src/main/java/org/sonar/javascript/checks/MisorderedParameterListCheck.java
```diff
@@ -0,0 +1,106 @@
...
+  key = "S2234",
+  name = "Parameters should be passed in the correct order",
+  priority = Priority.BLOCKER,
+  tags = {Tags.BUG})
+@ActivatedByDefault
+@SqaleSubCharacteristic(RulesDefinition.SubCharacteristics.LOGIC_RELIABILITY)
+@SqaleConstantRemediation("5min")
+public class MisorderedParameterListCheck extends BaseTreeVisitor {
+
+  private static final String MESSAGE_FORMAT = "Parameters to %s have the same names but not the same order as the function arguments.";
```
vilchik-elena: @pynicolas let's update message too

javascript-checks/src/main/java/org/sonar/javascript/checks/MisorderedParameterListCheck.java
```diff
@@ -0,0 +1,106 @@
...
+    TypeSet types = tree.callee().types();
+    FunctionType functionType = (FunctionType) types.getUniqueType(Type.Kind.FUNCTION);
+    return functionType == null ? null : functionType.functionTree();
+  }
+
+  private static boolean haveSameNamesAndDifferentOrders(List<String> argumentNames, List<String> parameterNames) {
+    return new HashSet<>(argumentNames).equals(new HashSet<>(parameterNames)) && !argumentNames.equals(parameterNames);
+  }
+
+  @CheckForNull
```
vilchik-elena: @pynicolas I usually use `@Nullable` in such cases. Is there a difference?

javascript-checks/src/test/resources/checks/misorderedParameterList.js
pynicolas: @vilchik-elena Fixed in 5f56749f3549c715ceac20e3cc27afbaeb88bb40 for the first case. In the 2nd case, the parameter has no name, I don't think that the rule can apply in such a case.

javascript-checks/src/main/java/org/sonar/javascript/checks/MisorderedParameterListCheck.java
pynicolas: @vilchik-elena Fixed in b601c8e920a18e4e8e07ebda6acf170b1f9823d0.

javascript-checks/src/main/java/org/sonar/javascript/checks/MisorderedParameterListCheck.java
pynicolas: @vilchik-elena I use `@CheckForNull` on methods. See http://stackoverflow.com/questions/12301570/javax-annotation-nullable-vs-checkfornull. I think that SQ Java rules are looking for `@CheckForNull` on methods, not `@Nullable`.

vilchik-elena: @pynicolas thanks!

PR #140, opened 2015-12-15
SONARJS-457 Rule: Multiline blocks should be enclosed in curly braces

pynicolas:


javascript-checks/src/main/java/org/sonar/javascript/checks/MultilineBlockCurlyBraceCheck.java
```diff
@@ -0,0 +1,134 @@
...
+@ActivatedByDefault
+@SqaleSubCharacteristic(RulesDefinition.SubCharacteristics.LOGIC_RELIABILITY)
+@SqaleConstantRemediation("5min")
+public class MultilineBlockCurlyBraceCheck extends BaseTreeVisitor {
+
+  private static final Kind[] NESTING_STATEMENT_KINDS = {
+    Kind.IF_STATEMENT,
+    Kind.FOR_STATEMENT,
+    Kind.FOR_IN_STATEMENT,
+    Kind.FOR_OF_STATEMENT
```
vilchik-elena: @pynicolas what about `while` loop?

pynicolas: @vilchik-elena Fixed in 5837115933304cd86fcf5f8d3c848fc54b048a5e. Thanks!

javascript-checks/src/test/resources/checks/multilineBlockCurlyBrace.js
```diff
@@ -0,0 +1,60 @@
...
+action2();
+
+if (condition)
+  action1();
+
+  action2(); // Noncompliant
+
+if (condition)
+action1();
+action2();
```
vilchik-elena: @pynicolas Do you think that it's not valuable to have issue here?

pynicolas: @vilchik-elena To me, it looks like false positives, e.g. on Java: http://nemo.sonarqube.org/issues/search#issues=93a6fddd-db3f-4a21-944a-c9e3c036756c or http://nemo.sonarqube.org/issues/search#issues=5bbc10a1-181a-40b8-acb2-0aa989f18428

vilchik-elena: @pynicolas indeed, agree. Another rule should care about such cases.

javascript-checks/src/test/resources/checks/multilineBlockCurlyBrace.js
```diff
@@ -0,0 +1,69 @@
...
+
+  if (condition)
+action1();
+action2(); // compliant, less indented
+
+if (condition) action1(); action2(); // compliant (?)
```
vilchik-elena: @pynicolas very interesting case. WDYT about it?

javascript-checks/src/test/resources/checks/multilineBlockCurlyBrace.js
```diff
@@ -0,0 +1,69 @@
...
+  action3();
+
+if (condition)
+  action1();
+action2();
+
+if (condition)
+  action1();
+
+  action2(); // Noncompliant
```
vilchik-elena: @pynicolas I would add one more test case when empty line after :
```
if (condition)
  action1();
  action2();  // Noncompliant [[secondary=+2]] {{This line will not be executed conditionally; only the first line of this 4-line block will be. The rest will execute unconditionally.}}

  action3();
```

javascript-checks/src/test/resources/checks/multilineBlockCurlyBrace.js
pynicolas: @vilchik-elena I think that this case is a bit different, it's not really about indentation, even if it's still about formatting. Another rule may be more appropriate. If we want to consider it as noncompliant, we have to be careful: I can very well imagine this kind of code inside a one-line function expression for example.

vilchik-elena: @pynicolas Agree, it should be another rule. I think we should ask Ann and remove this question mark to not be confused in the future

javascript-checks/src/test/resources/checks/multilineBlockCurlyBrace.js
pynicolas: @vilchik-elena fixed in 007e1c2.

javascript-checks/src/test/resources/checks/multilineBlockCurlyBrace.js
pynicolas: @vilchik-elena Removed in 3a3c467.

javascript-checks/src/main/java/org/sonar/javascript/checks/MultilineBlockCurlyBraceCheck.java
```diff
@@ -0,0 +1,135 @@
...
+    Kind.FOR_OF_STATEMENT,
+    Kind.WHILE_STATEMENT
+  };
+
+  public static final String IF_PRIMARY_MESSAGE = "This line will not be executed conditionally; "
+    + "only the first line of this %s-line block will be. The rest will execute unconditionally.";
+  public static final String LOOP_PRIMARY_MESSAGE = "This line will not be executed in a loop; "
+    + "only the first line of this %s-line block will be. The rest will execute only once.";
+  public static final String IF_SECONDARY_MESSAGE = "indented but not conditionally executed";
+  public static final String LOOP_SECONDARY_MESSAGE = "indented but not executed in a loop";
```
vilchik-elena: @pynicolas This message is too long, in SQ we see only "indented but not executed in a ...".

vilchik-elena: @pynicolas What about "indented as previous line"?

pynicolas: @vilchik-elena Please see c939ad3.

PR #142, opened 2015-12-16
SONARJS-119 Rule: Loop counters should not be assigned to from within the loop body

vilchik-elena:


its/ruling/src/test/expected/javascript-S2310.json
```diff
@@ -0,0 +1,293 @@
...
+9015,
+11778,
+],
+'project:ecmascript6/di.js/visualize/lib/angular.js':[
+6555,
+12108,
+12175,
+],
+'project:ecmascript6/di.js/visualize/lib/d3.v3.js':[
+1406,
```
pynicolas: @vilchik-elena can this case be considered as a loop counter?

javascript-checks/src/main/java/org/sonar/javascript/checks/CounterUpdatedInLoopCheck.java
```diff
@@ -0,0 +1,195 @@
...
+
+    writeUsagesOfCounters.addLast(writeUsages);
+  }
+
+  private void leaveLoopBody() {
+    writeUsagesOfCounters.removeLast();
+  }
+
+  private void raiseIssue(IdentifierTree writeUsage) {
+    getContext().addIssue(this, new IssueLocation(writeUsage, String.format(MESSAGE, writeUsage.name())), Collections.<IssueLocation>emptyList(), null);
```
pynicolas: @vilchik-elena I think it may be useful to have a secondary location on the loop counter, especially when there are some nested loops.

its/ruling/src/test/java/org/sonar/javascript/it/JavaScriptTest.java
```diff
@@ -59,7 +59,7 @@ public void test() throws Exception {
       .setLanguage("js")
       .setSourceDirs("./")
       .setSourceEncoding("utf-8")
-      .setProperty("sonar.analysis.mode", "preview")
+//      .setProperty("sonar.analysis.mode", "preview")
```
pynicolas: @vilchik-elena Is this an intentional change?

javascript-checks/src/main/java/org/sonar/javascript/checks/CounterUpdatedInLoopCheck.java
```diff
@@ -0,0 +1,195 @@
...
+      IdentifierTree identifier = (IdentifierTree) counterBlock;
+      if (identifier.symbol() != null) {
+        currentLoopCounters.add(identifier.symbol());
+      }
+
+    }
+
+  }
+
+  private void enterLoopBody(Set<Symbol> counterSymbols) {
```
pynicolas: @vilchik-elena This method is always called with currentLoopCounters which is a field. Why do we need a parameter in that case?

javascript-checks/src/main/java/org/sonar/javascript/checks/CounterUpdatedInLoopCheck.java
```diff
@@ -0,0 +1,195 @@
...
+  @Override
+  public void visitForOfStatement(ForOfStatementTree tree) {
+    scan(tree.variableOrExpression());
+    scan(tree.expression());
+    visitObjectIterationStatement(tree, tree.variableOrExpression());
+  }
+
+  @Override
+  public void visitIdentifier(IdentifierTree tree) {
+    Iterator<Set<IdentifierTree>> iterator = writeUsagesOfCounters.descendingIterator();
```
pynicolas: @vilchik-elena Why do we care about the order of iteration here?

javascript-checks/src/main/java/org/sonar/javascript/checks/CounterUpdatedInLoopCheck.java
```diff
@@ -0,0 +1,195 @@
...
+  @Override
+  public void visitAssignmentExpression(AssignmentExpressionTree tree) {
+    if (inUpdate) {
+      checkIdentifier(tree.variable());
+    }
+
+    super.visitAssignmentExpression(tree);
+  }
+
+  private void checkIdentifier(ExpressionTree tree) {
```
pynicolas: @vilchik-elena Suggested name: addCounterToCurrentLoop

javascript-checks/src/main/java/org/sonar/javascript/checks/CounterUpdatedInLoopCheck.java
```diff
@@ -0,0 +1,195 @@
...
+          currentLoopCounters.add(identifierTree.symbol());
+        }
+      }
+
+    } else if (counterBlock instanceof AssignmentExpressionTree) {
+      scanCounterBlock(((AssignmentExpressionTree) counterBlock).variable());
+
+    } else if (counterBlock instanceof IdentifierTree) {
+      IdentifierTree identifier = (IdentifierTree) counterBlock;
+      if (identifier.symbol() != null) {
```
pynicolas: @vilchik-elena This pattern seems to be duplicated 3 times in this class.

javascript-checks/src/main/java/org/sonar/javascript/checks/CounterUpdatedInLoopCheck.java
vilchik-elena: @pynicolas WDYT should be secondary location? usage in update clause?

its/ruling/src/test/java/org/sonar/javascript/it/JavaScriptTest.java
vilchik-elena: @pynicolas indeed, my bad.  Fixed in 7db40a82cef57222bab3fa7c805653f11da26808

javascript-checks/src/main/java/org/sonar/javascript/checks/CounterUpdatedInLoopCheck.java
vilchik-elena: @pynicolas historical reasons:) thanks! fixed in fbefb17fed01227e3092be002a31b0aef4c9432a

javascript-checks/src/main/java/org/sonar/javascript/checks/CounterUpdatedInLoopCheck.java
vilchik-elena: @pynicolas fixed in 0bb3846b1f0dcd124cf236cc9fbdabec99d77544

javascript-checks/src/main/java/org/sonar/javascript/checks/CounterUpdatedInLoopCheck.java
vilchik-elena: @pynicolas fixed in 0bb3846b1f0dcd124cf236cc9fbdabec99d77544

javascript-checks/src/main/java/org/sonar/javascript/checks/CounterUpdatedInLoopCheck.java
vilchik-elena: @pynicolas indeed, it doesn't matter. Fixed in b82aa62d5af53b6a76c53f12886bac1d810d9ceb

javascript-checks/src/main/java/org/sonar/javascript/checks/CounterUpdatedInLoopCheck.java
vilchik-elena: @pynicolas Please see ee74a7ca230d4d23898ec084ad279de87ddc7153

PR #145, opened 2015-12-17
SONARJS-450 Add Rule: Comparison operators should not be used with strings

vilchik-elena:


javascript-checks/src/main/java/org/sonar/javascript/checks/StringsComparisonCheck.java
```diff
@@ -0,0 +1,78 @@
...
+  priority = Priority.MAJOR,
+  tags = Tags.SUSPICIOUS)
+@ActivatedByDefault
+@SqaleSubCharacteristic(SubCharacteristics.INSTRUCTION_RELIABILITY)
+@SqaleConstantRemediation("5min")
+public class StringsComparisonCheck extends BaseTreeVisitor {
+
+  private static final String MESSAGE = "Convert operands of this use of \"%s\" to number type.";
+
+  private static final Kind[] COMPARATIVE_EXPRESSIONS = {
```
pynicolas: @vilchik-elena They are called relational operators in some documentations.

javascript-checks/src/main/java/org/sonar/javascript/checks/StringsComparisonCheck.java
```diff
@@ -0,0 +1,78 @@
...
+    Kind.LESS_THAN,
+    Kind.LESS_THAN_OR_EQUAL_TO,
+    Kind.GREATER_THAN,
+    Kind.GREATER_THAN_OR_EQUAL_TO
+  };
+
+  @Override
+  public void visitBinaryExpression(BinaryExpressionTree tree) {
+    if (tree.is(COMPARATIVE_EXPRESSIONS) && isString(tree.leftOperand()) && isString(tree.rightOperand())) {
```
pynicolas: @vilchik-elena What about expressions comparing a string and a non-string (e.g. a number)? Are they valid comparisons? Should we have separate rules for those cases?

javascript-checks/src/main/java/org/sonar/javascript/checks/StringsComparisonCheck.java
vilchik-elena: @pynicolas indeed, agree. Fixed in b1a81776f2bd4df5f0d670938af1e5cfe6df26b3

javascript-checks/src/main/java/org/sonar/javascript/checks/StringsComparisonCheck.java
vilchik-elena: @pynicolas When at least one operand not string both operands are casted to number. E.g. `"2" > true ` will be casted to `2 > 1` and will return true. For sure it's not the best practice but this rule covers cases when lexical comparison is made.

pynicolas: @vilchik-elena ok, thanks for the explanation!

PR #149, opened 2015-12-18
SONARJS-525 Add ability to exclude minified files from the analysis

vilchik-elena:


sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/JavaScriptSquidSensorTest.java
```diff
@@ -236,6 +229,55 @@ private void analyseFileWithException(JavaScriptCheck check, InputFile inputFile
   }
 
   @Test
+  public void not_analyse_minified_files_default_config() throws Exception {
+    settings.setProperty(JavaScriptPlugin.EXCLUDE_MINIFIED_FILES, true);
```
pynicolas: @vilchik-elena Why not have the same setup shared with analyse_minified_files_user_config, only with a difference in "settings"?

vilchik-elena: @pynicolas thanks! fixed a244517fbbb8a79044934ff6ba1f79cb61b542f4

PR #166, opened 2016-01-07
SONARJS-546 Fix highlighting of files with BOM

pynicolas:


javascript-frontend/src/main/java/org/sonar/javascript/highlighter/SourceFileOffsets.java
```diff
@@ -52,7 +53,8 @@ private static String fileContent(File file, Charset charset) {
   }
 
   private void initOffsets(String toParse) {
-    lineStartOffsets.add(0);
+    boolean hasByteOrderMark = toParse.startsWith(Character.toString(JavaScriptNodeBuilder.BYTE_ORDER_MARK));
```
dbolkensteyn: @pynicolas I think it might be better to keep all the logic in `JavaScriptNodeBuilder`: You'll need to subtract 1 to the column if the token is on line 1.

With the current solution, `getColumn()` on a token that is on line 1 with a BOM UTF-8 character will return an incorrect value.

javascript-frontend/src/main/java/org/sonar/javascript/highlighter/SourceFileOffsets.java
```diff
@@ -26,6 +26,7 @@
 import java.io.IOException;
 import java.nio.charset.Charset;
 import java.util.List;
+import org.sonar.javascript.parser.JavaScriptNodeBuilder;
 import org.sonar.plugins.javascript.api.tree.lexical.SyntaxToken;
 
 public class SourceFileOffsets {
```
dbolkensteyn: unrelated to this PR @pynicolas - but do you really need this class? Every token seems to know its start offset, and its end offset seems to always be its `start + length(value)`

pynicolas: @dbolkensteyn some tokens are in fact trivias (comments), and they don't have this start offset (for the moment).

pynicolas: @dbolkensteyn In fact, I don't see a way to set offsets on trivias without some new methods in SSLR.

javascript-frontend/src/main/java/org/sonar/javascript/highlighter/SourceFileOffsets.java
pynicolas: @dbolkensteyn I fixed columns in 9175675f620e1ebfc93592d0cb63e84404898cb4.

javascript-frontend/src/main/java/org/sonar/javascript/highlighter/SourceFileOffsets.java
```diff
@@ -52,6 +54,7 @@ private static String fileContent(File file, Charset charset) {
   }
 
   private void initOffsets(String toParse) {
+    boolean hasByteOrderMark = toParse.startsWith(Character.toString(JavaScriptNodeBuilder.BYTE_ORDER_MARK));
     lineStartOffsets.add(0);
```
dbolkensteyn: This works because 1 is being subtracted from all token columns on line 1 (and only on line 1)

dbolkensteyn:
LGTM, manually tested!

It's definitely tricky to get this right ;-)

PR #185, opened 2016-02-22
SONARJS-606 Rule: Attempts should not be made to update "const" variables

vilchik-elena:


vilchik-elena:
Merged in b8167e6d02205915950e49dffb4387b04ba70d99

PR #187, opened 2016-02-22
SONARJS-609 Rule: Function parameters with default values should be last

vilchik-elena:


javascript-checks/src/main/java/org/sonar/javascript/checks/DefaultParametersNotLastCheck.java
```diff
@@ -0,0 +1,104 @@
...
+import org.sonar.plugins.javascript.api.visitors.SubscriptionVisitorCheck;
+import org.sonar.squidbridge.annotations.ActivatedByDefault;
+import org.sonar.squidbridge.annotations.SqaleConstantRemediation;
+import org.sonar.squidbridge.annotations.SqaleSubCharacteristic;
+
+@Rule(
+  key = "S1788",
+  name = "Function parameters with default values should be last",
+  priority = Priority.CRITICAL,
+  tags = {Tags.ES2015, Tags.BUG, Tags.PSR2})
```
pynicolas: @vilchik-elena I don't think we should set the PSR2 tag: this is specific to PHP.

javascript-checks/src/test/resources/checks/DefaultParametersNotLast.js
```diff
@@ -0,0 +1,20 @@
...
+}
+
+foo((p1 = 1, p2) => {   // Noncompliant [[sc=6;ec=8]]
+});
+
+function foo({a, b} = {a:42, b:"hello"}, p1){ // Noncompliant {{Move parameter "{a, b}" after parameters without default value.}}
+}
+
+function foo(p1 = 1, ...restArgs) {  // Noncompliant
+}
```
pynicolas: @vilchik-elena I would add at least one compliant case.

vilchik-elena: @pynicolas indeed, thanks!

javascript-checks/src/main/java/org/sonar/javascript/checks/DefaultParametersNotLastCheck.java
vilchik-elena: @pynicolas Agree. BTW WDYT about "misra"?

pynicolas: @vilchik-elena Currently, we have 17 rules tagged with "misra".

vilchik-elena: @pynicolas And what does this statistics mean?:)

pynicolas: @vilchik-elena We have rules tagged with "misra", that's all. However, I don't want to add "psr2". What did you mean with your initial question: "BTW WDYT about "misra"?"?

PR #190, opened 2016-02-24
Identifier kinds refactoring

vilchik-elena:


pynicolas:
@vilchik-elena Looks good to me.

PR #196, opened 2016-02-25
SONARJS-625 Rule: Wildcard imports should not be used

vilchik-elena:


javascript-checks/src/test/resources/checks/WildcardImport.js
```diff
@@ -0,0 +1,6 @@
+import * as name from "module-name";   // Noncompliant {{Explicitly import the specific member needed.}}
+import defaultMember, * as name from "module-name"; // Noncompliant [[sc=23;ec=24]]
+
+export * from "module-name";  // Noncompliant
```
pynicolas: @vilchik-elena "export" is not mentioned in the title or in the description of the rule and the message says "Explicitly import the specific member needed.".

vilchik-elena: @pynicolas Indeed, we should update description. 
```
export * from "module";
// is the shortcut for
import * as Imported from "module"
export Imported
```

PR #210, opened 2016-03-11
Support for overall coverage

tisoft:
The java plugin combines the IT und UT coverage into an overall coverage value. This is my try to do the same for the javascript plugin. Seems to work fine in our builds.


vilchik-elena:
Hi @tisoft,
Thanks for the PR! I have created a ticket and assigned it to the next release (https://jira.sonarsource.com/browse/SONARJS-658).
But before merging we need some improvements:
1) there is no nor unit test nor integration test verifying this feature
2) build of plugin IT has failed (https://travis-ci.org/SonarSource/sonar-javascript/builds/115328759).
It's hard to define the reason for this failure as you moved implementation of UTCoverageSensor to LCOVCoverageSensor and git doesn't show diff.

tisoft:
I rebased to the latest master and the tests run through. I will try to add a new unit test to test the combined coverage feature now.

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/lcov/LCOVParser.java
```diff
@@ -34,6 +29,13 @@
 import org.sonar.api.batch.fs.InputFile;
 import org.sonar.api.measures.CoverageMeasuresBuilder;
 
+import javax.annotation.CheckForNull;
```
vilchik-elena: @tisoft Could you change project import formatting settings like here https://github.com/SonarSource/sonar-developer-toolset#code-style

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/lcov/CoverageSensorTest.java
```diff
@@ -90,28 +91,54 @@ public void report_not_found() throws Exception {
...
+  public void test_file_in_overall_coverage_report() {
+    DefaultFileSystem fs = newFileSystem();
+    fs.add(newSourceInputFile("fake_file1.js"));
+    fs.add(newSourceInputFile("fake_file2.js"));
+    fs.add(newSourceInputFile("fake_file3.js"));
+    newOverallSensor(fs, settings).analyse(project, context);
+
+    // 3 line coverage metrics for two files, 4 condition coverage metrics for 1 file for both the it and ut file
+    // minus 3 line coverages, that are in both files and so get merged
+    verify(context, times(17)).saveMeasure(any(Resource.class), (Measure) anyObject());
```
vilchik-elena: @tisoft It seems to me that explanation is not precise. WDYT about "3 line coverage metrics for 3 files, 4 condition coverage metrics for 2 files"?

vilchik-elena: @tisoft Can we also assert some metrics values? (e.g. the way it's done on line 143)

PR #213, opened 2016-03-24
SONARJS-666 Rule: Array indexes should be numeric

hbensalem:
Implementation for the new rule I have suggested
RSPEC-3579 Array indexes should be numeric
https://jira.sonarsource.com/browse/RSPEC-3579


javascript-checks/src/main/java/org/sonar/javascript/checks/AssociativeArraysCheck.java
```diff
@@ -0,0 +1,68 @@
...
+import org.sonar.javascript.tree.impl.expression.BracketMemberExpressionTreeImpl;
+import org.sonar.javascript.tree.impl.expression.IdentifierTreeImpl;
+import org.sonar.javascript.tree.impl.expression.LiteralTreeImpl;
+import org.sonar.plugins.javascript.api.symbols.Type.Kind;
+import org.sonar.plugins.javascript.api.tree.expression.AssignmentExpressionTree;
+import org.sonar.plugins.javascript.api.visitors.DoubleDispatchVisitorCheck;
+import org.sonar.squidbridge.annotations.SqaleConstantRemediation;
+import org.sonar.squidbridge.annotations.SqaleSubCharacteristic;
+
+@Rule(key = "AssociativeArrays", priority = Priority.MAJOR, name = "Associative arrays should not be used.", tags = {
```
vilchik-elena: @laftanza why this only line break. You better follow other rules style and put one parameter per line

javascript-checks/src/main/java/org/sonar/javascript/checks/AssociativeArraysCheck.java
```diff
@@ -0,0 +1,68 @@
...
+		"suspicious" })
+@SqaleSubCharacteristic(RulesDefinition.SubCharacteristics.DATA_RELIABILITY)
+@SqaleConstantRemediation("10min")
+
+public class AssociativeArraysCheck extends DoubleDispatchVisitorCheck {
+	private static final String MESSAGE = "Only use a numeric index for Arrays";
+
+	@Override
+	public void visitAssignmentExpression(AssignmentExpressionTree tree) {
+		if (tree.variable() instanceof BracketMemberExpressionTreeImpl) {
```
vilchik-elena: @laftanza Please use tree.variable().is(Kind.BRACKET_MEMBER_EXPRESSION) instead

javascript-checks/src/main/java/org/sonar/javascript/checks/AssociativeArraysCheck.java
```diff
@@ -0,0 +1,68 @@
...
+			if (((BracketMemberExpressionTreeImpl) tree.variable()).object().types().contains(Kind.ARRAY)) {
```
vilchik-elena: @laftanza you shouldn't use *Impl classes, but interfaces only, so cast `tree.variable()` to `BracketMemberExpressionTree`. The same for `LiteralTreeImpl` below.

vilchik-elena: @laftanza I think it's better to be safe and use `...types().containsOnly(Kind.ARRAY)` instead of `...types().contains(Kind.ARRAY)`. That way you avoid false positive here
```
var x = {};
x['name'] = 'John';  // False positive here
...
x = [];
...
```

javascript-checks/src/main/java/org/sonar/javascript/checks/AssociativeArraysCheck.java
```diff
@@ -0,0 +1,68 @@
...
+				if (((BracketMemberExpressionTreeImpl) tree.variable()).property() instanceof LiteralTreeImpl)
```
vilchik-elena: @laftanza You can write `((BracketMemberExpressionTreeImpl) tree.variable()).property().is(Kind.STRING_LITERAL)` and avoid check on line 52

javascript-checks/src/main/java/org/sonar/javascript/checks/AssociativeArraysCheck.java
```diff
@@ -0,0 +1,68 @@
...
+				{
```
vilchik-elena: @laftanza please move this brace on previous line

javascript-checks/src/main/java/org/sonar/javascript/checks/AssociativeArraysCheck.java
vilchik-elena: @laftanza you can extract `(BracketMemberExpressionTreeImpl) tree.variable()` to a local variable and not do this cast on lines 47, 50, 56 and 58

javascript-checks/src/main/java/org/sonar/javascript/checks/AssociativeArraysCheck.java
```diff
@@ -0,0 +1,68 @@
...
+import org.sonar.plugins.javascript.api.tree.expression.AssignmentExpressionTree;
+import org.sonar.plugins.javascript.api.visitors.DoubleDispatchVisitorCheck;
+import org.sonar.squidbridge.annotations.SqaleConstantRemediation;
+import org.sonar.squidbridge.annotations.SqaleSubCharacteristic;
+
+@Rule(key = "AssociativeArrays", priority = Priority.MAJOR, name = "Associative arrays should not be used.", tags = {
+		"suspicious" })
+@SqaleSubCharacteristic(RulesDefinition.SubCharacteristics.DATA_RELIABILITY)
+@SqaleConstantRemediation("10min")
+
```
vilchik-elena: @laftanza Please add `@ActivatedByDefault` as this rule is part of default profile (see https://jira.sonarsource.com/browse/RSPEC-3579)

javascript-checks/src/test/java/org/sonar/javascript/checks/AssociativeArraysCheckTest.java
```diff
@@ -0,0 +1,35 @@
...
+import java.io.File;
+import org.junit.Test;
+import org.sonar.javascript.checks.verifier.JavaScriptCheckVerifier;
+
+
+public class AssociativeArraysCheckTest {
+
+	@Test
+	  public void test() throws Exception {
+	    JavaScriptCheckVerifier.issues(new AssociativeArraysCheck(), new File("src/test/resources/checks/associativeArraysCheck.js"))
```
vilchik-elena: @laftanza You can use `JavaScriptCheckVerifier#verify` (you even already have comments in file, just extend them a bit)

javascript-checks/src/test/resources/checks/associativeArraysCheck.js
```diff
@@ -0,0 +1,5 @@
+let arr = [];
+arr[0] = 'a';
+arr['name'] = 'bob';  // Noncompliant
+arr[1] = 'foo';
+arr[lastname]="Ben"; //Noncompliant
```
vilchik-elena: @laftanza Could you add more test cases: when associative massive element is read, when object is used, when the same variable is used for array and object

javascript-checks/src/main/java/org/sonar/javascript/checks/AssociativeArraysCheck.java
```diff
@@ -0,0 +1,68 @@
...
+	public void visitAssignmentExpression(AssignmentExpressionTree tree) {
+		if (tree.variable() instanceof BracketMemberExpressionTreeImpl) {
+			if (((BracketMemberExpressionTreeImpl) tree.variable()).object().types().contains(Kind.ARRAY)) {
+				if (((BracketMemberExpressionTreeImpl) tree.variable()).property() instanceof LiteralTreeImpl)
+				{
+
+				LiteralTreeImpl al = (LiteralTreeImpl) ((BracketMemberExpressionTreeImpl) tree.variable()).property();
+
+				if (!NumberUtils.isNumber(al.token().text())) {
+					addLineIssue(tree, MESSAGE);
```
vilchik-elena: @laftanza Please instead use `addIssue(al)` to have correct highlighting of issue

vilchik-elena:
Hi @laftanza! 

Thanks for the pull request! 
Before we can merge please do couple of improvements:
1. Follow code style https://github.com/SonarSource/sonar-developer-toolset#code-style (in particular no tabs should be used and one indent is equal 2 spaces)
2. Description was created by Ann so please follow it (https://jira.sonarsource.com/browse/RSPEC-3579): key (S3579 instead of AssociativeArrays), description, message, SQALE etc.
3. Fix issues raised by me in comments

Finally, test is red because you didn't update ruling integration test results (see here https://github.com/SonarSource/sonar-javascript#ruling-test how to do that).

Do not hesitate to ask the things which are not clear for you!

laftanza:
Hi @vilchik-elena 
Thank you for your feedback! :)
When I run "mvn clean install" I get these 2 test failures even though I didn't chang "javascript-frontend":
 ```
 HighlightSymbolTableBuilderTest.sonar_symbol_table:62
Argument(s) are different! Wanted:
symbolTableBuilder.newReference(<any>, 29);
-> at org.sonar.javascript.highlighter.HighlightSymbolTableBuilderTest.sonar_symbol_table(HighlightSymbolTableBuilderTest.java:62)
Actual invocation has different arguments:
symbolTableBuilder.newReference(null, 33);
-> at org.sonar.javascript.highlighter.HighlightSymbolTableBuilder.highlightSymbol(HighlightSymbolTableBuilder.java:72)
  HighlightSymbolTableBuilderTest.test_properties:116
Argument(s) are different! Wanted:
symbolTableBuilder.newReference(<any>, 67);
-> at org.sonar.javascript.highlighter.HighlightSymbolTableBuilderTest.test_properties(HighlightSymbolTableBuilderTest.java:116)
Actual invocation has different arguments:
symbolTableBuilder.newReference(null, 83);
-> at org.sonar.javascript.highlighter.HighlightSymbolTableBuilder.highlightSymbol(HighlightSymbolTableBuilder.java:72)
```
As for the ruling integration test I get:
```
Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.665 sec <<< FAILURE! 
org.sonar.javascript.it.JavaScriptTest  Time elapsed: 0.663 sec  <<< ERROR!
java.lang.ExceptionInInitializerError: null  at com.sonar.orchestrator.locator.FileLocation.getOnlyFile(FileLocation.java:110)
at com.sonar.orchestrator.locator.FileLocation.byWildcardMavenFilename(FileLocation.java:105)
        at org.sonar.javascript.it.JavaScriptTest.<clinit>(JavaScriptTest.java:38)
Tests in error:
  JavaScriptTest.org.sonar.javascript.it.JavaScriptTest ╗ ExceptionInInitializer
```

File: javascript-checks/src/test/resources/checks/associativeArraysCheck.js [hunk 10]
```diff
@@ -1,5 +1,18 @@
-let arr = [];
-arr[0] = 'a';
-arr['name'] = 'bob';  // Noncompliant
-arr[1] = 'foo';
-arr[lastname]="Ben"; //Noncompliant
+let arr = []; 
+arr[0] = 'a'; // OK
+arr['name'] = 'bob'; // Noncompliant {{Make "arr" an object if it must have named properties; otherwise, use a numeric index here.}}
+arr[1] = 'foo'; // OK
+arr[lastname] = "Ben"; //  Noncompliant {{Make "arr" an object if it must have named properties; otherwise, use a numeric index here.}}
```
vilchik-elena: @hbensalem Can you explain why `arr[lastname]` is noncompliant? `lastname` could be a variable which holds a number

javascript-checks/src/test/resources/checks/associativeArraysCheck.js
```diff
@@ -1,5 +1,18 @@
...
+var x = {}; 
+x['name'] = 'John'; // OK
+x=[]; // OK
+let car = {
+  type : "Fiat",
+  model : "500",
+  color : "white"
+}; 
+car[type] = "BMW"; // OK
```
vilchik-elena: @hbensalem you can't access properties like that. 
```
car.type = ...
```
or
```
car["type"] = ...
```

javascript-checks/src/main/java/org/sonar/javascript/checks/AssociativeArraysCheck.java
```diff
@@ -0,0 +1,63 @@
...
+import org.sonar.plugins.javascript.api.visitors.DoubleDispatchVisitorCheck;
+import org.sonar.squidbridge.annotations.ActivatedByDefault;
+import org.sonar.squidbridge.annotations.SqaleConstantRemediation;
+import org.sonar.squidbridge.annotations.SqaleSubCharacteristic;
+
+@Rule(
+  key = "S3579",
+  priority = Priority.MAJOR,
+  name = "Array indexes should be numeric",
+  tags = {"suspicious"})
```
vilchik-elena: @hbensalem Could you use constant `Tags.SUSPICIOUS`?

javascript-checks/src/main/java/org/sonar/javascript/checks/AssociativeArraysCheck.java
```diff
@@ -0,0 +1,63 @@
...
+  private static final String MESSAGE = "Make \"%s\" an object if it must have named properties; otherwise, use a numeric index here.";
+
+  @Override
+  public void visitAssignmentExpression(AssignmentExpressionTree tree) {
+    if (tree.variable().is(Tree.Kind.BRACKET_MEMBER_EXPRESSION)) {
+      BracketMemberExpressionTree arrayObject = (BracketMemberExpressionTree) tree.variable();
+      if (arrayObject.object().types().containsOnly(Kind.ARRAY)) {
+        ExpressionTree arrayIndex = ((BracketMemberExpressionTreeImpl) tree.variable()).property();
+        if (arrayIndex.is(Tree.Kind.STRING_LITERAL)) {
+          addIssue(tree, String.format(MESSAGE, arrayObject.object().toString()));
```
vilchik-elena: @hbensalem this way the entire assignment expression will be highlighted as issue in UI. As specified in RSPEC we should highlight only property (see "Highlighting" field in https://jira.sonarsource.com/browse/RSPEC-3579), so you should pass `arrayIndex` instead of `tree`.

vilchik-elena: @hbensalem Please use `CheckUtils#asString()` instead of `toString()`.
Indeed now it works fine because `toString()` is implemented for IdentifierTree and on this moment we can infer type only for identifiers. But consider one day
```
var foo = { bar : []};
foo.bar['lala'] = 5; // you will call "toString" for "foo.bar" which is not implemented natively
```

vilchik-elena:
@hbensalem Hi!

Ruling integration test fails because it uses plugin .jar (which wasn't built due to the first exception).
But failure of plugin build is very weird. It's all fine on my computer. Could you double check that nothing is changed locally?

hbensalem:
@vilchik-elena Hi!
Sorry for the sloppy mistakes, this is my first attempt atcontributing. I have fixed the issues you have raised but I still have those 2 test failures when building the plugin. Even when i cloned the official repo I have the same failures coming from "javascript-frontend". If you have an idea on how to solve this issue, it would be really helpful independently of this pull request.

vilchik-elena:
@hbensalem Hi!
1. To run ruling you can build project without running tests 
`mvn clean install -DskipTests`
2. Please try this to fix test (if you on Windows):
Create file in the root named ".gitattributes". It should contain:
```
/javascript-frontend/src/test/resources/highlighter/* text eol=lf
```
And follow https://help.github.com/articles/dealing-with-line-endings/#refreshing-a-repository-after-changing-line-endings
Looks like test fails because you are on Windows where each line ending is 2 symbols.

javascript-checks/src/main/java/org/sonar/javascript/checks/AssociativeArraysCheck.java
```diff
@@ -52,9 +53,11 @@ public void visitAssignmentExpression(AssignmentExpressionTree tree) {
       if (arrayObject.object().types().containsOnly(Kind.ARRAY)) {
         ExpressionTree arrayIndex = ((BracketMemberExpressionTreeImpl) tree.variable()).property();
```
vilchik-elena: @hbensalem Try to not use `...Impl` when it's not required

javascript-checks/src/main/java/org/sonar/javascript/checks/AssociativeArraysCheck.java
```diff
@@ -52,9 +53,11 @@ public void visitAssignmentExpression(AssignmentExpressionTree tree) {
...
         if (arrayIndex.is(Tree.Kind.STRING_LITERAL)) {
-          addIssue(tree, String.format(MESSAGE, arrayObject.object().toString()));
+          addIssue(arrayIndex, String.format(MESSAGE, CheckUtils.asString(arrayObject.object())));
         } else if (arrayIndex instanceof IdentifierTree) {
-          addIssue(tree, String.format(MESSAGE, arrayObject.object().toString()));
+          if (arrayIndex.types().size() > 0 && !arrayIndex.types().contains(Kind.NUMBER)) {
```
vilchik-elena: @hbensalem I would prefer rule to be more strict and raise issue only when we are sure that index is string (see message text to make sure).
As string literal also has string type you can improve your code following way:
```
...
if (arrayObject.object().types().containsOnly(Kind.ARRAY)) { // btw "arrayObject" is not clear name, WDYT of "memberExpression"?
  ExpressionTree arrayIndex = arrayObject.property();
  if (arrayIndex.types().containsOnly(Kind.STRING)) {
    addIssue(...);
  } 
}
```

hbensalem:
@vilchik-elena Hi,
You were right, it was a problem with line endings.Thank you for the solution :)
However, ruling integration test is fine on my computer but it's failing here. I'd really appreciate your feedback!

javascript-frontend/src/main/java/org/sonar/javascript/highlighter/HighlightSymbolTableBuilder.java
```diff
@@ -71,8 +71,7 @@ private static void highlightSymbol(SymbolTableBuilder builder, Symbol symbol) {
       for (int i = 1; i < usagesList.size(); i++) {
         builder.newReference(
           reference,
-          getToken(usagesList.get(i).identifierTree()).startIndex()
-        );
+          getToken(usagesList.get(i).identifierTree()).startIndex());
```
vilchik-elena: @hbensalem Could you revert this change at it doesn't belong to the rule

vilchik-elena:
@hbensalem Ah! again the same - line endings
You see you have 1 new issue on another rule checking Line length, because on your machine each line 1 symbol longer than on ours (and on travis too, it uses linux).
So I would suggest you 
1) remove change to `javascript-LineLength.json` (which causes ruling failure on Travis)
2) remove `.gitattributes` (we will add it later with all understanding what it should contain)
3) rebase on master and fix conflicts

P.S. Your commit messages are awful! And they don't even correspond to the content of commit (see https://github.com/SonarSource/sonar-developer-toolset#commit-messages)

P.P.S. Sorry for all this mess with line endings, we should have configured that before on repo level, but nobody works on Windows so we are facing it for the first time.

hbensalem:
@vilchik-elena Thank you for all your feedback. This mess was a good learning opportunity :)

PR #217, opened 2016-04-04
SONARJS-665 Support exponentiation operator

vilchik-elena:


pynicolas:
@vilchik-elena Where do we test associativity? Exponentiation seems to be right-associative: https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/Operator_Precedence

vilchik-elena:
@pynicolas we test it in model test (we are taking operand and assert it). But indeed it should be right! thanks!
Fixed in 3e08507

PR #223, opened 2016-04-07
SONARJS-672 Tune rule StringConcatenationCheck (S3512)

vilchik-elena:


pynicolas:
@vilchik-elena It seems that we have false negatives because `visitBinaryExpression` does not call its `super` implementation.

vilchik-elena:
@pynicolas No, we shouldn't. We visit operands in `checkBinaryExpression`

pynicolas:
@vilchik-elena You're right. Sorry for the noise.

javascript-checks/src/main/java/org/sonar/javascript/checks/StringConcatenationCheck.java
```diff
@@ -62,28 +62,50 @@ public void visitBinaryExpression(BinaryExpressionTree tree) {
...
-        }
-      }
+    int operandsNum = operandList.size();
+    int literalsNum = stringLiteralsNumber(operandList);
+
+    if (operandsNum > 2 && literalsNum > 0) {
+      int shortLiteralsNum = smallLiteralsNumber(operandList);
 
-      return literalsNum > 0 && shortLiteralsNum <= literalsNum / 2.;
+      boolean isOneLiteralTreeOperandsCase = operandsNum == 3 && literalsNum == 1;
```
pynicolas: @vilchik-elena Typo? Shouldn't we have a "h" as in "Three"?

vilchik-elena: @pynicolas Thanks!:)

javascript-checks/src/main/java/org/sonar/javascript/checks/StringConcatenationCheck.java
```diff
@@ -62,28 +62,50 @@ public void visitBinaryExpression(BinaryExpressionTree tree) {
...
+      boolean atLeastHalfLongLiterals = shortLiteralsNum <= literalsNum / 2.;
+      boolean notTooManyNonLiterals = literalsNum - (operandsNum - literalsNum) >= -1;
```
pynicolas: @vilchik-elena Is it equivalent to `literalsNum > nonLiteralsNum`? That would be clearer I think.

javascript-checks/src/main/java/org/sonar/javascript/checks/StringConcatenationCheck.java
vilchik-elena: @pynicolas fixed 53df9786c8f7d4ec1c31ecf5923a0c9b60933ed1

javascript-checks/src/main/java/org/sonar/javascript/checks/StringConcatenationCheck.java
vilchik-elena: @pynicolas No, it's not equivalent. It's the same is `literalsNum > nonLiteralsNum - 2`. Is it clearer?:)

pynicolas: @vilchik-elena A bit.

PR #228, opened 2016-04-08
SONARJS-690 Exception on object binding pattern with both variable name and initializer

vilchik-elena:


javascript-frontend/src/main/java/org/sonar/javascript/tree/impl/declaration/ArrayBindingPatternTreeImpl.java
```diff
@@ -115,12 +115,13 @@ public void accept(DoubleDispatchVisitor visitor) {
    *   [a, , , b, ...c] // will return [a, b, c]
    * </pre>
    */
+  @Override
   public List<IdentifierTree> bindingIdentifiers() {
     List<IdentifierTree> bindingIdentifiers = Lists.newArrayList();
 
     for (Optional<BindingElementTree> element : elements) {
       if (element.isPresent()) {
-        bindingIdentifiers.addAll(bindingIdentifiers(element.get()));
```
pynicolas: @vilchik-elena Do we still need `bindingIdentifiers` defined at line 91?

vilchik-elena: @pynicolas thanks! fixed in f957c44d04794bf14eb725ba18536c66b208bcd2

PR #233, opened 2016-04-11
SONARJS-662 Remove useless dependencies in JAR

pynicolas:


sonar-javascript-plugin/pom.xml
```diff
@@ -87,8 +87,8 @@
             <configuration>
               <rules>
                 <requireFilesSize>
-                  <maxsize>8000000</maxsize>
-                  <minsize>720000</minsize>
+                  <maxsize>1400000</maxsize>
```
vilchik-elena: @pynicolas wow!:)

vilchik-elena:
@pynicolas do we need `cglib-nodep-3.1.jar`?

pynicolas:
@vilchik-elena You can try :) This is required by [SSLR `ActionParser`](https://github.com/SonarSource/sslr/blob/master/sslr-core/src/main/java/com/sonar/sslr/api/typed/ActionParser.java)

PR #235, opened 2016-04-18
Prevent IndexOutOfBoundsException on analysis

rhencke:
If a function is called on an Angular module, and the last parameter of
that function is an empty array, analysis previously failed with an
IndexOutOfBoundsException.


pynicolas:
Thanks a lot for the pull request! I created [a ticket](https://jira.sonarsource.com/browse/SONARJS-697) to track this issue.

rhencke:
Great - thank you! :)

PR #238, opened 2016-05-19
SONARJS-693 FP EqualInForLoopTerminationCheck: complex condition

vilchik-elena:


javascript-frontend/src/main/java/org/sonar/javascript/tree/TreeKinds.java
```diff
@@ -48,4 +55,8 @@ public static boolean isAssignment(Tree tree) {
     return tree.is(ASSIGNMENT_KINDS);
   }
 
+  public static boolean isIncDec(Tree tree) {
```
pynicolas: @vilchik-elena Naming: I would prefer something more explicit even if it's longer, e.g. `isIncrementOrDecrement`.

PR #240, opened 2016-05-19
SONARJS-692 FP BooleanEqualityComparisonCheck: not condition expression

vilchik-elena:


javascript-checks/src/main/java/org/sonar/javascript/checks/BooleanEqualityComparisonCheck.java
```diff
@@ -66,22 +57,14 @@ public void visitUnaryExpression(UnaryExpressionTree tree) {
 
   @Override
   public void visitBinaryExpression(BinaryExpressionTree tree) {
-    if (tree.is(BINARY_OPERATORS)) {
+    if (tree.is(Kind.EQUAL_TO, Kind.NOT_EQUAL_TO)) {
```
pynicolas: @vilchik-elena What about strict equal?

vilchik-elena: @pynicolas We don't have them here for a reason (there are even unit tests for that). We can't simplify strict equality.

PR #241, opened 2016-05-19
SONARJS-685 False Positive UnchangedLetVariable: for-in/of loop

vilchik-elena:


vilchik-elena:
Ticket is closed as "Not a Bug"

PR #254, opened 2016-05-30
SONARJS-714 Improve issue location for rules catching useless code

pynicolas:


javascript-checks/src/test/resources/checks/UselessExpressionStatement.js
```diff
@@ -22,8 +23,8 @@ var a = x == y;
 foo(x == y)
 
 // nok
-x == y    // NOK
-x == y;    // NOK
+x == y    // Noncompliant
```
vilchik-elena: @pynicolas we still have line issue here

PR #255, opened 2016-05-31
SONARJS-712 Improve issue location for statement-related rules

pynicolas:


javascript-checks/src/main/java/org/sonar/javascript/checks/NonCaseLabelInSwitchCheck.java
```diff
@@ -73,7 +73,7 @@ public void visitNode(Tree tree) {
     } else if (tree.is(Kind.LABELLED_STATEMENT)) {
 
       if (inCase()) {
-        addLineIssue(tree,
+        addIssue(((LabelledStatementTree) tree).label(),
```
vilchik-elena: @pynicolas It would be nice to extract `((LabelledStatementTree) tree).label()`

pynicolas: @vilchik-elena Of course! Fixed in 69452cc.

javascript-checks/src/test/resources/checks/returnOfBooleanExpression.js
```diff
@@ -1,18 +1,15 @@
...
     return false;
   }
 
-  if (something) {   // NOK
-    return false;
-  } else {
-    return true;
-  }
+  if (something) { return false; } else { return true; } // Noncompliant
+//^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
vilchik-elena: @pynicolas I would prefer to highlight only `if` (and Java does it https://nemo.sonarqube.org/issues/search#issues=f2a56d69-f00f-42e9-ac15-52b9a4164b92)

pynicolas: @vilchik-elena I highlighted the whole statement because the `return` statements inside are part of the problem. However, let's be consistent with java: changed in 58e001a.

PR #259, opened 2016-06-01
SONARJS-674 Improve issue location for rules with unwanted usage

vilchik-elena:


javascript-checks/src/main/java/org/sonar/javascript/checks/ConsoleLoggingCheck.java
```diff
@@ -59,7 +59,7 @@ public void visitNode(Tree tree) {
       DotMemberExpressionTree callee = (DotMemberExpressionTree) callExpression.callee();
 
       if (isCalleeConsoleLogging(callee)) {
-        addLineIssue(tree, MESSAGE);
+        addIssue(((CallExpressionTree) tree).callee(), MESSAGE);
```
pynicolas: @vilchik-elena why not re-use callExpression defined at line 56? In fact, we could probably change this check to extend `DoubleDispatchVisitorCheck`.

vilchik-elena: @pynicolas thanks! 10ec451d1e399d19e2dacf7e7ef3378584527662

PR #272, opened 2016-06-07
SONARJS-707 Use new CPD API

vilchik-elena:


javascript-frontend/src/main/java/org/sonar/javascript/tree/impl/lexical/InternalSyntaxTrivia.java
```diff
@@ -33,11 +33,35 @@
...
+
 
   public InternalSyntaxTrivia(String comment, int startLine, int column) {
     this.comment = comment;
     this.startLine = startLine;
     this.column = column;
+    calculateEndOffsets();
+  }
+
+  private void calculateEndOffsets() {
```
pynicolas: @vilchik-elena Could we avoid this duplication with `InternalSyntaxToken`?

vilchik-elena: @pynicolas At least I didn't find good solution to do that

PR #278, opened 2016-06-10
SONARJS-678 Extend DoubleDispatchVisitor: visitToken and visitComment

vilchik-elena:


javascript-frontend/src/test/java/org/sonar/plugins/javascript/api/visitors/DoubleDispatchVisitorTest.java
```diff
@@ -25,44 +25,68 @@
...
+import org.sonar.plugins.javascript.api.tree.lexical.SyntaxToken;
+import org.sonar.plugins.javascript.api.tree.lexical.SyntaxTrivia;
 
 import static org.fest.assertions.Assertions.assertThat;
 
 public class DoubleDispatchVisitorTest {
 
   @Test
+  public void test_visit_token() throws Exception {
+    assertNumberOfVisitedTokens("[1, 2, , ,]", 9);
```
pynicolas: @vilchik-elena I see only 8 tokens: what's the trick?

javascript-frontend/src/main/java/org/sonar/plugins/javascript/api/visitors/DoubleDispatchVisitor.java
```diff
@@ -125,439 +128,399 @@ protected void scan(@Nullable Tree tree) {
     }
   }
 
-  protected <T> void scan(List<T> trees) {
-    for (T tree : trees) {
+  protected void scanTree(Tree tree) {
```
pynicolas: @vilchik-elena Would `scanChildren` be a better name?

javascript-frontend/src/test/java/org/sonar/plugins/javascript/api/visitors/DoubleDispatchVisitorTest.java
vilchik-elena: @pynicolas yes :)

javascript-frontend/src/main/java/org/sonar/plugins/javascript/api/visitors/DoubleDispatchVisitor.java
vilchik-elena: @pynicolas indeed. Fixed in 0319ebf51c4aef3d9f71d29ff92779d3ffeea59c

PR #281, opened 2016-06-14
SONARJS-727 Symbolic execution: improve support for boolean expressions

pynicolas:


javascript-frontend/src/main/java/org/sonar/javascript/se/SymbolicExecution.java
```diff
@@ -288,15 +290,18 @@ private void pushConditionSuccessors(CfgBranchingBlock block, ProgramState curre
     }
   }
 
-  private ProgramState store(ProgramState currentState, Tree left, ExpressionTree right) {
-    Constraint constraint = Constraint.get(right);
-    return storeConstraint(currentState, left, constraint);
+  private ProgramState storeConstraint(ProgramState currentState, Tree left) {
```
vilchik-elena: @pynicolas I think it's worth to rename this method after removal of last parameter:)

javascript-frontend/src/test/resources/se/assignment.js
```diff
@@ -2,11 +2,15 @@ function main() {
 
   var x, y, z, untracked;
 
+  x = 0;  // PS x=FALSY
   x += 1; // PS x=UNKNOWN
   y = 5;  // PS y=TRUTHY
+  z = 0;  // PS z=FALSY
```
vilchik-elena: @pynicolas Why this line?

PR #287, opened 2016-06-20
ProgramState.equal() should take into account the constraint on the peek of the stack

pynicolas:


javascript-frontend/src/test/resources/se/conditions.js
```diff
@@ -96,4 +96,13 @@ function main() {
   }
 
   makeLive(y1, x);
+
+  x = foo();
+  var y = x && x.bar;
```
vilchik-elena: @pynicolas do we really need this line?

pynicolas: @vilchik-elena Without this line, the test is green without any change. I'm not sure how to improve this test case.

vilchik-elena:
LGTM

PR #304, opened 2016-07-26
SONARJS-743 Rule: Functions should not be called both with and without "new"

vilchik-elena:


javascript-checks/src/main/java/org/sonar/javascript/checks/InconsistentFunctionCallCheck.java
```diff
@@ -0,0 +1,134 @@
...
+    super.visitCallExpression(tree);
+  }
+
+  @Override
+  public void visitNewExpression(NewExpressionTree tree) {
+    visit(tree, tree.expression(), usedInCallExpression, usedInNewExpression, "out");
+    super.visitNewExpression(tree);
+  }
+
+  private <T1 extends Tree, T2 extends Tree> void visit(
```
ivandalbosco: Smart factorization! Small detail: I would not call it "visit", because we are now out of the visitor pattern. Call it "check"? Other detail: "withTail" is probably the naming of a first version of your code, where the arg was a boolean. Rename it into "tail"?

javascript-checks/src/main/java/org/sonar/javascript/checks/InconsistentFunctionCallCheck.java
```diff
@@ -0,0 +1,134 @@
...
+@ActivatedByDefault
+@SqaleConstantRemediation("15min")
+public class InconsistentFunctionCallCheck extends DoubleDispatchVisitorCheck {
+
+  private static final String MESSAGE = "Correct the use of this function; on line %s it was called with%s \"new\".";
+  private static final String SECONDARY_MESSAGE = "Called with%s \"new\"";
+
+  private Set<Symbol> hasIssue = new HashSet<>();
+
+  private ListMultimap<Symbol, CallExpressionTree> usedInCallExpression = LinkedListMultimap.create();
```
ivandalbosco: Looks like the ListMultimap is little of overshoot, as you only care of the last insertion. Why not using a HashMap and simplify a little method visit?

javascript-checks/src/test/resources/checks/InconsistentFunctionCallCheck.js
```diff
@@ -0,0 +1,22 @@
+function foo(){
+}
+
+var x = external();
+
+x();
```
ivandalbosco: How about duplicating this line, so that we check that "secondary" is -1, not -2?

javascript-checks/src/test/resources/checks/InconsistentFunctionCallCheck.js
```diff
@@ -0,0 +1,22 @@
...
+var xx = new x(); // Noncompliant [[secondary=-1]] {{Correct the use of this function; on line 6 it was called without "new".}}
+//       ^^^^^
+
```
ivandalbosco: I guess that adding again a line with 
x();
would be nice. Depending on how the rule is implemented, it might raise a spurious issue.

javascript-checks/src/test/resources/checks/InconsistentFunctionCallCheck.js
```diff
@@ -0,0 +1,22 @@
...
+function MyObj() {
+}
+
+var obj = new MyObj();
+  MyObj(); // Noncompliant [[secondary=-1]] {{Correct the use of this function; on line 17 it was called with "new".}}
+//^^^^^
+
+obj = new MyObj();
+MyObj();
```
ivandalbosco: Maybe you could add a test with a nested function, for example:
function bar() {
  function bar() {}
  var a = new bar();
}
var b = bar();  // OK

PR #306, opened 2016-07-27
SONARJS-729 Do not include complexity of nested functions into nestin…

ivandalbosco:
…g one complexity


ivandalbosco:
Close PR because only one check (travis-ci/push) was run.

PR #312, opened 2016-08-02
SONARJS-735 Add test verifying integration with SonarLint

ivandalbosco:


its/plugin/tests/pom.xml
```diff
@@ -27,17 +28,30 @@
...
+      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>org.easytesting</groupId>
       <artifactId>fest-assert</artifactId>
+      <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
```
pynicolas: @ivandalbosco Why not set the scope to "test" as you did for other dependencies?

pynicolas: Fixed in d04598ee7c45e00d6e50c38aff4a2415be554a5a.

PR #313, opened 2016-08-03
SONARJS-760 Support "new.target"

ivandalbosco:


javascript-frontend/src/main/java/org/sonar/plugins/javascript/api/tree/expression/NewTargetTree.java
```diff
@@ -0,0 +1,40 @@
...
+package org.sonar.plugins.javascript.api.tree.expression;
+
+import org.sonar.plugins.javascript.api.tree.lexical.SyntaxToken;
+
+/**
+ * <a href="http://www.ecma-international.org/ecma-262/6.0/#sec-left-hand-side-expressions">ECMA 6 New target</a>.
+ * <pre>
+ *   new . target
+ * </pre>
+ * Note: this tree handles only the <code>new.target</code> construct. It does not handle constructs such
```
pynicolas: @ivandalbosco This note seems confusing to me since `new.a` is not valid JavaScript. I would remove it.

javascript-frontend/src/test/java/org/sonar/javascript/parser/expressions/MemberExpressionTest.java
```diff
@@ -48,4 +48,13 @@ public void ok() {
       .matches("new super ( arguments ) ``");
   }
 
+  @Test
+  public void new_target() {
+    assertThat(JavaScriptLegacyGrammar.MEMBER_EXPRESSION)
+    .matches("new . target")
+    .notMatches("new . Target")
+    .notMatches("a = target")
```
pynicolas: @ivandalbosco I'm not sure of the value of this test.

javascript-frontend/src/main/java/org/sonar/plugins/javascript/api/tree/expression/NewTargetTree.java
ivandalbosco: @pynicolas Your're right.

PR #325, opened 2016-08-26
SONARJS-769 Synchonize metadata of rules with RSPECs

vilchik-elena:


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/rules/JavaScriptRulesDefinition.java
```diff
@@ -19,21 +19,100 @@
...
+      addMetadata(rule, metadataKey);
+    }
+
     repository.done();
   }
 
+  @Nullable
+  private static String readRuleDefinitionResource(String fileName) {
+    URL resource = JavaScriptRulesDefinition.class.getResource("/org/sonar/l10n/javascript/rules/javascript/" + fileName);
+    if (resource == null) {
```
ivandalbosco: @vilchik-elena Can " resource" really be null? If it is null, should not we raise an exception?

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/JavaScriptProfileTest.java
```diff
@@ -59,4 +76,19 @@ public Rule answer(InvocationOnMock invocation) {
     }).getMock();
   }
 
+  /**
+   * SonarLint will inject a rule finder containing only the rules coming from the abap repository
```
ivandalbosco: @vilchik-elena Can you please provide more information? It is not clear to me what ABAP has to see with the JS plugin.

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/rules/JavaScriptRulesDefinition.java
vilchik-elena: @ivandalbosco call `repository.done();` on line 57 will raise an issue if some of information is not provided, so no need for additional actions on our side

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/rules/JavaScriptRulesDefinitionTest.java
```diff
@@ -40,10 +43,20 @@ public void test() {
...
+  }
+
+  private void assertRuleProperties(Repository repository) {
+    Rule rule = repository.rule("ArrayAndObjectConstructors");
+    assertThat(rule).isNotNull();
+    assertThat(rule.name()).isEqualTo("Array and Object constructors should not be used");
+    assertThat(rule.debtRemediationFunction().type()).isEqualTo(Type.CONSTANT_ISSUE);
+    assertThat(rule.type()).isEqualTo(RuleType.BUG);
+    assertThat(repository.rule("CommentRegularExpression").template()).isTrue();
+  }
```
ivandalbosco: @vilchik-elena It would be nice to check every attribute of the param (picking a rule which has a param, such as S138). Typically param.type (a RuleParamType) may be incorrectly handled.

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/JavaScriptProfileTest.java
vilchik-elena: @ivandalbosco blind copy-paste. Should be javascript instead. fixed in 1757d594028f3645cf33767c0315e2485d967f3f

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/rules/JavaScriptRulesDefinitionTest.java
vilchik-elena: @ivandalbosco done bde1961986c48356cd4004b7faf3296620f5bd0a

javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S1697.html
```diff
@@ -36,5 +36,5 @@
 if (str == undefined || str.length &gt; 0) {
   console.log("String is not empty");
 }
-</pre>
+</pre> h2. Deprecated This rule is deprecated; use S2259 instead.
```
vilchik-elena: will be fixed

PR #331, opened 2016-09-01
=== and !== should not allow all paths when operands have incompatible constraints

pynicolas:


javascript-frontend/src/main/java/org/sonar/javascript/se/sv/RelationalSymbolicValue.java
```diff
@@ -60,7 +60,7 @@ private static SymbolicValue createEquality(Kind kind, SymbolicValue leftOperand
   @Override
   public List<ProgramState> constrain(ProgramState state, Constraint constraint) {
     ProgramState newState = state.constrainOwnSV(this, constraint);
-    if (newState == null) {
+    if (newState == null || state.getConstraint(this).isIncompatibleWith(constraint)) {
```
vilchik-elena: @pynicolas can we replace `state.getConstraint(this)` with `this.constraint(state)`?

PR #335, opened 2016-09-02
SONARJS-430 Extend FunctionNameCheck (S100): function expression and object property

vilchik-elena:


javascript-checks/src/main/java/org/sonar/javascript/checks/FunctionNameCheck.java
```diff
@@ -60,14 +62,39 @@ public void visitMethodDeclaration(MethodDeclarationTree tree) {
   }
 
   @Override
+  public void visitPairProperty(PairPropertyTree tree) {
+    if (isFunctionExpression(tree.value())) {
+      checkName(tree.key());
+    }
+    super.visitPairProperty(tree);
+  }
+
```
ivandalbosco: @vilchik-elena In visitMethodDeclaration, why do ignore get and set methods?

vilchik-elena: @ivandalbosco get and set names is in fact properties names (aka fields), so we don't want to check them in this rule (https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Functions/get)

PR #343, opened 2016-09-20
SONARJS-730 Extend the list of valid types for "typeof" expression

ivandalbosco:


javascript-frontend/src/main/java/org/sonar/javascript/se/Constraint.java
```diff
@@ -200,6 +200,10 @@ public boolean isIncompatibleWith(Constraint other) {
     return and(other).equals(NO_POSSIBLE_VALUE);
   }
 
+  public boolean isTrivial() {
```
vilchik-elena: @ivandalbosco Why add this method?

javascript-frontend/src/main/java/org/sonar/javascript/se/sv/TypeOfComparisonSymbolicValue.java
```diff
@@ -48,8 +49,18 @@
     .put("string", Constraint.STRING)
     .put("boolean", Constraint.BOOLEAN)
     .put("symbol", Constraint.OTHER_OBJECT)
+    // on IE, operator typeof returns either one of the standard values above, or "date", or "unknown"
+    .put("date", Constraint.ANY_VALUE)
```
vilchik-elena: @ivandalbosco why not narrow constraint here? E.g. OTHER_OBJECT

javascript-frontend/src/main/java/org/sonar/javascript/se/sv/TypeOfComparisonSymbolicValue.java
```diff
@@ -48,8 +49,18 @@
...
+    .put("unknown", Constraint.ANY_VALUE)
```
vilchik-elena: @ivandalbosco And here as well, i think it should be OTHER_OBJECT

javascript-frontend/src/main/java/org/sonar/javascript/tree/impl/expression/BinaryExpressionTreeImpl.java
```diff
@@ -88,4 +88,18 @@ public TypeSet types() {
   public void add(Type type) {
     types.add(type);
   }
+
+  @Override
+  public String toString() {
```
vilchik-elena: @ivandalbosco 
- this commit has nothing to do with ticket SONARJS-730, so you can just omit ticket in commit message
- i would again insist on removing these methods. Even if we put aside the motivation (which is true only for you) what is the logic behind this implementation? Why just these 2 trees has toString? Why this methods implementations look different? etc
 
Let's schedule the meeting for 3 of us to come to an aligned decision on this matter and to not come back to it again and again:)

javascript-frontend/src/main/java/org/sonar/javascript/se/Constraint.java
ivandalbosco: @vilchik-elena Not the most spectacular method ever, I admit. I wrote it order to define a behaviour of the constraints in class Constraint, not scattered in all client classes. I am not sure that every client class should define different behaviours according to the constraint at hand (ANY_VALUE, TRUE, BOOLEAN, etc.). Typically, enum classes (Constant could have been an enum) define their own behaviour. In my opinion, the fact that a method is used only once is not a criterion to get rid of it.

javascript-frontend/src/main/java/org/sonar/javascript/se/sv/TypeOfComparisonSymbolicValue.java
ivandalbosco: @vilchik-elena Good catch. Gonna try

javascript-frontend/src/main/java/org/sonar/javascript/se/sv/TypeOfComparisonSymbolicValue.java
ivandalbosco: Idem to above.

javascript-frontend/src/main/java/org/sonar/javascript/tree/impl/expression/BinaryExpressionTreeImpl.java
vilchik-elena: @ivandalbosco ping

PR #345, opened 2016-09-21
Refactor checks based on symbolic execution which check all paths

vilchik-elena:


vilchik-elena:
@ivandalbosco Could you review this PR?

javascript-checks/src/main/java/org/sonar/javascript/checks/AbstractAllPathSeCheck.java
```diff
@@ -0,0 +1,72 @@
...
+
+  @Override
+  public void beforeBlockElement(ProgramState currentState, Tree element) {
+    T tree = getTree(element);
+
+    if (tree != null) {
+      boolean isProblem = isProblem(tree, currentState);
+      if (!isProblem) {
+        problemMap.put(tree, true);
+      } else if (!problemMap.containsKey(tree)) {
```
ivandalbosco: @vilchik-elena It looks like problemMap is actually a noProblemMap. Suggestion: invert the meaning of the Boolean: true for problem, false for no problem.

javascript-checks/src/main/java/org/sonar/javascript/checks/AbstractAllPathSeCheck.java
```diff
@@ -0,0 +1,72 @@
...
+
+  @Override
+  public void startOfExecution(Scope functionScope) {
+    problemMap.clear();
+  }
+
+  @Override
+  public void endOfExecution(Scope functionScope) {
+    for (Entry<T, Boolean> entry : problemMap.entrySet()) {
+      if (!entry.getValue()) {
```
ivandalbosco: @vilchik-elena See my previous comment. It would be more natural to have if(entry.getValue()) here, without the "!"

javascript-checks/src/main/java/org/sonar/javascript/checks/AbstractAllPathSeCheck.java
```diff
@@ -0,0 +1,72 @@
...
+        raiseIssue(entry.getKey());
+      }
+    }
+  }
+
+  abstract void raiseIssue(T tree);
+
+}
```
ivandalbosco: I like the idea of factorizing all-paths behaviors into this superclass AbstractAllPathSeCheck. However I have a little concern over this design: nothing prevents subclasses to redefine the methods defined in class SeCheck. For example, class UnaryPlusMinusWithObjectCheck might redefine afterBlockElement; in that method it could create issues, using SeCheck#addIssue. In other words, class AbstractAllPathSeCheck is meant to be constraining, but it is easily circumvented by its subclasses.

So how about the following: in AbstractAllPathSeCheck,
- implement all methods (including addIssue) from SeCheck and make them final
- replace method "void raiseIssue(T tree)" with "Issue createIssue(T tree)". The call to SeCheck#addIssue will be made in AbstractAllPathSeCheck#endOfExecution

ivandalbosco:
@vilchik-elena Class AbstractAllPathCheck deserves to have its own JUnit test class, independently of the JUnit test classes of its subclasses. The logic in method beforeBlockElement is not trivial.

javascript-checks/src/main/java/org/sonar/javascript/checks/AbstractAllPathSeCheck.java
vilchik-elena: @ivandalbosco indeed:) thanks!

javascript-checks/src/main/java/org/sonar/javascript/checks/AbstractAllPathSeCheck.java
vilchik-elena: @ivandalbosco agree

javascript-checks/src/main/java/org/sonar/javascript/checks/AbstractAllPathSeCheck.java
vilchik-elena: @ivandalbosco I don't think it's required to limit overriding checks. They can create additional issues if they want. I would it's even unwanted behaviour to limit them

vilchik-elena:
@ivandalbosco Don't you think that unit test for 4 existing overriding classes is sufficient? It's quite tested already. And having dedicated test will require one more overriding class.

PR #347, opened 2016-09-23
SONARJS-731 Symbolic execution: false positive in array/object assignment pattern

ivandalbosco:


javascript-frontend/src/main/java/org/sonar/javascript/tree/impl/SeparateListUtils.java
```diff
@@ -0,0 +1,44 @@
...
+
+public class SeparateListUtils {
+  
+  private SeparateListUtils() {
+  }
+
+  /**
+   * Returns a new list containing the present (in the sense of Optional#isPresent) elements in <code>list</code>.
+   */
+  public static List<Tree> presentsOf(List<com.sonar.sslr.api.typed.Optional<Tree> > list) {
```
vilchik-elena: @ivandalbosco What about making it a `static` method of class `SeparatedList`?

javascript-frontend/src/main/java/org/sonar/javascript/tree/impl/SeparateListUtils.java
```diff
@@ -0,0 +1,44 @@
...
+    List<Tree> newList = new LinkedList<>();
+    for (com.sonar.sslr.api.typed.Optional<Tree> element : list) {
```
vilchik-elena: @ivandalbosco why full path to class?

javascript-frontend/src/test/java/org/sonar/javascript/cfg/ControlFlowGraphTest.java
```diff
@@ -564,6 +564,24 @@ public void assignment_with_or() throws Exception {
   }
 
   @Test
+  public void assignment_with_array_assignment_pattern_with_x() throws Exception {
```
vilchik-elena: @ivandalbosco Can we add test for empty element?

javascript-frontend/src/test/java/org/sonar/javascript/cfg/ControlFlowGraphTest.java
```diff
@@ -582,6 +582,18 @@ public void assignment_with_nested_array_assignment_pattern() throws Exception {
   }
 
   @Test
+  public void assignment_with_object_assignment_pattern_with_one_element() throws Exception {
+    ControlFlowGraph g = build("({p} = {p:10})", 1);
```
vilchik-elena: @ivandalbosco imho it's better to replace object literal in the lhs of assignment in these tests with just identifier variable, i.e. `{p} = obj`. We don't need to test cfg for object literal here, so it just musses the test

javascript-frontend/src/test/java/org/sonar/javascript/cfg/ControlFlowGraphTest.java
```diff
@@ -582,6 +582,18 @@ public void assignment_with_nested_array_assignment_pattern() throws Exception {
   }
 
   @Test
+  public void assignment_with_object_assignment_pattern_with_one_element() throws Exception {
```
vilchik-elena: @ivandalbosco I think we should provide more tests here:  e.g. `{p:p1} = obj`, `{p:p1[1] = 42}`

javascript-frontend/src/main/java/org/sonar/javascript/tree/impl/SeparateListUtils.java
ivandalbosco: @vilchik-elena Because this class Optional has the same name and almost the same methods as the new JDK 8 class Optional. By giving its long name, we provide a signal to the developer to be careful.

javascript-frontend/src/main/java/org/sonar/javascript/tree/impl/SeparateListUtils.java
ivandalbosco: @vilchik-elena Honestly I hesitated on this issue. As SeparateList is a generic class and this method use T = Tree, I chose to use a separate class.

javascript-frontend/src/test/java/org/sonar/javascript/cfg/ControlFlowGraphTest.java
ivandalbosco: @vilchik-elena Yes, thanks!

javascript-frontend/src/test/java/org/sonar/javascript/cfg/ControlFlowGraphTest.java
ivandalbosco: @vilchik-elena OK

javascript-frontend/src/main/java/org/sonar/javascript/se/ExpressionStack.java
```diff
@@ -62,10 +62,6 @@
 
   private final Deque<SymbolicValue> stack;
 
-  public static ExpressionStack emptyStack() {
```
vilchik-elena: @ivandalbosco can we revert the change for this file?

javascript-frontend/src/main/java/org/sonar/javascript/tree/impl/statement/VariableDeclarationTreeImpl.java
```diff
@@ -44,11 +44,9 @@
   private final SeparatedList<BindingElementTree> variables;
 
   public VariableDeclarationTreeImpl(Kind kind, InternalSyntaxToken token, SeparatedList<BindingElementTree> variables) {
-
```
vilchik-elena: @ivandalbosco Many files like this one contains just empty lines changes. Let's remove them from this PR

javascript-frontend/src/main/java/org/sonar/javascript/tree/impl/JavaScriptTree.java
```diff
@@ -95,7 +95,13 @@ public Tree getParent() {
 
   @Override
   public String toString() {
-    return getClass().getSimpleName();
+    StringBuilder sb = new StringBuilder();
+    Iterator<Tree> children = childrenIterator();
+    while (children.hasNext()) {
+      sb.append(children.next());
```
vilchik-elena: @ivandalbosco Don't you think we should skip children equal `null`?

Can we add unit test?

javascript-frontend/src/main/java/org/sonar/javascript/tree/impl/statement/VariableDeclarationTreeImpl.java
ivandalbosco: @vilchik-elena Done

javascript-frontend/src/main/java/org/sonar/javascript/tree/impl/JavaScriptTree.java
ivandalbosco: @vilchik-elena I found no such case when debugging. Should it come up, I guess the developer will be happy to see it

ivandalbosco: @vilchik-elena Good idea. Class JavaScriptTreeTest now unit tests all public methods (coverage not perfect, though).

PR #349, opened 2016-09-26
SONARJS-768 Rule: The output of functions that don't return anything should not be used

vilchik-elena:


javascript-checks/src/main/java/org/sonar/javascript/checks/UseOfEmptyReturnValueCheck.java
```diff
@@ -0,0 +1,152 @@
...
+
+    if (parent.is(CONDITIONAL_OR, CONDITIONAL_AND)) {
+      return ((BinaryExpressionTree) parent).leftOperand().equals(tree);
+    }
+
+    if (parent.is(CONDITIONAL_EXPRESSION)) {
+      return ((ConditionalExpressionTree) parent).condition().equals(tree);
+    }
+
+    return !parent.is(Kind.EXPRESSION_STATEMENT);
```
pynicolas: @vilchik-elena what if parent is a parenthesised expression?

javascript-checks/src/main/java/org/sonar/javascript/checks/UseOfEmptyReturnValueCheck.java
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  @CheckForNull
+  private static FunctionTree functionTree(CallExpressionTree tree, ProgramState currentState) {
+    if (tree.callee().is(Kind.IDENTIFIER_REFERENCE)) {
+      Symbol symbol = ((IdentifierTree) tree.callee()).symbol();
+      
+      if (symbol != null) {
+        SymbolicValue symbolicValue = currentState.getSymbolicValue(symbol);
```
pynicolas: @vilchik-elena Why not use the expression stack?

javascript-checks/src/main/java/org/sonar/javascript/checks/UseOfEmptyReturnValueCheck.java
```diff
@@ -0,0 +1,152 @@
...
+      }
+    }
+
+    return null;
+  }
+
+  @Override
+  void raiseIssue(CallExpressionTree tree) {
+    String functionName = ((IdentifierTree) tree.callee()).name();
+    addIssue(tree.callee(), String.format(MESSAGE, functionName, functionName));
```
pynicolas: @vilchik-elena Maybe we could add a secondary location on the function declaration (or a list of secondary locations as there may be more several function declarations depending on the execution path...).

javascript-checks/src/main/java/org/sonar/javascript/checks/UseOfEmptyReturnValueCheck.java
vilchik-elena: @pynicolas thanks! fixed

javascript-checks/src/main/java/org/sonar/javascript/checks/UseOfEmptyReturnValueCheck.java
vilchik-elena: @pynicolas fixed

javascript-checks/src/main/java/org/sonar/javascript/checks/UseOfEmptyReturnValueCheck.java
vilchik-elena: @pynicolas This is not easy to implement. We can keep the list of `FunctionTree`, which might be declaration or expression or arrow function. But it's not clear which part to highlight.

PR #352, opened 2016-09-28
SONARJS-732 Symbolic Execution: compound condition symbolic value

vilchik-elena:


javascript-checks/src/test/resources/checks/AlwaysTrueOrFalseCondition.js
```diff
@@ -225,3 +225,11 @@ function tro(x, y) {
   if (true || y) {} // Noncompliant
   if (false || y) {} // Noncompliant
 }
+
+function logical_and(p1, p2, p3) {
```
pynicolas: @vilchik-elena Doesn't this test exactly the same thing as what's added to conditions.js?

PR #354, opened 2016-09-29
SONARJS-738 Improve EqEqEq: ignore when operands of same type

vilchik-elena:


javascript-checks/src/main/java/org/sonar/javascript/checks/EqEqEqCheck.java
```diff
@@ -19,32 +19,98 @@
...
   }
 
+  private static class EqualityVisitor extends SubscriptionVisitor {
+
+    Set<BinaryExpressionTree> equalityExpressions = new HashSet<>();
+
+    @Override
+    public List<Kind> nodesToVisit() {
+      return ImmutableList.of(Kind.EQUAL_TO, Kind.NOT_EQUAL_TO);
+    }
```
ivandalbosco: @vilchik-elena You could factor the list of kinds into a static field Kind[] kinds = {Kind.EQUAL_TO, Kind.NOT_EQUAL_TO} in the enclosing class.

javascript-checks/src/test/resources/checks/eqEqEq.js
```diff
@@ -12,4 +15,36 @@ function sayHello() {
   }
   if (a == null) {  // OK
   }
+
+  foo(c);
+}
+
+function withTypes(a, b, c) {
```
ivandalbosco: @vilchik-elena Fundamental change required here: remove unused parameter "c"

javascript-checks/src/test/resources/checks/eqEqEq.js
```diff
@@ -12,4 +15,36 @@ function sayHello() {
...
+    if (a != b) {     // OK
+    }
+    if (a === b) {    // OK
+    }
+    if (a !== b) {    // OK
+    }
+    if (a != null) {  // OK
+    }
+    if (a == null) {  // OK
+    }
```
ivandalbosco: @vilchik-elena Why not adding symmetric case if (null == a) {}  and silly case if (null == null) {} ?

javascript-checks/src/test/resources/checks/eqEqEq.js
```diff
@@ -12,4 +15,36 @@ function sayHello() {
...
+  }
+}
```
ivandalbosco: @vilchik-elena  It would be nice to add a check where the types are known but they differ. For example:
  if (typeof a == "string" && typeof b == 'number') {
    if (a == b) {     // Noncompliant
    }
  }

ivandalbosco: @vilchik-elena And maybe also a test where one type is known and the other one is not. For example:
var d = foo();
if (typeof a == "string") {
    if (a == d) {     // Noncompliant
    }
}

javascript-checks/src/main/java/org/sonar/javascript/checks/EqEqEqCheck.java
vilchik-elena: @ivandalbosco done 468712af790f5b511026b943df4c722ff5ff6960

javascript-checks/src/test/resources/checks/eqEqEq.js
vilchik-elena: @ivandalbosco I hope it's irony:) Done in 468712af790f5b511026b943df4c722ff5ff6960

javascript-checks/src/test/resources/checks/eqEqEq.js
vilchik-elena: @ivandalbosco done 468712af790f5b511026b943df4c722ff5ff6960

javascript-checks/src/test/resources/checks/eqEqEq.js
vilchik-elena: @ivandalbosco Good idea! done in 468712af790f5b511026b943df4c722ff5ff6960

PR #358, opened 2016-09-30
SONARJS-784 Symbolic execution: track variables from binding pattern

vilchik-elena:


javascript-frontend/src/main/java/org/sonar/javascript/se/SymbolicExecution.java
```diff
@@ -300,7 +319,7 @@ public static boolean isUndefined(IdentifierTree tree) {
     return "undefined".equals(tree.name());
   }
 
-  private ProgramState createSymbolicValuesForTrackedVariables(List<Tree> trees, ProgramState state) {
+  private <T extends Tree> ProgramState createSymbolicValuesForTrackedVariables(List<T> trees, ProgramState state) {
```
pynicolas: @vilchik-elena Why have a parameterized type for this method? Why not: `private ProgramState createSymbolicValuesForTrackedVariables(List<? extends Tree> trees, ProgramState state)`?

vilchik-elena: @pynicolas fixed in d20abfc56e1a3576d43a89f3324784cb072c0be6

PR #360, opened 2016-09-30
SONARJS-776 Remove expression from issue message for ParenthesesCheck

vilchik-elena:


javascript-checks/src/main/java/org/sonar/javascript/checks/ParenthesesCheck.java
```diff
@@ -38,7 +38,7 @@
 @Rule(key = "Parentheses")
 public class ParenthesesCheck extends DoubleDispatchVisitorCheck {
 
-  private static final String MESSAGE = "Remove useless parentheses around \"%s\".";
+  private static final String MESSAGE = "Remove those useless parentheses.";
```
ivandalbosco: @vilchik-elena As you removed the parameter %s from the message, you can remove the two lines
      String expressingString = CheckUtils.asString(parenthesisedExpression.expression());
and simplify the two lines
      addIssue(expression, String.format(MESSAGE, expressingString));

vilchik-elena: @ivandalbosco My bad, fixed in 7b2f3ff3785a117c10318e5504d4f7bcd887f8f6

PR #372, opened 2016-11-01
SONARJS-807 Rule: Empty statements should be removed

ivandalbosco:


javascript-checks/src/test/resources/checks/emptyStatement.js
```diff
@@ -0,0 +1,131 @@
...
+  return 0;;                 // Noncompliant
+//         ^
+} ;                          // Noncompliant
+   ;                         // Noncompliant
+
+// ; this is a comment containing semicolons ;;  // OK
+
+/* ; this is a 
+   multi-line comment ; 
+   containing semicolons ;; */                   // OK
```
pynicolas: @ivandalbosco Is this line really useful? Generally speaking, I feel that this test file contains more cases than what we need.

PR #382, opened 2016-11-02
SONARJS-803 Rule: Comparison expressions should not always evaluate t…

ivandalbosco:
…o false


javascript-checks/src/test/resources/checks/comparisonReturningFalse.js
```diff
@@ -0,0 +1,97 @@
...
+  var date = new Date();
+  var nul = null;
+  var undef;
+  var symb = Symbol();
+
+  // 1. compare an object with the other types
+  obj > 0;                     // Noncompliant {{Change this value so that the comparison does not consistently evaluate to "false".}}
+//^^^
+  obj > "hello";               // Noncompliant
+//^^^
```
vilchik-elena: @ivandalbosco IMO it's enough to test secondary location just once in your case (or twice, where second test is testing 2 issues in one comparison)

javascript-checks/src/test/resources/checks/comparisonReturningFalse.js
```diff
@@ -0,0 +1,97 @@
...
+  var obj2 = new Object();
+  var date = new Date();
+  var nul = null;
+  var undef;
+  var symb = Symbol();
+
+  // 1. compare an object with the other types
+  obj > 0;                     // Noncompliant {{Change this value so that the comparison does not consistently evaluate to "false".}}
+//^^^
+  obj > "hello";               // Noncompliant
```
vilchik-elena: @ivandalbosco I just found out that it will not be always false. Looks like that they are converted to String, so we should ignore comparisons with Strings

javascript-checks/src/test/resources/checks/comparisonReturningFalse.js
```diff
@@ -0,0 +1,97 @@
...
+//^^^
+  obj > true;                  // Noncompliant
+//^^^
+  obj > false;                 // Noncompliant
+//^^^
+  obj >= symb;                 // Noncompliant
```
vilchik-elena: @ivandalbosco We don't support Symbol type, so I would suggest to replace it with something unknown (`unknown = foo();`) to not confuse anyone. Moreover, keeping in mind that comparison with string is fine (see comment above) we should not raise an issue when another operand is unknown

javascript-checks/src/test/resources/checks/comparisonReturningFalse.js
```diff
@@ -0,0 +1,97 @@
...
+// Noncompliant@+1
+  date < obj;                  // Noncompliant
+  nul < obj;                   // Noncompliant
+  undef < obj;                 // Noncompliant
+  
+  // 3. compare an object with itself
+  var objAlias = obj;
+// Noncompliant@+1
+  obj > obj;                   // Noncompliant
+  obj >= obj;                  // OK (returns true, actually)
```
vilchik-elena: @ivandalbosco Just of curiosity, did you find such issue on ruling?

javascript-checks/src/test/resources/checks/comparisonReturningFalse.js
```diff
@@ -0,0 +1,97 @@
...
+//^^^
+// Noncompliant@+1
+  obj >= date;                 // Noncompliant
+//       ^^^^
+  obj >= nul;                  // Noncompliant
+//^^^
+  obj >= undef;                // Noncompliant
+//^^^
+// Noncompliant@+1
+  obj > obj2;                  // Noncompliant
```
vilchik-elena: @ivandalbosco When there are 2 objects compared, they are converted to strings, no it's a FP issue. When we raise issue, we should check second operand as well, that's it's primitive

javascript-checks/src/test/resources/checks/comparisonReturningFalse.js
```diff
@@ -0,0 +1,97 @@
...
+  date < date2;                // OK
+  date <= date2;               // OK
+}
+
+function testStuffNotCoveredByThisCheck() {
+  var nul = null;
+  var tru = true;
+
+  1 < 2;                       // OK
+  nul > "hello";               // OK
```
vilchik-elena: @ivandalbosco IMO there is no use in creating variables for `true` and `null` values (especially, with such names), you can you literals directly

javascript-checks/src/test/resources/checks/comparisonReturningFalse.js
```diff
@@ -0,0 +1,97 @@
...
+  tru <= Symbol();             // OK
```
vilchik-elena: @ivandalbosco I don't think this case should be covered in future. `Symbol` is a special kind. You expression will raise `TypeError`

javascript-checks/src/test/resources/checks/comparisonReturningFalse.js
```diff
@@ -0,0 +1,97 @@
...
+  date >= date2;               // OK
+  date < date2;                // OK
+  date <= date2;               // OK
+}
+
+function testStuffNotCoveredByThisCheck() {
+  var nul = null;
+  var tru = true;
+
+  1 < 2;                       // OK
```
vilchik-elena: @ivandalbosco I don't think it's a right place for this test. This expression is compliant, and not "NotCoveredByThisCheck" as you say

javascript-checks/src/main/java/org/sonar/javascript/checks/ComparisonReturningFalseCheck.java
```diff
@@ -0,0 +1,116 @@
...
+
+  /**
+   * The operands for which an issue has already been raised.
+   */
+  private Set<ExpressionTree> operandsWithIssues = new HashSet<>();
+
+  @Override
+  public void startOfExecution(Scope functionScope) {
+    operandsWithIssues.clear();
+    super.startOfExecution(functionScope);
```
ivandalbosco: @vilchik-elena OK

javascript-checks/src/test/resources/checks/comparisonReturningFalse.js
ivandalbosco: @vilchik-elena good catch!

javascript-checks/src/test/resources/checks/comparisonReturningFalse.js
ivandalbosco: @vilchik-elena I tested secondary locations in several places, because the code behaves differently according to the type of the two operands, so I'd rather check several (left operand, right operand) pairs.

javascript-checks/src/test/resources/checks/comparisonReturningFalse.js
ivandalbosco: @vilchik-elena OK for the "unknown" case - I had missed it. I'd rather stick with Symbol, which may not be widely used but is a basic JS type.

vilchik-elena: @ivandalbosco Then the intend of the test case is not clear

javascript-checks/src/test/resources/checks/comparisonReturningFalse.js
ivandalbosco: @vilchik-elena 2 objects are converted to strings... unless one or both can be converted to numbers. I only know of Date as objects that convert to numbers. I am searching if other built-in object also convert to numbers

javascript-checks/src/test/resources/checks/comparisonReturningFalse.js
ivandalbosco: @vilchik-elena OK, I remove Symbol...

javascript-checks/src/test/resources/checks/comparisonReturningFalse.js
ivandalbosco: @vilchik-elena No, I didn't. I am going to fix the implementation and to remove this test, because comparison of variable to itself is out of the scope of this rule.

javascript-checks/src/test/resources/checks/comparisonReturningFalse.js
ivandalbosco: @vilchik-elena I am going to change the function's name.

javascript-checks/src/test/resources/checks/comparisonReturningFalse.js
ivandalbosco: @vilchik-elena OK

javascript-checks/src/test/resources/checks/comparisonReturningFalse.js
ivandalbosco: @vilchik-elena OK

javascript-checks/src/main/java/org/sonar/javascript/checks/ComparisonReturningFalseCheck.java
```diff
@@ -0,0 +1,113 @@
...
+      raiseIssue(element, leftIsUndefined, rightIsUndefined);
+    }
+  }
+  
+  private static boolean checkConvertibleToNumber(Constraint one, Constraint two) {
+    return isObjectButNotConvertibleToNumber(one) && isConvertibleToNumber(two);
+  }
+  
+  private static boolean isObjectButNotConvertibleToNumber(Constraint c) {
+    return c.isStricterOrEqualTo(Constraint.OBJECT)
```
vilchik-elena: @ivandalbosco Another way to do it is to create a constant constraint using `and`, `or`, `not`. See how PY did it (https://github.com/SonarSource/sonar-javascript/blob/master/javascript-checks/src/main/java/org/sonar/javascript/checks/ArithmeticOperationReturningNanCheck.java#L48) 

IMO it's a bit more clear way of doing this check

javascript-checks/src/main/java/org/sonar/javascript/checks/ComparisonReturningFalseCheck.java
```diff
@@ -0,0 +1,113 @@
...
+  
+  private static boolean isObjectButNotConvertibleToNumber(Constraint c) {
+    return c.isStricterOrEqualTo(Constraint.OBJECT) 
+       && !c.isStricterOrEqualTo(Constraint.BOOLEAN_OBJECT)
+       && !c.isStricterOrEqualTo(Constraint.NUMBER_OBJECT)
+       && !c.isStricterOrEqualTo(Constraint.DATE);
+  }
+  
+  private static boolean isConvertibleToNumber(Constraint c) {
+    return c.isStricterOrEqualTo(Constraint.ANY_BOOLEAN)
```
vilchik-elena: @ivandalbosco Same comment as for `isObjectButNotConvertibleToNumber`

javascript-checks/src/main/java/org/sonar/javascript/checks/ComparisonReturningFalseCheck.java
```diff
@@ -0,0 +1,113 @@
...
+    if (checkConvertibleToNumber(leftConstraint, rightConstraint)) {
+      raiseIssue(element, true, false);
+    } else if (checkConvertibleToNumber(rightConstraint, leftConstraint)) {
+      raiseIssue(element, false, true);
+    } else if (leftIsUndefined || rightIsUndefined) {
+      raiseIssue(element, leftIsUndefined, rightIsUndefined);
+    }
+  }
+  
+  private static boolean checkConvertibleToNumber(Constraint one, Constraint two) {
```
vilchik-elena: @ivandalbosco We should find a better name for this method, e.g. `checkObjectIsComparedNumerically`?

PR #383, opened 2016-11-03
Distinguish primitives and objects

pynicolas:


javascript-frontend/src/main/java/org/sonar/javascript/se/Type.java
```diff
@@ -53,7 +56,13 @@
   ;
 
   private static final List<Type> VALUES_REVERSED = Lists.reverse(Arrays.asList(Type.values()));
-  private static final EnumSet<Type> PRIMITIVE_TYPES = EnumSet.of(NUMBER, STRING, BOOLEAN);
+  private static final EnumSet<Type> PRIMITIVE_TYPES = EnumSet.of(
```
vilchik-elena: @pynicolas we should remove objects from this set

javascript-frontend/src/test/java/org/sonar/javascript/se/builtins/BuiltInPropertiesTest.java
```diff
@@ -36,25 +36,25 @@
 
   @Test
   public void test_string() throws Exception {
-    type = Type.STRING;
+    type = Type.STRING_OBJECT;
```
vilchik-elena: @pynicolas Why object? (in next tests you use primitive)

PR #388, opened 2016-11-04
SONARJS-813 Rule: Trailing commas should be used

ivandalbosco:


javascript-checks/src/test/resources/checks/missingTrailingComma.js
```diff
@@ -0,0 +1,84 @@
...
+  "Smith"];                                 // OK
+
+var arr = [
+  "joe",  "Smith"                           // Noncompliant  
+//        ^^^^^^^
+];
+
+var arr = [
+  "joe",
+  "Smith",                                  // OK
```
vilchik-elena: @ivandalbosco I think it's worth to test empty element:
```
var arr = [
  "joe",,
  "Smith"                                 // Noncompliant
];
```

PR #397, opened 2016-11-14
SONARJS-775 Set up default profile for SonarLint

pynicolas:


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/JavaScriptProfile.java
```diff
@@ -61,22 +59,22 @@ private void loadFromCommonRepository(RulesProfile profile) {
...
 
   private void loadActiveKeysFromJsonProfile(RulesProfile rulesProfile) {
-    URL profileUrl = JavaScriptProfile.class.getResource("/org/sonar/l10n/javascript/rules/javascript/Sonar_way_profile.json");
+    for (String ruleKey : activatedRuleKeys()) {
+      Rule rule = ruleFinder.findByKey(CheckList.REPOSITORY_KEY, ruleKey);
+      rulesProfile.activateRule(rule, null);
+    }
+  }
 
+  public static Set<String> activatedRuleKeys() {
```
ivandalbosco: @pynicolas @vilchik-elena Sorry for interfering in your dialogue. I suggest that method is renamed getDefaultRuleKeys, because activation has been deported out of it.

vilchik-elena: @ivandalbosco @pynicolas think we can keep it like that. Method returns not the rules, but just their keys, and the point of the name that the rules, which correspond to the returned rule keys, should be activated. IMO may be a better alternative is `activeRuleKeys` but that's really minor

PR #401, opened 2016-12-08
SONARJS-824 Extend CallabilityCheck: raise issue when at least one execution path leads to exception

vilchik-elena:


javascript-checks/src/main/java/org/sonar/javascript/checks/CallabilityCheck.java
```diff
@@ -19,38 +19,42 @@
...
 import org.sonar.plugins.javascript.api.tree.Tree;
 import org.sonar.plugins.javascript.api.tree.Tree.Kind;
 import org.sonar.plugins.javascript.api.tree.expression.CallExpressionTree;
 
 @Rule(key = "S2873")
-public class CallabilityCheck extends AbstractAllPathSeCheck<CallExpressionTree> {
+public class CallabilityCheck extends SeCheck {
 
-  private static final String MESSAGE = "This expression has a value which cannot be called; it is not a function.";
+  private static final String MESSAGE = "This expression might have a value which cannot be called; it is not a function.";
```
vilchik-elena: @ivandalbosco I've updated RSPEC (but there is no reason for regeneration of the doc)

PR #404, opened 2016-12-09
SONARJS-819 Rule: "in" should not be used with primitive types

vilchik-elena:


javascript-checks/src/test/resources/checks/InOperatorTypeError.js
```diff
@@ -0,0 +1,20 @@
...
+//S       ^^ 1
+             1; // Noncompliant [[id=1]] {{TypeError can be thrown as this operand might have primitive type.}}
+//           ^
+
+  unknown in "foo"; // Noncompliant
+  unknown in true;  // Noncompliant
+  unknown in null;  // Noncompliant
+  unknown in undefined;  // Noncompliant
+  unknown in new Number(1); // OK
+
```
vilchik-elena: @ivandalbosco it's not "in" operator if for-in loop, but we could have a test.

PR #408, opened 2016-12-12
SE: add assertion that at least one branch is executed

vilchik-elena:


its/ruling/src/test/expected/javascript-EqEqEq.json
```diff
@@ -5585,10 +5585,13 @@
 351,
 ],
 'project:dojo-1.8.1/dojox/json/schema.js.uncompressed.js':[
+181,
```
pynicolas: @vilchik-elena Why do we have this new issue?

javascript-frontend/src/test/resources/se/relations.js
```diff
@@ -58,4 +58,79 @@ function main() {
...
+
+  if (a === b) {
+    if (a) {
+      foo(b); // PS b=TRUTHY
+    }
+  }
+
+///------------- transitivity
+
+//  a = foo();
```
pynicolas: @vilchik-elena Why is this block commented?

javascript-frontend/src/main/java/org/sonar/javascript/se/ProgramState.java
```diff
@@ -158,6 +180,23 @@ public ProgramState newFunctionSymbolicValue(Symbol symbol, FunctionTree functio
     return value.constrainDependencies(newState, constraint);
   }
 
+  private Set<SymbolicValue> getEquivalentValues(SymbolicValue value) {
```
pynicolas: @vilchik-elena I'm a bit worried about the potential performance impact of this method.

javascript-frontend/src/test/resources/se/relations.js
```diff
@@ -58,4 +58,79 @@ function main() {
     }
   }
 
+///-------------- should not fail
```
pynicolas: @vilchik-elena Tests which don't assert anything are suspicious.

javascript-frontend/src/main/java/org/sonar/javascript/se/ProgramState.java
vilchik-elena: @pynicolas perf test is still green and I don't expect a lot of relations to be created at once

javascript-frontend/src/test/resources/se/relations.js
vilchik-elena: @pynicolas because it's not implemented:) I should drop it

javascript-frontend/src/main/java/org/sonar/javascript/se/Relation.java
```diff
@@ -53,7 +53,7 @@
       Kind.STRICT_NOT_EQUAL_TO,
       Kind.STRICT_EQUAL_TO,
       Kind.STRICT_NOT_EQUAL_TO,
-      ImmutableSet.of(Kind.NOT_EQUAL_TO, Kind.LESS_THAN, Kind.GREATER_THAN, Kind.LESS_THAN_OR_EQUAL_TO, Kind.GREATER_THAN_OR_EQUAL_TO)),
+      ImmutableSet.of(Kind.NOT_EQUAL_TO, Kind.LESS_THAN, Kind.GREATER_THAN, Kind.LESS_THAN_OR_EQUAL_TO, Kind.GREATER_THAN_OR_EQUAL_TO, Kind.EQUAL_TO)),
```
ivandalbosco: Correct. I also checked attribute compatibleKindsForSameOperands of all other enum values : they all seem to be correct.

javascript-frontend/src/test/resources/se/relations.js
```diff
@@ -58,4 +58,13 @@ function main() {
...
 
+  a = foo();
+  b = bar();
+
+  if (typeof a == "string" && typeof b == "number") {
+    if (a == b) {
+    }
+    if (a === b) {
+    }
+  }
```
ivandalbosco: @vilchik-elena I am a bit surprised that your change in class Relation incurs no change in its test class RelationTest. Shouldn't RelationTest#isCompatibleWith_and_same_operands test ALL combinations? A test method, isCompatibleOnlyWith(List<Relation>) would do that quite economically in terms of lines of code.

javascript-frontend/src/main/java/org/sonar/javascript/se/ProgramState.java
```diff
@@ -147,6 +149,26 @@ public ProgramState newFunctionSymbolicValue(Symbol symbol, FunctionTree functio
...
+
+        } else {
+          break;
+        }
+      }
+    }
+
+    return ps;
+  }
+
```
ivandalbosco: @vilchik-elena All that is probably wonderful, but couldn't we dream of something else, possibly simpler? The idea would be to merge symbolic values instead of keeping 2 (or more) instances and trying to synchronize them.

One option to do so would be to use a small class SymbolicValueRef which would be a simple wrapper on a SymbolicValue. Upon an assignment such as "b = a", SVRef(b) would no longer wrap SV(b) and would start wrapping SV(a). But the new indirection would mean some additional complexity.

Another option would be to clean everything (ProgramState#values, ProgramState#constraints, ProgramState#relations, and possibly also the ExpressionStack) at every Assignment operation. This would leave the program state in a clean state (which the first option above doesn't provide), possibly saving some execution time.

What do you think?

javascript-frontend/src/main/java/org/sonar/javascript/se/sv/EqualitySymbolicValue.java
```diff
@@ -59,13 +59,19 @@
...
       if (constraint.isStricterOrEqualTo(Constraint.TRUTHY)) {
         return state.constrain(receiving, constraintOnNullOrUndefined(constraintOnSending, false));
       } else if (constraint.isStricterOrEqualTo(Constraint.FALSY)) {
         return state.constrain(receiving, constraintOnNullOrUndefined(constraintOnSending, true));
       }
     }
+
+    if (Kind.STRICT_EQUAL_TO.equals(kind) && constraint.isStricterOrEqualTo(Constraint.TRUTHY)) {
+      return state.constrain(receiving, constraintOnSending);
+    }
```
ivandalbosco: @vilchik-elena This is correct but a little restrictive. Handles:

function bar(d) {
  let c = d;
  let a = (c === null);
  if (d === 'aaa') {
    // here we know that c = 'aaa' and a = false
    if (typeof(c) == "string") {  // OK, raises "always true" here
    }
    if (a) {                      // OK, raises "always false" here
    }
  }
  foo(c);
}

but does not handle: 

function baz(d) {
  let c = d;
  let a = (c === null);
  if (d !== 'aaa') {
  } else {
    // here we know that c = 'aaa' and a = false
    if (typeof(c) == "string") {  // KO, misses "always true" here
    }
    if (a) {                      // KO, misses "always false" here
    }
  }
  foo(c);
}

ivandalbosco: Sorry for the lost indentations!

javascript-frontend/src/test/resources/se/relations.js
vilchik-elena: @ivandalbosco indeed, thanks!

javascript-frontend/src/main/java/org/sonar/javascript/se/sv/EqualitySymbolicValue.java
vilchik-elena: @ivandalbosco indeed! good idea! I gonna implement it

PR #415, opened 2016-12-16
Add performance IT for SE

vilchik-elena:


its/performancing/src/test/resources/se_profile.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<profile>
+  <name>se-profile</name>
+  <language>js</language>
+  <rules>
+    <rule>
+      <repositoryKey>javascript</repositoryKey>
+      <key>S2583</key>  <!--Conditions should not unconditionally evaluate to "true" or to "false"-->
+      <priority>MAJOR</priority>g
```
pynicolas: @vilchik-elena What's this "g"?

its/performancing/src/test/java/JavaScriptPerformanceTest.java
```diff
@@ -38,36 +38,54 @@
...
   }
 
+  @Test
+  public void test_symbolic_engine_performance() throws IOException {
+    ORCHESTRATOR.getServer().provisionProject("se-project", "se-project");
+    ORCHESTRATOR.getServer().associateProjectToQualityProfile("se-project", "js", "se-profile");
+
+    SonarScanner build = getSonarScanner();
+
+    ORCHESTRATOR.executeBuild(getSonarScanner());
```
pynicolas: @vilchik-elena Why do we call `getSonarScanner()` twice?

its/performancing/src/test/java/JavaScriptPerformanceTest.java
```diff
@@ -38,36 +38,54 @@
...
+    double time = sensorTime(build.getProjectDir(), SENSOR);
+
+    double expected = 140.0;
+    Assertions.assertThat(time).isEqualTo(expected, offset(expected * 0.04));
```
pynicolas: @vilchik-elena Maybe we should also add some assertion on the number of issues to be sure that symbolic execution was run.

its/performancing/src/test/java/JavaScriptPerformanceTest.java
```diff
@@ -38,38 +40,61 @@
...
   @ClassRule
   public static final Orchestrator ORCHESTRATOR = Orchestrator.builderEnv()
-    .restoreProfileAtStartup(FileLocation.of("src/test/profile.xml"))
+    .restoreProfileAtStartup(FileLocation.of("src/test/resources/no_rules.xml"))
+    .restoreProfileAtStartup(FileLocation.of("src/test/resources/se_profile.xml"))
     .addPlugin(FileLocation.byWildcardMavenFilename(
       new File("../../sonar-javascript-plugin/target"), "sonar-javascript-plugin-*.jar"))
     .build();
 
+  private SonarScanner build;
```
pynicolas: @vilchik-elena Why a field? It seems that's only used as a local variable.

its/performancing/src/test/java/JavaScriptPerformanceTest.java
```diff
@@ -38,38 +40,61 @@
...
+
+  @Test
+  public void test_parsing_performance() throws IOException {
+    String projectKey = "parsing-project";
+    ORCHESTRATOR.getServer().provisionProject(projectKey, projectKey);
```
pynicolas: @vilchik-elena I think that we could still remove a bit of duplication between `test_parsing_performance` and `test_symbolic_engine_performance` :) That would highlight the differences.

PR #417, opened 2016-12-16
Add License and copyright files

tomverin:


third-party-licenses.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
```
ivandalbosco: @thomas @drautureau-sonarsource  I added a chmod u+x because there are no exec rights.

ivandalbosco: @tomverin Hmm, wrong thomas, sorry to both thomass!

PR #419, opened 2016-12-19
SONARJS-822 Rule: Jump statements should not be used unconditionally

pynicolas:


javascript-checks/src/main/java/org/sonar/javascript/checks/UnconditionalJumpStatementCheck.java
```diff
@@ -0,0 +1,105 @@
...
+
+@Rule(key = "S1751")
+public class UnconditionalJumpStatementCheck extends SubscriptionVisitorCheck {
+
+  private static final Kind[] LOOP_KINDS = {
+    Kind.WHILE_STATEMENT,
+    Kind.DO_WHILE_STATEMENT,
+    Kind.FOR_STATEMENT,
+    Kind.FOR_OF_STATEMENT
+  };
```
vilchik-elena: @pynicolas formatting: move this on previous line

javascript-checks/src/main/java/org/sonar/javascript/checks/UnconditionalJumpStatementCheck.java
```diff
@@ -79,12 +80,20 @@ private static boolean canExecuteMoreThanOnce(IterationStatementTree loopTree) {
...
-    for (CfgBlock loopPredecessor : block.predecessors()) {
-      StatementTree loopBody = loopTree.statement();
+  private static boolean hasPredecessorInsideLoopBody(CfgBranchingBlock conditionBlock, IterationStatementTree loopTree) {
+    for (CfgBlock loopPredecessor : conditionBlock.predecessors()) {
       List<Tree> predecessorElements = loopPredecessor.elements();
-      Tree lastElement = predecessorElements.get(predecessorElements.size() - 1);
-      if (isDescendant(lastElement, loopBody)) {
+      Tree predecessorLastElement = predecessorElements.get(predecessorElements.size() - 1);
+
+      if (loopTree.is(Kind.FOR_STATEMENT)) {
```
vilchik-elena: @pynicolas We can remove this "hack" if improve CFG and check not last but first element

javascript-checks/src/test/resources/checks/UnconditionalJumpStatement.js
```diff
@@ -81,3 +81,16 @@ while(foo()) {
...
+    continue;
+  }
+  baz();
+  break; // compliant
+}
+
+for (i = 0; foo();) {
+  baz();
+  break; // Noncompliant
+}
```
vilchik-elena: @pynicolas can we have a case for `for` loop with noncompliant `continue`?

javascript-checks/src/test/resources/checks/UnconditionalJumpStatement.js
```diff
@@ -0,0 +1,96 @@
...
+
+for(p in obj) {
+  bar();
+  break; // Compliant: often used to check whether an object is "empty"
+}
+
+for(p of obj) {
+  bar();
+  break; // Noncompliant
```
vilchik-elena: @pynicolas what about raising issue for 
```
for (p in obj) { foo(); continue; }
```
?

javascript-checks/src/main/java/org/sonar/javascript/checks/UnconditionalJumpStatementCheck.java
pynicolas: @vilchik-elena We could improve the CFG, but it's not easy: I don't think it's worth it.

javascript-checks/src/test/resources/checks/UnconditionalJumpStatement.js
pynicolas: @vilchik-elena Done in 977ed8c

javascript-checks/src/main/java/org/sonar/javascript/checks/UnconditionalJumpStatementCheck.java
pynicolas: @vilchik-elena Changed in e3226ef

javascript-checks/src/test/resources/checks/UnconditionalJumpStatement.js
pynicolas: @vilchik-elena Done in a22d781

PR #420, opened 2016-12-19
SONARJS-837 Rule: Functions should use "return" consistently

vilchik-elena:


javascript-checks/src/main/java/org/sonar/javascript/checks/ConsistentReturnsCheck.java
```diff
@@ -0,0 +1,126 @@
...
+      BlockTree body = (BlockTree) functionTree.body();
+      FunctionReturns functionReturns = getFunctionReturns(body);
+
+      if (functionReturns.containsReturnWithoutValue && functionReturns.containsReturnWithValue) {
+        raiseIssue(tree, functionReturns, body);
+      }
+    }
+  }
+
+  private void raiseIssue(Tree tree, FunctionReturns functionReturns, BlockTree body) {
```
pynicolas: @vilchik-elena I suggest to rename `tree` to `functionTree`.

javascript-checks/src/main/java/org/sonar/javascript/checks/ConsistentReturnsCheck.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    return functionReturns;
+  }
+
+  private static boolean isThrowStatement(Tree lastElement) {
+    return ((JavaScriptTree) lastElement).getParent().is(Kind.THROW_STATEMENT);
+  }
+
+  private static boolean isUnreachableBlock(CfgBlock cfgBlock, ControlFlowGraph cfg) {
```
pynicolas: @vilchik-elena This method seems to return the opposite of what its name suggests.

vilchik-elena: @pynicolas thanks!

javascript-checks/src/main/java/org/sonar/javascript/checks/ConsistentReturnsCheck.java
vilchik-elena: @pynicolas ok

PR #422, opened 2016-12-19
SONARJS-836 Rule: Callbacks of array methods should have a return statement

vilchik-elena:


javascript-checks/src/main/java/org/sonar/javascript/checks/ArrayCallbackWithoutReturnCheck.java
```diff
@@ -0,0 +1,153 @@
...
+    hasIssue.clear();
+  }
+
+  @Override
+  public void beforeBlockElement(ProgramState currentState, Tree element) {
+    if (element.is(Kind.DOT_MEMBER_EXPRESSION)) {
+      DotMemberExpressionTree memberExpression = (DotMemberExpressionTree) element;
+
+      if (isArrayPropertyExecuted(currentState) && METHODS_WITH_CALLBACK.contains(memberExpression.property().name())) {
+        checkArgumentToBeFunctionWithReturn(memberExpression, 0);
```
vilchik-elena: @ivandalbosco You always can remove comment and create a new one with same content for another line

javascript-checks/src/test/resources/checks/ArrayCallbackWithoutReturn.js
```diff
@@ -0,0 +1,60 @@
...
+  myArray.every(function(){}); // Noncompliant
+  myArray.filter(function(){}); // Noncompliant
+  myArray.find(function(){}); // Noncompliant
+  myArray.findIndex(function(){}); // Noncompliant
+  myArray.map(function(){}); // Noncompliant
+  myArray.reduce(function(){}); // Noncompliant
+  myArray.reduceRight(function(){}); // Noncompliant
+  myArray.some(function(){}); // Noncompliant
+  myArray.sort(function(){}); // Noncompliant
+
```
vilchik-elena: @ivandalbosco Ok

javascript-checks/src/test/resources/checks/ArrayCallbackWithoutReturn.js
```diff
@@ -0,0 +1,60 @@
...
+  myArray.reduce(function(){ return; }); // Noncompliant
+
+  myArray.reduce((a, b) => {foo();}); // Noncompliant
+//                      ^^
```
vilchik-elena: @ivandalbosco We prefer to avoid "uncontrolled" locations. Lambda could be very big and location will be ugly (and not clearer at all)

javascript-checks/src/test/resources/checks/ArrayCallbackWithoutReturn.js
```diff
@@ -0,0 +1,60 @@
...
+
+  myArray.reduce((a, b) => a + b);
+  myArray.reduce();
+  foo(myArray.reduce);
```
vilchik-elena: @ivandalbosco good idea! But I prefer that intent of test is obvious, so I've added `myArray.foobar()`

javascript-checks/src/main/java/org/sonar/javascript/checks/ArrayCallbackWithoutReturnCheck.java
vilchik-elena: @ivandalbosco What this will test? Note I have a test `myArray.reduce(42);` on line 53

javascript-checks/src/main/java/org/sonar/javascript/checks/ArrayCallbackWithoutReturnCheck.java
```diff
@@ -0,0 +1,153 @@
...
+      }
+
+      if (isArrayFromMethod(memberExpression, currentState)) {
```
vilchik-elena: @ivandalbosco done:)

javascript-checks/src/main/java/org/sonar/javascript/checks/ArrayCallbackWithoutReturnCheck.java
ivandalbosco: @vilchik-elena  Nothing functional, just technical. Would cover your "if" statement, using an Array method that is not in the list. Method reduce is in the list.

PR #423, opened 2016-12-20
SONARJS-840 Rule: Method parameters, caught exceptions and foreach variables should not be reassigned

vilchik-elena:


javascript-checks/src/test/resources/checks/ReassignedParameter.js
```diff
@@ -0,0 +1,105 @@
...
+  p1 = 42; // Noncompliant
+  foo(p1);
+}
+
+// Ok, function call
+var p1;
+p1 = 42;
+foo(p1, p2);
+
+
```
ivandalbosco: @vilchik-elena I checked that for IIFE no issue is detected. I think it should. For example:
(function(a) {
  a = 13;
})(1);

javascript-checks/src/test/resources/checks/ReassignedParameter.js
```diff
@@ -0,0 +1,105 @@
...
+// ---------- EXCEPTION in CATCH clause ---------
+
+try {
+  foo();
+} catch (e) {
+  e = foo(); // Noncompliant {{Introduce a new variable instead of reusing the caught exception "e".}}
+}
+
```
ivandalbosco: @vilchik-elena How about adding
e = foo();   // OK
again here?

javascript-checks/src/test/resources/checks/ReassignedParameter.js
```diff
@@ -0,0 +1,105 @@
...
+// ---------- VARIABLE in FOR-IN/FOR-OF loops ---------
+
+for (var x in obj) {
+  x = foo(); // Noncompliant {{Introduce a new variable instead of reusing the foreach variable "x".}}
+}
+
+for (var [a, b] in obj) {
+  a = foo(); // FN Noncompliant (SYMBOL TABLE should be improved)
+}
```
ivandalbosco: @vilchik-elena Could you please remove "Noncompliant" as it is confusing.

javascript-checks/src/test/resources/checks/ReassignedParameter.js
```diff
@@ -0,0 +1,105 @@
...
+
+for (let {prop1, prop2} in obj) {
+  prop1 = foo(); // FN Noncompliant (SYMBOL TABLE should be improved)
+}
```
ivandalbosco: Idem

javascript-checks/src/test/resources/checks/ReassignedParameter.js
vilchik-elena: @ivandalbosco I have issue for such code

javascript-checks/src/test/resources/checks/ReassignedParameter.js
vilchik-elena: @ivandalbosco "e" outside "catch" is another symbol. So it will be mainly test of symbol table and not of this rule

javascript-checks/src/test/resources/checks/ReassignedParameter.js
vilchik-elena: @ivandalbosco ok

javascript-checks/src/main/java/org/sonar/javascript/checks/utils/CheckUtils.java
```diff
@@ -93,5 +93,16 @@ public static ControlFlowGraph buildControlFlowGraph(Tree tree) {
     }
     return ControlFlowGraph.build((ScriptTree) parent);
   }
+
+  public static boolean isDescendant(Tree tree, Tree potentialParent) {
```
ivandalbosco: @vilchik-elena We have no unit test for this function.

javascript-checks/src/test/resources/checks/ReassignedParameter.js
```diff
@@ -0,0 +1,105 @@
...
+  foo(p3++); // Noncompliant
+//    ^^
+
+  foo(p2);
+
+  var x = p2 + 1;
+
+  var p4 = 42; // Noncompliant
+}
+
```
ivandalbosco: @vilchik-elena How about adding a test such as:
function updateArgument(person) {
  person.name = "Mary";   // OK
}

javascript-checks/src/main/java/org/sonar/javascript/checks/utils/CheckUtils.java
vilchik-elena: @ivandalbosco Thanks! added in 09c46df4498f9084d95f43fbe03c517729248b1a

javascript-checks/src/test/resources/checks/ReassignedParameter.js
vilchik-elena: @ivandalbosco good idea! Added in c6a5bd126a4956539218d77590593b553eb59cba

PR #425, opened 2016-12-22
SONARJS-834 Rule: Template literal placeholder syntax should not be used in regular strings

vilchik-elena:


its/ruling/src/test/expected/javascript-S3786.json
```diff
@@ -0,0 +1,1951 @@
...
+51,
+164,
+],
+'project:open-layers-2.12/OpenLayers/Layer/OSM.js':[
+51,
+52,
+53,
+],
+'project:open-layers-2.12/OpenLayers/Layer/WMTS.js':[
+423,
```
pynicolas: @vilchik-elena Isn't that a FP?

vilchik-elena: @pynicolas indeed. I wanted to avoid using regexps and was hoping there is no code like that:) Looks like I need to introduce them

pynicolas: @vilchik-elena Or maybe check `indexOf("${") < indexOf ("}")`.

javascript-checks/src/test/resources/checks/TemplateStringMisuse.js
```diff
@@ -4,3 +4,8 @@
 `Hello ${name}`; // OK
 `Hello name`; // OK
 `Hello {name}`; // OK
+
```
ivandalbosco: @vilchik-elena How about adding a test with two placeholders, e.g.,
'Hello ${firstName} ${lastName}';

ivandalbosco: @vilchik-elena Moreover RSPEC-3786 mentions secondary location. You should either update the RSPEC or add secondary location in the check class.

vilchik-elena: @ivandalbosco I removed it from RSPEC, thanks

vilchik-elena: @ivandalbosco Added unit test

PR #426, opened 2016-12-22
SONARJS-839 Rule: Destructuring patterns should not be empty

vilchik-elena:


javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S3799.html
```diff
@@ -0,0 +1,15 @@
+<p>Destructuring is a convenient way of extracting multiple values from data stored in (possibly nested) objects and arrays. However, it is possible
+to create an empty pattern that has no effect. When empty curly braces or brackets are used to the right of a property name most of the time the
+intent was to use a default value instead.</p>
+<p>This rule raises an issue when empty destructuring pattern is used.</p>
+<h2>Noncompliant Code Example</h2>
+<pre>
+var {a: {}} = foo; // Noncompliant
```
pynicolas: @vilchik-elena We should probably define our own code examples instead of copying some example from another tool.

javascript-checks/src/main/java/org/sonar/javascript/checks/EmptyDestructuringPatternCheck.java
```diff
@@ -0,0 +1,49 @@
...
+import org.sonar.plugins.javascript.api.visitors.DoubleDispatchVisitorCheck;
+
+@Rule(key = "S3799")
+public class EmptyDestructuringPatternCheck extends DoubleDispatchVisitorCheck {
+
+  private static final String MESSAGE = "Change this pattern to not be empty.";
+
+  @Override
+  public void visitArrayBindingPattern(ArrayBindingPatternTree tree) {
+    if (tree.elements().isEmpty()) {
```
pynicolas: @vilchik-elena This condition is not tested.

javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S3799.html
```diff
@@ -0,0 +1,19 @@
+<p>Destructuring is a convenient way of extracting multiple values from data stored in (possibly nested) objects and arrays. However, it is possible
+to create an empty pattern that has no effect. When empty curly braces or brackets are used to the right of a property name most of the time the
+intent was to use a default value instead.</p>
+<p>This rule raises an issue when empty destructuring pattern is used.</p>
+<h2>Noncompliant Code Example</h2>
+<pre>
+var {a: {}, b} = myObj; // Noncompliant
+function foo([first: {}, second]) { // Noncompliant
```
m-g-sonar: Are you sure about this syntax? I've got a parse error when I try it:
```
Parse error at line 5 column 16:

1: function myFunc() {
2:   var {a: {}, b} = myObj;
3:   return myObj;
4: }
5: function foo([a: {}, b]) {
                  ^
6: }
7: 
```

javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S3799.html
```diff
@@ -0,0 +1,19 @@
...
+  // ...
+}
+</pre>
+<h2>Compliant Solution</h2>
+<pre>
+var {a = {}, b} = myObj;
+function foo([first = {}, second]) {
```
m-g-sonar: same question about the brackets.

PR #427, opened 2016-12-22
SONARJS-835 Rule: Property setters should not be defined without getters

vilchik-elena:


javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S2376.html
```diff
@@ -0,0 +1,23 @@
+<p>In JavaScript when an object is created with setter for a property but without a corresponding getter it renders the property inaccessible, so it
```
ivandalbosco: @vilchik-elena We've just updated the RPSEC.

javascript-checks/src/main/java/org/sonar/javascript/checks/GetterSetterCheck.java
```diff
@@ -0,0 +1,90 @@
...
+@Rule(key = "S2376")
+public class GetterSetterCheck extends DoubleDispatchVisitorCheck {
+
+  private static final String MESSAGE = "Provide a %s matching this %s for '%s'.";
+
+  @Override
+  public void visitObjectLiteral(ObjectLiteralTree tree) {
+    checkProperties(tree.properties());
+
+    super.visitObjectLiteral(tree);
```
ivandalbosco: @vilchik-elena If someone removes this (necessary) line, the unit test will go on passing. Can't you enrich the unit test with a case featuring nested objects?

javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S2376.html
vilchik-elena: @ivandalbosco done

PR #429, opened 2016-12-22
SONARJS-67 Rule: Variables should not be declared and then immediately returned or thrown

vilchik-elena:


its/ruling/src/test/expected/javascript-S1488.json
```diff
@@ -32,10 +32,6 @@
 'project:dojo-1.8.1/dojo/dnd/Source.js.uncompressed.js':[
 68,
 ],
-'project:dojo-1.8.1/dojo/dojo.js.uncompressed.js':[
-2638,
-15035,
```
pynicolas: @vilchik-elena Why did we lose some issues?

javascript-checks/src/main/java/org/sonar/javascript/checks/ImmediatelyReturnedVariableCheck.java
```diff
@@ -0,0 +1,106 @@
...
+
+  private static final String MESSAGE = "Immediately %s this expression instead of assigning it to the temporary variable \"%s\".";
+
+  @Override
+  public void visitBlock(BlockTree tree) {
+    List<StatementTree> statements = tree.statements();
+
+    if (statements.size() > 1) {
+
+      StatementTree lastButOneStatement = statements.get(statements.size() - 2);
```
m-g-sonar: `penultimateStatement` ? :D

javascript-checks/src/main/java/org/sonar/javascript/checks/ImmediatelyReturnedVariableCheck.java
```diff
@@ -0,0 +1,106 @@
...
+  private void checkStatements(VariableDeclarationTree variableDeclaration, StatementTree lastStatement) {
+    SeparatedList<BindingElementTree> variables = variableDeclaration.variables();
+
+    if (variables.size() == 1 && variables.get(0).is(Kind.INITIALIZED_BINDING_ELEMENT)) {
+      InitializedBindingElementTree initializedBindingElementTree = (InitializedBindingElementTree) variables.get(0);
+
+      if (initializedBindingElementTree.left().is(Kind.BINDING_IDENTIFIER)) {
+        String name = ((IdentifierTree) initializedBindingElementTree.left()).name();
+
+        if (lastStatementReturnsVariable(lastStatement, name)) {
```
m-g-sonar: I would write call these methods `returnsVariableInStatement` and `throwsVariableInStatement`, in order to have the difference of naming at the begining of the name rather that in the middle.

javascript-checks/src/main/java/org/sonar/javascript/checks/ImmediatelyReturnedVariableCheck.java
```diff
@@ -0,0 +1,106 @@
...
+        }
+      }
+    }
+  }
+
+  private static boolean lastStatementReturnsVariable(StatementTree lastStatement, String variableName) {
+    if (lastStatement.is(Kind.RETURN_STATEMENT)) {
+      ReturnStatementTree returnStatement = (ReturnStatementTree) lastStatement;
+
+      if (returnStatement.expression() != null && returnStatement.expression().is(Kind.IDENTIFIER_REFERENCE)) {
```
m-g-sonar: I would factorize this part of the method, in order to have a method `isVariable(ExpressionTree expr, variableName)` that you can call from this method and from `lastStatementThrowsVariable`, instead of having two time nearly exactly the same thing.

PR #440, opened 2017-01-10
Sonarjs 871

inverno:


javascript-checks/src/main/java/org/sonar/javascript/checks/OpenCurlyBracesAtEOLCheck.java
```diff
@@ -147,21 +147,36 @@ private void checkConditional(Tree parent, SyntaxToken openCurly) {
...
     }
   }
 
   private void issueIfLineMismatch(SyntaxToken curlyBrace, SyntaxToken target) {
-    if (curlyBrace.line() != target.line()) {
+    CodeLine curlyBraceLine = new CodeLine(curlyBrace.line());
+    if (curlyBraceLine.isJustBelow(target.line())) {
       addIssue(new PreciseIssue(this, new IssueLocation(curlyBrace, "Move this open curly brace to the end of the previous line.")));
     }
+    if (curlyBraceLine.isFarBelow(target.line())) {
```
ivandalbosco: @inverno Why don't you use an "else" here? With an "else", without reading the implementation of methods isJust/FarBelow, we know that all cases are covered, that is, that we will always hit one of the 2 cases. Further advantage: either of the two methods isJust/FarBelow can be removed.

PR #445, opened 2017-01-13
SONARJS-861 Rule: Loops should not be infinite

inverno:


javascript-frontend/src/main/java/org/sonar/javascript/tree/impl/JavaScriptTree.java
```diff
@@ -85,6 +93,52 @@ public SyntaxToken getFirstToken() {
...
+  public Stream<Usage> allSymbolsUsages() {
+    return allSymbols().flatMap(symbol -> symbol.usages().stream()).filter(usage -> this.isAncestorOf((JavaScriptTree) usage.identifierTree()));
+  }
+
+  public boolean isAncestorOf(JavaScriptTree tree) {
+    Tree parentTree = tree.getParent();
+    if (this.equals(parentTree)) {
+      return true;
+    }
+    if (parentTree instanceof JavaScriptTree) {
```
vilchik-elena: @inverno remove this check and make `getParent` return JSTree

javascript-frontend/src/main/java/org/sonar/javascript/tree/impl/JavaScriptTree.java
```diff
@@ -85,6 +93,52 @@ public SyntaxToken getFirstToken() {
...
+      return this.isAncestorOf((JavaScriptTree) parentTree);
+    }
+    return false;
+  }
+
+  public Stream<JavaScriptTree> thisAndKin() {
```
vilchik-elena: @inverno `private`

javascript-frontend/src/main/java/org/sonar/javascript/tree/impl/JavaScriptTree.java
```diff
@@ -85,6 +93,52 @@ public SyntaxToken getFirstToken() {
     return ((JavaScriptTree) child).getFirstToken();
   }
 
+  public Stream<Symbol> allSymbols() {
```
vilchik-elena: @inverno move to check

javascript-frontend/src/main/java/org/sonar/javascript/tree/impl/JavaScriptTree.java
```diff
@@ -85,6 +93,52 @@ public SyntaxToken getFirstToken() {
...
+    return thisAndKin().filter(tree -> tree instanceof IdentifierTree)
+      .map(tree -> (IdentifierTree) tree)
+      .filter(identifierTree -> identifierTree.symbol() != null)
+      .map(IdentifierTree::symbol);
+  }
+
+  public Stream<Usage> allSymbolsUsages() {
```
vilchik-elena: @inverno move to check

javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S2189.html
```diff
@@ -0,0 +1,56 @@
...
+
+outer:
+while(true) {
+  while(true) {
+    break outer;
+  }
+}
+
+</pre>
+<p>Known limitations of this rule's implementation include :</p>
```
vilchik-elena: @inverno Could you ask Ann to validate such placing of information? (If you have not done this yet)

javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S2189.json
```diff
@@ -0,0 +1,15 @@
...
+  "type": "BUG",
+  "status": "ready",
+  "remediation": {
+    "func": "Constant\/Issue",
+    "constantCost": "15min"
+  },
+  "tags": [
+    "bug",
+    "cert",
+    "javascript"
```
vilchik-elena: @inverno We should not have `javascript` tag

javascript-checks/src/test/resources/checks/loopsShouldNotBeInfinite.js
```diff
@@ -0,0 +1,171 @@
+/*
```
vilchik-elena: @inverno What for this empty comment?

javascript-checks/src/test/resources/checks/loopsShouldNotBeInfinite.js
```diff
@@ -0,0 +1,171 @@
...
+*/
+for ( ; ; ) {       // Noncompliant {{Add an end condition for this loop}}
+//    ^^^
+}
+
+for (;;) {          // Noncompliant {{Add an end condition for this loop}}
```
vilchik-elena: @inverno Is it important to test spaced and not spaced semicolons?

javascript-checks/src/test/resources/checks/loopsShouldNotBeInfinite.js
```diff
@@ -0,0 +1,171 @@
+/*
+*/
+for ( ; ; ) {       // Noncompliant {{Add an end condition for this loop}}
```
vilchik-elena: @inverno I would put issue location on loop keyword `for` (which makes sense with such message)

javascript-checks/src/test/resources/checks/loopsShouldNotBeInfinite.js
```diff
@@ -0,0 +1,171 @@
...
+}
+
+for (;;) {          // Noncompliant {{Add an end condition for this loop}}
+//   ^^
+}
+
+var a;
+var b;
+for (;a > b;) {     // OK
+ a = true;
```
vilchik-elena: @inverno indentation is broken here and on kine 21

javascript-checks/src/test/resources/checks/loopsShouldNotBeInfinite.js
```diff
@@ -0,0 +1,171 @@
...
+    }
+}
+
+while(true) {           // OK
+    while(true) {
+        throw new String("Some Object");
+    }
+}
+
+function someFunction() {
```
vilchik-elena: @inverno what do we test?

javascript-checks/src/test/resources/checks/loopsShouldNotBeInfinite.js
```diff
@@ -0,0 +1,171 @@
...
+        }
+    }
+}
+
+
+do {
+    foo();
+} while (true);         // Noncompliant
+
+var j=0,k=3;
```
vilchik-elena: @inverno why `k`?

javascript-checks/src/test/resources/checks/loopsShouldNotBeInfinite.js
```diff
@@ -0,0 +1,171 @@
...
+    foo();
+}
+
+var l = 45;
+while (l = bar()) {      // OK
+    boo();
+}
+
+var m = 100;
+for (var n=0; n < m; n--) {     // OK False negative
```
vilchik-elena: @inverno I think we can target this case for https://jira.sonarsource.com/browse/RSPEC-2704 and not consider it as FN for infinite-loop rule

javascript-checks/src/test/resources/checks/loopsShouldNotBeInfinite.js
```diff
@@ -0,0 +1,171 @@
...
+    m++;
+}
+
+var o = 5;
+while (o < 10) {        // OK False negative
+    o = o % 5;
+}
+
+while(false) {          // Noncompliant In fact this should be detected by a dead code check.
```
vilchik-elena: @inverno We should not raise issue here, its FP (there is no infinite loop, whatever the message).

javascript-checks/src/test/resources/checks/loopsShouldNotBeInfinite.js
inverno: I tried as much as possible not to declare twice the same single-letter variable in the test file. K comes after J, hence here you find K!

javascript-checks/src/test/resources/checks/loopsShouldNotBeInfinite.js
```diff
@@ -0,0 +1,171 @@
...
+}
+
+var q = 0
+while(q < 3) {          // OK False negative
```
vilchik-elena: @inverno Why we don't raise issue here?

javascript-checks/src/test/resources/checks/loopsShouldNotBeInfinite.js
```diff
@@ -0,0 +1,171 @@
...
+}
+
+function * generator() {
+    while(true) {               // Noncompliant False POSITIVE!! Needs evolution in CFG to support yield.
+        foo();
+        yield;
+    }
+}
+/*
+*/
```
vilchik-elena: @inverno I suggest to work more on this file:
- reorder cases so that there is some logic
- the point of some cases is hard to understand, it can be improved with variables renamings, comments

javascript-checks/src/test/resources/checks/loopsShouldNotBeInfinite.js
vilchik-elena: @inverno I was meaning that `k` is not used, what is the point of having it?

javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S2189.html
vilchik-elena: @inverno I would suggest to drop this block, it's not common to put limitations of the rule in the description. FYI you might create paragraph `Exceptions` listing cases which we ignore for some reason.

javascript-checks/src/test/resources/checks/loopsShouldNotBeInfinite.js
inverno: Just making sure that variable declarations are accounted for even when they are multiple. I had a ruling failure that I feared was related to symbols not being properly identified in that case. It turned up not to be. I can probably delete this whole case.

javascript-checks/src/test/resources/checks/loopsShouldNotBeInfinite.js
inverno: Probably not, but when I wrote this I was not sure how robust we were against stupid things like overlapping column indexes in PreciseLocation.

javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S2189.html
inverno: Ok, I'll put them in an Exceptions paragraph.

sonar-javascript-plugin/pom.xml
```diff
@@ -74,6 +74,7 @@
 
       <plugin>
         <artifactId>maven-shade-plugin</artifactId>
+          <version>2.4.3</version>
```
vilchik-elena: @inverno indentation

javascript-checks/src/main/java/org/sonar/javascript/checks/LoopsShouldNotBeInfiniteCheck.java
```diff
@@ -0,0 +1,160 @@
...
+
+  @Override
+  public void visitDoWhileStatement(DoWhileStatementTree tree) {
+    visitConditionalIterationStatement(tree, (JavaScriptTree) tree.condition());
+    super.visitDoWhileStatement(tree);
+  }
+
+  private void visitConditionalIterationStatement(IterationStatementTree tree, JavaScriptTree condition) {
+    ControlFlowGraph controlFlow = CheckUtils.buildControlFlowGraph(tree);
+    Map<Tree, CfgBlock> treesOfFlowGraph = treesToBlocs(controlFlow);
```
vilchik-elena: @inverno `controlFlow` or `FlowGraph`? Choose one for referencing cgf :)

javascript-checks/src/main/java/org/sonar/javascript/checks/LoopsShouldNotBeInfiniteCheck.java
```diff
@@ -0,0 +1,160 @@
...
+      return loopBlocks.stream().filter(this::isAnExit).collect(Collectors.toSet());
+    }
+
+    private boolean isAnExit(CfgBlock cfgBlock) {
+      return cfgBlock.successors().stream().anyMatch(successor -> !loopBlocks.contains(successor));
+    }
+
+  }
+
+  private static Map<Tree, CfgBlock> treesToBlocs(ControlFlowGraph graph) {
```
vilchik-elena: @inverno `k` is missed in `treesToBlocs`

javascript-checks/src/main/java/org/sonar/javascript/checks/LoopsShouldNotBeInfiniteCheck.java
vilchik-elena: @inverno You are building this map twice: here and in `Loop`

inverno: Yes. I was not convinced that the performance gain was worth the noice of an extra parameter. I'll think what can be done, originally it was calculated and stored in Loop, maybe I'll move back to something like that.

javascript-checks/src/main/java/org/sonar/javascript/checks/LoopsShouldNotBeInfiniteCheck.java
```diff
@@ -0,0 +1,160 @@
...
+      return iterationStatement.kin().map(treesOfFlowGraph::get).filter(Objects::nonNull).collect(Collectors.toSet());
+    }
+
+    public Set<CfgBlock> jumpingExits() {
+      Set<CfgBlock> exits = exits();
+      branchBlock.ifPresent(exits::remove);
+      return exits;
+    }
+
+    public final Set<CfgBlock> exits() {
```
vilchik-elena: @inverno Why public?

inverno: It's from when I was planning to move the class together with the other CFG related classes.

javascript-frontend/src/main/java/org/sonar/javascript/tree/impl/JavaScriptTree.java
vilchik-elena: @inverno Why having `Stream` and not `Set` as return?

inverno: Stream is more powerful and less constrained (it does not imply you have to have all of the elements up front) and thus I prefer it in a public interface on a generic data structure. I believe it's the same reason why the tree returns a children Iterator, rather than a Set. Besides, this way all navigation methods (descendants, symbols and symbolsUsages) all return the same data structure. I would have changed children Iterator too, but it was a bit too much.

sonar-javascript-plugin/pom.xml
inverno: I left that indentation on purpose. It needs to disappear as soon as the pom is updated. I'll instead indent it correctly and add a comment.

javascript-checks/src/main/java/org/sonar/javascript/checks/LoopsShouldNotBeInfiniteCheck.java
inverno: I remember now why I removed it from the Loop. The cohesion had moved to 2 because of it. It has no correlation with the rest of Loop's methods. As things stand I would rather keep it the way it is.

javascript-checks/src/main/java/org/sonar/javascript/checks/LoopsShouldNotBeInfiniteCheck.java
inverno: Moved to default access.

javascript-checks/src/main/java/org/sonar/javascript/checks/LoopsShouldNotBeInfiniteCheck.java
```diff
@@ -0,0 +1,220 @@
...
+  private boolean isInfiniteLoop(IterationStatementTree tree, @Nullable JavaScriptTree condition) {
+    if (isNeverExecutedLoop(condition)) {
+      return false;
+    }
+    ControlFlowGraph flowGraph = CheckUtils.buildControlFlowGraph(tree);
+    Map<Tree, CfgBlock> treesOfFlowGraph = treesToBlocks(flowGraph);
+    if (isBrokenLoop(tree, flowGraph)) {
+      return false;
+    }
+    if (condition != null && conditionIsUpdated(condition, (JavaScriptTree) tree, treesOfFlowGraph)) {
```
SonarTech: MINOR Replace this if-then-else statement by a single return statement. [rule](https://next.sonarqube.com/sonarqube/coding_rules#rule_key=squid%3AS1126)

javascript-checks/src/main/java/org/sonar/javascript/checks/LoopsShouldNotBeInfiniteCheck.java
```diff
@@ -0,0 +1,220 @@
...
+    super.visitDoWhileStatement(tree);
+  }
+
+  private void visitLoop(IterationStatementTree tree, @Nullable Tree condition) {
+    if (isInfiniteLoop(tree, (JavaScriptTree) condition)) {
+      tree.accept(new LoopIssueCreator());
+    }
+  }
+
+  private boolean isInfiniteLoop(IterationStatementTree tree, @Nullable JavaScriptTree condition) {
```
SonarTech: MAJOR Make "isInfiniteLoop" a "static" method. [rule](https://next.sonarqube.com/sonarqube/coding_rules#rule_key=squid%3AS2325)

SonarTech:
SonarQube analysis reported 2 issues
* MAJOR 1 major
* MINOR 1 minor

vilchik-elena:
Done in https://github.com/SonarSource/sonar-javascript/pull/448

PR #446, opened 2017-01-13
SONAR-8574 Replace usage of api/resources/index by api/measures/component in ITs

julienlancelot:
…nent in ITs

### DIFF
--- a/its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/DefaultProfileTest.java
+++ b/its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/DefaultProfileTest.java
@@ -19,17 +19,18 @@
  */
 package com.sonar.javascript.it.plugin;
 
-import com.google.common.collect.ImmutableMap;
 import com.sonar.orchestrator.Orchestrator;
-import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 import org.junit.ClassRule;
 import org.junit.Test;
-import org.sonar.wsclient.SonarClient;
-import org.sonar.wsclient.jsonsimple.JSONValue;
+import org.sonarqube.ws.QualityProfiles.SearchWsResponse.QualityProfile;
+import org.sonarqube.ws.Rules;
+import org.sonarqube.ws.client.rule.SearchWsRequest;
 
+import static com.sonar.javascript.it.plugin.Tests.newWsClient;
+import static java.util.Collections.singletonList;
 import static org.assertj.core.api.Assertions.assertThat;
 
 public class DefaultProfileTest {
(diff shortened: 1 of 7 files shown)


its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/DefaultProfileTest.java
```diff
@@ -40,40 +38,14 @@
...
-  private static List<String> ruleActivations(Map actives, Map<String, String> jsProfileNameById, String ruleKey) {
-    List<String> profileNames = new ArrayList<>();
-    List<Map> activations = (List) actives.get(ruleKey);
-    for (Map activation : activations) {
-      String profileId = (String) activation.get("qProfile");
-      profileNames.add(jsProfileNameById.get(profileId));
-    }
-    return profileNames;
+    assertThat(actives.get("common-js:InsufficientCommentDensity").getActiveListList()).isEmpty();
+    assertThat(actives.get("common-js:DuplicatedBlocks").getActiveListList()).hasSize(1);
```
vilchik-elena: @julienlancelot I hardly understand what is happening here, but can we add assert on profile? (unless we already doing that)

julienlancelot: @vilchik-elena Here the goal is to check that the rule "common-js:DuplicatedBlocks" is activated on one profile.
It's the using the WS api/rules/search that allow to returned the list of profiles where a rule is activated (using the parameter fields to "actives").
Does it answer your question ?

its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/MetricsTest.java
```diff
@@ -51,92 +48,90 @@ public static void prepare() {
...
-    assertThat(getFileMeasure("functions").getIntValue()).isEqualTo(5);
-    assertThat(getFileMeasure("files").getIntValue()).isEqualTo(1);
+    assertThat(getFileMeasureAsDouble("functions")).isEqualTo(5);
+    assertThat(getFileMeasureAsDouble("files")).isEqualTo(1);
     // Documentation
-    assertThat(getFileMeasure("comment_lines").getIntValue()).isEqualTo(1);
-    assertThat(getFileMeasure("commented_out_code_lines")).isNull();
-    assertThat(getFileMeasure("comment_lines_density").getValue()).isEqualTo(4.8);
+    assertThat(getFileMeasureAsDouble("comment_lines")).isEqualTo(1);
+    assertThat(getFileMeasureAsDouble("commented_out_code_lines")).isZero();
```
vilchik-elena: @julienlancelot Why this change?

its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/MetricsTest.java
```diff
@@ -51,92 +48,90 @@ public static void prepare() {
...
+    assertThat(getFileMeasureAsDouble("comment_lines_density")).isEqualTo(4.8);
     // Complexity
-    assertThat(getFileMeasure("file_complexity_distribution")).isNull();
+    assertThat(getFileMeasureAsDouble("file_complexity_distribution")).isNull();
     // Duplication
-    assertThat(getFileMeasure("duplicated_lines")).isNull();
-    assertThat(getFileMeasure("duplicated_blocks")).isNull();
-    assertThat(getFileMeasure("duplicated_files")).isNull();
-    assertThat(getFileMeasure("duplicated_lines_density")).isNull();
+    assertThat(getFileMeasureAsDouble("duplicated_lines")).isZero();
```
vilchik-elena: @julienlancelot And here as well null->zero

its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/DefaultProfileTest.java
vilchik-elena: @julienlancelot Ok, then with new implementation we asserting that rule is activated at ONE profile (whatever it is), but previously we were asserting that it's activated on "Sonar way" profile. 
So how can we check on which profile rule is activated?
Thanks!

its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/MetricsTest.java
julienlancelot: Because 'commented_out_code_lines' is using best values => When there's no measure, it's returning the best value, which is 0 for this measure.
Old WS api/resources/index was not handling best values, by new one is handling them.

its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/MetricsTest.java
julienlancelot: Same answer, duplicated_lines is using best values (1).

its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/DefaultProfileTest.java
julienlancelot: Good catch ! For sure it's possible to check that the rule is activated on the Sonar way profile.
I'm fixing it.

julienlancelot: @vilchik-elena It should be better now

vilchik-elena:
LGTM, Thanks!

PR #447, opened 2017-01-16
SONARJS-860 Rule: Ternary operators should not be nested

ivandalbosco:


javascript-checks/src/main/java/org/sonar/javascript/checks/NestedConditionalOperatorsCheck.java
```diff
@@ -0,0 +1,79 @@
...
+
+@Rule(key = "S3358")
+public class NestedConditionalOperatorsCheck extends DoubleDispatchVisitorCheck {
+
+  private static final String MESSAGE = "Extract this nested ternary operation into an independent statement.";
+
+  private int nestingLevel;
+
+  @Override
+  public List<Issue> scanFile(TreeVisitorContext context) {
```
pynicolas: @ivandalbosco I would prefer to override `visitScript` rather than `scanFile`.

javascript-checks/src/main/java/org/sonar/javascript/checks/NestedConditionalOperatorsCheck.java
```diff
@@ -0,0 +1,79 @@
...
+    while (!parent.is(Kind.CONDITIONAL_EXPRESSION)) {
+      if (breaksRecursion(parent)) {
+        return true;
+      }
+      parent = ((JavaScriptTree) parent).getParent();
+    }
+    return false;
+  }
+
+  private static boolean breaksRecursion(Tree tree) {
```
pynicolas: @ivandalbosco This method name looks strange to me. Why not `breaksNesting` to be consistent with `isNestingBroken`?

javascript-checks/src/main/java/org/sonar/javascript/checks/NestedConditionalOperatorsCheck.java
```diff
@@ -0,0 +1,79 @@
...
+   * Returns true if, between the ConditionalExpression and its nesting ConditionExpression,
+   * there is a construct that functionally "disconnects" the two ConditionExpressions. 
+   */
+  private static boolean isNestingBroken(ConditionalExpressionTree conditionalExpression) {
+    Tree parent = ((JavaScriptTree) conditionalExpression).getParent();
+    while (!parent.is(Kind.CONDITIONAL_EXPRESSION)) {
+      if (breaksRecursion(parent)) {
+        return true;
+      }
+      parent = ((JavaScriptTree) parent).getParent();
```
pynicolas: @ivandalbosco `CheckUtils.parent(parent)` would avoid the cast.

javascript-checks/src/main/java/org/sonar/javascript/checks/NestedConditionalOperatorsCheck.java
ivandalbosco: @pynicolas OK

javascript-checks/src/main/java/org/sonar/javascript/checks/NestedConditionalOperatorsCheck.java
ivandalbosco: @pynicolas You're right. This was a remnant a previous version of my class.

javascript-checks/src/main/java/org/sonar/javascript/checks/NestedConditionalOperatorsCheck.java
ivandalbosco: @pynicolas OK, thanks

PR #448, opened 2017-01-18
SONARJS-861 Rule: Loops should not be infinite

vilchik-elena:
New methods in JavaScriptTree : descendants, isAncestorOf


javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S2189.html
```diff
@@ -0,0 +1,60 @@
...
+  while(true) {
+    break outer;
+  }
+}
+
+</pre>
+
+<h2>Known Limitations</h2>
+<ul>
+    <li>False positives: <tt>yield</tt> is used - <a
```
vilchik-elena: @inverno tt->code

PR #454, opened 2017-01-26
Update to sonar.version 6.2

janos-ss:
Just the minimum changes to compile and pass tests.

### DIFF
--- a/sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/JavaScriptSquidSensorTest.java
+++ b/sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/JavaScriptSquidSensorTest.java
@@ -40,12 +40,10 @@ import org.sonar.api.batch.fs.internal.FileMetadata;
 import org.sonar.api.batch.rule.ActiveRules;
 import org.sonar.api.batch.rule.CheckFactory;
 import org.sonar.api.batch.rule.internal.ActiveRulesBuilder;
-import org.sonar.api.batch.sensor.coverage.CoverageType;
 import org.sonar.api.batch.sensor.internal.DefaultSensorDescriptor;
 import org.sonar.api.batch.sensor.internal.SensorContextTester;
 import org.sonar.api.batch.sensor.issue.Issue;
 import org.sonar.api.config.MapSettings;
-import org.sonar.api.config.Settings;
 import org.sonar.api.internal.SonarRuntimeImpl;
 import org.sonar.api.internal.google.common.base.Charsets;
 import org.sonar.api.issue.NoSonarFilter;
@@ -394,7 +392,7 @@ public class JavaScriptSquidSensorTest {
   }
 
   @Test
-  public void sq_greater_6_1_still_honor_it_ut_coverage_reports() throws Exception {
+  public void sq_greater_6_1_still_honor_coverage_reports() throws Exception {
     baseDir = new File("src/test/resources/coverage");
     context = SensorContextTester.create(baseDir);
     context.setRuntime(SONAR_RUNTIME_6_2);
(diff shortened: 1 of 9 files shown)


sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/lcov/CoverageSensorTest.java
```diff
@@ -179,46 +173,38 @@ public void test_unresolved_path() {
...
   public void test_no_report_path_no_force_zero() {
     context.setSettings(new MapSettings());
     utCoverageSensor.execute(context, linesOfCode, RUN_WITH_SQ_6_1);
-    assertThat(context.lineHits("moduleKey:file1.js", CoverageType.UNIT, 1)).isNull();
+    assertThat(context.lineHits("moduleKey:file1.js", 1)).isNull();
 
     context.setSettings(new MapSettings().setProperty(JavaScriptPlugin.LCOV_UT_REPORT_PATH, UT_LCOV));
     itCoverageSensor.execute(context, linesOfCode, RUN_WITH_SQ_6_1);
-    assertThat(context.lineHits("moduleKey:file1.js", CoverageType.UNIT, 1)).isNull();
-    overallCoverageSensor.execute(context, linesOfCode, RUN_WITH_SQ_6_1);
```
pynicolas: @janos-ss Why was this line removed?

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/JavaScriptSquidSensorTest.java
```diff
@@ -406,9 +404,8 @@ public void sq_greater_6_1_still_honor_it_ut_coverage_reports() throws Exception
     inputFile("file1.js");
     createSensor().execute(context);
 
-    // note that these coverages will be aggregated on platform side.
-    assertThat(context.lineHits("moduleKey:file1.js", CoverageType.UNIT, 1)).isEqualTo(2);
-    assertThat(context.lineHits("moduleKey:file1.js", CoverageType.IT, 1)).isEqualTo(1);
+    // note: as of SQ 6.2, there is no more distinction of coverage types
+    assertThat(context.lineHits("moduleKey:file1.js", 1)).isEqualTo(6);
```
janos-ss: @pynicolas `(2 + 1) * 2`, where the `* 2` is caused by the `OverallCoverageSensor`. How about we delete that class?

vilchik-elena: @janos-ss Actually that's a bug in our plugin, it should be `3`, but I tend to think it's not critical at all

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/lcov/CoverageSensorTest.java
```diff
@@ -96,46 +96,8 @@ public void report_not_found() throws Exception {
     assertThat(context.lineHits("moduleKey:file1.js", 1)).isNull();
   }
 
-  // note: as of SQ 6.2, there is no more distinction of coverage types
   @Test
-  public void test_ut_coverage() {
```
pynicolas: @janos-ss I think these tests are still relevant. I would keep them.

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/JavaScriptSquidSensorTest.java
pynicolas: @janos-ss I see. Let's keep it like that. We may change that behavior in the future.

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/lcov/CoverageSensorTest.java
janos-ss: @pynicolas yeah I concluded that too, added them back now

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/lcov/CoverageSensorTest.java
janos-ss: @pynicolas Partly in anticipation of removing the `OverallCoverageSensor`, and partly because it made the next assertion about line hits in UTs confusing, due to the doubled counting we've seen in another test case too.

I improved that now, by changing the order of executions and assertions of coverage sensors, to this order: IT, UT, overall, instead of IT, overall, UT.

pynicolas:
Thanks @janos-ss !

PR #455, opened 2017-01-26
SONARJS-891 Fix Wrong number of jumping exits in composite conditions loops

inverno:


its/ruling/src/test/expected/javascript-S2189.json
```diff
@@ -1,12 +1,12 @@
...
-],
-'project:google-closure-library/iter/iter.js':[
-164,
-],
+  'project:ecmascript6/di.js/visualize/lib/d3.v3.js': [
+    8711,
+    8716,
+  ],
+  'project:ecmascript6/yield-expression.js': [
+    4,
```
ivandalbosco: @inverno Don't you think this is a FP? Infinite generators such as this Fibonacci sequence might be excluded from the rule. The exclusion criterion might be the presence of a "yield" instruction in the loop.

its/ruling/src/test/expected/javascript-S2189.json
```diff
@@ -1,12 +1,12 @@
...
+  ],
+  'project:google-closure-library/iter/iter.js': [
+    164,
```
ivandalbosco: @inverno OK, this is an infinite loop with no jump exit. However, it is enclosed in try/catch block. Do you really think we should raise an issue here?

ivandalbosco: @inverno Sorry I just realized that all of these issues are simultaneously removed and added, so they are not relevant to this commit. However, please just take a look at my above comments.

its/ruling/src/test/expected/javascript-S2189.json
inverno: Yes, it is a false positive. I added an improvement for supporting Yield in the CFG. See : https://jira.sonarsource.com/browse/SONARJS-890

javascript-checks/src/main/java/org/sonar/javascript/checks/LoopsShouldNotBeInfiniteCheck.java
```diff
@@ -155,7 +164,7 @@ private static boolean isWriteUsage(Usage usage) {
 
     Set<CfgBlock> jumpingExits() {
       Set<CfgBlock> exits = exits();
-      branchBlock.ifPresent(exits::remove);
+      branchBlocks.forEach(exits::remove);
       return exits;
     }
```
inverno: Done

javascript-checks/src/main/java/org/sonar/javascript/checks/LoopsShouldNotBeInfiniteCheck.java
```diff
@@ -131,22 +130,32 @@ private static boolean isWriteUsage(Usage usage) {
...
 
-    private static Optional<CfgBranchingBlock> findRootBranchingBlock(ControlFlowGraph flowGraph, IterationStatementTree iteration) {
+    private static Set<CfgBranchingBlock> findRootBranchingBlocks(ControlFlowGraph flowGraph, IterationStatementTree iteration, @Nullable JavaScriptTree condition) {
       return flowGraph.blocks().stream()
         .filter(cfgBlock -> cfgBlock instanceof CfgBranchingBlock)
         .map(cfgBlock -> (CfgBranchingBlock) cfgBlock)
-        .filter(cfgBranchingBlock -> iteration.equals(cfgBranchingBlock.branchingTree()))
-        .findAny();
+        .filter(cfgBranchingBlock -> {
```
inverno: Done

PR #460, opened 2017-01-30
SONARJS-887 Rule: super() must be invoked appropriately

ivandalbosco:


javascript-checks/src/main/java/org/sonar/javascript/checks/SuperInvocationCheck.java
```diff
@@ -0,0 +1,343 @@
...
+      int line = superTree.getFirstToken().line();
+      int column = superTree.getFirstToken().column();
+      MethodDeclarationTree method = getEnclosingConstructor(superTree);
+      Set<IssueLocation> secondaryLocations = new HashSet<>();
+
+      // get the usages of "super" before super()
+      Set<SuperTreeImpl> superTrees = new SuperDetector().detectIn(method);
+      superTrees.stream()
+        .filter(s -> !s.getParent().is(Kind.CALL_EXPRESSION))
+        .filter(s -> isBefore(s.getFirstToken(), line, column))
```
pynicolas: @ivandalbosco This is certainly a corner case but this approach may lead to FPs, e.g.:
```
class B51 extends A51 {
  constructor() {
    for (let i = 0; i < 1; super.x = 42) {
      super();
      i++;
    }
  }
}
```

javascript-checks/src/main/java/org/sonar/javascript/checks/SuperInvocationCheck.java
```diff
@@ -0,0 +1,343 @@
...
+    }
+  }
+
+  private void checkSuperHasCorrectNumberOfArguments(SuperTreeImpl superTree) {
+    if (isInConstructor(superTree) && !isInBaseClass(getEnclosingConstructor(superTree))) {
+      ClassTree classTree = getEnclosingClass(superTree);
+      ExpressionTree superClassTree = classTree.superClass();
+      // we consider only the simple case "class A extends B", not cases such as "class A extends class {} ..."
+      if (superClassTree.is(Kind.IDENTIFIER_REFERENCE)) {
+        IdentifierTree superClassId = (IdentifierTree) superClassTree;
```
pynicolas: @ivandalbosco The 3 lines below can be replaced with `Symbol superClassSymbol = superClassId.symbol();`

javascript-checks/src/main/java/org/sonar/javascript/checks/SuperInvocationCheck.java
```diff
@@ -0,0 +1,343 @@
...
+      }
+    }
+    return false;
+  }
+
+  private static ClassTree getEnclosingClass(Tree tree) {
+    return (ClassTree) CheckUtils.getFirstAncestor(tree, Kind.CLASS_DECLARATION, Kind.CLASS_EXPRESSION);
+  }
+
+  private void raiseIssue(SuperTreeImpl superTree, SuperTreeImpl firstSuperTree) {
```
pynicolas: @ivandalbosco Why do we have a separate method for that? It's invoked only once and its name does not really say what it does.

javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S3854.html
```diff
@@ -0,0 +1,93 @@
...
+  }
+}
+
+class FilaBrasileiro extends Dog {
+  constructor(name) {
+    super(name);
+  }
+}
+</pre>
+<p>&lt;h2&gt;Known Limitations&lt;/h2&gt;</p>
```
pynicolas: @ivandalbosco That does not render correctly.

javascript-checks/src/main/java/org/sonar/javascript/checks/SuperInvocationCheck.java
```diff
@@ -0,0 +1,343 @@
...
+  private void compareNumberOfArguments(SuperTreeImpl superTree, Symbol superClassSymbol) {
+    Optional<ClassTree> baseClassTree = getDeclarationTree(superClassSymbol);
+    if (baseClassTree.isPresent()) {
+      Optional<MethodDeclarationTree> baseClassConstructor = getConstructor(baseClassTree.get());
+      if (baseClassConstructor.isPresent()) {
+        Integer nbParams = baseClassConstructor.get().parameterList().size();
+        int nbArguments = ((CallExpressionTree) superTree.getParent()).arguments().parameters().size();
+        if (nbArguments != nbParams) {
+          String message = String.format(MESSAGE_SUPER_WITH_CORRECT_NUMBER_OF_ARGUMENTS, nbParams, nbParams == 1 ? "" : "s");
+          addIssue(superTree, message);
```
pynicolas: @ivandalbosco We could add a secondary location on the signature of the constructor of the super class.

javascript-checks/src/main/java/org/sonar/javascript/checks/SuperInvocationCheck.java
```diff
@@ -0,0 +1,343 @@
...
+
+  @Override
+  public void visitFunctionExpression(FunctionExpressionTree tree) {
+    superInvocations.push(new LinkedList<>());
+    super.visitFunctionExpression(tree);
+    superInvocations.pop();
+  }
+
+  private void checkSuperOnlyInvokedInDerivedClassConstructor(SuperTreeImpl superTree) {
+    if (!isInConstructor(superTree) || isInBaseClass(getEnclosingConstructor(superTree))) {
```
ivandalbosco: @pynicolas If my understanding is correct, you'd like me to move up the test on isConstructor() and make it cover the 3 invocations of checkSuperXXX() methods in visitSuper(). But this does not work, as in one case (checkSuperOnlyInvokedInDerivedClassConstructor), the test is negated (with the "!" operator).

javascript-checks/src/main/java/org/sonar/javascript/checks/SuperInvocationCheck.java
```diff
@@ -0,0 +1,343 @@
...
+      addIssue(superTree, MESSAGE_SUPER_ONLY_IN_DERIVED_CLASS_CONSTRUCTOR);
+    }
+  }
+
+  private void checkSuperInvokedInAnyDerivedClassConstructor(MethodDeclarationTree method) {
+    if (isConstructor(method) && !isInBaseClass(method) && !isInDummyDerivedClass(method)) {
+      Set<SuperTreeImpl> superTrees = new SuperDetector().detectIn(method);
+      if (!superTrees.stream().anyMatch(s -> s.getParent().is(Kind.CALL_EXPRESSION))) {
```
ivandalbosco: @pynicolas You're right, thanks.

javascript-checks/src/main/java/org/sonar/javascript/checks/SuperInvocationCheck.java
ivandalbosco: @pynicolas OK, done,

javascript-checks/src/main/java/org/sonar/javascript/checks/SuperInvocationCheck.java
ivandalbosco: @pynicolas I think I will live with that FP threat above my head. I assume that a constructor is expected to have a simple implementation, where super() is invoked almost immediately.

javascript-checks/src/main/java/org/sonar/javascript/checks/SuperInvocationCheck.java
ivandalbosco: @pynicolas Great, thanks.

javascript-checks/src/main/java/org/sonar/javascript/checks/SuperInvocationCheck.java
```diff
@@ -0,0 +1,343 @@
...
+  /**
+   * Returns true if the class of the specified method extends "null", else returns false.
+   * It is assumed that the class is a derived class.
+   */
+  private static boolean isInDummyDerivedClass(MethodDeclarationTree method) {
+    ClassTree classTree = getEnclosingClass(method);
+    return classTree.superClass().is(Kind.NULL_LITERAL);
+  }
+
+  private static MethodDeclarationTree getEnclosingConstructor(Tree tree) {
```
ivandalbosco: @pynicolas OK

javascript-checks/src/main/java/org/sonar/javascript/checks/SuperInvocationCheck.java
```diff
@@ -0,0 +1,343 @@
...
+    return null;
+  }
+
+  private static boolean isConstructor(FunctionTree tree) {
+    if (tree.is(Kind.METHOD)) {
+      MethodDeclarationTree constructor = (MethodDeclarationTree) tree;
+      Tree nameTree = constructor.name();
+      if (nameTree.is(Kind.IDENTIFIER_NAME)) {
+        String name = ((IdentifierTree) nameTree).name();
+        if ("constructor".equals(name)) {
```
ivandalbosco: @pynicolas OK

javascript-checks/src/main/java/org/sonar/javascript/checks/SuperInvocationCheck.java
ivandalbosco: @pynicolas You're right. Done.

javascript-checks/src/main/java/org/sonar/javascript/checks/utils/CheckUtils.java
```diff
@@ -106,5 +106,17 @@ public static boolean isDescendant(Tree tree, Tree potentialParent) {
     }
     return false;
   }
+
+  public static Tree getFirstAncestor(Tree tree, Kind... kind) {
```
ivandalbosco: @pynicolas Well, I don't know. Annotations Nullable, CheckForNull, etc, are currently such a mess that they almost make more noise than help. They are in JSR 305, not in Java8. Looks more like a Findbugs stuff that like a Java spec. The Javadoc contract itself is hard to find on Google. I suggest we first decide what our group's policy is.

javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S3854.html
ivandalbosco: @pynicolas Thanks, fixed.

javascript-checks/src/main/java/org/sonar/javascript/checks/SuperInvocationCheck.java
ivandalbosco: @pynicolas OK. I didn't find how to get the signature, so I defined the secondary location on the parameter list of the super constructor.

PR #461, opened 2017-01-30
SONARJS-892 ReferenceErrorCheck should raise an issue when a let variable is used before it's declared

pynicolas:


javascript-frontend/src/test/resources/ast/resolve/usage.js
```diff
@@ -67,6 +67,12 @@ for(let i = i + 1; i < 20; i++) {}
 var j = 10;
 for (var j = j + 1; j < 20; j++) {}
 
+if (true) {
```
vilchik-elena: @pynicolas what about adding such case:
```
var x;
if (condition) {
  x;
  let x = 42;
}
```

pynicolas: @vilchik-elena done in 4121506

PR #464, opened 2017-01-31
Refresh of the rule metadata

ivandalbosco:


javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S104.json
```diff
@@ -1,5 +1,5 @@
 {
-  "title": "Files should not have too many lines",
+  "title": "Files should not have too many lines of code",
```
vilchik-elena: @ivandalbosco We should not include this change before https://jira.sonarsource.com/browse/SONARJS-876 is done

javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S138.json
```diff
@@ -1,5 +1,5 @@
 {
-  "title": "Functions should not have too many lines",
+  "title": "Functions should not have too many lines of code",
```
vilchik-elena: @ivandalbosco We should not include this change before https://jira.sonarsource.com/browse/SONARJS-876 is done

javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S104.json
ivandalbosco: @vilchik-elena OK

javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S138.json
ivandalbosco: @vilchik-elena OK

PR #492, opened 2017-02-17
SONARJS-903 With SQ < 6.2 use executable_lines_data for zero coverage

vilchik-elena:


javascript-frontend/src/main/java/org/sonar/javascript/metrics/MetricsVisitor.java
```diff
@@ -171,8 +170,11 @@ private void saveLineMetrics(TreeVisitorContext context) {
       fileLinesContext.setIntValue(CoreMetrics.NCLOC_DATA_KEY, line, isCodeLine);
       fileLinesContext.setIntValue(CoreMetrics.COMMENT_LINES_DATA_KEY, line, commentLines.contains(line) ? 1 : 0);
     }
+
+    Set<Integer> executableLines = new ExecutableLineVisitor(context.getTopTree()).getExecutableLines();
```
inverno: IMO we should avoid having this visitor run all its logic running in the constructor. But I understand it was not the scope of the change.

PR #510, opened 2017-03-15
SONARJS-936 Improve NonEmptyCaseWithoutBreak

vilchik-elena:
Ignore when ends with comment that fall-through is intentional


its/ruling/src/test/expected/javascript-NonEmptyCaseWithoutBreak.json
```diff
@@ -199,7 +82,6 @@
 'project:google-closure-library/crypt/hash32.js':[
 112,
 113,
-114,
```
ivandalbosco: @vilchik-elena This is a FP or, at the very least, a noisy issue. Could we additionally check for some comment patterns? For example, if words like "fall*" come out in the comments (possibly above the switch), as in this case, we don't raise issues.

javascript-checks/src/main/java/org/sonar/javascript/checks/NonEmptyCaseWithoutBreakCheck.java
```diff
@@ -74,7 +74,10 @@ public void visitSwitchStatement(SwitchStatementTree tree) {
           firstBlockInClauseBody = caseClauseBlocksByTree.get(lastCaseClause).falseSuccessor();
         }
 
-        if (canBeFallenInto(switchClause, firstBlockInClauseBody, caseExpressions) || hasOnlyEmptyStatements(previousClauseWithStatement)) {
+        boolean canBeFallenInto = canBeFallenInto(switchClause, firstBlockInClauseBody, caseExpressions);
+        boolean hasOnlyEmptyStatements = hasOnlyEmptyStatements(previousClauseWithStatement);
+
+        if ((canBeFallenInto || hasOnlyEmptyStatements) && !hasCommentFallThroughIntentional(previousClauseWithStatement, tree)) {
```
ivandalbosco: @vilchik-elena Generally speaking we cannot avoid to have FPs or noisy issues. However, could we reduce the noise in case of FP? For example, by raising one issue at most per switch. We could have the primary location on the switch and the secondary locations on the faulty cases.

PR #511, opened 2017-03-15
SONARJS-944 Improve OneStatementPerLineCheck: ignore single statement without block

vilchik-elena:


javascript-checks/src/test/resources/checks/oneStatementPerLine.js
```diff
@@ -35,3 +35,17 @@ function sayHello() {
...
+
+  if (a) doSomething();
+  while (a) doSomething();
+  do doSomething(); while(a);
+  for(x in obj) doSomething();
+  for(x of obj) doSomething();
+  for(;;) doSomething();
+
+  if (a) { doSomething(); } // Noncompliant
+  if (a) doSomething(); doSomethingElse(); // Noncompliant
```
ivandalbosco: @vilchik-elena The following yields a FN:
```
if (a
  && b) doSomething(); doSomethingElse();     // I would expect an issue on doSomethingElse()
```
I understand this is not part of this ticket, but I think this rule is a very useful one, so we should strive to avoid FNs.

ivandalbosco:
@vilchik-elena Shouldn't we complete the Exception section of the RSPEC?

vilchik-elena:
@ivandalbosco indeed, thanks!

vilchik-elena:
done

javascript-checks/src/test/resources/checks/oneStatementPerLine.js
vilchik-elena: @ivandalbosco Ok, i think i can fix it now

PR #514, opened 2017-03-16
SONARJS-947 Improve PrimitiveWrappersCheck: ignore when argument of same type

vilchik-elena:


javascript-checks/src/main/java/org/sonar/javascript/checks/PrimitiveWrappersCheck.java
```diff
@@ -19,26 +19,46 @@
...
 import org.sonar.plugins.javascript.api.visitors.DoubleDispatchVisitorCheck;
 
 @Rule(key = "PrimitiveWrappers")
 public class PrimitiveWrappersCheck extends DoubleDispatchVisitorCheck {
 
   private static final String MESSAGE = "Use a literal value for this.";
-  private static final Set<String> WRAPPERS = ImmutableSet.of("Boolean", "Number", "String");
+
+  private static final Map<String, Kind> ALLOWED_ARGUMENT_PER_WRAPPER = ImmutableMap.of(
+    "Boolean", Kind.BOOLEAN_LITERAL,
```
inverno: I would only allow `true` to be passed to `new Boolean`, because if you do `new Boolean(false)` and then you use the boolean in a condition it evaluates to `true`, which is very misleading : `if(new Boolean(false)) alert('this gets printed!');`

javascript-checks/src/test/resources/checks/primitiveWrappers.js
```diff
@@ -1,17 +1,30 @@
-function sayHello() {
-  var x = new Boolean(false); // Noncompliant {{Use a literal value for this.}}
-//        ^^^^^^^^^^^^^^^^^^
-  if (x) {
-    alert('hi');  // Shows 'hi'.
-  }
+function sayHello(y) {
```
inverno: Wouldn't it be safer to not raise an issue if we don't know the type of `y`?

javascript-checks/src/main/java/org/sonar/javascript/checks/PrimitiveWrappersCheck.java
vilchik-elena: ok

PR #515, opened 2017-03-16
SONARJS-946 Improve ImmediatelyReturnedVariableCheck: ignore if variable is used

vilchik-elena:


javascript-checks/src/test/resources/checks/ImmediatelyReturnedVariable.js
```diff
@@ -118,3 +118,11 @@ var arrow_function_ok = (a, b) => {
 }
 
 var arrow_function_no_block = (a, b) =>  a + b;
+
```
ivandalbosco: @vilchik-elena Your implementation is very robust! How about adding a unit test with a wrapping array:
```
function variable_is_used() {
  const bar = [             // OK
     {
       doSomethingElse(p) { },
       doSomething() { bar.doSomethingElse(1); }
     },
     "hello" ];
  return bar;
}
```

ivandalbosco: @vilchik-elena I also suggest a unit test with the usage of a homonymous local variable "bar":
```
function homonymous_is_used() {
  const bar = {           // OK
     doSomethingElse(p) { var bar = 2; return p + bar; },
     doSomething() { return doSomethingElse(1); }
  };
  return bar;
}
```

its/ruling/src/test/expected/javascript-S1488.json
```diff
@@ -190,7 +156,6 @@
 'project:dojo-1.8.1/dojox/jq.js.uncompressed.js':[
 706,
 785,
-1755,
```
ivandalbosco: @elena Isn't this is new FN?
```
var dfd = dojo.io.script.get(args);
return dfd;
```

its/ruling/src/test/expected/javascript-S1488.json
```diff
@@ -273,9 +235,6 @@
 'project:ecmascript6/sonar-web/src/main/js/apps/coding-rules/facets/status-facet.js':[
 9,
 ],
-'project:ecmascript6/watchtower.js/test/watchgroup.spec.js':[
-593,
```
ivandalbosco: @vilchik-elena New FN?

its/ruling/src/test/expected/javascript-S1488.json
vilchik-elena: @ivandalbosco fixed

its/ruling/src/test/expected/javascript-S1488.json
vilchik-elena: @ivandalbosco same nature as previous one. Thanks for noticing!

javascript-checks/src/test/resources/checks/ImmediatelyReturnedVariable.js
vilchik-elena: @ivandalbosco I added second one, but first one is IMO redundant

PR #518, opened 2017-03-17
SONARJS-928 Improve BitwiseOperatorsCheck: detect suspicious cases

vilchik-elena:


javascript-checks/src/main/java/org/sonar/javascript/checks/BitwiseOperatorsCheck.java
```diff
@@ -55,15 +73,56 @@
...
+      bitwiseAndOrOperators.stream()
+        .filter(BitwiseOperatorsCheck::insideCondition)
+        .forEach(syntaxToken ->
+          addIssue(syntaxToken, String.format(MESSAGE, syntaxToken.text())));
+    }
+  }
+
+  private static boolean insideCondition(SyntaxToken token) {
+    Tree treeWithCondition = CheckUtils.getFirstAncestor(token, Kind.IF_STATEMENT, Kind.WHILE_STATEMENT, Kind.DO_WHILE_STATEMENT, Kind.FOR_STATEMENT, Kind.CONDITIONAL_EXPRESSION);
+
```
inverno: If would rather check for a null treeWithCondition here and return false rather than passing it into the `condition` method as a nullable.

javascript-checks/src/main/java/org/sonar/javascript/checks/BitwiseOperatorsCheck.java
```diff
@@ -20,20 +20,38 @@
...
+import org.sonar.plugins.javascript.api.tree.statement.DoWhileStatementTree;
+import org.sonar.plugins.javascript.api.tree.statement.ForStatementTree;
+import org.sonar.plugins.javascript.api.tree.statement.IfStatementTree;
+import org.sonar.plugins.javascript.api.tree.statement.WhileStatementTree;
 import org.sonar.plugins.javascript.api.visitors.SubscriptionVisitorCheck;
 
 @Rule(key = "BitwiseOperators")
 public class BitwiseOperatorsCheck extends SubscriptionVisitorCheck {
 
   private static final String MESSAGE = "Remove the use of \"%s\" operator.";
```
inverno: Considering the change in the nature of the rule we could change the message to `"Suspicious use of \"%s\" operator."`

javascript-checks/src/main/java/org/sonar/javascript/checks/BitwiseOperatorsCheck.java
```diff
@@ -55,15 +73,56 @@
 
   @Override
   public void visitNode(Tree tree) {
-    SyntaxToken operator;
-    if (tree.is(Tree.Kind.BITWISE_COMPLEMENT)) {
-      operator = ((UnaryExpressionTree) tree).operator();
-    } else if (tree instanceof BinaryExpressionTree) {
-      operator = ((BinaryExpressionTree) tree).operator();
+    if (tree.is(Kind.BITWISE_AND, Kind.BITWISE_OR) && !((BinaryExpressionTree) tree).rightOperand().is(Kind.NUMERIC_LITERAL)) {
+      SyntaxToken operator = ((BinaryExpressionTree) tree).operator();
```
inverno: You can inline this operator variable now that it's just added to `bitwiseAndOrOperators`.

javascript-checks/src/main/java/org/sonar/javascript/checks/BitwiseOperatorsCheck.java
```diff
@@ -55,15 +73,56 @@
...
     } else {
-      operator = ((AssignmentExpressionTree) tree).operator();
+      fileContainsBitwiseOperations = true;
     }
-    addIssue(operator, String.format(MESSAGE, operator.text()));
   }
 
+  @Override
+  public void leaveFile(Tree scriptTree) {
+    if (!fileContainsBitwiseOperations && bitwiseAndOrOperators.size() == 1) {
```
inverno: I would try to avoid the set entirely and more explicitly show the concept of the bitwise and/or being alone. I would have a single instance field called `LonelyBitwiseAndOr` which gets set only if an and/or bw operator appears first, and unset (to null) if any other bw operator appears later in the file. Maybe worth a talk when you see this comment?

javascript-checks/src/test/java/org/sonar/javascript/checks/BitwiseOperatorsCheckTest.java
```diff
@@ -25,9 +25,24 @@
...
 public class BitwiseOperatorsCheckTest {
 
+  private static File DIR = new File("src/test/resources/checks/BitwiseOperatorsCheck");
+  private BitwiseOperatorsCheck check = new BitwiseOperatorsCheck();
+
   @Test
-  public void test() {
-    JavaScriptCheckVerifier.verify(new BitwiseOperatorsCheck(), new File("src/test/resources/checks/bitwiseOperators.js"));
+  public void test_ok() {
+    JavaScriptCheckVerifier.verify(check, new File(DIR, "file_ok1.js"));
```
inverno: Could you name these files more meaningfully? I had to open each and every one to find out what they were about :)

javascript-checks/src/main/java/org/sonar/javascript/checks/BitwiseOperatorsCheck.java
vilchik-elena: @inverno updated

javascript-checks/src/main/java/org/sonar/javascript/checks/BitwiseOperatorsCheck.java
vilchik-elena: done

javascript-checks/src/main/java/org/sonar/javascript/checks/BitwiseOperatorsCheck.java
vilchik-elena: @inverno done, thanks!

javascript-checks/src/main/java/org/sonar/javascript/checks/BitwiseOperatorsCheck.java
vilchik-elena: @inverno done

javascript-checks/src/test/java/org/sonar/javascript/checks/BitwiseOperatorsCheckTest.java
vilchik-elena: done:)

PR #528, opened 2017-03-22
SONARJS-925 Update DeadStoreCheck: ignore initializations to basic values

vilchik-elena:


javascript-checks/src/main/java/org/sonar/javascript/checks/DeadStoreCheck.java
```diff
@@ -132,10 +144,39 @@ private void checkUsage(Usage usage, Set<Symbol> liveSymbols, Usages usages) {
...
+    return false;
+  }
+
+  private static boolean isBasicValue(ExpressionTree expression) {
+    if (expression.is(Kind.BOOLEAN_LITERAL, Kind.NUMERIC_LITERAL, Kind.STRING_LITERAL, Kind.NULL_LITERAL)) {
+      return BASIC_VALUES.contains(((LiteralTree) expression).value());
+
+    } else if (expression.is(Kind.UNARY_MINUS)) {
+      ExpressionTree operand = ((UnaryExpressionTree) expression).expression();
+      return operand.is(Kind.NUMERIC_LITERAL) && "1".equals(((LiteralTree) operand).value());
```
inverno: I've refactored this bit to avoid the magic `1` and use `BASIC_VALUES` instead

PR #546, opened 2017-03-31
SONARJS-971 NPE in SuperInvocationCheck

inverno:
constructor inside object literal

### DIFF
--- a/javascript-checks/src/main/java/org/sonar/javascript/checks/SuperInvocationCheck.java
+++ b/javascript-checks/src/main/java/org/sonar/javascript/checks/SuperInvocationCheck.java
@@ -164,7 +164,7 @@ public class SuperInvocationCheck extends DoubleDispatchVisitorCheck {
   }
 
   private void checkSuperHasCorrectNumberOfArguments(SuperTreeImpl superTree) {
-    getEnclosingClass(superTree).ifPresent(classTree -> {
+    getEnclosingType(superTree).ifPresent(classTree -> {
       ExpressionTree superClassTree = classTree.superClass();
       // we consider only the simple case "class A extends B", not cases such as "class A extends class {} ..."
       if (superClassTree.is(Kind.IDENTIFIER_REFERENCE)) {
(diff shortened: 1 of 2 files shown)


javascript-checks/src/main/java/org/sonar/javascript/checks/SuperInvocationCheck.java
```diff
@@ -284,7 +284,12 @@ private static boolean isConstructor(FunctionTree tree) {
   }
 
   private static Optional<ClassTree> getEnclosingClass(Tree tree) {
```
ivandalbosco: @inverno How about renaming this method? Into getEnclosingType()?

javascript-checks/src/test/resources/checks/superInvocation.js
```diff
@@ -295,7 +295,35 @@ class B40 extends A40 {
...
+class B41 extends A40 {
+
   constructor() {
+    super();
+  }
+
+  createObject() {
+    var obj = {
+      constructor() {       // OK
+      }
```
ivandalbosco: @inverno I would add a noncompliant case. With a compliant case, we don't know if the the case is really compliant or simply ignored. Suggestion: add a line with a call to super().

ivandalbosco: @inverno How about adding a test where the constructor is defined separately, e.g.,
```
createObject2() {
    var obj = { };
    obj.constructor = function() {
      super();      // Noncompliant ...
    };
    return obj;
  }

createObject3() {
    var obj = { };
    obj.constructor = function() {
      // OK
    };
    return obj;
  }
```
OK, I know, people already told me I add too many tests...

javascript-checks/src/main/java/org/sonar/javascript/checks/SuperInvocationCheck.java
inverno: Done

javascript-checks/src/test/resources/checks/superInvocation.js
inverno: It's a very good idea, thanks. Done

inverno: I did already use `super()` in the `top_level_object` test. Now I also do it in the new cases you suggested below. I think it should be enough.

PR #548, opened 2017-04-04
Patch 2.21

vilchik-elena:


inverno:
Change already ported to master

PR #555, opened 2017-04-13
SONARJS-927 Declare properties to have a category of "JavaScript"

vilchik-elena:


sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/JavaScriptPluginTest.java
```diff
@@ -38,6 +40,22 @@ public void count_extensions_for_sonarqube_server_5_6() throws Exception {
   }
 
   @Test
+  public void test_properties() throws Exception {
```
inverno: Maybe split this test method into `should_have_javascript_as_category` and `should_contain_all_default_properties` ?

PR #565, opened 2017-04-20
SONARJS-970 Make SonarJS resistant to Exceptions

vilchik-elena:


inverno:
Don't forget to create a Jira issue to update the progress report

Issue #727, opened 2017-09-12
Incorrect detection of async functions


fxOne:
**I want to report a bug.**

**SonarJS version:** 3.1.1<br>
**SonarQube version:** Version 6.5 (build 27846)<br>

**Rule key:** javascript:Semicolon

**Reproducer** 
```
async function outerFunction() {
  async function innerAsyncFunction() {

  }
  function innerSyncFunction() {

  }
  innerSyncFunction();
  await innerAsyncFunction();
}
```
I'm getting the error `Add a semicolon at the end of this statement.` for the `innerAsyncFunction` function.

**Expected behavior**
I expected no error as it makes no sense to add an semicolon at the end of the async function definition. I don't get the ticket for the `outerFunction` or `innerSyncFunction` functions.

vilchik-elena:
Nice catch, @fxOne thanks for reporting!

Looks like we are mistakenly parsing `innerAsyncFunction` as function expression, while it should be function declaration.

inverno:
@fxOne fix is now part of : https://github.com/SonarSource/sonar-javascript/releases/tag/3.2-RC1

PR #732, opened 2017-09-12
Allow usage of the promise returned by async functions

### LINKED ISSUE #731: Allow use of async return value promise
**I want to report a bug.**

**SonarJS version:** 3.1.1<br>

**Rule key:** S3699 aka UseOfEmptyReturnValue

**Reproducer** 
```
var arrowReturnsPromise = async () => {
   var x = () => {return 1}
};
x = arrowReturnsPromise(); // False-positive
```

**Expected behavior**
All async functions return a promise, regardless of if the body has an empty return value, so this rule should allow use of the return value.

NilSet:
closes #731

### DIFF
--- a/javascript-checks/src/main/java/org/sonar/javascript/checks/UseOfEmptyReturnValueCheck.java
+++ b/javascript-checks/src/main/java/org/sonar/javascript/checks/UseOfEmptyReturnValueCheck.java
@@ -29,6 +29,7 @@ import org.sonar.javascript.se.sv.SymbolicValue;
 import org.sonar.javascript.tree.KindSet;
 import org.sonar.plugins.javascript.api.tree.Tree;
 import org.sonar.plugins.javascript.api.tree.Tree.Kind;
+import org.sonar.plugins.javascript.api.tree.declaration.FunctionDeclarationTree;
 import org.sonar.plugins.javascript.api.tree.declaration.FunctionTree;
 import org.sonar.plugins.javascript.api.tree.expression.ArrowFunctionTree;
 import org.sonar.plugins.javascript.api.tree.expression.BinaryExpressionTree;
@@ -139,6 +140,12 @@ public class UseOfEmptyReturnValueCheck extends AbstractAllPathSeCheck<CallExpre
       if (tree.is(Kind.FUNCTION_EXPRESSION)) {
         FunctionExpressionTree functionExpression = (FunctionExpressionTree) tree;
         if (functionExpression.asyncToken() != null) {
+          return true;
+        }
+      }
+      if (tree.is(Kind.FUNCTION_DECLARATION)) {
+        FunctionDeclarationTree functionDeclaration = (FunctionDeclarationTree) tree;
+        if (functionDeclaration.asyncToken() != null) {
           return true;
         }
       }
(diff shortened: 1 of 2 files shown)


javascript-checks/src/main/java/org/sonar/javascript/checks/UseOfEmptyReturnValueCheck.java
```diff
@@ -131,7 +132,17 @@ private static boolean hasReturnValue(FunctionTree tree) {
...
+          return true;
+        }
+      }
+      if (tree.is(Kind.FUNCTION_EXPRESSION)) {
+        FunctionExpressionTree functionExpression = (FunctionExpressionTree) tree;
+        if (functionExpression.asyncToken() != null) {
+          return true;
+        }
       }
+
```
inverno: We should also support `FUNCTION_DECLARATION`, but the test is green

javascript-checks/src/test/resources/checks/UseOfEmptyReturnValue.js
```diff
@@ -31,7 +31,23 @@ function main() {
...
+
+  x = arrowReturnsNothing(); // Noncompliant
+
+  var arrowReturnsPromise = async () => {
+    var x = () => {return 1}
+  };
+
+  x = arrowReturnsPromise(); // OK
+
+  async function statementReturnsPromise() {
```
inverno: For some reason the async keyword is making this declaration appear as a function expression. I need to investigate

javascript-checks/src/main/java/org/sonar/javascript/checks/UseOfEmptyReturnValueCheck.java
NilSet: I noticed something strange about that, but wasn't sure what was going on and the behavior was correct. Sounds related to #727

NilSet: Should I just add that case in for when the parsing is fixed?

NilSet:
Do you think this fix supersedes the exception put in for `await` in #600 ? It should cover the example given in the issue at least.

inverno:
@NilSet you are likely right regarding #600 but I first want to integrate your change. I've fixed the parsing bug, moved your commit to a branch directly on SonarSource/sonar-javascript and added the code to manage implicit promises in function declarations. I'm just waiting for the QA tests to finish then I'll merge and close this PR.

inverno:
@NilSet done. Your code is now on master : https://github.com/SonarSource/sonar-javascript/commit/6ca446253b987336cc763d8717e47f7330f7b333 . Thanks a lot for the great contribution.

NilSet:
@inverno Im glad it got merged in. If you don't mind could you explain the process you used to merge it in? You somehow managed to add your commit on to my branch in my fork, which I didn't think was possible! Also, it somehow lost my second commit to the branch https://github.com/nilset/sonar-javascript/commit/afe7b51f68348b85e07e162bcf0292869f2e42fd which I think may be a slightly better solution.

inverno:
@NilSet Indeed I was surprised that it allowed me to push to your branch, but since it worked I thought it would be more consistent to have your branch aligned with what I finally merged. Regarding your missing commit : I was traveling and didn't have connectivity all the time, you might have added that second commit after I fetched on my laptop and before I rebased on master and force-pushed your branch (to account for the fix in the grammar). The push --force  would account for your dangling commit, which I didn't see until now. My bad.
I'll cherry-pick your commit when I push the revert of #600 later this week.

PR #815, opened 2017-11-01
Update metadata to contain SQ key

vilchik-elena:


javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S1226.html
```diff
@@ -1,29 +1,15 @@
...
-  //...
-  return a;
-}
-</pre>
-<h2>Compliant Solution</h2>
-<pre>
-function MyClass(name) {
-  this.name = name;
-}
+function MyClass(name, strings) {
```
inverno: @vilchik-elena I don't get why @ganncamp removed the compliant section. Do you know what's the reason?

PR #816, opened 2017-11-01
Fix clearing of stack for parenthesised expression in for loop

### LINKED ISSUE #631: Clear stack of parenthesized update expressions
in Data Flow Analyss, `assignment` program points (not an actual object yet, just a line in `ExecutionStack` ) leave the assigned value in the stack for assignment chaining.
When this assignment appears in a `for` statement (like in an update expression) we clean up the stack before proceeding.
If the assignment expression is wrapped into a parenthesized expression `isProducingUnconsumedValue` is wrongly returning false because it's not finding `FOR_STATEMENT` as parent.

Reproducer:
```
  for(let k = 0; ; (bar())) {
    foo();
  }

  for((k = 0); ; bar()) {
    foo();
  }
```

vilchik-elena:
Fixes #631

### DIFF
--- a/javascript-checks/src/main/java/org/sonar/javascript/checks/utils/CheckUtils.java
+++ b/javascript-checks/src/main/java/org/sonar/javascript/checks/utils/CheckUtils.java
@@ -40,8 +40,6 @@ import org.sonar.plugins.javascript.api.tree.statement.BlockTree;
 import org.sonar.plugins.javascript.api.visitors.JavaScriptFile;
 
 public class CheckUtils {
-  
-  public static final String CHECK_DIR = "src/test/resources/checks";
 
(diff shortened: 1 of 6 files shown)


javascript-checks/src/main/java/org/sonar/javascript/checks/utils/CheckUtils.java
```diff
@@ -40,8 +40,6 @@
 import org.sonar.plugins.javascript.api.visitors.JavaScriptFile;
 
 public class CheckUtils {
-  
-  public static final String CHECK_DIR = "src/test/resources/checks";
```
inverno: Nothing against this, but I wonder why.

PR #832, opened 2017-11-09
Extract all bindings in variable declarations

### LINKED ISSUE #828: False positive: let with flow annotation not counted as declaration
**I want to report a bug.**
`let` variable declarations annotated with a flow type are seemingly not tracked as a declaration.

**SonarJS version:** 3.2<br>

**Rule key:** S2703 aka VariableDeclarationWithoutVarCheck

**Reproducer** 
```js
let unflowed;
unflowed = true; // ok

let flowed: boolean;
flowed = true; //false positive
```

**Expected behavior**
Recognize the annotated `let` statement as the declaration of `flowed` instead of assuming the usage of `flowed` is an implicit global declaration.

NilSet:
Fixes #828 and also some other false negatives from ReassignedParameter

### DIFF
--- a/javascript-frontend/src/main/java/org/sonar/javascript/tree/symbols/HoistedSymbolVisitor.java
+++ b/javascript-frontend/src/main/java/org/sonar/javascript/tree/symbols/HoistedSymbolVisitor.java
@@ -284,7 +284,6 @@ public class HoistedSymbolVisitor extends DoubleDispatchVisitor {
       scope = getFunctionScope();
     }
 
-    // todo Consider other BindingElementTree types
     for (BindingElementTree bindingElement : tree.variables()) {
       Symbol.Kind variableKind = getVariableKind(tree);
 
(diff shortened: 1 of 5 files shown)


NilSet:
Does the CI run the integration tests? I was not able to figure out how to run them myself.

inverno:
@NilSet thanks for the PR! It looks like it's just what was needed. See the section "Integration Tests" here : https://github.com/SonarSource/SonarJS/blob/master/docs/DEV.md#testing

javascript-frontend/src/main/java/org/sonar/javascript/tree/symbols/HoistedSymbolVisitor.java
```diff
@@ -293,11 +292,11 @@ private void addUsages(VariableDeclarationTree tree) {
           symbolModel.declareSymbol(identifier.name(), variableKind, scope)
             .addUsage(identifier, Usage.Kind.DECLARATION_WRITE);
         }
-      }
-      if (bindingElement.is(Kind.BINDING_IDENTIFIER)) {
-        IdentifierTree identifierTree = (IdentifierTree) bindingElement;
-        symbolModel.declareSymbol(identifierTree.name(), variableKind, scope)
-          .addUsage(identifierTree, insideForLoopVariable ? Usage.Kind.DECLARATION_WRITE : Usage.Kind.DECLARATION);
+      } else {
+        for (IdentifierTree identifier : bindingElement.bindingIdentifiers()) {
```
vilchik-elena: @NilSet Could you add unit test for this change to symbol model? It probably should be in `SymbolModelImplTest` (add code sample to `src/test/resources/ast/resolve/symbolModel.js` and assert that corresponding symbols exist)

NilSet: Sure

NilSet:
@inverno Had some difficulty getting it to actually run those. Do I need a local instance of sonarqube running? How do I point the tests at it?

NilSet:
Sorry for the delay @vilchik-elena

vilchik-elena:
@NilSet cool! thank you for fix!

inverno:
@NilSet sorry, I forgot to answer your question. You don't need a SQ installed, normally you just follow those instructions and then run the integration test you need from your IDE (at the class level, not a single method). If you would provide the actual error message you get it might be useful to spot some missing point in our documentation.

PR #847, opened 2017-11-29
Parse Flow 'export type' statements

### LINKED ISSUE #833: Parse Flow "export type" statements
relates to #771 

```
export type { Foo, Bar }
export type {foo} from 'bar'
```

vilchik-elena:
Fixes #833

### DIFF
--- a/javascript-frontend/src/main/java/org/sonar/javascript/parser/TreeFactory.java
+++ b/javascript-frontend/src/main/java/org/sonar/javascript/parser/TreeFactory.java
@@ -1214,14 +1214,18 @@ public class TreeFactory {
       asToken = nameSpaceExport.get().first;
       synonymIdentifier = nameSpaceExport.get().second;
     }
-    return new NameSpaceExportDeclarationTreeImpl(exportToken, flowTypeKeywordToken.orNull(), starToken, asToken, synonymIdentifier, fromClause, nullableSemicolonToken(semicolonToken));
+    return new NameSpaceExportDeclarationTreeImpl(
+      exportToken,
+      flowTypeKeywordToken.orNull(),
+      starToken, asToken, synonymIdentifier,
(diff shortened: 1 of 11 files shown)


javascript-frontend/src/main/java/org/sonar/javascript/parser/TreeFactory.java
```diff
@@ -1214,14 +1214,18 @@ public NameSpaceExportDeclarationTree namespaceExportDeclaration(
...
+    return new NameSpaceExportDeclarationTreeImpl(
+      exportToken,
+      flowTypeKeywordToken.orNull(),
+      starToken, asToken, synonymIdentifier,
+      fromClause, nullableSemicolonToken(semicolonToken));
   }
 
-  public ExportClauseTree exportClause(NamedImportExportClauseTree exportList, Optional<FromClauseTree> fromClause, Tree semicolonToken) {
-    if (fromClause.isPresent()) {
-      return new ExportClauseTreeImpl(exportList, fromClause.get(), nullableSemicolonToken(semicolonToken));
```
inverno: @vilchik-elena was this `fromClause.get()` generating parsing errors?

vilchik-elena: @inverno hah? nope!
It's pure refactoring here.
Parsing error was because we were not supporting the syntax. 
https://github.com/SonarSource/SonarJS/pull/847/files#diff-d1591ffd5825fa188dee6bf25e165588R1232 fixed it

inverno: @vilchik-elena of course I know that we were missing the syntax. I only wondered if this `get()` actually throws exceptions when the token is not there or it is just a bit less explicit than `orNull()` but behaves the same. I'll check the source code.

PR #853, opened 2017-12-04
Extend S2137: include other names and deprecate BoundOrAssignedEvalOrArguments

### LINKED ISSUE #687: Extend S2137: include other names and deprecate BoundOrAssignedEvalOrArguments

vilchik-elena:
Fixes #687


javascript-checks/src/main/java/org/sonar/javascript/checks/GlobalsShadowingCheck.java
```diff
@@ -23,28 +23,40 @@
 import java.util.List;
 import org.sonar.check.Rule;
 import org.sonar.plugins.javascript.api.symbols.Symbol;
+import org.sonar.plugins.javascript.api.symbols.Usage;
 import org.sonar.plugins.javascript.api.tree.ScriptTree;
 
 @Rule(key = "S2137")
-public class UndefinedShadowingCheck extends AbstractSymbolNameCheck {
+public class GlobalsShadowingCheck extends AbstractSymbolNameCheck {
```
inverno: @vilchik-elena I think that the hierarchy of `AbstractSymbolNameCheck` doesn't make sense (anymore)

vilchik-elena: @inverno why? there are still 3 rules using this class.

PR #888, opened 2017-12-13
repair FileHeaderCheckTest and CoverageSensorTest for Winrows environment

Nebotov:


Nebotov:
After, I change .gitattributes file to "*.js text eol=lf" and all test success

PR #902, opened 2018-01-29
Fix parsing of JSX member expressions

### LINKED ISSUE #900: FP S3827 when using property of global var in ReactJS
When we used property of variable declared in `sonar.javascript.globals` we still get issue on S3827 . See https://jira.sonarsource.com/browse/SUPPORT-8649

saberduck:
Fixes #900


vilchik-elena:
Manually tested

PR #910, opened 2018-02-01
Fix S3854 when checking argument count on calling super in constructor

### LINKED ISSUE #886: [S3854] Raise wrong "must be invoked with 2 arguments." error on destructuring
**I want to report a bug.**

**SonarJS version:**  3.3.0.5702<br>
**SonarQube version:** 5.6.7<br>

**Rule key:** S3854

**Reproducer** 
```javascript
const defaultOptions = {};
class Foo {
    constructor(name, options = {}) {
        this.name = name;
        this.options = Object.assign({}, defaultOptions, options);
    }
}
class Bar extends Foo {
    constructor(...args) {
        super(...args);
    }
}
function getBar(name, options = {}) {
    return new Bar(name, options);
}
```

**Logs**
```
super() must be invoked with 2 arguments.
Rule: javascript:S3854
```

**Expected behavior**

There should be no error as super will correctly receive the right number of arguments

saberduck:
Fixes #886

### DIFF
--- a/javascript-checks/src/main/java/org/sonar/javascript/checks/SuperInvocationCheck.java
+++ b/javascript-checks/src/main/java/org/sonar/javascript/checks/SuperInvocationCheck.java
@@ -195,17 +195,17 @@ public class SuperInvocationCheck extends DoubleDispatchVisitorCheck {
       return;
     }
     getDeclarationTree(superClassSymbol)
-        .flatMap(SuperInvocationCheck::getConstructor)
-        .ifPresent(baseClassConstructor -> {
-          int paramCount = baseClassConstructor.parameterList().size();
-          long nonDefaultParamCount = baseClassConstructor.parameterList().stream()
-              .filter(b -> !b.is(Kind.INITIALIZED_BINDING_ELEMENT)).count();
-          int nbArguments = superCallTree.argumentClause().arguments().size();
-          if (nbArguments < nonDefaultParamCount || nbArguments > paramCount) {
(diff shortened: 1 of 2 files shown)


javascript-checks/src/main/java/org/sonar/javascript/checks/SuperInvocationCheck.java
```diff
@@ -189,18 +189,23 @@ private void pushSuperInvocation(SuperTree tree) {
...
-          addIssue(superTree.parent(), message).secondary(baseClassConstructor.get().parameterClause());
-        }
-      }
+    CallExpressionTree superCallTree = (CallExpressionTree) superTree.parent();
+    boolean hasSpreadArgument = superCallTree.argumentClause().arguments().stream().anyMatch(t -> t.is(Kind.SPREAD_ELEMENT));
+    if (hasSpreadArgument) {
+      return;
     }
+    getDeclarationTree(superClassSymbol)
+        .flatMap(SuperInvocationCheck::getConstructor)
```
vilchik-elena: @saberduck reformat

javascript-checks/src/test/resources/checks/superInvocation.js
```diff
@@ -345,3 +345,22 @@ var top_level_object = {
...
+        this.options = Object.assign({}, defaultOptions, options);
+    }
+}
+
+class Bar extends Foo {
+    constructor(...args) {
+        super(...args);
+    }
+}
+class Baz extends Foo {
```
vilchik-elena: @saberduck more tests!!

saberduck:
Fixed formatting and added more tests, please have a look @vilchik-elena

vilchik-elena:
@saberduck 👍

PR #914, opened 2018-02-02
Add rule S3863: Imports from the same modules should be merged

### LINKED ISSUE #645: Rule: Imports from the same modules should be merged
Implements [RSPEC-3863](https://jira.sonarsource.com/browse/RSPEC-3863)

vilchik-elena:
Fixes #645


javascript-checks/src/main/java/org/sonar/javascript/checks/SameModuleImportsCheck.java
```diff
@@ -0,0 +1,91 @@
...
+
+    if (flowTypeToken == null) {
+      addImportModule(tree, simpleImportsByModule);
+
+    } else if (flowTypeToken.text().equals("type")) {
+      addImportModule(tree, typeImportsByModule);
+
+    } else {
+      addImportModule(tree, typeOfImportsByModule);
+
```
saberduck: @vilchik-elena  why so much spacing?

vilchik-elena: @saberduck I like empty lines:) If you don't mind I will not reformat

PR #916, opened 2018-02-02
Improve FunctionDefinitionInsideLoop: ignore all safe Array functions

### LINKED ISSUE #915: FunctionDefinitionInsideLoop: do not warn using safe Array functions
**I want to improve a rule, PR will follow soon.**

**SonarJS version:** master branch (`3.2 (build 5506)`, but does not matter)

**Rule key:** javascript:FunctionDefinitionInsideLoop

**Reproducer** 
```
for (const items of something) {
    var item = items.find(x => x.Name === name);
    ...
}
```

**Expected behavior**

`Define this function outside of a loop.` / `Functions should not be defined inside loops` warning is not shown.

**Expected PR content**

I want put the following methods to the safe function list: `find`, `findIndex`, `every`, `some`, `reduce`, `reduceRight`, `sort` (based on https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array)

koczkatamas:
Adds `find`, `findIndex`, `every`, `some`, `reduce`, `reduceRight` and `sort` to the safe list (`ALLOWED_CALLBACKS`).

Fixes #915


vilchik-elena:
Migrated to #919 to fix Ruling Test and formatting

PR #1003, opened 2018-07-10
Support ESLint issues import

### LINKED ISSUE #1001: Support ESLint external issues report
See https://jira.sonarsource.com/browse/MMF-1231

vilchik-elena:
fixes #1001


javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/eslint/react.json
```diff
@@ -0,0 +1,474 @@
...
+  {
+    "key": "react/jsx-closing-tag-location",
+    "name": "Validate closing tag location in JSX (fixable)",
+    "url":
+    "https://github.com/yannickcr/eslint-plugin-react/blob/HEAD/docs/rules/jsx-closing-tag-location.md"
+  },
+  {
+    "key": "react/jsx-curly-spacing",
+    "name":
+    "Enforce or disallow spaces inside of curly braces in JSX attributes and expressions (fixable)",
```
saberduck: Should we have  "fixable" in the title?

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/JsonProfileReader.java
```diff
@@ -32,7 +32,7 @@ private JsonProfileReader() {
   }
 
   public static Set<String> ruleKeys(String pathToFile) {
-    URL profileUrl = JsonProfileReader.class.getResource(pathToFile);
+    URL profileUrl = JsonProfileReader.class.getClassLoader().getResource(pathToFile);
```
saberduck: I don't understand why this change was necessary?

its/plugin/projects/eslint_report/readme
```diff
@@ -0,0 +1,4 @@
+The report was generated with
+npx eslint --ext ts -f json . > report.json
```
saberduck: nice to have: I didn't manage to get the same report as in the example, it would be good to have instructions how to get exactly same behavior. We can even generate two separate reports to test import of multiple reports

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/JsonProfileReader.java
vilchik-elena: looks like `Class#getResource` and `ClassLoader#getResource` don't do the same thing and in analyzer-commons we use `ClassLoader#getResource`

PR #1004, opened 2018-07-10
Fix parsing of imports when identifier matches flow keywords

### LINKED ISSUE #995: E6 Modules named import variable "types" treated as non-existent
**SonarJS version:** 4.1 (build 6085)
**SonarQube version:** Version 7.1 (build 11001)<br>

**Rule key:** javascript:S3827

**Reproducer** 
```
import { types } from 'some/path/here';

export default (state = {}, action) => {
    switch (action.type) {
    case types.SOMETHING:
        return { test: true };
    default:
        return state;
    }
};
```

**Result**
Line 5: "types" does not exist. Change its name or declare it so that its usage doesn't result in a "ReferenceError".

**Expected behavior**
No errors at all.

If i rename the import like the example below Sonarqube finds no errors.
```
import { types as someTypes } from 'some/path/here';
```

saberduck:
Fixes #995

### DIFF
--- a/javascript-frontend/src/main/java/org/sonar/javascript/parser/EcmaScriptLexer.java
+++ b/javascript-frontend/src/main/java/org/sonar/javascript/parser/EcmaScriptLexer.java
@@ -460,7 +460,7 @@ public enum EcmaScriptLexer implements GrammarRuleKey {
 
     // Flow
     b.rule(OPAQUE).is(SPACING, b.token(GenericTokenType.IDENTIFIER, "opaque"));
-    b.rule(TYPE).is(SPACING, b.token(GenericTokenType.IDENTIFIER, "type"));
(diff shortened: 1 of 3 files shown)


javascript-frontend/src/main/java/org/sonar/javascript/parser/EcmaScriptLexer.java
```diff
@@ -460,7 +460,7 @@ private static void lexical(LexerlessGrammarBuilder b) {
 
     // Flow
     b.rule(OPAQUE).is(SPACING, b.token(GenericTokenType.IDENTIFIER, "opaque"));
-    b.rule(TYPE).is(SPACING, b.token(GenericTokenType.IDENTIFIER, "type"));
+    b.rule(TYPE).is(SPACING, b.token(GenericTokenType.IDENTIFIER, "type"), b.nextNot(LETTER_OR_DIGIT));
```
vilchik-elena: @saberduck we should use `word` everywhere

vilchik-elena:
@saberduck could you also add test that there is no FP for javascript:S3827?

PR #1009, opened 2018-07-11
Remove dependency on sslr-squid-bridge

### LINKED ISSUE #948: Remove usages of sslr-squid-bridge
Note that sslr-squid-bridge is used for definition of custom rules as well.

saberduck:
Fixes #948

### DIFF
--- a/its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/CustomRulesTests.java
+++ b/its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/CustomRulesTests.java
@@ -21,25 +21,28 @@ package com.sonar.javascript.it.plugin;
 
 import com.sonar.orchestrator.Orchestrator;
 import com.sonar.orchestrator.build.SonarScanner;
+import java.util.Collections;
 import java.util.List;
 import org.junit.BeforeClass;
 import org.junit.ClassRule;
 import org.junit.Test;
-import org.sonar.wsclient.issue.Issue;
-import org.sonar.wsclient.issue.IssueClient;
-import org.sonar.wsclient.issue.IssueQuery;
+import org.sonarqube.ws.Issues;
+import org.sonarqube.ws.Rules;
+import org.sonarqube.ws.client.HttpConnector;
+import org.sonarqube.ws.client.WsClient;
+import org.sonarqube.ws.client.WsClientFactories;
+import org.sonarqube.ws.client.rules.SearchRequest;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
 public class CustomRulesTests {
 
+  public static final String CUSTOM_RULES_WHILE_CHECK = "javascript-custom-rules:whileCheck";
   @ClassRule
   public static Orchestrator orchestrator = Tests.ORCHESTRATOR;
 
-  private static IssueClient issueClient;
-
   @BeforeClass
-  public static void prepare() throws InterruptedException {
+  public static void prepare() {
     orchestrator.resetData();
     SonarScanner build = Tests.createScanner()
       .setProjectDir(TestUtils.projectDir("custom_rules"))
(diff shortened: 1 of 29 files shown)


its/plugin/plugins/javascript-custom-rules-plugin/src/main/java/org/sonar/samples/javascript/BaseTreeVisitorCheck.java
```diff
@@ -23,14 +23,12 @@
...
 import org.sonar.plugins.javascript.api.tree.expression.FunctionExpressionTree;
 import org.sonar.plugins.javascript.api.visitors.DoubleDispatchVisitorCheck;
-import org.sonar.squidbridge.annotations.SqaleConstantRemediation;
 
 @Rule(
   key = "base",
   name = "Base tree visitor check",
   description = "desc",
   priority = Priority.MINOR)
-@SqaleConstantRemediation("5min")
```
vilchik-elena: did you check the docs?

its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/CustomRulesTests.java
```diff
@@ -63,7 +63,6 @@ public void base_tree_visitor_check() {
     Issue issue = issues.get(0);
     assertThat(issue.line()).isEqualTo(2);
     assertThat(issue.message()).isEqualTo("Function expression.");
-    assertThat(issue.debt()).isEqualTo("5min");
```
vilchik-elena: IMO it's better keep explicit assertion (is it null?)

javascript-checks/src/test/java/org/sonar/javascript/checks/UselessIncrementCheckTest.java
```diff
@@ -20,16 +20,11 @@
...
 import java.io.File;
-import org.junit.Rule;
 import org.junit.Test;
 import org.sonar.javascript.checks.verifier.JavaScriptCheckVerifier;
-import org.sonar.squidbridge.checks.CheckMessagesVerifierRule;
 
 public class UselessIncrementCheckTest {
 
-  @Rule
-  public CheckMessagesVerifierRule checkMessagesVerifier = new CheckMessagesVerifierRule();
```
vilchik-elena: was it unused?

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/rules/JavaScriptRulesDefinition.java
```diff
@@ -53,8 +53,10 @@ public void define(Context context) {
       .createRepository(CheckList.REPOSITORY_KEY, JavaScriptLanguage.KEY)
       .setName(CheckList.REPOSITORY_NAME);
 
-    new AnnotationBasedRulesDefinition(repository, JavaScriptLanguage.KEY)
-      .addRuleClasses(/* don't fail if no SQALE annotations */ false, CheckList.getChecks());
+    RulesDefinitionAnnotationLoader rulesDefinitionAnnotationLoader = new RulesDefinitionAnnotationLoader();
```
vilchik-elena: it will be nice to migrate this class to use analyzer-commons (in a separate PR)

btw I tried to drop `rule.setInternalKey(metadataKey)` and everything seems to work.

sonar-javascript-plugin/pom.xml
```diff
@@ -44,6 +44,18 @@
       <groupId>com.google.code.gson</groupId>
       <artifactId>gson</artifactId>
     </dependency>
+    <dependency>
```
vilchik-elena: why these new dependencies? (guava and sslr-core)

javascript-checks/pom.xml
```diff
@@ -25,6 +21,18 @@
       <groupId>org.sonarsource.sonarqube</groupId>
       <artifactId>sonar-plugin-api</artifactId>
     </dependency>
+    <dependency>
```
vilchik-elena: these 3 dependencies were transitive before?

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/rules/JavaScriptRulesDefinition.java
saberduck: Ok, the same would be probably possible for the profile. I created #1012, maybe we will have time for it in this sprint

sonar-javascript-plugin/pom.xml
saberduck: These were before provided implicitly because they were transitive dependencies of `sslr-squid-bridge`

javascript-checks/pom.xml
saberduck: yes

its/plugin/plugins/javascript-custom-rules-plugin/src/main/java/org/sonar/samples/javascript/BaseTreeVisitorCheck.java
saberduck: there is nothing specific in the docs here https://docs.sonarqube.org/display/PLUG/Custom+Rules+for+SonarJS#CustomRulesforSonarJS-ImplementaRule   
However Squale annotations are used in the example plugin, I will also remove them from there

sonar-javascript-plugin/pom.xml
saberduck: I prefer to keep those dependencies explicitly even though now they are again provided transitevely through sslr-squid-bridge

its/plugin/tests/src/test/resources/profile-javascript-custom-rules.xml
```diff
@@ -13,6 +13,11 @@
       <key>subscription</key>
       <priority>MAJOR</priority>
     </rule>
+    <rule>
+      <repositoryKey>javascript-custom-rules-2</repositoryKey>
```
vilchik-elena: I'm not a bit fan of this `2` everywhere. Especially it will not make sense when deprecated rules will be removed. How about renaming old repo and rules with prefixing `deprecated`?

javascript-frontend/src/main/java/org/sonar/plugins/javascript/api/CustomJavaScriptRulesDefinition.java
```diff
@@ -28,10 +28,13 @@
 
 /**
  * Extension point to create custom rule repository for JavaScript.
+ *
+ * @deprecated  implement {@link RulesDefinition} directly
```
vilchik-elena: Add deprecation version

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/JavaScriptChecks.java
```diff
@@ -57,11 +59,17 @@ public JavaScriptChecks addChecks(String repositoryKey, Iterable<Class> checkCla
     return this;
   }
 
-  public JavaScriptChecks addCustomChecks(@Nullable CustomJavaScriptRulesDefinition[] customRulesDefinitions) {
+  public JavaScriptChecks addCustomChecks(@Nullable CustomJavaScriptRulesDefinition[] customRulesDefinitions,
```
vilchik-elena: can we add warning log here to not use deprecated `CustomJavaScriptRulesDefinition`?

PR #1010, opened 2018-07-12
Fix sqKey and add rules' scope

### LINKED ISSUE #958: Fix SQ key in rule metadata
A lot of rules with legacy keys contains incorrect value for `sqKey` in their metadata files (e.g. `UnunsedVariable`, `WithStatement`). RSPECs should be updated (copy legacy key to subtask) in order to generate correct metadata.

vilchik-elena:
Fixes #958


vilchik-elena:
@andrea-guarino-sonarsource merge next time if no adjustments are required

PR #1015, opened 2018-07-13
Add 'onlyWhenConfiguration' to CoverageSensor

### LINKED ISSUE #987: Set `onlyWhenConfiguration` on SensorDescriptor where possible
`onlyWhenConfiguration` allows completely skip sensor thus avoid confusing logs when sensor does nothing:
```
[INFO] Sensor TypeScript LCOV Coverage Sensor [typescript]
[INFO] Sensor TypeScript LCOV Coverage Sensor [typescript] (done) | time=3ms
```

vilchik-elena:
Fixes #987

**REBASE** when merging!

### DIFF
--- a/its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/Tests.java
+++ b/its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/Tests.java
@@ -58,7 +58,7 @@ public final class Tests {
 
   @ClassRule
   public static final Orchestrator ORCHESTRATOR = Orchestrator.builderEnv()
-    .setSonarVersion(System.getProperty("sonar.runtimeVersion", "7.2-RC1"))
+    .setSonarVersion(System.getProperty("sonar.runtimeVersion", "LATEST_RELEASE"))
     .addPlugin(JAVASCRIPT_PLUGIN_LOCATION)
     .restoreProfileAtStartup(FileLocation.ofClasspath("/empty-profile.xml"))
     .addPlugin(FileLocation.byWildcardMavenFilename(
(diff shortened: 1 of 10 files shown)


its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/Tests.java
```diff
@@ -58,7 +58,7 @@
 
   @ClassRule
   public static final Orchestrator ORCHESTRATOR = Orchestrator.builderEnv()
-    .setSonarVersion(System.getProperty("sonar.runtimeVersion", "7.2-RC1"))
+    .setSonarVersion(System.getProperty("sonar.runtimeVersion", "7.2"))
```
saberduck: Rather than using concrete version we should use alias "LATEST_RELEASE"

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/JavaScriptPlugin.java
```diff
@@ -70,6 +70,8 @@
   public static final String EXTERNAL_ANALYZERS_SUB_CATEGORY = "JavaScript/TypeScript";
   public static final String ESLINT_REPORT_PATHS = "sonar.eslint.reportPaths";
 
+  public static final String DEPRECATED_SONARTS_PROPERTY = "sonar.typescript.eslint.reportPaths";
```
saberduck: Maybe we should have `ESLINT` in the name, like `DEPRECATED_ESLINT_PROPERTY`. SonarTS can be mentioned in the comment.

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/JavaScriptSensor.java
```diff
@@ -325,6 +329,12 @@ public void execute(SensorContext context) {
     analyseFiles(context, treeVisitors, inputFiles, executor, progressReport);
   }
 
+  private static void checkDeprecatedTsProperty(SensorContext context) {
```
saberduck: "Ts" in the name is bit confusing, maybe `checkDeprecatedEslintProperty`   and just a comment that this property is from SonarTS

PR #1023, opened 2018-07-17
Add docs for ESLint import

vilchik-elena:


docs/DOC.md
```diff
@@ -65,6 +65,17 @@ Property | Description
 ---------|------------
 `sonar.javascript.lcov.reportPaths` | comma-separated list of paths to LCOV coverage report files
 
+### <a name="eslint"></a>Import of ESLint issues
+SonarJS provides an import into SonarQube of issues generated by ESLint. For that run your ESLint with JSON formatter and forward the results to some file. Here is an example of command generating required report.
```
nicolas-harraudeau-sonarsource: I see a few spots where we could improve readability
SonarJS enables the import of ESLint reports into SonarQube. In order to do this, run eslint with the JSON formatter and write the resulting report into a file. Here is an example of such a command:

docs/DOC.md
```diff
@@ -65,6 +65,17 @@ Property | Description
...
+```
+> eslint -f json . > report.json
```
nicolas-harraudeau-sonarsource: What about using `$` instead of `>`. It is usually more common for shell prompts. But this is a minor detail.

docs/DOC.md
```diff
@@ -65,6 +65,17 @@ Property | Description
...
+```
+Property | Description
```
nicolas-harraudeau-sonarsource: Before this I would add:
```
One this is done, add the following property to the sonnarQube scanner configuration:
```

docs/DOC.md
```diff
@@ -65,6 +65,17 @@ Property | Description
...
+---------|------------
+`sonar.eslint.reportPaths` | comma-separated list of paths (absolute or relative) to the JSON files with ESLint issues.
+
```
nicolas-harraudeau-sonarsource: Here we could add the following:
"""
You can now run the sonarQube-scanner and all the issues will be imported.
Note that the next time you run the scanner, all issues which disappeared from the eslint report will also disappear from the SonarQube reports.
"""

I deduced the last sentence from my current tests. Correct me if this is not the expected behavior.

docs/DOC.md
nicolas-harraudeau-sonarsource: Also the command could be simplified as:
```
$ eslint -f json -o report.json .
```

docs/DOC.md
```diff
@@ -65,6 +65,21 @@ Property | Description
 ---------|------------
 `sonar.javascript.lcov.reportPaths` | comma-separated list of paths to LCOV coverage report files
 
+### <a name="eslint"></a>Import of ESLint issues
+SonarJS enables the import of ESLint reports into SonarQube. In order to do this, run eslint with the JSON formatter and write the resulting report into a file. Here is an example of such a command:
+```
+$ eslint -f json -o report.json .
+```
+Once the report is created, add the following property to the sonnarQube scanner configuration:
```
nicolas-harraudeau-sonarsource: Sorry I noticed that I suggested sonnarQube instead of SonnarQube (with a capital "S").

PR #1038, opened 2018-08-03
Fix API for custom rules

### LINKED ISSUE #1037: Fix broken API for custom plugins
In the latest 4.2 release we accidentally broke API for custom rules plugins, because `JavaScriptSensor` cannot be instantiated properly.

saberduck:
Fixes #1037


saberduck:
Oups, sorry for that irrelavant file which I commited, good catch!

PR #1066, opened 2018-09-24
Connect eslint-bridge with SonarJS plugin

### LINKED ISSUE #1048: Connect eslint-bridge with SonarJS plugin
Open questions:
* which technology to use to communicate between node and java processes (see comment below)
* what is the protocol of communication
* how to package `eslint-bridge` inside SonarJS SQ plugin

Relies on #1059 

Do not forget to check for NodeJS presence (just quit with warning if it's not available).

vilchik-elena:
Fixes #1048


eslint-bridge/tests/server.test.ts
```diff
@@ -0,0 +1,74 @@
...
+import { promisify } from "util";
+import * as getPort from "get-port";
+
+describe("server", () => {
+  let port;
+  let server;
+  let close;
+
+  beforeEach(async () => {
+    port = await getPort();
```
saberduck: I am not sure we need getPort library, you can just make port default to 0 in `start` and then you can use `server.address().port` to get the allocated port

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBasedRulesSensor.java
```diff
@@ -0,0 +1,250 @@
...
+      try {
+        return file.contents();
+      } catch (IOException e) {
+        throw new IllegalStateException(e);
+      }
+    }
+  }
+
+  static class AnalysisResponseIssue {
+    Integer line;
```
saberduck: Can these fields be `null`? If not I would use `int`

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/NetUtils.java
```diff
@@ -0,0 +1,67 @@
...
+public class NetUtils {
+
+  private NetUtils(){
+  }
+
+  public static int findOpenPort() {
+    try (ServerSocket socket = new ServerSocket(0)) {
+      return socket.getLocalPort();
+    } catch (IOException e) {
+      return 3000;
```
saberduck: I don't think it makes sense to try to recover from this exception, we should just log and fail in the sensor (i.e. skip analysis)

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBasedRulesSensor.java
```diff
@@ -0,0 +1,250 @@
...
+
+  private void startEslintBridgeServerProcess() {
+    port = findOpenPort();
+    ProcessBuilder processBuilder = new ProcessBuilder(nodeExecutable, startServerScript, String.valueOf(port));
+    try {
+      LOG.debug("Starting node process to start eslint-bridge server at port " + port);
+      process = processBuilder.start();
+      setProcessStreamConsumers();
+
+      if (!waitServerToStart("localhost", port, timeoutSec * 1000)) {
```
saberduck: actually we don't need to wait for server to startup here, we can just set the timeout with okhttp . See here https://stackoverflow.com/questions/25953819/how-to-set-connection-timeout-with-okhttp

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/NetUtils.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+  }
+
+  public static boolean waitServerToStart(String host, int port, int timeoutMs) {
```
saberduck: you can probably remove this code, if we use timeout in okhttp

sonar-javascript-plugin/src/test/resources/eslint-bridge/logging.js
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env node
+
+console.log(`DEBUG starting server`)
```
saberduck: this is a nice "hack" to get the `DEBUG`, I like it

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBasedRulesSensor.java
vilchik-elena: yep, they can, it's possible to have issues with only start line and column

PR #1067, opened 2018-09-24
Fix description of external eslint plugin rules

### LINKED ISSUE #1035: Fix description of external eslint plugin rules
image

Note that this description is provided by [analyser-commons](https://github.com/SonarSource/sonar-analyzer-commons/blob/ceded2e42bb85ad3cf30f684609c3c0c8cfe7838/commons/src/main/java/org/sonarsource/analyzer/commons/ExternalRuleLoader.java#L55).

One possible fix is: “See the description of ESLint rule react/jsx-sort-props [here](url).” 
but this way we are not so explicit that description is provided by third party tool and not us. And this change should be done in analyzer-commons thus it will effect other language plugins. 

Another option is to provide `description` instead of `url`. That way to don't need to update analyzer-commons and can put to description what we want. For example: 
`See description of %s rule <code>%s</code> in the <a href=\"%s\">ESLint plugin for React documentation</a>.`

vilchik-elena:
Fixes #1035

### DIFF
--- a/javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/eslint/angular.json
+++ b/javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/eslint/angular.json
@@ -3,368 +3,358 @@
     "key": "angular/avoid-scope-typos",
     "type": "BUG",
     "name": "Avoid mistakes when naming methods defined on the scope object",
-    "url":
-      "https://github.com/Gillespie59/eslint-plugin-angular/blob/HEAD/docs/rules/avoid-scope-typos.md"
+    "url": "https://github.com/Gillespie59/eslint-plugin-angular/blob/HEAD/docs/rules/avoid-scope-typos.md",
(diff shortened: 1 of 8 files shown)


javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/eslint/angular.json
```diff
@@ -3,368 +3,358 @@
     "key": "angular/avoid-scope-typos",
     "type": "BUG",
     "name": "Avoid mistakes when naming methods defined on the scope object",
-    "url":
-      "https://github.com/Gillespie59/eslint-plugin-angular/blob/HEAD/docs/rules/avoid-scope-typos.md"
+    "url": "https://github.com/Gillespie59/eslint-plugin-angular/blob/HEAD/docs/rules/avoid-scope-typos.md",
+    "description": "See description of ESLint rule <code>angular/avoid-scope-typos</code> at the <a href=\"https://github.com/Gillespie59/eslint-plugin-angular/blob/HEAD/docs/rules/avoid-scope-typos.md\">ESLint AngularJS plugin website</a>."
```
stas-vilchik: I suggest to always use the plugin name, `eslint-plugin-angular` in this case.

PR #1071, opened 2018-09-26
Integrate issues coming from eslint-bridge into analysis results

### LINKED ISSUE #1060: Integrate issues coming from eslint-bridge into analysis results
When #1048 is fixed, we will be able to get issues coming from eslint-based rules in Java SQ sensor. Then we should save these issues into SQ, while not duplicating them with issues from corresponding java-based rules. 

Note that there will be 15 rules to integrate, rules which are available in [eslint-sonarjs 0.1.1](https://github.com/SonarSource/eslint-plugin-sonarjs).

vilchik-elena:
Fixes #1060


eslint-bridge/src/server.ts
```diff
@@ -29,7 +29,7 @@ export function start(port = 0): Promise<Server> {
     const app = express();
 
     // for parsing application/json requests
-    app.use(bodyParser.json());
+    app.use(bodyParser.json({ limit: "50mb" }));
```
andrea-guarino-sonarsource: I think we should put "50mb" in a constant

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/JavaScriptPlugin.java
```diff
@@ -84,7 +86,10 @@ public void define(Context context) {
       JavaScriptRulesDefinition.class,
       SonarWayRecommendedProfile.class,
       SonarWayProfile.class,
-      NodeCommand.builder());
+      // fixme replace path to script
+      new EslintBridgeServerImpl(NodeCommand.builder(), 5, "/Users/elenavilchik/Projects/LANG/SonarJS/eslint-bridge/bin/server"),
```
andrea-guarino-sonarsource: Can you store "5" in a constant? (to make it explicit it is regarding timeout)

vilchik-elena: done in Tibor's PR

vilchik-elena:
merged in https://github.com/SonarSource/SonarJS/pull/1073

PR #1073, opened 2018-09-27
Integrate eslint-bridge

### LINKED ISSUE #1060: Integrate issues coming from eslint-bridge into analysis results

### LINKED ISSUE #1070: package eslint-bridge into jar

saberduck:
Fixes #1070 and fixes #1060


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
```diff
@@ -41,14 +40,14 @@
...
-@ScannerSide
 public class EslintBridgeServerImpl implements EslintBridgeServer {
 
   private static final Logger LOG = Loggers.get(EslintBridgeServerImpl.class);
 
   private static final int DEFAULT_TIMEOUT_SECONDS = 10;
   private static final String DEFAULT_STARTUP_SCRIPT = "eslint-bridge/bin/server";
-  private static final String DEFAULT_BUNDLE = "node_modules";
+  private static final String NODE_MODULES = "node_modules";
+  private static final String DEFAULT_BUNDLE = NODE_MODULES;
```
saberduck: I don't understand this change, the constant `NODE_MODULES` doesn't have any meaning by itself

PR #1098, opened 2018-10-02
Improve exception handling in eslint-based sensor

vilchik-elena:


sonar-javascript-plugin/src/test/resources/mock-eslint-bridge/logging.js
```diff
@@ -1,6 +1,13 @@
 #!/usr/bin/env node
 
-console.log(`DEBUG starting server`)
-console.log(`server is started`)
+const http = require('http')
```
saberduck: why starting the server here? I don't think it's needed, the test should just verify port redirection

PR #1099, opened 2018-10-03
Use xz compression for eslint-bridge bundle

saberduck:


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
```diff
@@ -68,19 +72,23 @@ public EslintBridgeServerImpl(NodeCommandBuilder nodeCommandBuilder, TempFolder
     this.nodeCommandBuilder = nodeCommandBuilder;
     this.timeoutSeconds = timeoutSeconds;
     this.startServerScript = startServerScript;
-    this.bundleLocation = bundleLocation;
     this.client = new OkHttpClient.Builder()
       .readTimeout(timeoutSeconds, TimeUnit.SECONDS)
       .build();
     // name has to be "node_modules" for correct dependency resolution in NodeJS
```
vilchik-elena: comment can be removed now

Issue #1141, opened 2018-12-20
Arrow function parsing error


napztah:
**I want to report a bug.**

Previously I was using 

**SonarJS version:** 3.3 <br>
**SonarQube version:** 6.7.2<br>

And I was able to successfully run sonar-scanner. But after I have upgraded to following versions

**SonarJS version:** 5.0 <br>
**SonarQube version:** 7.4<br>

I get the following error while executing sonar-scanner

**Logs**
```
ERROR: Failed to parse file [file:///C:/Users/uperer9/Documents/duplo/socket-thin-client/src/components/CourseInfo/Avatar/Avatar.js] at line 28: Unexpected token = (with espree parser in module mode)
```

Line 28 mentioned in the log is as follows
```
    keyPressHandler = (e) => {
        (e.keyCode === 13) && this.props.clicked();
    };
```

vilchik-elena:
Are you sure it's not assignment `=`? 
Can you give more context of your file? Is it class field may be?

napztah:
Following is the file

```
import React, {Component} from 'react';
import classNames from 'classnames';
import PropTypes from 'prop-types';
import { defineMessages, injectIntl, intlShape } from 'react-intl';

import styles from './Avatar.scss';

class Avatar extends Component {

    static messages = defineMessages({
        imageAltText: {
            id: 'avatarImage.alternative',
            defaultMessage: 'Product Image',
        },
        openRevel: {
            id: 'avatarImage.anchorTag',
            defaultMessage: 'Link to open',
        },
    });

    constructor(props) {
        super(props);
    }

    keyPressHandler = (e) => {
        (e.keyCode === 13) && this.props.clicked();
    };

    render() {

        const { formatMessage } = this.props.intl;

        const avatarOrientationBaseClass = (this.props.avatarOrientation === 'portrait' ?
            styles.AvatarContainerPortrait :
            styles.AvatarContainerLandscape);
        const disableTab = (this.props.disable && {tabIndex: '-1'});

        return (
            <div className={classNames(avatarOrientationBaseClass)}>
                <div className={classNames(styles.ContentBorderTreatment, styles.DashboardContentBorderTreatment)}>
                    <a href="#" tabIndex='-1' aria-label={formatMessage(Avatar.messages.openRevel)} aria-disabled={this.props.disable}>
                        <img className={classNames(styles.Avatar)} src={this.props.url} onClick={this.props.clicked}
[...]

PR #1143, opened 2019-01-10
Use only babel-eslint to parse

### LINKED ISSUE #1100: Support props in ESLint-based rules

simonedavico:
Update tests to check for babel-eslint error messages. I was forced to comment out the duplicate identifier test, because babel does not throw errors for duplicate identifiers (see [this issue](https://github.com/babel/babel/issues/6722)).

Fixes #1100 #1141 

Would this be a viable approach? I initially wanted to add a config flag to allow switching between `espree` and `babel-eslint`, but it would complicate some cases (e.g., `Flow` files cannot be parsed by `espree`).

### DIFF
--- a/eslint-bridge/package.json
+++ b/eslint-bridge/package.json
@@ -42,7 +42,6 @@
     "body-parser": "1.18.3",
     "eslint": "5.6.0",
     "eslint-plugin-sonarjs": "0.2.0",
-    "espree": "4.0.0",
     "express": "4.16.3"
   },
   "prettier": {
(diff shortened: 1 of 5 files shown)


vilchik-elena:
@simonedavico thanks for PR! it's all ok from the implementation point of view, but I'm not yet sure that's the way we want to go for parsing.
We will consider it more carefully next time we will have a sprint for SonarJS (hopefully in Feb).

simonedavico:
Sure, I just needed a fix for our internal infrastructure since we are using `stage-0` features, and I thought that while I was at it I could contribute back. Hope you will take into consideration my observations on issue #1100 when you decide how much JS ecosystem you want to support :) thanks!

slikts:
I'd like to try running this patched version of SonarJS, but not sure how to proceed after building the .jar files; what would be the way to run the patched version with sonar-scanner?

PR #1154, opened 2019-01-31
Remove creation of test tars during build

vilchik-elena:
Also makes usage of '--force-local' option conditional (only for windows)


eslint-bridge/scripts/package.js
```diff
@@ -25,7 +25,11 @@ async function jar() {
     fs.ensureDirSync(target);
     const bundle = path.join(target, 'eslint-bridge.tar.xz');
     // --force-local is required so windows path starting with C: is recognized as local
-    const cmd = `tar --xz --force-local -cf '${bundle}' *`;
+    let forceLocal = "";
+    if (process.platform == "win32" || process.platform == "win64") {
```
saberduck: It seems that platform can only be `win32`, `win64` doesn't exists. See https://nodejs.org/api/os.html#os_os_platform  or https://stackoverflow.com/questions/8683895/how-do-i-determine-the-current-operating-system-with-node-js

PR #1159, opened 2019-02-05
Add hotspot rule S2255 (Cookies)

### LINKED ISSUE #1139: Rule S2255: Using cookies is security-sensitive
Implements [RSPEC-2255](https://jira.sonarsource.com/browse/RSPEC-5084)

vilchik-elena:
Fixes #1139


eslint-bridge/src/rules/cookies.ts
```diff
@@ -0,0 +1,81 @@
...
+    let usingExpressFramework = false;
+
+    return {
+      Program() {
+        // init flag for each file
+        usingExpressFramework = false;
+      },
+
+      Literal(node: estree.Node) {
+        if ((node as estree.Literal).value === "express") {
```
andrea-guarino-sonarsource: IMO, we need to be more precise: I think we should check for `import` or `require` statements that contains "express"

eslint-bridge/src/rules/cookies.ts
```diff
@@ -0,0 +1,81 @@
...
+            message,
+            node,
+          });
+        }
+      },
+
+      CallExpression(node: estree.Node) {
+        const { callee, arguments: args } = node as estree.CallExpression;
+        if (
+          callee.type === "MemberExpression" &&
```
andrea-guarino-sonarsource: what about having a `usingHttp` flag as well?

eslint-bridge/src/rules/cookies.ts
vilchik-elena: Doing that will overcomplexify implementation without a lot of return value. I believe we will not have FP even like this.

eslint-bridge/src/rules/cookies.ts
vilchik-elena: We are checking first argument value, so pattern is pretty narrow, I don't think we need more specialization

PR #1160, opened 2019-02-06
Add hotspot rule S4797 (File Handling)

### LINKED ISSUE #1129: Rule S4797: Handling files is security-sensitive
Implements [RSPEC-4797](https://jira.sonarsource.com/browse/RSPEC-5076)

andrea-guarino-sonarsource:
Fixes #1129


eslint-bridge/src/rules/file-handling.ts
```diff
@@ -0,0 +1,101 @@
...
+    callee.type === "MemberExpression" &&
+    callee.object.type === "Identifier" &&
+    isQuestionable(callee.property, args, getModuleNameOfIdentifier(callee.object, context))
+  ) {
+    context.report({ message: MESSAGE, node: callee });
+  } else if (
+    callee.type === "Identifier" &&
+    isQuestionable(callee, args, getModuleNameOfImportedIdentifier(callee, context))
+  ) {
+    context.report({ message: MESSAGE, node: callExpression });
```
vilchik-elena: why do we report here on entire expression and above on callee only ?

eslint-bridge/src/rules/file-handling.ts
```diff
@@ -0,0 +1,101 @@
...
+
+function isQuestionable(
+  expression: estree.Expression,
+  args: Argument[],
+  moduleName?: estree.Literal,
+) {
+  if (!moduleName || moduleName.value !== FS_MODULE) {
+    return false;
+  }
+  if (expression.type === "Identifier") {
```
vilchik-elena: it's always identifier, you can change parameter declared type

eslint-bridge/src/rules/file-handling.ts
andrea-guarino-sonarsource: Forgot that while refactoring, I will report only on callee

eslint-bridge/src/rules/file-handling.ts
andrea-guarino-sonarsource: As discussed, “MemberExpression” can have property that is not an identifier (like `fs[getFunctionName()]('foo.txt');`)

PR #1189, opened 2019-02-18
Enable eslint based sensor in SonarLint - start server lazily

### LINKED ISSUE #1095: Enable ESLint-based rules in SonarLint
Since we are having some performance problem, we are not executing ESLint-based rules in SonarLint context so far.
In order to improve performance we should use ContextualServer as in SonarTS and activate ESLint-based analysis on SonarLint.

saberduck:
Fixes #1095


saberduck:
Replaced by #1192

PR #1192, opened 2019-02-18
Activate ESLint-based rules in SonarLint

### LINKED ISSUE #1191: Launch node via `sh -c`
On MacOS when process is launched from the system dock, the environment is not the same as if it is launched from the terminal, because shell configuration files (`.bashrc`, ...) are not loaded. 

This might cause the issue when launching Node.js process, because `PATH` variable will be most likely configured in `.bashrc` or similar file. 

As a workaround we can launch `node` via shell like `/bin/sh -c node` , this way the configuration should be loaded.

saberduck:
Fixes #1191 , #1095


its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/SonarLintTest.java
```diff
@@ -75,75 +70,43 @@ public static void stop() {
...
+      "  doSomething();" +
+      "} catch (ex) {" +
+      "  throw ex;" +
+      "}}";
+    List<Issue> issues = analyze(FILE_PATH, sourceCode);
+    assertThat(issues).extracting(Issue::getRuleKey).containsExactly("javascript:S2737");
+    // let's analyze again
+    issues = analyze(FILE_PATH, sourceCode);
+    assertThat(issues).extracting(Issue::getRuleKey).containsExactly("javascript:S2737");
+    assertThat(logs).contains("Node.js process is already started, no need to start.");
```
vilchik-elena: let's be more precise here, which process is started

its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/SonarLintTestCustomNodeJS.java
```diff
@@ -0,0 +1,88 @@
...
+import org.sonarsource.sonarlint.core.StandaloneSonarLintEngineImpl;
+import org.sonarsource.sonarlint.core.client.api.common.analysis.ClientInputFile;
+import org.sonarsource.sonarlint.core.client.api.common.analysis.Issue;
+import org.sonarsource.sonarlint.core.client.api.standalone.StandaloneAnalysisConfiguration;
+import org.sonarsource.sonarlint.core.client.api.standalone.StandaloneGlobalConfiguration;
+import org.sonarsource.sonarlint.core.client.api.standalone.StandaloneSonarLintEngine;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class SonarLintTestCustomNodeJS {
```
vilchik-elena: I think you should add this test to test suite in `Tests`

nodejs-utils/src/main/java/org/sonarsource/nodejs/MockCommandBuilder.java
```diff
@@ -107,7 +108,7 @@ public NodeCommand build() {
   private class MockCommand extends NodeCommand {
```
vilchik-elena: Is `MockCommandBuilder` used? I can't find it's usages

saberduck: indeed, seems like it's not, I don't remember why we needed it. I deleted it

PR #1204, opened 2019-02-22
Update rule metadata

saberduck:
Fixes #

### DIFF
--- a/javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S104.json
+++ b/javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S104.json
@@ -1,5 +1,5 @@
 {
-  "title": "Files should not have too many lines of code",
+  "title": "Files should not have too many lines",
   "type": "CODE_SMELL",
   "status": "ready",
   "remediation": {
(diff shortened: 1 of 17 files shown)


javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S104.json
```diff
@@ -1,5 +1,5 @@
 {
-  "title": "Files should not have too many lines",
+  "title": "Files should not have too many lines of code",
```
vilchik-elena: should be reverted

javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S138.json
```diff
@@ -1,5 +1,5 @@
 {
-  "title": "Functions should not have too many lines",
+  "title": "Functions should not have too many lines of code",
```
vilchik-elena: should be reverted

javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S2228.json
```diff
@@ -1,7 +1,7 @@
 {
   "title": "Console logging should not be used",
   "type": "VULNERABILITY",
-  "status": "ready",
+  "status": "superseded",
```
vilchik-elena: it's a bug in rule-api, should be reverted

PR #1208, opened 2019-02-27
In case of failure log only once

### LINKED ISSUE #1207: ESLint sensor should not attempt to start server if it failed previously
ESLint sensor will try to start server on every analysis if it is not started yet. In case of missing Node.js this means that we will attempt to start (and fail) on every keystroke in VSCode. See also https://jira.sonarsource.com/browse/SLVSCODE-63

saberduck:
Fixes #1207

### DIFF
--- a/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBasedRulesSensor.java
+++ b/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBasedRulesSensor.java
@@ -102,6 +102,10 @@ public class EslintBasedRulesSensor implements Sensor {
         progressReport.nextFile();
       }
       progressReport.stop();
+    } catch (ServerAlreadyFailedException e) {
+      LOG.debug("Skipping start of eslint-bridge server due to the failure during first analysis");
+      LOG.debug("Skipping execution of eslint-based rules due to the problems with eslint-bridge server");
+
     } catch (NodeCommandException e) {
       LOG.error(e.getMessage(), e);
       if (analysisWarnings != null) {
--- a/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
+++ b/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
@@ -36,6 +36,7 @@ import org.sonar.api.utils.log.Loggers;
 import org.sonarsource.nodejs.BundleUtils;
 import org.sonarsource.nodejs.NodeCommand;
 import org.sonarsource.nodejs.NodeCommandBuilder;
+import org.sonarsource.nodejs.NodeCommandException;
 
 import static org.sonar.plugins.javascript.eslint.NetUtils.findOpenPort;
 import static org.sonar.plugins.javascript.eslint.NetUtils.waitServerToStart;
@@ -59,6 +60,7 @@ public class EslintBridgeServerImpl implements EslintBridgeServer {
   private String startServerScript;
   private String bundleLocation;
   private Path deployLocation;
+  private boolean failedToStart;
 
   // Used by pico container for dependency injection
   @SuppressWarnings("unused")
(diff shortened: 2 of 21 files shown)


sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/EslintBasedRulesSensorTest.java
```diff
@@ -318,14 +321,31 @@ public void should_detect_unknown_rule_key() throws Exception {
 
   @Test
   public void handle_missing_node() throws Exception {
-    doThrow(new NodeCommandException("Exception Message", new IOException())).when(eslintBridgeServerMock).startServerLazily(any());
+    EslintBridgeServerImpl eslintBridgeServer = new EslintBridgeServerImpl(NodeCommand.builder(), tempFolder);
+    eslintBridgeServer = spy(eslintBridgeServer);
+    doThrow(new NodeCommandException("Exception Message", new IOException())).when(eslintBridgeServer).deploy();
```
vilchik-elena: why did you change throw method from `startServerLazily` for `deploy`?

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/EslintBasedRulesSensorTest.java
```diff
@@ -318,14 +321,31 @@ public void should_detect_unknown_rule_key() throws Exception {
 
   @Test
   public void handle_missing_node() throws Exception {
-    doThrow(new NodeCommandException("Exception Message", new IOException())).when(eslintBridgeServerMock).startServerLazily(any());
```
vilchik-elena: I think we can inline `eslintBridgeServerMock` into `createSensor`

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBasedRulesSensor.java
```diff
@@ -103,7 +103,7 @@ public void execute(SensorContext context) {
       }
       progressReport.stop();
     } catch (NodeCommandException e) {
-      LOG.error(e.getMessage(), e);
+      LOG.debug(e.getMessage());
```
vilchik-elena: So we will log twice the same thing? Can we think of a better solution?

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
```diff
@@ -129,12 +131,21 @@ private void initNodeCommand(SensorContext context, File scriptFile) {
 
   @Override
   public void startServerLazily(SensorContext context) throws IOException {
-    if (isAlive()) {
-      LOG.debug("SonarJS eslint-bridge server is up, no need to start.");
-      return;
+    if (failedToStart) {
+      throw new NodeCommandException("Failed to start SonarJS eslint-bridge before, will not attempt again.");
```
vilchik-elena: How about rephrasing? e.g. `Skipping start of eslint-bridge server due to the failure during first analysis`

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/EslintBasedRulesSensorTest.java
saberduck: because exception is now logged inside `EslintBridgeServerImpl` , so it can't be mocked anymore and we need a real instance. So the failure is mocked in the `deploy` method. I agree that this is not a unit test anymore, but more like integration test between sensor and server, but I didn't find better solution

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
saberduck: I think I will just let you write all the messages for me, it's gonna be faster :)

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/EslintBasedRulesSensorTest.java
saberduck: I don't understand what you mean, inline how?

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBasedRulesSensor.java
saberduck: well, I didn't find anything better, without making it much more complicated. And note that we will log twice only in very specific case, when we fail first time the analysis, on subsequent invocation it will not happen (an once it's on error and otherwise on debug level)

PR #1211, opened 2019-03-11
update travis utils to version 55

benzonico:
Fixes #


vilchik-elena:
@benzonico I'm taking care of ITs (metrics removed)

PR #1215, opened 2019-03-14
Fix unclear UntrustedContentCheck rule description

CidTori:
The current description make it unclear if dots are considered as litteral or special characters.
Since the source code consider that every piece of the list is a regex, litteral dots should be escaped.


CidTori:
This is my first PR, so I hope I did it the right way, but that rule description was bugging me, so when I found the source code on GitHub, I couldn't refrain from fixing it.

vilchik-elena:
merged in #1218

vilchik-elena:
thanks @CidTori for the contribution!

PR #1217, opened 2019-03-18
Update IT: remove assertions on metrics computed on SQ side

christophe-zurn-sonarsource:


vilchik-elena:
Thanks @christophe-zurn-sonarsource

PR #1219, opened 2019-03-20
Update ITs to not use deprecated sonar.profile property

saberduck:


its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/NoSonarTest.java
```diff
@@ -1,63 +1,61 @@
-/*
```
saberduck: This file is changed because of incorrect EOL (windows)

PR #1221, opened 2019-04-01
S2692: add compatibility notes and string example

acdha:
This updates the suggestion to use `includes` with compatibility information and an example of `String.includes` usage.

### DIFF
--- a/javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S2692.html
+++ b/javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S2692.html
@@ -1,6 +1,18 @@
 <p>Most checks against an <code>indexOf</code> call against a string or array compare it with -1 because 0 is a valid index. Any checks which look for
 values &gt;0 ignore the first element, which is likely a bug. If you're merely checking the presence of the string, consider using
 <code>includes</code> instead.</p>
+
+<p>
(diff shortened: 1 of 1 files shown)


vilchik-elena:
Hi @acdha! 
Thanks for your PR, but I'm not really happy with the way you put it (IMHO no need for extra example and it's not clear what does `are available in browsers newer than Internet Explorer 11` mean).

On my side I've updated RSPEC for this rule https://jira.sonarsource.com/browse/RSPEC-2708.
We generate the file you've changed based on this RSPEC, so it will be updated in next version

acdha:
My rationale for including the extra example is that it looks like a bug in the scanner when the alert is raised on a string `indexOf()` invocation but the examples are all for arrays. 

Re: “what does ‘are available in browsers newer than Internet Explorer 11‘ mean” — I wanted to note that this was an ES 2015 addition so the suggestion can only be followed if you either don't support legacy browsers or have some sort of polyfill or compilation process which provides it. Part of my rationale for linking to MDN was that they have generally very high-quality browser compatibility information.

vilchik-elena:
In fact there is an example for string (`arr[0].indexOf...`), but it's not really explicit. I've updated RSPEC

acdha:
Thanks — I figure things like this can always be as clear as possible since people are often triaging a fairly large list. I'm trying to reduce  cases where someone marks a finding as a false positive in a hurry.

PR #1256, opened 2019-06-07
Use NodeCommandException when invalid arguments in NodeCommand

vilchik-elena:


saberduck:
Why do we need this change actually? The idea to use `IllegalArgumentException` comes from the fact, that these errors are programmer errors, it is wrong to try to build `NodeCommand` without providing a script. The `NodeCommandException` is used for errors due to the environment - i.e. failed to run process, etc.

PR #1261, opened 2019-06-10
Implement sonarSource/SonarJS #1260 fix

### LINKED ISSUE #1260: Custom JSX, expression on attribute value in *.jsx files crash JavaScript analysis
**I want to request a feature.**

When we have a custom JSX with expression in the attribute value, the JavaScript analysis crash.
Eg: `<repeat collection={ MapObject } as={ key, value }>`

This is a following issue of: https://github.com/SonarSource/SonarJS/issues/1031

holblin:
Fixes #1260


vilchik-elena:
See comment on #1260

PR #1295, opened 2019-08-19
Declare TypeScript language

### LINKED ISSUE #1288: Declare TypeScript language

saberduck:
Fixes #1288

### DIFF
--- a/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/JavaScriptLanguage.java
+++ b/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/JavaScriptLanguage.java
@@ -26,6 +26,8 @@ import org.sonar.api.resources.AbstractLanguage;
 public class JavaScriptLanguage extends AbstractLanguage {
 
   public static final String KEY = "js";
+  static final String FILE_SUFFIXES_KEY = "sonar.javascript.file.suffixes";
+  static final String FILE_SUFFIXES_DEFVALUE = ".js,.jsx,.vue";
 
   private Configuration configuration;
 
@@ -36,9 +38,9 @@ public class JavaScriptLanguage extends AbstractLanguage {
 
   @Override
   public String[] getFileSuffixes() {
-    String[] suffixes = configuration.getStringArray(JavaScriptPlugin.FILE_SUFFIXES_KEY);
+    String[] suffixes = configuration.getStringArray(FILE_SUFFIXES_KEY);
     if (suffixes == null || suffixes.length == 0) {
-      suffixes = StringUtils.split(JavaScriptPlugin.FILE_SUFFIXES_DEFVALUE, ",");
+      suffixes = StringUtils.split(FILE_SUFFIXES_DEFVALUE, ",");
     }
     return suffixes;
   }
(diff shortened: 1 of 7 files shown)


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/JavaScriptPlugin.java
```diff
@@ -74,6 +73,9 @@
 
   public static final String DEPRECATED_ESLINT_PROPERTY = "sonar.typescript.eslint.reportPaths";
 
+  private static final String TYPESCRIPT_CATEGORY = "TypeScript";
+  private static final String GENERAL_SUBCATEGORY = "General";
```
vilchik-elena: we already have a constant for this one

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/JavaScriptPlugin.java
```diff
@@ -177,5 +179,22 @@ public void define(Context context) {
...
 
+    context.addExtensions(typeScriptExtensions());
+  }
+
+  private static Collection<?> typeScriptExtensions() {
+    return Arrays.asList(
+      TypeScriptLanguage.class,
+
+      PropertyDefinition.builder(TypeScriptLanguage.FILE_SUFFIXES_KEY)
+      .defaultValue(TypeScriptLanguage.FILE_SUFFIXES_DEFVALUE)
```
vilchik-elena: why there is no indentation here?

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/JavaScriptLanguage.java
```diff
@@ -1,46 +1,48 @@
...
+
+  private Configuration configuration;
+
+  public JavaScriptLanguage(Configuration configuration) {
+    super(KEY, "JavaScript");
+    this.configuration = configuration;
+  }
+
+  @Override
+  public String[] getFileSuffixes() {
```
vilchik-elena: should we simplify this implementation the same way it's done for TS?

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/JavaScriptPlugin.java
```diff
@@ -99,8 +101,8 @@ public void define(Context context) {
         .multiValues(true)
         .build(),
 
-      PropertyDefinition.builder(FILE_SUFFIXES_KEY)
-        .defaultValue(FILE_SUFFIXES_DEFVALUE)
+      PropertyDefinition.builder(JavaScriptLanguage.FILE_SUFFIXES_KEY)
+        .defaultValue(JavaScriptLanguage.FILE_SUFFIXES_DEFVALUE)
         .name("File Suffixes")
         .description("List of suffixes for files to analyze.")
```
quentin-jaquier-sonarsource: We should also update this description, to match the one at line [192](https://github.com/SonarSource/SonarJS/pull/1295/files#diff-75c0e6bce6afa14a48c230e8f54b880bR192), in order to be clear with the user if he can or cannot use comma.

vilchik-elena: I would say the opposite: this description is ok, the one for TS should be updated as this description is only seen in UI where commas are not allowed

saberduck: I think I am with Lena here, I updated both descriptions to be "List of suffixes for files to analyze."

PR #1299, opened 2019-08-21
Import of LCOV coverage report for TypeScript

### LINKED ISSUE #1287: Import of LCOV coverage report for TypeScript
We should deprecate `sonar.typescript.lcov.reportPaths`, still relying on it for couple of SonarJS releases. Print warning log when its value is provided.

When importing report we save coverage for both JS and TS file

quentin-jaquier-sonarsource:
Fixes #1287


sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/lcov/CoverageSensorTest.java
```diff
@@ -97,6 +97,20 @@ public void report_not_found() throws Exception {
   public void test_coverage() {
     coverageSensor.execute(context);
 
+    testTwoFilesContext();
+  }
+
+  @Test
+  public void should_work_and_log_warning_when_deprecated_key() throws Exception {
+    settings.setProperty(JavaScriptPlugin.TS_LCOV_REPORT_PATHS, UT_LCOV + ", " + IT_LCOV);
```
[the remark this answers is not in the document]
quentin-jaquier-sonarsource: Do you mean when the same report is mentioned in both property?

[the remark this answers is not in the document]
quentin-jaquier-sonarsource: Ho, nvm, I got what you mean, I will add more tests!

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/lcov/CoverageSensorTest.java
```diff
@@ -97,6 +97,20 @@ public void report_not_found() throws Exception {
...
+    coverageSensor.execute(context);
+
+    testTwoFilesContext();
+
+    assertThat(logTester.logs(LoggerLevel.WARN)).contains("The use of sonar.typescript.lcov.reportPaths for coverage import is deprecated, use sonar.javascript.lcov.reportPaths instead!");
+  }
+
+  private void testTwoFilesContext() {
```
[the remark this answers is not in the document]
quentin-jaquier-sonarsource: I will try 😅

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/JavaScriptPlugin.java
```diff
@@ -204,7 +204,7 @@ public void define(Context context) {
       PropertyDefinition.builder(TS_LCOV_REPORT_PATHS)
         .defaultValue(TS_LCOV_REPORT_PATHS_DEFAULT_VALUE)
         .name("LCOV Files")
-        .description("DEPRECATED - Use " + LCOV_REPORT_PATHS + " (Javascript tab) instead! \n" +
+        .description("DEPRECATED - Use " + LCOV_REPORT_PATHS + " (Javascript tab) instead. \n" +
```
vilchik-elena: remove `(Javascript tab)`, we never use such "helpers" and I think it should be clear for this property where it is

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/lcov/CoverageSensorTest.java
```diff
@@ -96,21 +97,40 @@ public void report_not_found() throws Exception {
...
+    settings.setProperty(JavaScriptPlugin.TS_LCOV_REPORT_PATHS, IT_LCOV);
+    coverageSensor.execute(context);
+
+    testUtAndItReportsCoverageData();
+    assertThat(logTester.logs(LoggerLevel.WARN)).contains(DEPRECATED_MESSAGE);
+  }
+
+  @Test
+  public void both_properties_reports_paths_overlap() throws Exception {
+    // LCOV_REPORT_PATHS is set with UT_LCOV and IT_LCOV during initialisation.
```
vilchik-elena: I suggest to move this line from initializion here (and everywhere else it's required). I will be more explicit

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/lcov/CoverageSensorTest.java
```diff
@@ -47,6 +47,7 @@
 
   private static final String UT_LCOV = "reports/report_ut.lcov";
```
vilchik-elena: I believe `UT` and `IT` are here since the times we had separate properties for difference tests. So I suggest to rename them into `report1` and `report2`

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/lcov/CoverageSensorTest.java
```diff
@@ -96,7 +97,40 @@ public void report_not_found() throws Exception {
   @Test
   public void test_coverage() {
     coverageSensor.execute(context);
+    testUtAndItReportsCoverageData();
```
vilchik-elena: it's better to start this method name with `assert` verb, as that's what happens inside

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/lcov/CoverageSensorTest.java
```diff
@@ -45,8 +45,10 @@
 
 public class CoverageSensorTest {
 
-  private static final String UT_LCOV = "reports/report_ut.lcov";
-  private static final String IT_LCOV = "reports/report_it.lcov";
+  private static final String REPORT1 = "reports/report_ut.lcov";
```
vilchik-elena: lcov files should be renamed as well

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/lcov/CoverageSensorTest.java
```diff
@@ -45,8 +45,10 @@
...
+  private static final String REPORT2 = "reports/report_it.lcov";
+  private static final String REPORT1AND2 = REPORT1 + ", " + REPORT2;
```
vilchik-elena: `REPORTS` or `TWO_REPORTS` or `BOTH_REPORTS`

PR #1314, opened 2019-08-23
Add rule S4325 "no-unnecessary-type-assertion" 

### LINKED ISSUE #1308: Add rule S4325 "no-unnecessary-type-assertion" [TS]
`@typescript-eslint/no-unnecessary-type-assertion`

vilchik-elena:
Fixes #1308


vilchik-elena:
There are differences in results, but this rule is heavily based on compiler so it's hard to estimate if issues diff is good or bad. Looks like some of missing issues were FP, so it's good.

its/ruling/src/test/resources/convert-expectations.js
```diff
@@ -1,22 +1,40 @@
...
+});
+
+projects.forEach(projectKey => {
+  const result = resultsByProject[projectKey];
+  const json = JSON.stringify(result, null, 2);
+  const dir = destination + projectKey;
+  if (!fs.existsSync(dir)) {
+    fs.mkdirSync(dir);
+  }
+  fs.writeFileSync(destination + projectKey + "/typescript-" + ruleKey + ".json", json);
```
yassin-kammoun-sonarsource: You can replace `destination + projectKey` with `dir`.

eslint-bridge/tests/analyzer.test.ts
```diff
@@ -80,16 +89,19 @@ describe("#analyzeTypeScript", () => {
   it("should report issue running eslint", () => {
     const { issues } = analyzeTypeScript({
       filePath: filePath,
-      fileContent: codeToTest,
+      fileContent: codeToTest + `let x = 4; x as number;`,
```
saberduck: This is a bit ugly, could we create a separate test for typechecker based rule?

javascript-checks/src/main/java/org/sonar/javascript/checks/JavaScriptRule.java
```diff
@@ -0,0 +1,30 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.javascript.checks;
```
saberduck: Let's move these annotation classes to separate package, like `org.sonar.javascript.checks.annotations`

eslint-bridge/typings/typescipt-eslint.d.ts
```diff
@@ -0,0 +1,3 @@
+declare module '@typescript-eslint/eslint-plugin' {
```
saberduck: I am surprised we need this, something they could fix :(

its/ruling/src/test/expected/ts/Joust/typescript-S4325.json
```diff
@@ -0,0 +1,15 @@
+{
+'Joust:ts/Launcher.tsx':[
```
saberduck: pitty that formatting between lits and your script is different, hard to see the diff. we should fix lits to produce proper json

javascript-checks/src/test/java/org/sonar/javascript/checks/CheckListTest.java
```diff
@@ -107,6 +107,7 @@ public void testTypeScriptChecks() {
   public void testJavaScriptChecks() {
     List<Class> javaScriptChecks = CheckList.getJavaScriptChecks();
     assertThat(javaScriptChecks).isNotEmpty();
+    assertThat(javaScriptChecks).isNotEqualTo(CheckList.getAllChecks());
```
saberduck: Can we add the test that every check is either JavaScriptCheck or TypeScriptCheck, ie at least one annotation is required?

eslint-bridge/typings/typescipt-eslint.d.ts
vilchik-elena: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/tsconfig.build.json#L5 looks like they are doing this on purpose

its/ruling/src/test/expected/ts/Joust/typescript-S4325.json
vilchik-elena: I updated converter to produce the same formatting, no diff should be more meaningful

PR #1315, opened 2019-08-23
Provide syntax highlighting for TypeScript

### LINKED ISSUE #1306: Provide syntax highlighting for TypeScript

yassin-kammoun-sonarsource:
Fixes #1306


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/TypeScriptSensor.java
```diff
@@ -83,14 +88,28 @@ protected void analyze(InputFile file, SensorContext context) {
...
+    for (AnalysisResponseIssue issue : issues) {
+      new EslintBasedIssue(issue).saveIssue(context, file, checks);
+    }
+  }
+
+  private static void saveHighlights(InputFile file, SensorContext context, AnalysisResponseHighlight[] highlights) {
+    NewHighlighting highlighting = context.newHighlighting().onFile(file);
+    for (AnalysisResponseHighlight highlight : highlights) {
+      highlighting.highlight(highlight.startLine, highlight.startCol, highlight.endLine, highlight.endCol,
+        TypeOfText.valueOf(highlight.textType.toUpperCase(Locale.ENGLISH)));
```
saberduck: we should avoid this conversion by sending uppercase constants in the response directly

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServer.java
```diff
@@ -100,6 +101,14 @@ private static String fileContent(InputFile file) {
     String message;
   }
 
+  class AnalysisResponseHighlight {
+    Integer startLine;
```
saberduck: can this be `null`?  if not we should use primitive `int` type to make it clear

eslint-bridge/tests/analyzer.test.ts
```diff
@@ -46,16 +119,26 @@ describe("#analyze", () => {
...
   });
 
+  it("should report syntax highlights", () => {
+    const highlights = analyzeJavaScript({
+      filePath,
+      fileContent: codeToTest,
+      rules: [{ key: "no-all-duplicated-branches", configurations: [] }],
+    }).highlights;
+    expect(highlights).toHaveLength(10);
+    expect(highlights).toEqual(highlighting);
```
vilchik-elena: IMO we don't need this assertion, that's too detailed for this test. Just asserting number of highlighting should be good

eslint-bridge/tests/runner/highlighter.test.ts
```diff
@@ -0,0 +1,85 @@
...
+});
+
+it("should highlight strings", () => {
+  expect(actual("'str'")).toContainEqual(token(1, 0, 1, 5, "STRING"));
+  expect(actual('"str"')).toContainEqual(token(1, 0, 1, 5, "STRING"));
+
+  expect(actual("`str`")).toContainEqual(token(1, 0, 1, 5, "STRING"));
+  expect(actual("`line1\nline2`")).toContainEqual(token(1, 0, 2, 6, "STRING"));
+
+  expect(actual("`start ${x} middle ${y} end`")).toContainEqual(token(1, 0, 1, 9, "STRING"));
```
vilchik-elena: let's extract this string to make it obvious that it's the same input in these 3 assertions

eslint-bridge/src/runner/highlighter.ts
```diff
@@ -0,0 +1,81 @@
...
+  const highlights: Highlight[] = [];
+  const nodes: any[] = [...sourceCode.ast.tokens, ...sourceCode.ast.comments];
+  for (const node of nodes) {
+    let highlighted: Highlight | undefined;
+    switch (node.type) {
+      case "Keyword":
+        highlighted = highlight(node, "keyword");
+        break;
+      case "String":
+      case "Template":
```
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: Yes, I did, but I will add a test case for that.

eslint-bridge/src/runner/highlighter.ts
```diff
@@ -0,0 +1,81 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+import { AST, SourceCode } from "eslint";
+import * as ESTree from "estree";
+
+export default function getHighlighting(sourceCode: SourceCode) {
+  const highlights: Highlight[] = [];
+  const nodes: any[] = [...sourceCode.ast.tokens, ...sourceCode.ast.comments];
+  for (const node of nodes) {
+    let highlighted: Highlight | undefined;
```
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: This symbol is already used by a function `highlight`, but I will rename it.

eslint-bridge/src/runner/highlighter.ts
```diff
@@ -0,0 +1,88 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+import { AST, SourceCode } from "eslint";
+import * as ESTree from "estree";
+
+export default function getHighlighting(sourceCode: SourceCode) {
+  const highlights: Highlight[] = [];
+  for (const token of sourceCode.ast.tokens) {
+    let highlight: Highlight | undefined;
```
vilchik-elena: I think we can drop this variable and push `highlight` directly in each `case`

eslint-bridge/src/runner/highlighter.ts
```diff
@@ -0,0 +1,88 @@
...
+    }
+  }
+  for (const comment of sourceCode.ast.comments) {
+    let highlight: Highlight | undefined;
+    switch (comment.type) {
+      case "Block":
+        if (comment.value.startsWith("*")) {
+          highlight = highlightTokenOrComment(comment, "STRUCTURED_COMMENT");
+          break;
+        }
```
vilchik-elena: we fallthrough here, I suppose on purpose but that's not the most explicit way to structure this logic. Could you think of some refactoring to avoid falling through?

eslint-bridge/tests/analyzer.test.ts
```diff
@@ -28,11 +28,84 @@ const noDuplicateStringIssue = {
   secondaryLocations: [],
 };
 
-describe("#analyze", () => {
+const highlighting = [
```
vilchik-elena: this value can be dropped

vilchik-elena:
@yassin-kammoun-sonarsource I pushed one more refactoring, feel free to merge if it's ok for you

PR #1319, opened 2019-08-27
Provide copy-paste detection and size metrics for TypeScript

### LINKED ISSUE #1303: Provide copy-paste detection for TypeScript

### LINKED ISSUE #1304: Provide size metrics for TypeScript
ncloc lines, commentLines, nosonarLines, executableLines, functions number,  statements number, classes number

yassin-kammoun-sonarsource:
Fixes #1303, fixes #1304

### DIFF
--- a/eslint-bridge/package.json
+++ b/eslint-bridge/package.json
@@ -45,6 +45,7 @@
     "body-parser": "1.18.3",
     "eslint": "5.6.0",
     "eslint-plugin-sonarjs": "0.4.0",
+    "eslint-visitor-keys": "1.1.0",
     "espree": "5.0.0",
     "express": "4.16.3"
   },
--- a/eslint-bridge/src/analyzer.ts
+++ b/eslint-bridge/src/analyzer.ts
@@ -17,20 +17,24 @@
  * along with this program; if not, write to the Free Software Foundation,
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
-import { parseSourceFile, parseTypeScriptSourceFile } from "./parser";
+import { parseJavaScriptSourceFile, parseTypeScriptSourceFile, Parse } from "./parser";
 import getHighlighting, { Highlight } from "./runner/highlighter";
+import getMetrics, { Metrics, EMPTY_METRICS } from "./runner/metrics";
+import getCpdTokens, { CpdToken } from "./runner/cpd";
 import * as linter from "./linter";
 
-const EMPTY_RESPONSE: AnalysisResponse = { issues: [], highlights: [] };
+const EMPTY_RESPONSE: AnalysisResponse = {
+  issues: [],
+  highlights: [],
+  metrics: EMPTY_METRICS,
+  cpdTokens: [],
+};
 
 export interface AnalysisInput {
   filePath: string;
   fileContent: string;
   rules: Rule[];
-}
-
-export interface TypeScriptAnalysisInput extends AnalysisInput {
-  tsConfigs: string[];
+  tsConfigs?: string[];
 }
 
 // eslint rule key
@@ -43,6 +47,8 @@ export interface Rule {
 export interface AnalysisResponse {
   issues: Issue[];
   highlights: Highlight[];
+  metrics: Metrics;
+  cpdTokens: CpdToken[];
 }
 
 export interface Issue {
@@ -65,27 +71,22 @@ export interface IssueLocation {
 }
 
 export function analyzeJavaScript(input: AnalysisInput): AnalysisResponse {
-  const { filePath, fileContent, rules } = input;
-  if (fileContent) {
-    const sourceCode = parseSourceFile(fileContent, filePath);
-    if (sourceCode) {
-      return {
-        issues: linter.analyze(sourceCode, rules, filePath).issues,
-        highlights: getHighlighting(sourceCode).highlights,
-      };
-    }
-  }
-  return EMPTY_RESPONSE;
+  return analyze(input, parseJavaScriptSourceFile);
 }
 
-export function analyzeTypeScript(input: TypeScriptAnalysisInput): AnalysisResponse {
-  const { filePath, fileContent, rules, tsConfigs } = input;
-  if (fileContent) {
-    const sourceCode = parseTypeScriptSourceFile(fileContent, filePath, tsConfigs);
+export function analyzeTypeScript(input: AnalysisInput): AnalysisResponse {
+  return analyze(input, parseTypeScriptSourceFile);
+}
+
+function analyze(input: AnalysisInput, parse: Parse): AnalysisResponse {
+  if (input.fileContent) {
+    const sourceCode = parse(input.fileContent, input.filePath, input.tsConfigs);
     if (sourceCode) {
       return {
-        issues: linter.analyze(sourceCode, rules, filePath).issues,
+        issues: linter.analyze(sourceCode, input.rules, input.filePath).issues,
         highlights: getHighlighting(sourceCode).highlights,
+        metrics: getMetrics(sourceCode),
+        cpdTokens: getCpdTokens(sourceCode).cpdTokens,
       };
     }
   }
--- a/eslint-bridge/src/parser.ts
+++ b/eslint-bridge/src/parser.ts
@@ -41,7 +41,16 @@ export const PARSER_CONFIG_SCRIPT: Linter.ParserOptions = {
   sourceType: "script",
 };
 
-export function parseSourceFile(fileContent: string, filePath: string): SourceCode | undefined {
+export type Parse = (
+  fileContent: string,
+  filePath: string,
+  tsConfigs?: string[],
+) => SourceCode | undefined;
+
+export function parseJavaScriptSourceFile(
+  fileContent: string,
+  filePath: string,
+): SourceCode | undefined {
   let parseFunctions = [espree.parse, babel.parse];
   if (fileContent.includes("@flow")) {
     parseFunctions = [babel.parse];
@@ -67,7 +76,7 @@ export function parseSourceFile(fileContent: string, filePath: string): SourceCo
 export function parseTypeScriptSourceFile(
   fileContent: string,
   filePath: string,
-  tsConfigs: string[],
+  tsConfigs?: string[],
 ): SourceCode | undefined {
   try {
     // we load the typescript parser dynamically, so we don't need typescript dependency when analysing pure JS project
--- a/eslint-bridge/src/runner/metrics.ts
+++ b/eslint-bridge/src/runner/metrics.ts
@@ -1,0 +1,176 @@
+/*
+ * SonarQube JavaScript Plugin
+ * Copyright (C) 2011-2019 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+import { SourceCode } from "eslint";
+import * as estree from "estree";
+import visit from "../utils/visitor";
+
+const EXECUTABLE_NODES = [
+  "ExpressionStatement",
+  "IfStatement",
+  "LabeledStatement",
+  "BreakStatement",
+  "ContinueStatement",
+  "WithStatement",
+  "SwitchStatement",
+  "ReturnStatement",
+  "ThrowStatement",
+  "TryStatement",
+  "WhileStatement",
+  "DoWhileStatement",
+  "ForStatement",
+  "ForInStatement",
+  "DebuggerStatement",
+  "VariableDeclaration",
+  "ForOfStatement",
+];
+
+const STATEMENT_NODES = [
+  "ExpressionStatement",
+  "IfStatement",
+  "LabeledStatement",
+  "BreakStatement",
+  "ContinueStatement",
+  "WithStatement",
+  "SwitchStatement",
+  "ReturnStatement",
+  "ThrowStatement",
+  "TryStatement",
+  "WhileStatement",
+  "DoWhileStatement",
+  "ForStatement",
+  "ForInStatement",
+  "DebuggerStatement",
+  "FunctionDeclaration",
+  "VariableDeclaration",
+  "ForOfStatement",
+  "ClassDeclaration",
+];
+
+const FUNCTION_NODES = ["FunctionDeclaration", "FunctionExpression", "ArrowFunctionExpression"];
+
+const CLASS_NODES = ["ClassDeclaration", "ClassExpression"];
+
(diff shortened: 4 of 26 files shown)


eslint-bridge/src/analyzer.ts
```diff
@@ -86,6 +97,8 @@ export function analyzeTypeScript(input: TypeScriptAnalysisInput): AnalysisRespo
       return {
         issues: linter.analyze(sourceCode, rules, filePath).issues,
         highlights: getHighlighting(sourceCode).highlights,
+        metrics: getMetrics(sourceCode),
+        cpdTokens: getCpdTokens(sourceCode).cpdTokens,
```
vilchik-elena: it seems that there is more and more duplicating between `analyzeTypeScript` and `analyzeJavaScript`. Can we refactor that?
Also I don't understand why there is no `return EMPTY_RESPONSE;` at the end of this function

eslint-bridge/tests/runner/cpd.test.ts
```diff
@@ -0,0 +1,117 @@
+import getCpdTokens, { CpdToken } from "../../src/runner/cpd";
+import { parseTypeScriptSourceFile } from "../../src/parser";
+import { join } from "path";
+
+it("should not skip any token", () => {
+  const result = actual(
+    `class A {
```
vilchik-elena: I don't think we should use code from highlighting here. First there are too many tokens and it's hard to count them, second it's more interesting to test tokens specific for CPD

eslint-bridge/tests/fixtures/tsx-project/sample.lint.tsx
```diff
@@ -0,0 +1,7 @@
+function foo(b: number) {
```
vilchik-elena: I would remove content and leave comment that content of this file does not matter

eslint-bridge/tests/utils/visitor.test.ts
```diff
@@ -55,11 +55,7 @@ it("should visit a node and its children", () => {
     functions: 1,
     literals: 3,
     returns: 2,
-<<<<<<< HEAD
```
vilchik-elena: why build didn't fail.. that's weird

yassin-kammoun-sonarsource: I probably fixed it and rebased while you were reviewing.

eslint-bridge/src/analyzer.ts
yassin-kammoun-sonarsource: > it seems that there is more and more duplicating between analyzeTypeScript and analyzeJavaScript. Can we refactor that?

+1

> Also I don't understand why there is no return EMPTY_RESPONSE; at the end of this function

The return statement is there. GitHub truncated the file view for some reason.

yassin-kammoun-sonarsource: > it seems that there is more and more duplicating between analyzeTypeScript and analyzeJavaScript. Can we refactor that?

+1

> Also I don't understand why there is no return EMPTY_RESPONSE; at the end of this function

The return statement is there; GitHub truncated the file view for some reason.

eslint-bridge/package.json
```diff
@@ -41,9 +41,10 @@
   "dependencies": {
     "@typescript-eslint/parser": "2.0.0",
     "babel-eslint": "10.0.1",
-    "body-parser": "1.18.3", 
+    "body-parser": "1.18.3",
     "eslint": "5.6.0",
     "eslint-plugin-sonarjs": "0.4.0",
+    "eslint-visitor-keys": "^1.1.0",
```
saberduck: we should avoid using caret, and have exact version (makes the lock file more stable)

eslint-bridge/src/runner/metrics.ts
```diff
@@ -0,0 +1,176 @@
...
+  "DoWhileStatement",
+  "ForStatement",
+  "ForInStatement",
+  "DebuggerStatement",
+  "VariableDeclaration",
+  "ForOfStatement",
+];
+
+const STATEMENT_NODES = [
+  "ExpressionStatement" /*, "BlockStatement" ??? */,
```
saberduck: what does this comment mean?

yassin-kammoun-sonarsource: Ah, yes. I didn't know if I had to take into account BlockStatement - i.e. `{}` - for the metrics. SonarTS doesn't, but it's because it does not have a node subtype for this. Should I ignore block statements ?

vilchik-elena: @yassin-kammoun-sonarsource indeed, I believe we should not count it (for JS we also don't count it, it's in CounterVisitor.java)

eslint-bridge/src/utils/visitor.ts
```diff
@@ -0,0 +1,49 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+import * as estree from "estree";
+import * as evk from "eslint-visitor-keys";
+
+export default function visit(root: estree.Node, visitor: (node: estree.Node) => void): void {
```
vilchik-elena: I'm not sure that `visitor` is a good name here. How about `callBack`?

eslint-bridge/src/utils/visitor.ts
```diff
@@ -0,0 +1,49 @@
...
+  const stack: estree.Node[] = [root];
+  const toVisit = [];
+  while (stack.length) {
+    const node = stack.pop() as estree.Node;
+    toVisit.push(node);
+    stack.push(...childrenOf(node).filter(Boolean));
```
vilchik-elena: what does `.filter(Boolean)` do?

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/TypeScriptSensor.java
```diff
@@ -91,6 +106,7 @@ protected void analyze(InputFile file, SensorContext context) {
       AnalysisResponse response = eslintBridgeServer.analyzeTypeScript(request);
       saveIssues(file, context, response.issues);
       saveHighlights(file, context, response.highlights);
+      saveMetrics(file, context, response.metrics);
```
vilchik-elena: move it before saving issues, otherwise `NOSONAR` will not work (btw having IT to test this is a good idea)

eslint-bridge/src/analyzer.ts
```diff
@@ -74,28 +71,19 @@ export interface IssueLocation {
...
+function analyze(input: AnalysisInput, parser: Parser): AnalysisResponse {
+  if (input.fileContent) {
+    const sourceCode = parser(input.fileContent, input.filePath, input.tsConfigs);
     if (sourceCode) {
       return {
-        issues: linter.analyze(sourceCode, rules, filePath).issues,
+        issues: linter.analyze(sourceCode, input.rules, input.filePath).issues,
         highlights: getHighlighting(sourceCode).highlights,
         metrics: getMetrics(sourceCode),
         cpdTokens: getCpdTokens(sourceCode).cpdTokens,
```
vilchik-elena: I just realized that we compute all this data not only for TS, but for JS as well.
That means we throw away this data in JavaScriptSensor. May be we can consider this or next week saving those metrics for JS.

eslint-bridge/src/parser.ts
```diff
@@ -41,6 +41,12 @@ export const PARSER_CONFIG_SCRIPT: Linter.ParserOptions = {
   sourceType: "script",
 };
 
+export type Parser = (
```
vilchik-elena: I think it should be verb as a name

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServer.java
```diff
@@ -37,7 +38,7 @@
 
   AnalysisResponse analyzeJavaScript(AnalysisRequest request) throws IOException;
 
-  AnalysisResponse analyzeTypeScript(TypeScriptAnalysisRequest request) throws IOException;
+  AnalysisResponse analyzeTypeScript(AnalysisRequest request) throws IOException;
```
vilchik-elena: I don't get why we should change this file (as not sending tsconfigs for JS, will just make it `undefined` on Nodejs side, which is ok for us)

eslint-bridge/src/utils/visitor.ts
yassin-kammoun-sonarsource: Node.js style? I like it.

eslint-bridge/src/utils/visitor.ts
yassin-kammoun-sonarsource: Specifically, `.filter(Boolean)` filters out falsy values (`null`, `undefined`, `0`, `false`, `''`). It was shorter than using a lambda with explicit `null` and `undefined` checks.

Now, the idea behind this filter is to get rid of `undefined`/`null` children of a given node. For example, the children of an `if` node without an `else` clause would be the `then` node and `null`.

I think I will move this filter in `childrenOf`; makes more sense to return children that actually exist.

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/TypeScriptSensor.java
yassin-kammoun-sonarsource: Out of curiosity, could you tell me why `NOSONAR ` won't work?

> (btw having IT to test this is a good idea)

(Testing the `NOSONAR `thing only or the whole set of metrics ?)

eslint-bridge/src/analyzer.ts
yassin-kammoun-sonarsource: We do indeed. I planned to do this while I was working on the syntax highlighting (we even discussed it), but I didn't. I am actually wondering if we might face some problems. If I'm not wrong, we currently parse JavaScript source files from both plugin and eslint-bridge which means that both sides compute highlights, metrics, cpd, etc. I don't have the full picture of how things completely work yet, but is it possible that a given JS file is not sent to, analysed by eslint-bridge? If this is the case, this would mean that those files will never have highlights, metrics, cpd, etc until we fully migrate the analysis to eslint-bridge.

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServer.java
yassin-kammoun-sonarsource: I am not sure I understand your comment. I just wanted to have a single DTO on both side. This is why I dropped `TypeScriptAnalysisRequest` and added a `tsConfigs` field to `AnalysisRequest`. This field will be empty for JS requests (it would be actually be better to make it `null` so, as you said, Node.js will make it `undefined`) and possibly non-empty for TS requests. Does it make sense ? Do you still prefer to keep `TypeScriptAnalysisRequest` ?

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/TypeScriptSensor.java
vilchik-elena: I think we should test all metrics, you can extend existing `TypeScriptAnalysisTest`.
NOSONAR will not work in this case because filtering of issues by NOSONAR lines happens when saving issues.

eslint-bridge/src/analyzer.ts
vilchik-elena: I don't think we have such files, set of files analyzed with ESLint and with Java frontend should be the same. I will create tickets to rely on ESLint frontend for JS to generate metrics and highlighting

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServer.java
vilchik-elena: Ok, so you just wanted to make it more consistent with what's on NodeJS side. In other words it is not required (which was not clear to me), but indeed that probably makes sense.

eslint-bridge/tests/utils/visitor.test.ts
```diff
@@ -0,0 +1,63 @@
...
+    `function factorial(n: number): number {
+      if (n < 2) {
+        return 1;
+      }
+      return n * factorial(n - 1);
+    }`,
+    "foo.ts",
+    [],
+  );
+  const visited = {
```
vilchik-elena: I think you should have a better test, asserting the order of children, e.g.:
```
  visit(sourceCode.ast, node =>  list.push(node.type + " " + node.loc.start.line));
```
If you do so, you will see that the order is not so nice, I think we should improve here. 
Also doing so will reveal that we don't visit types (I'm not sure how critical it's but that was unexpected to me)

yassin-kammoun-sonarsource: > Also doing so will reveal that we don't visit types (I'm not sure how critical it's but that was unexpected to me)

I expected that since we visit AST from `estree` which does not include type information.

its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/TypeScriptAnalysisTest.java
```diff
@@ -64,9 +64,22 @@ public void test() throws Exception {
     orchestrator.executeBuild(build);
 
     SearchRequest request = new SearchRequest();
-    request.setComponentKeys(Collections.singletonList(TSPROJECT)).setRules(ImmutableList.of("typescript:S3923"));
+    request
```
vilchik-elena: You you didn't test NOSONAR

its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/TypeScriptAnalysisTest.java
```diff
@@ -64,9 +64,22 @@ public void test() throws Exception {
...
     assertThat(issuesList).hasSize(1);
     assertThat(issuesList.get(0).getLine()).isEqualTo(2);
+
+    assertThat(Tests.getMeasureAsInt(TSPROJECT, "ncloc")).isEqualTo(7);
+    assertThat(Tests.getMeasureAsInt(TSPROJECT, "classes")).isEqualTo(0);
+    assertThat(Tests.getMeasureAsInt(TSPROJECT, "functions")).isEqualTo(1);
+    assertThat(Tests.getMeasureAsInt(TSPROJECT, "statements")).isEqualTo(4);
+    assertThat(Tests.getMeasureAsInt(TSPROJECT, "comment_lines")).isEqualTo(1);
+
+    assertThat(Tests.getMeasureAsDouble(TSPROJECT, "duplicated_lines")).isEqualTo(0.0);
```
vilchik-elena: It worths testing properly, create 2nd identical file. Then it will be better assert metrics only for one file, not entire project (and btw in Tibor PR he adds one more file, so it will be useful even there)

its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/TypeScriptAnalysisTest.java
```diff
@@ -58,52 +58,25 @@ public void test() throws Exception {
...
-      .setSourceEncoding("UTF-8")
-      .setSourceDirs(".")
-      .setProjectDir(PROJECT_DIR);
-
-    Tests.setProfile(projectKey, "eslint-based-rules-profile", "ts");
-
-    TestUtils.npmInstall(PROJECT_DIR);
-
-    orchestrator.executeBuild(build);
+    assertThat(Tests.getMeasureAsInt(projectKey + ":sample.lint.ts", "ncloc")).isEqualTo(7);
```
vilchik-elena: extract into variable file key

vilchik-elena:
I think we can squash this PR into single PR

PR #1323, opened 2019-08-28
Fix typescript installation in ITs

saberduck:


its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/SonarLintTestCustomNodeJS.java
```diff
@@ -82,8 +82,8 @@ public void should_use_set_nodejs() throws Exception {
       .addInputFile(inputFile)
       .putAllExtraProperties(properties).build();
     sonarlintEngine.analyze(configuration, issues::add, null, null);
+    sonarlintEngine.stop();
```
vilchik-elena: why did you add this call?

Jenkinsfile
```diff
@@ -45,7 +45,9 @@ pipeline {
             label 'linux'
           }
           steps {
-            runITs("ruling","LATEST_RELEASE")            
+            nodejs(configId: 'npm-artifactory', nodeJSInstallationName: 'NodeJS latest') {
```
vilchik-elena: I don't understand why we need this

saberduck: we need to have NodeJS available for ruling

its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/SonarLintTestCustomNodeJS.java
saberduck: because if sonarlintEngine adds something to the log after the call to analyze it can throw `ConcurrentModificationException` when we are checking the logs (you can't iterate the collection when it is modified). This is causing false positives on the QA

Jenkinsfile
vilchik-elena: how it was working before?)

saberduck: it wasn't, sometimes we were lucky that we were on agent with NodeJS and sometimes it failed

PR #1325, opened 2019-08-28
Use NodeJS provided via Jenkinsfile

saberduck:
… of node

### DIFF
--- a/eslint-bridge/pom.xml
+++ b/eslint-bridge/pom.xml
@@ -47,7 +47,6 @@
             </goals>
             <configuration>
               <nodeVersion>${node.version}</nodeVersion>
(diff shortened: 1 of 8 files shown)


eslint-bridge/pom.xml
```diff
@@ -47,7 +47,6 @@
             </goals>
             <configuration>
               <nodeVersion>${node.version}</nodeVersion>
-              <npmVersion>${npm.version}</npmVersion>
```
vilchik-elena: Are you sure we don't need it anymore?

saberduck: yes, if it's not defined it will use the one provided with node, which is better than installing some other version.

PR #1331, opened 2019-08-29
Report issue for ParsingError rule when (and only) failed to parse with ESLint frontend

### LINKED ISSUE #1330: Report issue for ParsingError rule when (and only) failed to parse with ESLint frontend

vilchik-elena:
Fixes #1330


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/AbstractEslintSensor.java
```diff
@@ -45,6 +50,9 @@
   final Rule[] rules;
   final JavaScriptChecks checks;
 
+  // parsingErrorRuleKey equals null if ParsingErrorCheck is not activated
+  protected RuleKey parsingErrorRuleKey = null;
```
saberduck: why protected? Also, it can be final with little change

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/AbstractEslintSensor.java
```diff
@@ -56,6 +64,11 @@
 
     this.eslintBridgeServer = eslintBridgeServer;
     this.analysisWarnings = analysisWarnings;
+
+    checks.all().stream()
+      .filter(check -> check instanceof ParsingErrorCheck)
+      .findFirst()
+      .ifPresent(parsingErrorCheck -> parsingErrorRuleKey = checks.ruleKeyFor(parsingErrorCheck));
```
saberduck: if we rewrite this into expression like `.map(checks::ruleKeyFor).orElse(null);` we can make field `final` which is nicer.

PR #1336, opened 2019-09-03
Don't send file content to NodeJS process in analysis request

saberduck:


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServer.java
```diff
@@ -57,22 +57,11 @@
 
     AnalysisRequest(InputFile file, Rule[] rules, @Nullable List<String> tsConfigs) {
       this.filePath = file.absolutePath();
-      this.fileContent = fileContent(file);
-      if (this.fileContent.startsWith("#!")) {
```
vilchik-elena: hm-hm, how ESLint will now parse js files with shebang? (I'm surprised nothing fails in ITs due to this)
Should we move this logic to eslint-bridge, to where we read file?

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/JavaScriptEslintBasedSensor.java
```diff
@@ -63,8 +63,8 @@ protected void analyze(InputFile file, SensorContext context) {
       LOG.debug("Skipping analysis of Vue.js file {}", file.uri());
       return;
     }
-    AnalysisRequest analysisRequest = new AnalysisRequest(file, rules);
     try {
+      AnalysisRequest analysisRequest = new AnalysisRequest(file, rules);
```
vilchik-elena: did you do this move because it can throw or just for a better flow?

saberduck: mostly for consistency with other sensor, and also it could throw in the future when we provide content for SonarLint

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServer.java
saberduck: I added the test on the eslint-bridge side, and it seems to be working without issue. So I am not sure why this was needed originally, but it seems that it's not necessary anymore

eslint-bridge/src/analyzer.ts
```diff
@@ -44,7 +45,7 @@ export const SYMBOL_HIGHLIGHTING_RULE = {
 
 export interface AnalysisInput {
   filePath: string;
-  fileContent: string;
+  fileContent: string | undefined;
```
yassin-kammoun-sonarsource: You can use the optional operator instead, i.e. `fileContent?:`.

eslint-bridge/src/analyzer.ts
```diff
@@ -98,27 +99,28 @@ export function analyzeTypeScript(input: AnalysisInput): AnalysisResponse {
...
-      };
-    } else {
-      return {
-        ...EMPTY_RESPONSE,
-        parsingError: result,
-      };
-    }
+  let fileContent = input.fileContent;
+  if (!fileContent) {
+    fileContent = fs.readFileSync(input.filePath, { encoding: "utf8" });
```
yassin-kammoun-sonarsource: An exception is raised if the file does not exist. Are we okay with that ?

eslint-bridge/src/analyzer.ts
saberduck: I prefer to be explicit that the property is not set, I believe it makes the test a bit more clean

eslint-bridge/src/analyzer.ts
saberduck: I was thinking about this, the basic assumption is that it should not happen. If it happens anyway, it will be handled by default error handling of express and we will print the error in the console. This is ok imo

PR #1342, opened 2019-09-03
Add rule S107 (based on 'max-params') for TS/JS

### LINKED ISSUE #1339: Add rule S107 (based on 'max-params') for TS/JS
As this rule for JS has a legacy key (`ExcessiveParameterList`), we should consider changing the key to an RSPEC based.

vilchik-elena:
Fixes #1339


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/rules/JavaScriptRulesDefinition.java
```diff
@@ -29,24 +28,20 @@
...
       .setName(CheckList.REPOSITORY_NAME);
 
-    RuleMetadataLoader ruleMetadataLoader = new RuleMetadataLoader(METADATA_LOCATION, JavaScriptProfilesDefinition.SONAR_WAY_JSON, sonarRuntime);
+    RuleMetadataLoader ruleMetadataLoader = new RuleMetadataLoader(METADATA_LOCATION, JavaScriptProfilesDefinition.SONAR_WAY_JSON);
     ruleMetadataLoader.addRulesByAnnotatedClass(repository, CheckList.getJavaScriptChecks());
 
     NewRule commentRegularExpression = repository.rule("CommentRegularExpression");
     commentRegularExpression.setTemplate(true);
 
+
```
yassin-kammoun-sonarsource: Superfluous newline.

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/rules/JavaScriptRulesDefinitionTest.java
```diff
@@ -58,10 +54,7 @@ public void sonarlint() {
   }
 
   private RulesDefinition.Repository buildRepository() {
-    JavaScriptRulesDefinition rulesDefinition = new JavaScriptRulesDefinition(
-      SonarRuntimeImpl.forSonarQube(Version.create(7, 9),
```
yassin-kammoun-sonarsource: Why was it there in the first place ?

vilchik-elena: @yassin-kammoun-sonarsource we needed SQ version for rules definition for security-hotspot rules. This concept was added in some version of SQ and if we would use it on an earlier SQ, it would fail

PR #1352, opened 2019-09-09
Add rule S113 (based on 'eol-last') for TS/JS

### LINKED ISSUE #1343: Add rule S113 (based on 'eol-last') for TS/JS
As this rule for JS has a legacy key (MissingNewlineAtEndOfFile), we should consider changing the key to an RSPEC based.

yassin-kammoun-sonarsource:
Fixes #1343 

For the record, the ESLint-based implementation for this rule differs from ours in the reported location. We used to report issues at the beginning of the file while ESLint reports them at the end.


javascript-checks/src/main/java/org/sonar/javascript/checks/MissingNewlineAtEndOfFileCheck.java
```diff
@@ -19,28 +19,36 @@
...
 
-  private static final String MESSAGE = "Add a new line at the end of this file.";
+  private static final String DEFAULT_PRESENCE = "always";
 
-  @Override
-  public void visitScript(ScriptTree tree) {
-    if (tree.items() != null) {
-      int lastLine = tree.EOFToken().line();
-      int lastTokenLine = tree.items().lastToken().endLine();
+  @RuleProperty(
```
vilchik-elena: why did you decide to add this parameter? we don't need it, just pass `always` in `configurations()`

javascript-checks/src/main/java/org/sonar/javascript/checks/MissingNewlineAtEndOfFileCheck.java
```diff
@@ -19,28 +19,36 @@
...
+    key = "presence",
+    description = "Presence (or absence) of at least one newline at the end of non-empty files",
```
quentin-jaquier-sonarsource: This description is not clear for me, you "always" check the presence or absence?
From what I understood, we can use "never", to report when a new line is present, I don't think this rule should report such cases!

quentin-jaquier-sonarsource: Never mind, you already removed the property, it's better this way!

PR #1353, opened 2019-09-09
Add rule S108 (based on 'no-empty') for TS/JS

### LINKED ISSUE #1340: Add rule S108 (based on 'no-empty') for TS/JS
As this rule for JS has a legacy key (`EmptyBlock`), we should consider changing the key to an RSPEC based.

yassin-kammoun-sonarsource:
Fixes #1340


javascript-checks/src/main/java/org/sonar/javascript/checks/EmptyBlockCheck.java
```diff
@@ -19,33 +19,36 @@
...
+    defaultValue = "" + DEFAULT)
+  boolean allowEmptyCatch = DEFAULT;
 
-    if (!parent.is(KindSet.FUNCTION_KINDS, Kind.CATCH_BLOCK) && tree.statements().isEmpty() && !hasComment(tree.closeCurlyBraceToken())) {
-      addIssue(tree, MESSAGE);
-    }
-    super.visitBlock(tree);
+  @Override
+  public List<Object> configurations() {
+    return Collections.singletonList(allowEmptyCatch);
```
vilchik-elena: we need an object here

its/ruling/src/test/expected/ts/angular/typescript-S108.json
```diff
@@ -0,0 +1,8 @@
+{
+'angular:packages/compiler/src/expression_parser/parser.ts':[
```
vilchik-elena: it's weird to have last commit in the PR to have `wip` message

yassin-kammoun-sonarsource: Indeed... I pushed it yesterday before leaving and forgot about it.

PR #1354, opened 2019-09-09
Add cognitive complexity metric (TS) and rule (TS/JS) based on 'cognitive-complexity' from eslint-plugin-sonarjs

### LINKED ISSUE #1081: Make CognitiveComplexity(S3776) eslint-based rule (cognitive-complexity) (TS/JS)
Currently we don't have a way to support secondary issue locations for eslint-based rules.
Since those are fundamental for cognitive complexity, we will replace this rule when we will implement secondary locations for eslint-based rules.

### LINKED ISSUE #1326: Provide cognitive complexity metrics for TypeScript

vilchik-elena:
Fixes #1326 
Fixes #1081


eslint-bridge/src/linter.ts
```diff
@@ -49,14 +55,16 @@ export function analyze(
   sourceCode: SourceCode,
   inputRules: Rule[],
   fileUri: string,
-  additionalRule?: { ruleId: string; ruleModule: ESLintRule.RuleModule },
+  ...additionalRules: AdditionalRule[]
 ) {
-  if (additionalRule) {
-    linter.defineRule(additionalRule.ruleId, additionalRule.ruleModule);
+  if (additionalRules.length > 0) {
```
yassin-kammoun-sonarsource: We can remove the `if` statement.

eslint-bridge/src/linter.ts
```diff
@@ -110,16 +118,19 @@ function removeIrrelevantProperties(eslintIssue: Linter.LintMessage): Issue | nu
...
 ) {
   const ruleConfig: Linter.Config = { rules: {}, parserOptions: { sourceType: "module" } };
   inputRules.forEach(inputRule => {
     const ruleModule = linter.getRules().get(inputRule.key);
     ruleConfig.rules![inputRule.key] = ["error", ...getRuleConfig(ruleModule, inputRule)];
   });
 
-  if (additionalRule) {
-    ruleConfig.rules![additionalRule.ruleId] = ["error"];
+  if (additionalRules.length > 0) {
```
yassin-kammoun-sonarsource: We can remove the `if` statement here too.

eslint-bridge/src/analyzer.ts
```diff
@@ -123,15 +124,54 @@ function analyze(input: AnalysisInput, parse: Parse): AnalysisResponse {
...
+    input.filePath,
+    SYMBOL_HIGHLIGHTING_RULE,
+    COGNITIVE_COMPLEXITY_RULE,
+  ).issues;
+  const highlightedSymbols = getHighlightedSymbols(issues);
+  const metrics = getMetrics(sourceCode);
+  metrics.cognitiveComplexity = getCognitiveComplexity(issues);
+  return {
+    issues,
+    highlights: getHighlighting(sourceCode).highlights,
```
saberduck: wdyt about extracting all properties to variables in this object literal?

eslint-bridge/src/analyzer.ts
```diff
@@ -123,15 +124,54 @@ function analyze(input: AnalysisInput, parse: Parse): AnalysisResponse {
...
+  } else {
+    console.log("DEBUG Failed to retrieve symbol highlighting from analysis results");
+    return [];
+  }
+}
+
+// exported for testing
+export function getCognitiveComplexity(issues: Issue[]) {
+  const issue = findAndRemoveFirstIssue(issues, COGNITIVE_COMPLEXITY_RULE_ID);
+  if (issue && !isNaN(Number(issue.message))) {
```
saberduck: can we use `Number.isNan` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN ?

eslint-bridge/src/linter.ts
```diff
@@ -49,14 +55,16 @@ export function analyze(
   sourceCode: SourceCode,
   inputRules: Rule[],
   fileUri: string,
-  additionalRule?: { ruleId: string; ruleModule: ESLintRule.RuleModule },
+  ...additionalRules: AdditionalRule[]
```
saberduck: some comment about purpose of this extra parameter could be helpfull

eslint-bridge/src/analyzer.ts
vilchik-elena: In fact I believe we should use just `isNaN` (`Number.isNan` results false if argument is not number)

vilchik-elena: in fact both these function in TS require number as argument, so it's ok this way

PR #1358, opened 2019-09-09
Add rule S103 (based on 'max-len') for TS/JS

### LINKED ISSUE #1337: Add rule S103 (based on 'max-len') for TS/JS
As this rule for JS has a legacy key (`LineLength`), we should consider changing the key to an RSPEC based.

quentin-jaquier-sonarsource:
Fixes #1337


its/ruling/src/test/expected/js/js-project/javascript-S103.json
```diff
@@ -2,13 +2,6 @@
 'js-project:angular-1.0.3.js':[
 14531,
 ],
-'js-project:babylon/src/util/identifier.js':[
-25,
-37,
-39,
```
quentin-jaquier-sonarsource: 3 issues removed since the implementation of the rule on ESlint is not reporting string literals.

its/ruling/src/test/expected/js/js-project/javascript-S103.json
```diff
@@ -2,13 +2,6 @@
...
-56,
-58,
```
quentin-jaquier-sonarsource: These two issues are preceded by 
```
// eslint-disable-next-line comma-spacing
```

and therefore not reported by eslint.

its/ruling/src/test/expected/js/js-project/javascript-S103.json
yassin-kammoun-sonarsource: If we really want ESLint to consider string literals, we can do so by explicitly setting the option `ignoreStrings` to `false`. Is it worth doing it ?

saberduck: I think this is good, let's keep like it is

its/ruling/src/test/expected/js/js-project/javascript-S103.json
saberduck: this will be fix by PR #1356 , depending on who merges first you will need to update

its/ruling/src/test/expected/js/js-project/javascript-S103.json
quentin-jaquier-sonarsource: In fact, these issues disappeared because of the first comment:
```
/* eslint max-len: 0 */
```
And not because the rule is ignoring strings! (`ignoreStrings` is false by default...)

Luckily, this is not a problem, this means that the rule behave the same way as the old implementation, we can keep it that way.

PR #1360, opened 2019-09-10
Add rule S1066 (based on 'eslint-plugin-sonarjs/no-collapsible-if') for TS

### LINKED ISSUE #1345: Add rule S1066 (based on 'eslint-plugin-sonarjs/no-collapsible-if') for TS
JS implementation is already using eslint-based rule

yassin-kammoun-sonarsource:
Fixes #1345


yassin-kammoun-sonarsource:
eslint-plugin-sonarjs reports issue location from outer `if` token while TSLint reports location from inner `if`'s one. Do we want to adapt eslint-plugin-sonarjs's collapsible-if rule ?

its/ruling/src/test/expected/ts/Joust/typescript-S1066.json
```diff
@@ -0,0 +1,14 @@
+{
```
quentin-jaquier-sonarsource: You could have put the addition of the expected typescript ruling in a separate commit to be able to spot the differences in a glance!

vilchik-elena:
implementations of TS and JS were different, so there is no way to make everyone happy here. I think it's ok to keep as it is.

PR #1363, opened 2019-09-10
Add rule S104 (based on 'max-lines') for TS/JS

### LINKED ISSUE #1338: Add rule S104 (based on 'max-lines') for TS/JS

quentin-jaquier-sonarsource:
Fixes #1338


quentin-jaquier-sonarsource:
Two kinds of differences:

1. We are now reporting the issue on the first line, and not the line 0.
2. We skip blank lines and comments to match as closely as possible the [ncloc metrics](https://docs.sonarqube.org/latest/user-guide/metric-definitions/).

javascript-checks/src/test/java/org/sonar/javascript/checks/TooManyLinesInFileCheckTest.java
```diff
@@ -1,44 +1,46 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+import org.junit.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class TooManyLinesInFileCheckTest {
+
+  private Map<String, Object> expectedConfigurationsMap = new HashMap<>();
```
vilchik-elena: can be local var

javascript-checks/src/test/java/org/sonar/javascript/checks/TooManyLinesInFileCheckTest.java
```diff
@@ -1,44 +1,46 @@
...
+
+  @Test
+  public void testDefault() {
```
vilchik-elena: `default` can be removed from name

PR #1365, opened 2019-09-10
Split Ruling IT into 2 stages

vilchik-elena:


Jenkinsfile
```diff
@@ -40,14 +40,22 @@ pipeline {
             runITsNoSubmodules("plugin","LATEST_RELEASE[7.9]")
           }
         }
-        stage('ruling/LATEST_RELEASE/linux') {
+        stage('ruling/JS/LATEST_RELEASE/linux') {
           agent {
-            label 'multicpu'
+            label 'linux'
```
saberduck: let's use `multicpu` also for JS

PR #1366, opened 2019-09-10
Support TypeScript analysis in SonarLint context

### LINKED ISSUE #1302: Support TypeScript analysis in SonarLint context

saberduck:
Fixes #1302

### DIFF
--- a/eslint-bridge/src/analyzer.ts
+++ b/eslint-bridge/src/analyzer.ts
@@ -17,19 +17,20 @@
  * along with this program; if not, write to the Free Software Foundation,
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
-import { parseJavaScriptSourceFile, parseTypeScriptSourceFile, Parse } from "./parser";
+import { Parse, parseJavaScriptSourceFile, parseTypeScriptSourceFile } from "./parser";
 import getHighlighting, { Highlight } from "./runner/highlighter";
-import getMetrics, { Metrics, EMPTY_METRICS } from "./runner/metrics";
+import getMetrics, { EMPTY_METRICS, Metrics } from "./runner/metrics";
 import getCpdTokens, { CpdToken } from "./runner/cpd";
 import * as linter from "./linter";
 import { SourceCode } from "eslint";
 import {
   HighlightedSymbol,
-  symbolHighlightingRuleId,
   rule as symbolHighlightingRule,
+  symbolHighlightingRuleId,
 } from "./runner/symbol-highlighter";
 import * as fs from "fs";
 import { rules as sonarjsRules } from "eslint-plugin-sonarjs";
+import * as path from "path";
 
 const COGNITIVE_COMPLEXITY_RULE_ID = "internal-cognitive-complexity";
 
@@ -113,7 +114,8 @@ function analyze(input: AnalysisInput, parse: Parse): AnalysisResponse {
   if (!fileContent) {
     fileContent = fs.readFileSync(input.filePath, { encoding: "utf8" });
   }
-  const result = parse(fileContent, input.filePath, input.tsConfigs);
+  const filePath = path.normalize(input.filePath);
+  const result = parse(fileContent, filePath, input.tsConfigs);
   if (result instanceof SourceCode) {
     return analyzeFile(result, input);
   } else {
--- a/eslint-bridge/tests/analyzer.test.ts
+++ b/eslint-bridge/tests/analyzer.test.ts
@@ -1,8 +1,8 @@
 import {
   analyzeJavaScript,
   analyzeTypeScript,
-  getHighlightedSymbols,
   getCognitiveComplexity,
+  getHighlightedSymbols,
 } from "../src/analyzer";
 import { join } from "path";
 import * as fs from "fs";
@@ -149,6 +149,25 @@ describe("#analyzeTypeScript", () => {
     expect(issues).toHaveLength(2);
     expect(issues).toContainEqual(noOneIterationIssue);
     expect(issues).toContainEqual(noDuplicateStringIssue);
+  });
+
+  it("should normalize provided path", () => {
+    let result = analyzeTypeScript({
+      filePath: __dirname + "/./fixtures/ts-project/sample.lint.ts",
+      fileContent: "true ? 42 : 42",
+      rules: [{ key: "no-all-duplicated-branches", configurations: [] }],
+      tsConfigs: [tsConfig],
+    });
+    console.log(result);
+    expect(result.issues).toHaveLength(1);
+
+    result = analyzeTypeScript({
+      filePath: __dirname + "/././fixtures/ts-project/sample.lint.ts",
+      fileContent: "true ? 42 : 24",
+      rules: [{ key: "no-all-duplicated-branches", configurations: [] }],
+      tsConfigs: [tsConfig],
+    });
+    expect(result.issues).toHaveLength(0);
   });
 
   it("should report syntax highlights", () => {
--- a/eslint-bridge/tests/analyzer.test.ts
+++ b/eslint-bridge/tests/analyzer.test.ts
@@ -158,7 +158,6 @@ describe("#analyzeTypeScript", () => {
       rules: [{ key: "no-all-duplicated-branches", configurations: [] }],
       tsConfigs: [tsConfig],
     });
-    console.log(result);
     expect(result.issues).toHaveLength(1);
 
     result = analyzeTypeScript({
@@ -167,6 +166,7 @@ describe("#analyzeTypeScript", () => {
       rules: [{ key: "no-all-duplicated-branches", configurations: [] }],
       tsConfigs: [tsConfig],
     });
+    // fileContent doesn't have the issue anymore, without path normalization we receive the AST from the first request
     expect(result.issues).toHaveLength(0);
   });
 
--- a/its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/SonarLintTest.java
+++ b/its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/SonarLintTest.java
@@ -21,7 +21,10 @@ package com.sonar.javascript.it.plugin;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import org.junit.AfterClass;
 import org.junit.BeforeClass;
@@ -35,6 +38,7 @@ import org.sonarsource.sonarlint.core.client.api.standalone.StandaloneAnalysisCo
 import org.sonarsource.sonarlint.core.client.api.standalone.StandaloneGlobalConfiguration;
 import org.sonarsource.sonarlint.core.client.api.standalone.StandaloneSonarLintEngine;
 
+import static java.util.Collections.singleton;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.tuple;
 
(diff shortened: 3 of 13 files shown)


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
```diff
@@ -145,10 +148,13 @@ private void initNodeCommand(SensorContext context, File scriptFile) throws IOEx
...
-    getTypeScriptLocation(context).ifPresent(typeScriptLocation -> {
-      LOG.debug("Using typescript at: '{}'", typeScriptLocation);
-      nodeCommandBuilder.addToNodePath(typeScriptLocation.toAbsolutePath());
-    });
+    Optional<Path> typeScriptLocation = getTypeScriptLocation(context);
+    if (typeScriptLocation.isPresent()) {
+      LOG.debug("Using TypeScript at: '{}'", typeScriptLocation.get());
+      nodeCommandBuilder.addToNodePath(typeScriptLocation.get().toAbsolutePath());
+    } else {
+      LOG.debug("TypeScript dependency was not found, TypeScript analysis might fail.");
```
vilchik-elena: how about being absolutely explicit here and say `TypeScript dependency was not found inside project directory, TypeScript will be looked based on NODE_PATH; if NODE_PATH does not reference directory with TypeScript, analysis will fail.`

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
```diff
@@ -258,11 +264,15 @@ public void stop() {
     clean();
   }
 
-  private static Optional<Path> getTypeScriptLocation(SensorContext context) throws IOException {
-    Optional<String> typeScriptLocationProperty = context.config().get(TYPESCRIPT_DEPENDENCY_LOCATION_PROPERTY);
+  private Optional<Path> getTypeScriptLocation(SensorContext context) throws IOException {
```
vilchik-elena: how about passing `context.baseDir()` instead of whole context? that way it will be more explicit what is input data

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
```diff
@@ -258,11 +264,15 @@ public void stop() {
...
+    // we have to use global Configuration and not SensorContext#config to lookup typescript set from vscode extension
+    // see https://jira.sonarsource.com/browse/SLCORE-250
+    Optional<String> typeScriptLocationProperty = configuration.get(TYPESCRIPT_DEPENDENCY_LOCATION_PROPERTY);
     if (typeScriptLocationProperty.isPresent()) {
+      LOG.debug("TypeScript location set via property {}={}", TYPESCRIPT_DEPENDENCY_LOCATION_PROPERTY, typeScriptLocationProperty.get());
```
vilchik-elena: do you think it will be better to use `info` level here? I'm not sure how sonarlint displays debug logs

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/JavaScriptEslintBasedSensorTest.java
```diff
@@ -356,6 +361,27 @@ public void should_not_create_parsing_issue_when_no_rule() throws IOException {
...
+    ArgumentCaptor<AnalysisRequest> captor = ArgumentCaptor.forClass(AnalysisRequest.class);
+    createSensor().execute(ctx);
+    verify(eslintBridgeServerMock).analyzeJavaScript(captor.capture());
+    assertThat(captor.getValue().fileContent).isEqualTo("if (cond)\n" +
+      "doFoo(); \n" +
+      "else \n" +
+      "doFoo();");
+
+    clearInvocations(eslintBridgeServerMock);
+    ctx = SensorContextTester.create(tempFolder.newDir());
```
vilchik-elena: how about using existing `SensorContextTester` but reseting its runtime? that way it will be more explicit what we are testing here

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/TypeScriptSensorTest.java
```diff
@@ -267,6 +270,27 @@ public void should_raise_a_parsing_error_without_line() throws IOException {
...
+    ArgumentCaptor<AnalysisRequest> captor = ArgumentCaptor.forClass(AnalysisRequest.class);
+    createSensor().execute(ctx);
+    verify(eslintBridgeServerMock).analyzeTypeScript(captor.capture());
+    assertThat(captor.getValue().fileContent).isEqualTo("if (cond)\n" +
+      "doFoo(); \n" +
+      "else \n" +
+      "doFoo();");
+
+    clearInvocations(eslintBridgeServerMock);
+    ctx = SensorContextTester.create(tempFolder.newDir());
```
vilchik-elena: same here

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
saberduck: ok, I will update. But even your formulation is not exactly precise - typescript was also not found via property in sonarlint context, and there is bunch of other locations as described in https://nodejs.org/api/modules.html#modules_loading_from_the_global_folders . So to be precise means to have Tolstoy novel in the log message (which few people will read anyway) :)

eslint-bridge/src/analyzer.ts
```diff
@@ -17,19 +17,20 @@
  * along with this program; if not, write to the Free Software Foundation,
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
-import { parseJavaScriptSourceFile, parseTypeScriptSourceFile, Parse } from "./parser";
```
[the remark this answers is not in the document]
saberduck: IntelliJ orders alphabetically by default, sorry

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
saberduck: they are displayed in the output window

saberduck: I changed to `info` in the caller, I think that's sufficient for now

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/JavaScriptEslintBasedSensorTest.java
saberduck: I think it's better to create fresh context, because the one we used has already measures reported and other changed state. This way I am getting "default" context without specifying the runtime at all.

eslint-bridge/src/analyzer.ts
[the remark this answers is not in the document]
vilchik-elena: I think we can configure prettier to do that

eslint-bridge/src/analyzer.ts
```diff
@@ -113,7 +114,8 @@ function analyze(input: AnalysisInput, parse: Parse): AnalysisResponse {
   if (!fileContent) {
     fileContent = fs.readFileSync(input.filePath, { encoding: "utf8" });
   }
-  const result = parse(fileContent, input.filePath, input.tsConfigs);
+  const filePath = path.normalize(input.filePath);
```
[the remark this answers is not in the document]
saberduck: added test

eslint-bridge/src/analyzer.ts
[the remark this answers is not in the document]
saberduck: not sure, see https://github.com/prettier/prettier/issues/2277#issuecomment-311216366 , but there is eslint rule about this https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md

eslint-bridge/tests/analyzer.test.ts
```diff
@@ -151,6 +151,25 @@ describe("#analyzeTypeScript", () => {
...
   });
 
+  it("should normalize provided path", () => {
+    let result = analyzeTypeScript({
+      filePath: __dirname + "/./fixtures/ts-project/sample.lint.ts",
+      fileContent: "true ? 42 : 42",
+      rules: [{ key: "no-all-duplicated-branches", configurations: [] }],
+      tsConfigs: [tsConfig],
+    });
+    console.log(result);
```
vilchik-elena: forgotten?

eslint-bridge/tests/analyzer.test.ts
```diff
@@ -151,6 +151,25 @@ describe("#analyzeTypeScript", () => {
...
+    expect(result.issues).toHaveLength(1);
+
+    result = analyzeTypeScript({
+      filePath: __dirname + "/././fixtures/ts-project/sample.lint.ts",
```
vilchik-elena: I don't get why we don't have issue here. should not it be normalized?

eslint-bridge/src/analyzer.ts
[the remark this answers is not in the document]
saberduck: btw this is a bug on their side, I will try to report it

eslint-bridge/tests/analyzer.test.ts
saberduck: fileContent is different, that's the point of the test, without path normalization we would have an issue, because we would get AST from first invocation

eslint-bridge/tests/analyzer.test.ts
saberduck: 🤦‍♂

eslint-bridge/tests/analyzer.test.ts
vilchik-elena: ah:) I didn't notice the difference

its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/SonarLintTest.java
```diff
@@ -96,6 +100,40 @@ public void should_run_eslint_based_rules() throws Exception {
...
+    StandaloneGlobalConfiguration sonarLintConfig = StandaloneGlobalConfiguration.builder()
+      .addPlugin(Tests.JAVASCRIPT_PLUGIN_LOCATION.getFile().toURI().toURL())
+      .setSonarLintUserHome(temp.newFolder().toPath())
+      .setLogOutput((formattedMessage, level) -> logs.add(formattedMessage))
+      .setExtraProperties(properties)
+      .build();
+    sonarlintEngine = new StandaloneSonarLintEngineImpl(sonarLintConfig);
+
+    ClientInputFile inputFile = TestUtils.prepareInputFile(baseDir, "foo.ts", "true ? 42 : 42");
+    // we have to provide tsconfig.json
```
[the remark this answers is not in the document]
saberduck: We need to do some work to enable it, it's not possible yet

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
```diff
@@ -145,13 +149,23 @@ private void initNodeCommand(SensorContext context, File scriptFile) throws IOEx
...
-      LOG.debug("Using typescript at: '{}'", typeScriptLocation);
-      nodeCommandBuilder.addToNodePath(typeScriptLocation.toAbsolutePath());
-    });
+    if (shouldDetectTypeScript(context.fileSystem())) {
+      Optional<Path> typeScriptLocation = getTypeScriptLocation(context.fileSystem().baseDir());
+      if (typeScriptLocation.isPresent()) {
+        LOG.info("Using TypeScript at: '{}'", typeScriptLocation.get());
+        nodeCommandBuilder.addToNodePath(typeScriptLocation.get().toAbsolutePath());
+      } else {
+        LOG.info("TypeScript dependency was not found inside project directory, NodeJs will search TypeScript using " +
```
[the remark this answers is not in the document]
saberduck: I will fix this in another PR

PR #1378, opened 2019-09-11
Rely on visitor-keys from SourceCode

vilchik-elena:


eslint-bridge/tests/utils/visitor.test.ts
```diff
@@ -11,14 +12,18 @@ it("should visit a node and its children", () => {
...
+  visit(sourceCode, node => visited.push(node.type + " " + node.loc.start.line));
   expect(visited).toEqual([
     "Program 1",
     "FunctionDeclaration 1",
     "Identifier 1",
     "Identifier 1",
+    "TSTypeAnnotation 1",
+    "TSNumberKeyword 1",
+    "TSTypeAnnotation 1",
+    "TSNumberKeyword 1",
```
yassin-kammoun-sonarsource: Very nice to see that we also visit those nodes.

PR #1385, opened 2019-09-11
Rely on ESLint frontend to generate metrics and highlighting for JavaScript

### LINKED ISSUE #1327: Rely on ESLint frontend to generate metrics and highlighting for JavaScript
In #1303, #1304, #1305, #1306, #1307 and #1326 we added computation of different metrics and highlighting for TS based on ESLint frontend. These metrics are already computed for JS as well, yet there are ignored by JS sensor. We should save it for JS and drop all the code computing them based on Java frontend.

yassin-kammoun-sonarsource:
Fixes #1327 

There are three differences with regard to generating metrics between JavaScript frontend and ESLint-bridge (originally SonarTS's implementation):

* ESLint-bridge counts both class and function declarations as statements while JavaScript frontend doesn't.
* ESLint-bridge ignores empty statements and counts labeled ones while JavaScript frontend does the opposite.
* ESLint-bridge counts empty comments `// <empty>` while JavaScript frontend doesn't.
* JavaScript frontend considers plugin's property `sonar.javascript.ignoreHeaderComments` for comment metrics while ESLint-bridge doesn't.


eslint-bridge/tests/runner/metrics.test.ts
```diff
@@ -23,21 +24,80 @@ it("should return lines of code and comment lines", () => {
...
+    //   
+
+    /*  
+        */
+       
+    /**   */`,
+      true,
+    ),
+  ).toEqual([5, 6]);
+  expect(
```
vilchik-elena: how about empty line between assertions to improve readability?

quentin-jaquier-sonarsource:
I just had a quick look at the PR and saw that you still have 10 code smells, you should probably have a look at them!

eslint-bridge/tests/runner/metrics.test.ts
```diff
@@ -23,21 +24,80 @@ it("should return lines of code and comment lines", () => {
...
+    ),
+  ).toEqual([3, 4, 5]);
+  expect(
+    comments(
+      `/* header */
+    x;
+    y; /*
+    some coment
+    */`,
+      false,
```
vilchik-elena: add test with `false` and no header comments

eslint-bridge/tests/runner/metrics.test.ts
vilchik-elena: I would even split these assertions in difference tests so that you could give them titles.

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/JavaScriptPlugin.java
```diff
@@ -83,6 +83,9 @@
   public static final String TS_LCOV_REPORT_PATHS = "sonar.typescript.lcov.reportPaths";
   public static final String TS_LCOV_REPORT_PATHS_DEFAULT_VALUE = "";
 
+  public static final String TS_IGNORE_HEADER_COMMENTS = "sonar.typescript.ignoreHeaderComments";
```
vilchik-elena: Drop this property. It didn't exist, so it's a new feature for TS and we better rely on one propery.

PR #1386, opened 2019-09-11
Handle missing typescript more gracefully

### LINKED ISSUE #1382: Handle missing typescript more gracefully
```
 LOG.info("Install TypeScript in the project directory or use NODE_PATH env. variable to set TypeScript location, " +
        "if it's located outside of project directory.");
```

saberduck:
Fixes #1382


its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/TypeScriptAnalysisTest.java
```diff
@@ -109,6 +110,27 @@ public void should_use_custom_tsconfig() throws Exception {
 
   }
 
+  @Test
+  public void test_missing_typescript() throws Exception {
```
vilchik-elena: what if order of tests is different and it will not be last?
In https://github.com/SonarSource/SonarJS/pull/1384 I created a test project without typescript. You might use it

Issue #1377, opened 2019-09-11
Parse vue files in ESLint-based sensor


vilchik-elena:
Looks like standard eslint parser is able to parse Vue files.

PR #1393, opened 2019-09-12
Add rule S1441 (based on 'quotes') for TS/JS

### LINKED ISSUE #1375: Add rule S1441 (based on 'quotes') for TS/JS
Rule has legacy rule key `SingleQuote`

yassin-kammoun-sonarsource:
Fixes #1375

### Changes in the rule

* JSX attributes are not ignored anymore.

### Changes in ruling

* Issues for `.vue` files will no longer be reported until #1377 is done.


javascript-checks/src/main/java/org/sonar/javascript/checks/SingleQuoteStringLiteralsCheck.java
```diff
@@ -19,32 +19,41 @@
...
+@Rule(key = "S1441")
+public class SingleQuoteStringLiteralsCheck extends EslintBasedCheck {
+
+  private static final boolean DEFAULT = true;
 
-  private static final String MESSAGE = "Replace double quotes by single quotes";
+  @RuleProperty(
+    key = "singleQuotes",
+    description = "Set to true to require single quotes, false for double quotes.",
+    defaultValue = "" + DEFAULT)
```
quentin-jaquier-sonarsource: You should not add parameters, the [RSPEC](https://jira.sonarsource.com/browse/RSPEC-1441) states that you should use single quotes, not that you can choose between two!

yassin-kammoun-sonarsource: Ah, indeed. I looked at the [TypeScript's RSPEC](https://jira.sonarsource.com/browse/RSPEC-4206) for this rule instead.

quentin-jaquier-sonarsource: Alternatively, you can also change the RSEPC, just make sure that everything refers to the same!

yassin-kammoun-sonarsource: @vilchik-elena What do you suggest ?

vilchik-elena: I think we can add parameter (otherwise it kills the rule for TS users who configured it for double quotes). That way it will be a new feature for JS. And for sure description should be updated (like in TS)

javascript-checks/src/main/java/org/sonar/javascript/checks/SingleQuoteStringLiteralsCheck.java
```diff
@@ -19,32 +19,41 @@
...
+import org.sonar.javascript.checks.annotations.TypeScriptRule;
+import org.sonarsource.analyzer.commons.annotations.DeprecatedRuleKey;
 
 @JavaScriptRule
-@Rule(key = "SingleQuote")
-public class SingleQuoteStringLiteralsCheck extends DoubleDispatchVisitorCheck {
+@TypeScriptRule
+@DeprecatedRuleKey(ruleKey = "SingleQuote")
+@Rule(key = "S1441")
+public class SingleQuoteStringLiteralsCheck extends EslintBasedCheck {
```
saberduck: Can we update class name to reflect the possibility to configure the rule?

PR #1440, opened 2019-09-16
Add S1440 (based on 'eqeqeq') for TS/JS

### LINKED ISSUE #1410: Add S1440 (based on 'eqeqeq') for TS/JS
:exclamation:  might be tricky for JS lang, as we had a lot of exclusions

vilchik-elena:
Fixes #1410 

### Changes for JS
No more reporting on `typeof x == "string"` as we know that `typeof` returns string. 

That some issues back as we were able to know type of operands thanks to Symbolic execution.


vilchik-elena:
@saberduck it was activated before.. so I believe we can keep it this way (as for TS new rule impl does not raise more)

PR #1453, opened 2019-09-16
Add rule S1656 (based on 'no-self-assign') for TS/JS

### LINKED ISSUE #1427: Add rule S1656 (based on 'no-self-assign') for TS/JS

saberduck:
Fixes #1427 

Reason for differences
* eslint implementation doesn't consider mutating functions on arrays `sort` and `reverse`, ie. `a = a.sort()` is not raised
* eslint implementations detects equivalent expressions which differ due to parenthesis `a = (a)`
* `no-self-assign` doesn't handle member expressions using `this`, so  `this.x = this.x` is not detected


its/ruling/src/test/expected/ts/vscode/typescript-S1656.json
```diff
@@ -0,0 +1,52 @@
+{
+'vscode:build/lib/nls.ts':[
+355,
+],
+'vscode:src/vs/base/browser/ui/list/listWidget.ts':[
+664,
```
vilchik-elena: you are saying array mutating methods are not reported but these issues are exactly that

saberduck: my bad, expectations were not updated correctly, that's why QA failed. These issues are not raised anymore

its/ruling/src/test/expected/ts/angular/typescript-S1656.json
```diff
@@ -0,0 +1,5 @@
+{
+'angular:packages/compiler-cli/src/path_mapped_compiler_host.ts':[
+134,
```
vilchik-elena: `sf.fileName = sf.fileName;` here and you say that we don't report on properties

saberduck: indeed, I was wrong, what's not working is detecting self assignment with `ThisExpression` as can be seen in https://github.com/eslint/eslint/blob/master/lib/rules/no-self-assign.js#L66 , maybe we can contribute a fix for that ( `lobj.type` is `ThisExpression` )

PR #1454, opened 2019-09-16
Validate dir when installing typescript in tests

saberduck:


vilchik-elena:
@saberduck can we close it?

saberduck:
@vilchik-elena Maybe we can merge it? Imo, these sanity checks make sense anyway

PR #1492, opened 2019-09-17
Add rule S3504 (based on 'no-var') for TS/JS

### LINKED ISSUE #1448: Add rule S3504 (based on 'no-var') for TS/JS

yassin-kammoun-sonarsource:
Fixes #1448

[This unit test](https://github.com/SonarSource/SonarJS/blob/4adfa8c78b500dc4193157426cc54cd795cc94a5/sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/JavaScriptSensorTest.java#L203) is kind of annoying. Whenever we migrate the rule that is used here, we need to find another one that hasn't been migrated yet. Any idea on how to avoid that? Should we drop the test?


vilchik-elena:
@yassin-kammoun-sonarsource I think we should put there a rule which we know will be migrated in the last place. For example S3757

yassin-kammoun-sonarsource:
@vilchik-elena I forgot to run the ruling tests for JavaScript. Running them lead to an interesting situation as some issues were not raised anymore.

* [repl.js](https://github.com/SonarSource/javascript-test-sources/blob/79128d2a36552f24da36b20c5c01063fc7b907dd/src/jest/packages/jest-repl/src/cli/repl.js#L14) uses TypeScript's typing and `declare` construct. ESLint parser fails to parse it, but Babel's manages to do so. However, Babel's represents such a statement as a `DeclareVariable` node instead of a `VariableDeclaration` one. Because of that, ESLint's `no-var` rule never visits such a node, and thus does not raise any issue. As a result, three issues ended up not being reported.
* [2-const.js](https://github.com/SonarSource/javascript-test-sources/blob/79128d2a36552f24da36b20c5c01063fc7b907dd/src/ecmascript6/ecmascript6-today/2-const.js) declares the symbol `PI` more than once. While trying to parse those declarations, ESLint parser produces a syntax error: `Identifier 'PI' has already been declared`. In this case also, Babel's parser succeeds in parsing the file. Even though the returned AST seems well-formed, the issue is still not raised, though. I have to admit I have no idea why on this one. I tried the rule locally with the same code, and the issue is raised.

Do we really want to keep using Babel's parser ? It's really weird to parse JavaScript code with Babel's parser when ESLint's parser fails, and then feed the parsed trees to ESLint's analyzer.

PR #1501, opened 2019-09-17
Add rule S2228 (based on 'no-console') for TS/JS

### LINKED ISSUE #1434: Add rule S2228 (based on 'no-console') for TS/JS

quentin-jaquier-sonarsource:
Fixes #1434 

### Changes in the rule
* extra issues when console is used in others context than a console call.

Ex:
```
console.log = myFunction();
console.warn.mock.calls[0][0];
```

Ideally, we don't want to report such issues, but the rule is not enabled by default, and using similar code means that he will eventually use console and therefore not enable the rule in the first place.
So I think we can keep it as it is.


its/ruling/src/test/expected/js/js-project/javascript-S106.json
```diff
@@ -308,12 +315,12 @@
...
 175,
 ],
+'js-project:ecmascript6/sonar-web/src/main/js/helpers/handlebars-helpers.js':[
+14,
+],
 'js-project:ecmascript6/templating/scripts/read-pkg-url.js':[
 8,
 ],
-'js-project:google-closure-library/debug/console.js':[
-196,
```
vilchik-elena: do you why this one is removed?

quentin-jaquier-sonarsource: It is removed because the console is shadowed by the parameter of the function.

```
function f(console) {
  console.log("123")
}
```

quentin-jaquier-sonarsource:
Fixes https://github.com/SonarSource/SonarTS/issues/723

PR #1534, opened 2019-09-17
Add rule S3257 (based on '@typescript-eslint/no-inferrable-types') for TS

### LINKED ISSUE #1444: Add rule S3257 (based on '@typescript-eslint/no-inferrable-types') for TS

vilchik-elena:
Fixes #1444


javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S3257.html
```diff
@@ -0,0 +1,24 @@
+<p>Unnecessarily verbose declarations and initializations make it harder to read the code, and should be simplified. Specifically, primitive
+(<code>number</code>, <code>string</code>, and <code>boolean</code>) types should be omitted from variable and parameter declaration when they can be
```
quentin-jaquier-sonarsource: By looking at the [rule description](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-inferrable-types.md), I see that more than just the three type stated are reported!

vilchik-elena: I updated RSPEC a bit, IMO we don't need to be super detailed here.

PR #1542, opened 2019-09-18
Fix 2 issues differences

quentin-jaquier-sonarsource:


Jenkinsfile
```diff
@@ -24,14 +24,17 @@ pipeline {
     }
     stage('QA') {
       parallel {
+        /*
```
vilchik-elena: don't forget to bring back tests

quentin-jaquier-sonarsource:
To enable the ongoing work to continue in good condition, we will temporary exclude this file from the analysis, and investigate the problem separately.

PR #1544, opened 2019-09-18
Fix flaky test - SonarLintTestCustomNodeJS

saberduck:
This failure happens because we are running the analysis twice in the test, we cannot simply stop the engine, because the second analysis would fail. By using concurrent collection we can avoid `ConcurrentModificationException` even if logs are modified asynchronously.


its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/SonarLintTestCustomNodeJS.java
```diff
@@ -90,6 +91,7 @@ public void should_use_set_nodejs() throws Exception {
 
   @Test
   public void should_log_failure_only_once() throws Exception {
+    List<Issue> issues = new ArrayList<>();
```
saberduck: this additional assertion is just to improve the test a bit, not related to the flakiness of the test

PR #1545, opened 2019-09-18
Add rule S3353 (based on 'prefer-const') for TS/JS

### LINKED ISSUE #1445: Add rule S3353 (based on 'prefer-const') for TS/JS

saberduck:
Fixes #1445 
Previous Java implementation had some undocumented exceptions which are not considered in the eslint implemenation. Rule raises a lot of issues on some projects. 

It was in Sonar way recommended in TS, but not in JS.


javascript-checks/src/test/resources/checks/UnchangedLetVariable.js
```diff
@@ -1,41 +0,0 @@
...
-
-let [x1, x2, x3] = obj;  // OK, although x2 is not updated (we make an exception for array/object binding pattern)
-x1 = 1;
-x3 = 2;
-foo(x1, x2, x3);
-
-let [x4] = obj;          // Noncompliant {{Make "x4" "const".}} (we make no exception for array/object binding pattern of size 1)
-foo(x4);
-
-let {y1, y2, y3} = obj;  // OK, although y2 is not updated (we make an exception for array/object binding pattern)
```
vilchik-elena: to be consistent we should enable ` {"destructuring": "all"}` in eslint rule.

its/ruling/src/test/expected/js/js-project/javascript-S3353.json
```diff
@@ -3,6 +3,7 @@
 51,
 ],
 'js-project:ecmascript6/Ghost/core/client/app/adapters/embedded-relation-adapter.js':[
+32,
```
vilchik-elena: it's a nice improvement. I think we should document it

its/ruling/src/test/expected/js/js-project/javascript-S3353.json
```diff
@@ -420,6 +446,8 @@
 ],
 'js-project:ecmascript6/Ghost/core/client/app/helpers/gh-count-down-characters.js':[
 10,
+11,
```
vilchik-elena: this will disappear with rule property

its/ruling/src/test/expected/js/js-project/javascript-S3353.json
vilchik-elena: by documenting I meant mentioning it in the ticket

saberduck: I added example in the RSPEC :)

quentin-jaquier-sonarsource:
I see that the rule is still in Sonar way recommended in TS, but not in JS, do you plan to keep it that way?

saberduck:
@quentin-jaquier-sonarsource thanks, I updated the profile

PR #1548, opened 2019-09-18
Add rule S3512 (based on 'prefer-template') for TS/JS

### LINKED ISSUE #1449: Add rule S3512 (based on 'prefer-template') for TS/JS

quentin-jaquier-sonarsource:
Fixes #1449

### Changes in the rule:

- Rule was in sonar-way-recommended for TS but not for JS: they are now present for both.

New issues when:

*For JS and TS:*
- more than 2 operands in concatenation expression.

*Only for JS:*
- litteral with less than 6 characters
- others specific cases not documented.


quentin-jaquier-sonarsource:
Description updated.

For the rule activation in the recommended profile, we reconsidered this rule and decided to be consistent and enabling it for both TS and JS. It is justified since templates are now available commonly for already few years.

thw0rted:
This landed in `master` almost a year ago, but I just got flagged for

````ts
        let output = result.columns.map(col => JSON.stringify(col)).join(",") + "\n";
````

Note that there's only one concatenation operator.  I'm trying to figure out if this is a bug in the rule implementation, or if the version I'm using (SQ 8.3, scanner 4.3.0, TS plugin 2.1, JS plugin 6.2.1) is too old.  If it's the former, I guess this is a bug report.  If it's the latter, I'd like to know how to tell what version this PR landed in so I can request an update from my Sonar server's owner.

I hope you don't mind my asking here, but I've made a bunch of posts in the Sonar Community over the past few weeks without getting any answers there.  If I need to make a new issue (or a post in the Community, I guess) instead, please let me know.

thw0rted:
The help text for the rule on my server says "Available Since Nov 15, 2018" and gives the example

````ts
function sayHello(name) {
  console.log("hello " + name);  // Noncompliant
}
````

which is the same example that is marked "complaint" on the [spec page](https://jira.sonarsource.com/browse/RSPEC-3512) since Sep. 2016, if I'm reading the Jira history correctly.

quentin-jaquier-sonarsource:
Hello @thw0rted 

>If I need to make a new issue (or a post in the Community, I guess) instead, please let me know.

Yes, a post in the [Community ](https://community.sonarsource.com/) is the way to go. Topics there are indeed not always answered right away but have way bigger chance to not get forgotten than a comment on this one-year-old closed PR.

thw0rted:
Thanks Quentin, I *knew* I had posted about this recently but for the life of me I couldn't remember where.  I did search for issues I had commented on using [global issue search](https://github.com/issues/), but the default filter included `is:issue` which ignored this thread because it's a PR.

I don't know if you noticed or the timing was just a coincidence, but I made [this Community post](https://community.sonarsource.com/t/template-string-rule-does-not-agree-with-original-spec/31335) just a few minutes ago.  Please, uh, like and subscribe? 😁

PR #1550, opened 2019-09-18
Add rule S4144 (based on 'eslint-plugin-sonarjs/no-identical-functions') for TS

### LINKED ISSUE #1470: Add rule S4144 (based on 'eslint-plugin-sonarjs/no-identical-functions') for TS

yassin-kammoun-sonarsource:
Fixes #1470


vilchik-elena:
I'm not so sure that issues which are removed are FPs. These differences are related to literal consideration, in JS we are strict, so if two funcs are identical but have different string literals we are not reporting. 
But let's try that, at least now we know explicitly the difference with previous rule implementation.

PR #1555, opened 2019-09-19
Add rule S3699 (based on 'eslint-plugin-sonarjs/no-use-of-empty-return-value') for TS

### LINKED ISSUE #1452: Add rule S3699 (based on 'eslint-plugin-sonarjs/no-use-of-empty-return-value') for TS

quentin-jaquier-sonarsource:
Fixes #1452


quentin-jaquier-sonarsource:
@yassin-kammoun-sonarsource My bad, I forgot to add the old expected ruling.

Both implementation are different, TS was using return type (void), while eslint-plugin-sonarjs is looking for return in functions.

It adds and remove issue and it's not trivial to describe all the differences.

@vilchik-elena, do you think it's fine to keep only one implementation for this one?

vilchik-elena:
I checked previous implementation snapshot and it seems that most of the issues are FPs (related to Promises). I suggest that we merge this PR but we create a ticket in eslint-plugin-sonarjs to improve rule based on type-checker.

quentin-jaquier-sonarsource:
[Issue created](https://github.com/SonarSource/eslint-plugin-sonarjs/issues/124) in eslint-plugin-sonarjs.

PR #1560, opened 2019-09-19
Add rules from SonarTS to the default TS profiles

saberduck:


sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/JavaScriptProfilesDefinitionTest.java
```diff
@@ -88,6 +92,9 @@ public void no_legacy_Key_in_profile_json() {
       Annotation ruleAnnotation = c.getAnnotation(Rule.class);
       return ((Rule) ruleAnnotation).key();
     }).collect(Collectors.toSet());
+    allKeys.addAll(RULES_PROVIDED_BY_SONARTS);
+    // TODO remove this 3 rules once they are implemented in SonarJS
+    allKeys.addAll(Arrays.asList("S4136", "S4139", "S4137"));
```
quentin-jaquier-sonarsource: It's done right? You can remove them.

its/ruling/src/test/java/org/sonar/javascript/it/JavaScriptRulingTest.java
```diff
@@ -54,11 +56,72 @@
 
   private static final Logger LOG = LoggerFactory.getLogger(JavaScriptRulingTest.class);
 
+  private static final Location TS_PLUGIN_LOCATION = MavenLocation.of("org.sonarsource.typescript", "sonar-typescript-plugin", "DOGFOOD");
+
+  private static final List<String> RULES_PROVIDED_BY_SONARTS = Arrays.asList(
```
quentin-jaquier-sonarsource: Does the order (and separation) matter in this list? If yes, what is the meaning? If not, you may want to sort it.

saberduck: Indeed, I sorted the list, it's going to be easier to maintain

PR #1561, opened 2019-09-19
Add rule S3854 (based on 'constructor-super') for TS

### LINKED ISSUE #1458: Add rule S3854 (based on 'constructor-super') for TS

vilchik-elena:
Fixes #1458 

No issues on ruling (which is ok as this problem produces runtime error)


javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S3854.html
```diff
@@ -12,82 +12,23 @@ <h3>Known Limitations</h3>
 </ul>
 <h2>Noncompliant Code Example</h2>
 <pre>
-class Animal {
```
vilchik-elena: I removed a lot of examples as they don't make sense for TS.

its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/ExpectedIssues.java
```diff
@@ -0,0 +1,110 @@
...
+    String ruleKey = file.toFile().getName().split("\\.")[0];
+    String fileKey = projectKey + ":" + projectDirPath.relativize(file).toString().replaceAll(Pattern.quote(File.separator), "/");
+    List<String> lines = getLines(file);
+    for (int lineNumber = 1; lineNumber <= lines.size(); lineNumber++) {
+      String currentLine = lines.get(lineNumber - 1);
+
+      if (!NON_COMPLIANT_PATTERN.matcher(currentLine).find() || currentLine.trim().startsWith("//")) {
+        // continue if no "// Noncompliant" in the line,
+        // or the entire line is commented (to not consider commented out code)
+        continue;
```
quentin-jaquier-sonarsource: Do you need this continue? Why not inverting the condition of the `if` and put line 77 inside the body?

vilchik-elena: nice idea for a rule:)

vilchik-elena: I will merge as it is as I plan to improve this test further.

PR #1564, opened 2019-09-20
Add rule S3696 (based on 'no-throw-literal') for TS/JS (new rule for JS)

### LINKED ISSUE #1451: Add rule S3696 (based on 'no-throw-literal') for TS/JS (new rule for JS)

yassin-kammoun-sonarsource:
Fixes #1451


quentin-jaquier-sonarsource:
The RSPEC needs to be updated:

- Update TypeScript sub-task to include literals and not only strings. 
- Add JavaScript sub-task.

Also make sure that both have the same Default Quality Profiles.

PR #1571, opened 2019-09-20
Upgrade typescript-eslint to 2.5.0

yassin-kammoun-sonarsource:
Changes:
* __S2933__ (`prefer-readonly`) looks like issues with properties of array or object type are excluded now
* __S3257__ (`no-inferrable-types`), FPs removed, see https://github.com/typescript-eslint/typescript-eslint/pull/918


its/ruling/src/test/expected/ts/emission/typescript-S3257.json
```diff
@@ -2,10 +2,4 @@
 'emission:src/lib/containers/gene.tsx':[
 78,
 ],
-'emission:src/lib/containers/home.tsx':[
-35,
```
vilchik-elena: optional properties are ignored now. See https://github.com/typescript-eslint/typescript-eslint/pull/918

quentin-jaquier-sonarsource:
Before continuing this effort, we should consider updating to 2.3.2 instead, since it contains a useful fix for `no-unnecessary-type-arguments`.

vilchik-elena:
Replaced by #1702

PR #1579, opened 2019-09-23
Parse vue files in ESLint-based sensor

### LINKED ISSUE #1114: Analyze Vue.js single file components in eslint-based context
Currently SonarJS is analyzing Vue.js thanks to its own parser. As we migrate all rules to eslint-based architecture, Vue.js are not supported out of the box, eslint-based rules are currently simply disabled by file extension `.vue`. 
We should find a way to make eslint-based rules to run on Vue.js single file components.

### LINKED ISSUE #1377: Parse vue files in ESLint-based sensor

yassin-kammoun-sonarsource:
Fixes #1114
Fixes #1377

The standard ESLint parser isn't actually able to parse all kind of Vue files. For example, it fails to parse Single Files Components. To overcome that, we kill two birds with one stone by using [vue-eslint-parser](https://github.com/mysticatea/vue-eslint-parser). For the record, this parser is used under the hood by Vue's official ESLint plugin, whose author is one of the main contributors. It is worth mentioning that this plugin has specific ESLint-based rules for Vue files that we may want to use also at some point.

This seems to be a safe solution and avoids us the effort of parsing ourself Vue files.


quentin-jaquier-sonarsource:
Was the goal of the last commit to parse Vue content in .js files? (may be the origin of the red QA?)
If yes, we may want to add a test for it! 
If it was just a refactoring, I prefer the other way, where you don't go though 3 parser for each .vue files!

yassin-kammoun-sonarsource:
Just refactoring, I was wondering if it was better to group them together, but I guess I just got the answer. I will undo this last commit.

PR #1584, opened 2019-09-24
Support 'sonar.typescript.node' property from SonarTS with deprecation warning

### LINKED ISSUE #1572: Support `sonar.typescript.node` property from SonarTS with deprecation warning
Support `sonar.typescript.node` property from SonarTS with deprecation warning

quentin-jaquier-sonarsource:
Fixes #1572

### DIFF
--- a/nodejs-utils/src/main/java/org/sonarsource/nodejs/NodeCommandBuilderImpl.java
+++ b/nodejs-utils/src/main/java/org/sonarsource/nodejs/NodeCommandBuilderImpl.java
@@ -211,7 +211,7 @@ class NodeCommandBuilderImpl implements NodeCommandBuilder {
         LOG.info("Using Node.js executable {} from property {}.", file.getAbsoluteFile(), usedProperty);
         return nodeExecutable;
       } else {
-        LOG.error("Provided Node.js executable file does not exist. Property '{}' was to '{}'", NODE_EXECUTABLE_PROPERTY, nodeExecutable);
+        LOG.error("Provided Node.js executable file does not exist. Property '{}' was to '{}'", usedProperty, nodeExecutable);
         throw new NodeCommandException("Provided Node.js executable file does not exist.");
       }
     }
--- a/nodejs-utils/src/main/java/org/sonarsource/nodejs/NodeCommandBuilderImpl.java
+++ b/nodejs-utils/src/main/java/org/sonarsource/nodejs/NodeCommandBuilderImpl.java
@@ -43,6 +43,7 @@ class NodeCommandBuilderImpl implements NodeCommandBuilder {
   private static final String NODE_EXECUTABLE_DEFAULT = "node";
 
   private static final String NODE_EXECUTABLE_PROPERTY = "sonar.nodejs.executable";
+  private static final String NODE_EXECUTABLE_PROPERTY_TS = "sonar.typescript.node";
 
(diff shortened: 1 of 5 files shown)


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/JavaScriptPlugin.java
```diff
@@ -217,6 +221,16 @@ private static void addTypeScriptExtensions(Context context) {
         .multiValues(true)
         .build(),
 
+      PropertyDefinition.builder(NODE_EXECUTABLE_PROPERTY_TS)
+        .defaultValue(NODE_EXECUTABLE_DEFAULT)
+        .name("Node.js executable")
+        .description("DEPRECATED - Use " + NODE_EXECUTABLE_PROPERTY + " instead. \n" +
```
yassin-kammoun-sonarsource: Just for my own curiosity, but how come `PropertyDefinition.builder(NODE_EXECUTABLE_PROPERTY)` isn't defined  ? Maybe it's somewhere else, but I couldn't find it.

nodejs-utils/src/main/java/org/sonarsource/nodejs/NodeCommandBuilderImpl.java
```diff
@@ -189,11 +190,25 @@ private String getVersion(String nodeExecutable) throws NodeCommandException {
...
+        nodeExecutable = configuration.get(NODE_EXECUTABLE_PROPERTY).get();
+      }
+
       File file = new File(nodeExecutable);
       if (file.exists()) {
-        LOG.info("Using Node.js executable {} from property {}.", file.getAbsoluteFile(), NODE_EXECUTABLE_PROPERTY);
+        LOG.info("Using Node.js executable {} from property {}.", file.getAbsoluteFile(), usedProperty);
         return nodeExecutable;
       } else {
         LOG.error("Provided Node.js executable file does not exist. Property '{}' was to '{}'", NODE_EXECUTABLE_PROPERTY, nodeExecutable);
```
yassin-kammoun-sonarsource: Can we also replace `NODE_EXECUTABLE_PROPERTY ` with `usedProperty` ? If yes, there is probably a test that needs to be updated.

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/JavaScriptPlugin.java
quentin-jaquier-sonarsource: You are correct, on my side, I understand it as it should not be set inside the UI.
It is the same for the TS property that is hidden.

@saberduck Any clue? Should we add it?

saberduck: Properties declared with `PropertyDefinition` will be available in the SQ UI, however it is **not** required for the property to be declared in order to be used.
We had feedback from security audit company that this is potentially security issue because it could be used to "hack" the machine running the scanner by setting this property by a malicious user (yeah, I know :-/ ).  Anyhow it makes little sense to configure this kind of setting from the SQ UI, that's why removed it to make everyone happy.

saberduck: So to be clear, this property should not be declared in the plugin also.

PR #1585, opened 2019-09-24
Create default configuration if no tsconfig file is found

### LINKED ISSUE #1392: Make it possible analysis of TypeScript without tsconfig
when no tsconfig is provided to ESLint, rules are run without ParserServices

saberduck:
Fixes #1392 
I refactored logic around tsconfig to separate class to make unit testing easier and also to simplify TypeScriptSensor

### DIFF
--- a/its/plugin/projects/missing-tsconfig/package-lock.json
+++ b/its/plugin/projects/missing-tsconfig/package-lock.json
@@ -1,0 +1,12 @@
+{
+  "requires": true,
+  "lockfileVersion": 1,
+  "dependencies": {
+    "typescript": {
+      "version": "3.5.3",
+      "resolved": "https://registry.npmjs.org/typescript/-/typescript-3.5.3.tgz",
+      "integrity": "sha512-ACzBtm/PhXBDId6a6sDJfroT2pOWt/oOnk4/dElG5G33ZL776N3Y6/6bKZJBFpd+b05F3Ct9qDjMeJmRWtE2/g==",
+      "dev": true
+    }
+  }
+}
--- a/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/TsConfigProvider.java
+++ b/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/TsConfigProvider.java
@@ -33,6 +33,7 @@ import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
+import org.sonar.api.SonarProduct;
 import org.sonar.api.batch.fs.FileSystem;
 import org.sonar.api.batch.fs.InputFile;
 import org.sonar.api.batch.sensor.SensorContext;
@@ -69,8 +70,7 @@ class TsConfigProvider {
         return tsconfigs;
       }
     }
-    // this should be impossible, DefaultTsConfigProvider always creates the file
-    throw new IllegalStateException("Failed to provide tsconfig.json file");
+    return emptyList();
   }
 
   static class PropertyTsConfigProvider implements Provider {
@@ -125,6 +125,10 @@ class TsConfigProvider {
 
     @Override
     public List<String> tsconfigs(SensorContext context) throws IOException {
+      if (context.runtime().getProduct() == SonarProduct.SONARLINT) {
+        // we don't support per analysis temporary files in SonarLint see https://jira.sonarsource.com/browse/SLCORE-235
+        return emptyList();
+      }
       Iterable<InputFile> inputFiles = context.fileSystem().inputFiles(TypeScriptSensor.filePredicate(context.fileSystem()));
       TsConfig tsConfig = new TsConfig(inputFiles);
       File tsconfigFile = writeToJsonFile(tsConfig);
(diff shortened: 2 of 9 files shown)


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/TypeScriptSensor.java
```diff
@@ -40,33 +32,33 @@
...
   private static final Logger LOG = Loggers.get(TypeScriptSensor.class);
   private List<String> tsconfigs;
+  private final TempFolder tempFolder;
 
   /**
    * Required for SonarLint
    */
-  public TypeScriptSensor(CheckFactory checkFactory, NoSonarFilter noSonarFilter, FileLinesContextFactory fileLinesContextFactory, EslintBridgeServer eslintBridgeServer) {
-    this(checkFactory, noSonarFilter, fileLinesContextFactory, eslintBridgeServer, null);
+  public TypeScriptSensor(CheckFactory checkFactory, NoSonarFilter noSonarFilter, FileLinesContextFactory fileLinesContextFactory, EslintBridgeServer eslintBridgeServer, TempFolder tempFolder) {
```
quentin-jaquier-sonarsource: I see that this constructor is required for SonarLint, can SonarLint provide this tempFolder? Where can I find this information?

saberduck: good question! in fact I went to check, and I found https://jira.sonarsource.com/browse/SLCORE-235 , so they dropped this ability recently :) I will rework this to use temp files from JVM, and I will also add IT which tests this feature

saberduck:
I need to change the approach here, because SonarLint doesn't allow to use `TempFolder` per analysis components, it will fail with runtime exception

```
[Error - 10:53:23] java.lang.UnsupportedOperationException: Don't create temp files during analysis
	at org.sonarsource.sonarlint.core.container.analysis.AnalysisTempFolderProvider$NoTempFilesDuringAnalysis.throwUOEFiles(AnalysisTempFolderProvider.java:56)
	at org.sonarsource.sonarlint.core.container.analysis.AnalysisTempFolderProvider$NoTempFilesDuringAnalysis.newFile(AnalysisTempFolderProvider.java:52)
	at org.sonar.plugins.javascript.eslint.TsConfigProvider$DefaultTsConfigProvider.writeToJsonFile(TsConfigProvider.java:137)
	at org.sonar.plugins.javascript.eslint.TsConfigProvider$DefaultTsConfigProvider.tsconfigs(TsConfigProvider.java:130)
	at org.sonar.plugins.javascript.eslint.TsConfigProvider.tsconfigs(TsConfigProvider.java:67)
	at org.sonar.plugins.javascript.eslint.TypeScriptSensor.tsConfigs(TypeScriptSensor.java:104)
	at org.sonar.plugins.javascript.eslint.TypeScriptSensor.analyze(TypeScriptSensor.java:94)
	at org.sonar.plugins.javascript.eslint.AbstractEslintSensor.execute(AbstractEslintSensor.java:112)
	at org.sonar.plugins.javascript.eslint.TypeScriptSensor.execute(TypeScriptSensor.java:44)
	at org.sonarsource.sonarlint.core.analyzer.sensor.SensorsExecutor.executeSensor(SensorsExecutor.java:81)
	at org.sonarsource.sonarlint.core.analyzer.sensor.SensorsExecutor.execute(SensorsExecutor.java:72)
	at org.sonarsource.sonarlint.core.container.analysis.AnalysisContainer.doAfterStart(AnalysisContainer.java:134)
	at org.sonarsource.sonarlint.core.container.ComponentContainer.startComponents(ComponentContainer.java:125)
	at org.sonarsource.sonarlint.core.container.ComponentContainer.execute(ComponentContainer.java:110)
```

saberduck:
After discussion with SonarLint, we decided that it would be better to not support analysis without `tsconfig.json` in SonarLint context

its/plugin/projects/missing-tsconfig/package-lock.json
```diff
@@ -0,0 +1,12 @@
+{
```
yassin-kammoun-sonarsource: Just to be sure, do we need to push this file ?

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/TsConfigProvider.java
```diff
@@ -0,0 +1,155 @@
...
+    private final TempFolder folder;
+
+    DefaultTsConfigProvider(TempFolder folder) {
+      this.folder = folder;
+    }
+
+    @Override
+    public List<String> tsconfigs(SensorContext context) throws IOException {
+      if (context.runtime().getProduct() == SonarProduct.SONARLINT) {
+        // we don't support per analysis temporary files in SonarLint see https://jira.sonarsource.com/browse/SLCORE-235
```
yassin-kammoun-sonarsource: We can maybe add a log here because we stop the analysis in the sensor saying we didn't find any tsconfig file, but it only happens in the context of SonarLint.

its/plugin/projects/missing-tsconfig/package-lock.json
saberduck: we have it also for other test projects, so it's ok imo

Issue #1419, opened 2019-09-13
Add rule S1526 ('no-variable-usage-before-declaration') for TS/JS


vilchik-elena:

quentin-jaquier-sonarsource:
### Changes in the rule
- We are now reporting on `const` and `let` variables used before declaration 
- We are now reporting an issue on all usages (and not once per symbol).
- We are now reporting Classes and Enums used before declaration (only if outside of a function scope).

vilchik-elena:
Reverted and renamed to `Add rule S1526 ('no-variable-usage-before-declaration' in tslint) for TS/JS`

quentin-jaquier-sonarsource:
Due to the revert, we lost the matching between the implementation behavior, and the [RSPEC](https://jira.sonarsource.com/browse/RSPEC-1526) description.

When tackling this ticket, make sure that the title (for sub-tasks JS and TS), match the implemented behavior.

nicolas-harraudeau-sonarsource:
IMHO we shouldn't try to detect `let`, `const`, `class` or `enum` with this rule. Using them before declaring them will result in a bug, whereas S1526 is a code-smell rule because using a variable declared with `var` before it is used works but is confusing.

We already have S3827 which raises an issue when a variable is undefined. We could later extend it to take into account `let` and `const` variable, as well as classes and enums.

Does it make sense?

PR #1591, opened 2019-09-26
DO-NOT-MERGE: Improve issue 1419

quentin-jaquier-sonarsource:
In #1419, we added rule S1526, after dogfooding, we realized that this rule raises 800 new issues, that may seems like false positive.

This PR tries to change the configuration of the rule, setting `variables` to false.

For information, here are the changes in the rule:
- We are now reporting on `const` and `let` variables used before declaration 
- We are now reporting an issue on all usages (and not once per symbol).
- We are now reporting Classes and Enums used before declaration, only if outside of a function scope.


nicolas-harraudeau-sonarsource:
@quentin-jaquier-sonarsource after reviewing the effect on Peach. The number of issues raised makes it more difficult to see where the variable is actually defined.

The ESLint rule is not enabled by default and now I see why. IMHO we should go back to the previous implementation for now and think about how we can improve it.

PR #1594, opened 2019-09-27
Stop analysis if TypeScript version is below 3.2.1

### LINKED ISSUE #1583: Stop analysis if TS version is  below 3.2.1
Otherwise we just print ERROR for every file, like here 

https://cirrus-ci.com/task/5738479914319872?command=analyze#L69

which doesn't make much sense for the user.

yassin-kammoun-sonarsource:
Fixes #1583

### DIFF
--- a/eslint-bridge/src/parser.ts
+++ b/eslint-bridge/src/parser.ts
@@ -22,12 +22,12 @@ import * as babel from "babel-eslint";
 import { Linter, SourceCode } from "eslint";
 import { ParsingError } from "./analyzer";
 import * as VueJS from "vue-eslint-parser";
-import * as semver from "semver";
 
 // this value is taken from typescript-estree
 // still we might consider extending this range
 // if everything which we need is working on older/newer versions
-const TYPESCRIPT_MINIMUM_VERSION = ">=3.2.1";
+const TYPESCRIPT_MINIMUM_VERSION = "3.2.1";
+const TYPESCRIPT_MAXIMUM_VERSION = "3.6.0";
 
 export const PARSER_CONFIG_MODULE: Linter.ParserOptions = {
   tokens: true,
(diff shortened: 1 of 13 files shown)


eslint-bridge/src/parser.ts
```diff
@@ -22,11 +22,12 @@ import * as babel from "babel-eslint";
 import { Linter, SourceCode } from "eslint";
 import { ParsingError } from "./analyzer";
 import * as VueJS from "vue-eslint-parser";
+import * as semver from "semver";
 
 // this value is taken from typescript-estree
 // still we might consider extending this range
 // if everything which we need is working on older/newer versions
-const COMPATIBLE_TS_VERSIONS = ">=3.2.1 <3.6.0";
```
quentin-jaquier-sonarsource: Did you check that everything was working on version >=3.6.0?

If not, you are missing to log a warning when we are above the range, since the [original error](https://github.com/SonarSource/SonarJS/pull/1384) say that:
> [...]
> SUPPORTED TYPESCRIPT VERSIONS: >=3.2.1 <3.6.0
> You may find that it works just fine, or you may not.

eslint-bridge/src/parser.ts
```diff
@@ -64,11 +65,11 @@ export function loggerFn(msg: string) {
     const currentVersionMatch = msg.match(/YOUR TYPESCRIPT VERSION: (.+)\n/);
     const currentVersion = currentVersionMatch ? currentVersionMatch[1] : "";
 
-    // to make warning coming from typescipt-eslint parser less importunate
-    console.log(
-      `WARN You are using version of TypeScript ${currentVersion} which is not officially supported; supported versions ` +
-        COMPATIBLE_TS_VERSIONS,
-    );
+    if (!semver.satisfies(currentVersion, TYPESCRIPT_MINIMUM_VERSION)) {
```
quentin-jaquier-sonarsource: Don't you think that using an external dependency is overkill here?
I may miss a case, but why not a lexicographical comparison? (provided by default when you compare two string with `>`)

yassin-kammoun-sonarsource: Indeed.

eslint-bridge/src/parser.ts
yassin-kammoun-sonarsource: Added a log.

eslint-bridge/src/parser.ts
```diff
@@ -63,12 +64,15 @@ export function loggerFn(msg: string) {
   ) {
     const currentVersionMatch = msg.match(/YOUR TYPESCRIPT VERSION: (.+)\n/);
     const currentVersion = currentVersionMatch ? currentVersionMatch[1] : "";
-
-    // to make warning coming from typescipt-eslint parser less importunate
-    console.log(
-      `WARN You are using version of TypeScript ${currentVersion} which is not officially supported; supported versions ` +
-        COMPATIBLE_TS_VERSIONS,
-    );
+    if (currentVersion > TYPESCRIPT_MAXIMUM_VERSION) {
```
quentin-jaquier-sonarsource: 3.6.0 is not supported, you can use `>=`

PR #1596, opened 2019-09-27
Split analysis by individual tsconfig files if there are more than 2

### LINKED ISSUE #1586: Analyze projects with lot of tsconfig files piece-by-piece
To avoid issues with OutOfMemory we should analyze projects with many `tsconfig.json` files by single tsconfig only

saberduck:
Fixes #1586 

Lost issues from ruling are due to fact that we skip analysis for files which don't belong to any `tsconfig.json` . We also lost parsing error issues, because those files don't fail anymore.


sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/TsConfigFileTest.java
```diff
@@ -0,0 +1,144 @@
...
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+import org.sonar.api.batch.fs.InputFile;
+import org.sonar.api.batch.fs.internal.TestInputFileBuilder;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.entry;
+
+public class TsConfigFileTest {
```
quentin-jaquier-sonarsource: From the [doc](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#details), I see a few specific behaviors that are not tested:

- > If the "files" and "include" are both left unspecified, the compiler defaults to including all TypeScript (.ts, .d.ts and .tsx)
- > If the "files" or "include" properties are specified, the compiler will instead include the union of the files included by those two properties.
- > Files included using "include" can be filtered using the "exclude" property.
- > However, files included explicitly using the "files" property are always included regardless of "exclude".

By looking at the code, it seems to work as expected, but it's a good idea to test them explicitly.

saberduck: I added more tests to improve the coverage, thanks for listing the cases!

PR #1599, opened 2019-09-30
Revert #1419: Add rule S1526 (based on `no-use-before-define`) for TS/JS

### LINKED ISSUE #1597: Revert #1419: Add rule S1526 (based on `no-use-before-define`) for TS/JS
In #1419 we enabled the rule `no-use-before-define`  as the implementation of S1526.

For one case of variable defined after its use, ESLint rule `no-use-before-define` raises an issue on every use of the variable, i.e. dozens of times. This makes the resulting code difficult to read and when there is no issue every one has to be marked as "Won't Fix".

The rule also does not point to where the definition is actually made, thus the developer has to search blindly in the code.

This ESLint rule is not enabled by default in ESLint. => "*some believe* it is best to always declare variables and functions before using them." (see [documentation](https://eslint.org/docs/rules/no-use-before-define))

Thus we should revert to previous implementation and try to improve it.

quentin-jaquier-sonarsource:
Fixes #1597


saberduck:
We need to add the rule to the lists in https://github.com/SonarSource/SonarJS/blob/issue-1597/its/ruling/src/test/java/org/sonar/javascript/it/JavaScriptRulingTest.java#L61 and https://github.com/SonarSource/SonarJS/blob/issue-1597/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/JavaScriptProfilesDefinition.java#L45

I know, it's a bit painful to maintain these lists manually, we should get rid of them rather quickly though.

PR #1612, opened 2019-10-03
Add rule S105 (`indent` in tslint) for TS/JS

### LINKED ISSUE #1486: Add rule S105 (`indent` in tslint) for TS/JS

quentin-jaquier-sonarsource:
Fixes #1486


quentin-jaquier-sonarsource:
### Changes in the rule
_For TS:_
- We are now reporting only one issue per file.

I think this is better, no need to report hundreds of issues if one file use tab, one is enough.
In addition, the issue message clearly implies that there is possibly more than one tab in the file.

eslint-bridge/tests/rules/no-tab.test.ts
```diff
@@ -0,0 +1,70 @@
...
+    },
+  ],
+  invalid: [
+    {
+      code: `\t`,
+      errors: [
+        {
+          message: "Replace all tab characters in this file by sequences of white-spaces.",
+          line: 1,
+          column: 1,
```
vilchik-elena: I don't get why we are reporting on column `0` but in test it is `1`

quentin-jaquier-sonarsource: When reporting ([doc](https://eslint.org/docs/developer-guide/working-with-rules#contextreport)):
> column - the 0-based column number at which the problem occurred.

When testing ([doc](https://eslint.org/docs/developer-guide/nodejs-api#ruletester)):

> column (number): The 1-based column number of the reported location

... 👎

PR #1613, opened 2019-10-03
Add rule S881 for TS/JS (`no-nested-incdec` in tslint)

### LINKED ISSUE #1487: Add rule S881 for TS/JS (`no-nested-incdec` in tslint)

quentin-jaquier-sonarsource:
Fixes #1487


its/ruling/src/test/expected/ts/angular/typescript-S881.json
```diff
@@ -147,10 +144,6 @@
 'angular:packages/core/testing/src/test_bed.ts':[
 442,
 ],
-'angular:packages/examples/common/ngIf/ts/module.ts':[
```
quentin-jaquier-sonarsource: Differences in angular are due to the fact that these are now ignored.

vilchik-elena: which "these" issues? what special about them that they are ignored?

eslint-bridge/tests/rules/no-nested-incdec.test.ts
```diff
@@ -0,0 +1,232 @@
...
+    {
+      code: `foo[++i] = 0;`,
+      errors: 1,
+    },
+    {
+      code: `foo[--i] = 0;`,
+      errors: 1,
+    },
+    {
+      code: `() => {if (i == 1) {
```
vilchik-elena: what is the point of this big test?

eslint-bridge/src/rules/no-nested-incdec.ts
```diff
@@ -0,0 +1,68 @@
...
+        switch (firstAncestor.type) {
+          case "ExpressionStatement":
+            return true;
+          case "ForStatement":
+            return firstAncestor.update === node;
+          case "SequenceExpression": {
+            const secondAncestor = ancestors.pop();
+            return (
+              secondAncestor !== undefined &&
+              secondAncestor.type === "ForStatement" &&
```
vilchik-elena: I didn't find a test when this condition is false

eslint-bridge/tests/rules/no-nested-incdec.test.ts
```diff
@@ -0,0 +1,232 @@
...
+    {
+      code: `foo[-i] = 0;`,
+    },
+    {
+      code: `let j = 0;
+            for (i = 0; i < 10; i++, j++, k++) {
+            }`,
+    },
+  ],
+  invalid: [
```
vilchik-elena: tbh these tests are a mess. I believe we should not mix tests with and without location assertion, it makes it unreadable. And it feels like these tests are redundant

its/ruling/src/test/expected/ts/angular/typescript-S881.json
quentin-jaquier-sonarsource: I wanted to mean that the **files** in which the issues were appearing are now ignored during the ruling.

eslint-bridge/src/rules/no-nested-incdec.ts
quentin-jaquier-sonarsource: Test added

eslint-bridge/tests/rules/no-nested-incdec.test.ts
quentin-jaquier-sonarsource: I agree, I re-organized the file.

PR #1614, opened 2019-10-03
Add rule S1821 ('no-nested-switch' in tslint) for TS/JS 

### LINKED ISSUE #1495: Add rule S1821 ('no-nested-switch' in tslint) for TS/JS (new rule for JS)

quentin-jaquier-sonarsource:
Fixes #1495

New rule for JS.


eslint-bridge/src/rules/no-nested-switch.ts
```diff
@@ -0,0 +1,38 @@
...
+
+const message = 'Refactor the code to eliminate this nested "switch".';
+
+export const rule: Rule.RuleModule = {
+  create(context: Rule.RuleContext) {
+    return {
+      "SwitchStatement SwitchStatement": function(node: estree.Node) {
+        context.report({
+          message,
+          node,
```
christophe-zurn-sonarsource: I think it's a bit overkill to highlight the whole switch statement, highlighting the `switch` keyword should be enough. (This is what is done in java & slang-based analyzers)

eslint-bridge/src/rules/no-nested-switch.ts
```diff
@@ -31,13 +31,11 @@ export const rule: Rule.RuleModule = {
...
-          context.report({
-            message,
-            node,
-            loc: switchToken.loc,
-          });
-        }
+        context.report({
+          message,
+          node,
+          loc: switchToken!.loc,
```
christophe-zurn-sonarsource: It's a bit confusing to have both the `node` and `loc` in the `ReportDescriptor` as it is defined as `type ReportDescriptorLocation = { node: ESTree.Node } | { loc: AST.SourceLocation | { line: number, column: number } };`, so I would expect only to see `loc` in this case.
However, from the test, we see that the `loc` property takes precedence over the `node` for reporting location

christophe-zurn-sonarsource:
See minor comment about `node` property of the `ReportDescriptor` object before merging.

eslint-bridge/src/rules/no-nested-switch.ts
quentin-jaquier-sonarsource: It is also stated clearly in the [doc](https://eslint.org/docs/developer-guide/working-with-rules#contextreport) aswell.
At first, I saw it as the problematic node, and the range we want to highlight, if we report only  the location, we loose the node itself.
But at the same time, in our context, we only need the location, and will not use extra-information, so I also think it's better to drop it to avoid confusion.

PR #1617, opened 2019-10-04
Add rule S1439 for JS/TS (before `label-position`)

### LINKED ISSUE #1374: Add rule S1439 for JS/TS (before `label-position`)
Rule has legacy key "LabelPlacement"

christophe-zurn-sonarsource:
Fixes #1374 

Note that rule logic changed slightly for JS implementation, as the logic taken was the one from tslint, which also consider labels on switch statements as valid.
Result is that there is one issue missing in expected JS ruling, which was a FP IMHO


its/ruling/src/test/expected/js/js-project/javascript-S1439.json
```diff
@@ -9,9 +9,6 @@
 2370,
 2377,
 ],
-'js-project:open-layers-2.12/OpenLayers/Format/XML.js':[
-773,
```
christophe-zurn-sonarsource: This was a label on a switch. Looking at the code, it was indeed a FP, as the switch label was used to break out of the switch from inside a loop

PR #1618, opened 2019-10-04
Add rule S101 ('class-name' in tslint) for TS/JS

### LINKED ISSUE #1484: Add rule S101 ('class-name' in tslint) for TS/JS
with parameter, default value the same as in JS

yassin-kammoun-sonarsource:
Fixes #1484


its/ruling/src/test/expected/js/js-project/javascript-S101.json
```diff
@@ -25,7 +25,4 @@
 'js-project:ecmascript6/watchtower.js/test/observer.spec.js':[
 132,
 ],
-'js-project:jest/types/Environment.js':[
```
yassin-kammoun-sonarsource: We fail to parse this file because it uses TypeScript `declare` construct. Hence, the issue is no longer raised.

vilchik-elena: it's not TypeScript, it's FlowJS. That's weird that babel can't parse it.

eslint-bridge/tests/rules/class-name.test.ts
```diff
@@ -7,106 +7,68 @@ const ruleTester = new RuleTester({
...
 ruleTester.run("Class and interface names should comply with a naming convention", rule, {
   valid: [
     {
       code: `
-      class MyClass {
-
-      }
+      class MyClass {}
 
       var x = class y { // Compliant, rule doesn't check class expressions
```
vilchik-elena: here as well...

PR #1622, opened 2019-10-04
Declare dependency on SonarTS plugin in manifest

saberduck:
I am not sure if we are able to write a test for this setting. I tried with orchestrator, but it's not very straightforward what to assert.


sonar-javascript-plugin/pom.xml
```diff
@@ -99,6 +99,7 @@
           <skipDependenciesPackaging>true</skipDependenciesPackaging>
           <sonarLintSupported>true</sonarLintSupported>
           <sonarQubeMinVersion>${sonarQubeMinVersion}</sonarQubeMinVersion>
+          <requirePlugins>typescript:2.1.0</requirePlugins>
```
vilchik-elena: what is `typescript` here? key of what?

vilchik-elena: @saberduck ping

saberduck: it's key of the plugin, it's computed from artifactId here https://github.com/SonarSource/sonar-packaging-maven-plugin/blob/master/src/main/java/org/sonarsource/pluginpackaging/AbstractSonarMojo.java#L90

vilchik-elena: should not it be `sonar-typescript-plugin` then? https://github.com/SonarSource/SonarTS/blob/master/sonarts-sq-plugin/sonar-typescript-plugin/pom.xml#L11

saberduck: No, when you look deeper in sonar mojo, it will remove `sonar` and `plugin` https://github.com/SonarSource/sonar-packaging-maven-plugin/blob/master/src/main/java/org/sonarsource/pluginpackaging/AbstractSonarMojo.java#L176

btw, if it was wrong, integration tests would start to fail.

PR #1624, opened 2019-10-07
Add rule S2123 ('no-useless-increment' in tslint) for TS/JS

### LINKED ISSUE #1499: Add rule S2123 ('no-useless-increment' in tslint) for TS/JS
This rule has low value, it's rare that developer writes:
```
x = x++;
```

But if we could detect this where we know that `x` is local var, that could increase rule value
```
return x++;
```

quentin-jaquier-sonarsource:
Fixes #1499


eslint-bridge/src/rules/no-useless-increment.ts
```diff
@@ -0,0 +1,62 @@
...
+            reportUpdateExpression(rhs);
+          }
+        }
+      },
+    };
+  },
+};
+
+function isLocalIdentifier(id: estree.Identifier, scope: Scope.Scope) {
+  return scope.variables.find(v => v.identifiers.find(i => i.name === id.name)) !== undefined;
```
vilchik-elena: can't we use `some` instead of `find`?

eslint-bridge/src/rules/no-useless-increment.ts
```diff
@@ -0,0 +1,62 @@
...
+      context.report({
+        message: `Remove this ${updateOperator}rement or correct the code not to waste it.`,
+        node: updateExpression,
+      });
+    }
+
+    return {
+      "ReturnStatement > UpdateExpression": function(node: estree.Node) {
+        const argument = (node as estree.UpdateExpression).argument;
+        if (argument.type === "Identifier" && isLocalIdentifier(argument, context.getScope())) {
```
christophe-zurn-sonarsource: You should also check here that it is not the prefixed operator, as
```
function f1() {
              let i = 1;
              return ++i;
            }`,
```
is allowed (it is even given as example in the compliant part of the RSPEC)

eslint-bridge/src/rules/no-useless-increment.ts
```diff
@@ -0,0 +1,62 @@
...
+          reportUpdateExpression(node as estree.UpdateExpression);
```
christophe-zurn-sonarsource: nitpick: you could extract `node as estree.UpdateExpression` as you're using it in two places

its/ruling/src/test/java/org/sonar/javascript/it/JavaScriptRulingTest.java
```diff
@@ -58,57 +58,6 @@
...
 
-  private static final List<String> RULES_PROVIDED_BY_SONARTS = Arrays.asList(
-    "S117",
-    "S125",
-    "S1110",
-    "S1121",
-    "S1226",
-    "S1451",
-    "S1526",
-    "S1533",
```
christophe-zurn-sonarsource: The 2 changes in this file would deserve to be in a separate commit. You can leave it inside this PR, but maybe you should squash the other changes so that the PR can be rebased with 2 clean commits: 1 for the rule, the other one for removing this from the ITs

PR #1625, opened 2019-10-07
Add rule S117 ('variable-name') for TS/JS (new rule for JS)

### LINKED ISSUE #1485: Add rule S117 (`variable-name` in tslint) for TS/JS (new rule for JS)
with parameter, default value corresponding to `variable-name` in tslint

yassin-kammoun-sonarsource:
Fixes #1485

(We could also raise issues on interface properties, which tslint's rule doesn't.)


its/ruling/src/test/expected/ts/TypeScript/typescript-S117.json
```diff
@@ -42,7 +42,4 @@
 720,
 726,
 ],
-'TypeScript:src/server/builder.ts':[
```
yassin-kammoun-sonarsource: The file is not analysed because it is not part of any tsconfig file.

eslint-bridge/src/rules/variable-name.ts
```diff
@@ -0,0 +1,129 @@
...
+      raiseOnInvalidIdentifier(id, "local variable", context),
+    ),
+  );
+}
+
+function checkFunction(func: FunctionLike, context: Rule.RuleContext) {
+  if (func.declare) {
+    return;
+  }
+  func.params.forEach(param =>
```
vilchik-elena: can't we directly subscribe to `Parameter`?

eslint-bridge/tests/rules/variable-name.test.ts
```diff
@@ -0,0 +1,222 @@
...
+        let PascalCase;
+        let UPPER_CASE;`,
+        options: [{ format: DEFAULT_FORMAT }],
+      },
+      {
+        code: `let [ foo, [ bar, [ baz, ...qux ] ] ] = arr;`,
+        options: [{ format: DEFAULT_FORMAT }],
+      },
+      {
+        code: `let { foo, bar: bar, baz: { qux: { quux: Quux }, ...corge } } = obj;`,
```
vilchik-elena: it will be nice to test we are not reporting on default value: `let {bar: foo = foo_bar} = obj` (same for array)

vilchik-elena: should we test we don't report on shorthand? `let {foo_foo} = obj`

javascript-checks/src/main/java/org/sonar/javascript/checks/VariableNameCheck.java
```diff
@@ -0,0 +1,63 @@
...
+import org.sonar.javascript.checks.annotations.TypeScriptRule;
+
+@JavaScriptRule
+@TypeScriptRule
+@Rule(key = "S117")
+public class VariableNameCheck extends EslintBasedCheck {
+
+  private static final String PASCAL_CAMEL_CASED = "^[_$A-Z][$A-Za-z0-9]*$";
+  private static final String LOWER_CAMEL_CASED = "^[_$a-z][$A-Za-z0-9]*$";
+  private static final String UPPER_CASED = "^[_$A-Z0-9][_$A-Z0-9]+$";
```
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: Nope.

its/ruling/src/test/expected/ts/emission/typescript-S117.json
```diff
@@ -0,0 +1,5 @@
+{
+'emission:src/lib/components/artist/shows/metadata.tsx':[
```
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: The default format does not include snake case pattern; raising this issue makes sense then. However, it's weird that SonarTS didn't use to raise it.

eslint-bridge/src/rules/variable-name.ts
```diff
@@ -0,0 +1,129 @@
...
+  declare?: boolean;
+  params: TSESTree.Parameter[];
+}
+
+export const rule: Rule.RuleModule = {
+  create(context: Rule.RuleContext) {
+    return {
+      VariableDeclaration: (node: estree.Node) =>
+        checkVariable(node as TSESTree.VariableDeclaration, context),
+      "FunctionDeclaration, FunctionExpression, ArrowFunctionExpression, TSMethodSignature, TSConstructSignatureDeclaration, TSEmptyBodyFunctionExpression": (
```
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: Good catch.

eslint-bridge/src/rules/variable-name.ts
yassin-kammoun-sonarsource: There is a way, but it doesn't work for TypeScript interfaces and empty methods. I will propose a slightly different implementation that makes easier the retrieval of variables and parameters, but we'll still need the recursive identifiers resolution for the aforementioned TypeScript constructs.

yassin-kammoun-sonarsource: Okay, it was supposed to be easier and cleaner, but it's actually uglier and more error-prone so I will keep the current approach.

For the record, and to answer the initial question, subscribing to `Parameter` like the following is not possible:

```typescript
    return {
      "Parameter": (node: estree.Node) => {
        ...
      }
    };
  },
};
```

Parameters are never visited because `Parameter` is neither a subtype of `estree.Node` nor a subtype of `TSESTree.Node`, though there is `TSESTree.Parameter`. Selectors don't help here either unfortunately.

Nevertheless, there is a way to visit parameters (and variable declarations). For instance, one can do the following:

```typescript
    return {
      "*": (node: estree.Node) => {
        const variables = context.getDeclaredVariables(node);
        for (const variable of variables) {
          for (const def of variable.defs) {
            switch (def.type) {
              case "Variable":
                ...
                break;
              case "Parameter":
                ...
                break;
            }
          }
        }
      },
    };
  },
};
```

Even though this seems convenient, there are limitations and problems from what I observed:
* `Variable` can be visited twice: one for `VariableDeclaration`, one for `VariableDeclarator`,
* Identifiers `a`, `b`, and `c` from array patterns like `let [ a [ b [ c ] ] ] = arr;` are automatically resolved, but they aren't when those appear as parameters, which means we need to explicitly resolve them (Same situation for object patterns).
* Class properties are not considered declared variables; which means we need to subscribe to `ClassProperty` and resolve explicitly property identifiers.
* Parameters from interface methods and constructor aren't considered declared variables either.

javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S117.html
```diff
@@ -0,0 +1,14 @@
...
+<code>UPPER_CASED</code>. This rule also allows leading underscore.</p>
+<h2>Noncompliant Code Example</h2>
+<pre>
+const foo_bar = 1;
+const baz_ = 2;
+</pre>
+<h2>Compliant Solution</h2>
+<pre>
+const fooBar = 1;
+const _baz = 2;{code}
```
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: Fixed

javascript-checks/src/main/java/org/sonar/javascript/checks/VariableNameCheck.java
```diff
@@ -32,11 +32,10 @@
 @Rule(key = "S117")
 public class VariableNameCheck extends EslintBasedCheck {
 
-  private static final String PASCAL_CAMEL_CASED = "^[_$A-Z][$A-Za-z0-9]*$";
-  private static final String LOWER_CAMEL_CASED = "^[_$a-z][$A-Za-z0-9]*$";
+  private static final String CAMEL_CAMEL_CASED = "^[_$A-Za-z][$A-Za-z0-9]*$";
   private static final String UPPER_CASED = "^[_$A-Z][_$A-Z0-9]+$";
 
-  private static final String DEFAULT_FORMAT = PASCAL_CAMEL_CASED + "|" + LOWER_CAMEL_CASED + "|" + UPPER_CASED;
+  private static final String DEFAULT_FORMAT = CAMEL_CAMEL_CASED + "|" + UPPER_CASED;
```
vilchik-elena: `CAMEL_CAMEL_CASED`? why 2 times camel?

eslint-bridge/src/rules/variable-name.ts
vilchik-elena: Great investigation!

PR #1626, opened 2019-10-07
Add rule S1110 for TS/JS (`no-redundant-parentheses` in tslint)

### LINKED ISSUE #1488: Add rule S1110 for TS/JS (`no-redundant-parentheses` in tslint)

christophe-zurn-sonarsource:
Fixes #1488


its/ruling/src/test/expected/ts/rxjs/typescript-S1110.json
```diff
@@ -0,0 +1,5 @@
+{
+'rxjs:src/observable/combineLatest.ts':[
+141,
```
christophe-zurn-sonarsource: This is a new issue, as the implementation looks at all kind of node, not only expression, so here we raise redundant parentheses error on TS type parameter, that indeed has an extra unneeded parenthesis
```
export function combineLatest<T, R>(...observables: Array<any | ObservableInput<any> |
                                                    Array<ObservableInput<any>> |
                                                    (((...values: Array<any>) => R)) | // Here we raise new issue
                                                    IScheduler>): Observable<R> { // ...
```

eslint-bridge/src/rules/no-redundant-parentheses.ts
```diff
@@ -0,0 +1,110 @@
...
+      {
+        // internal parameter for rules having secondary locations
+        enum: ["sonar-runtime"],
+      },
+    ],
+  },
+  create(context: Rule.RuleContext) {
+    return {
+      "*": function(node: estree.Node) {
+        checkRedundantParenthesis(context.getSourceCode(), node, context);
```
vilchik-elena: checkRedundantParentheses (plural)

eslint-bridge/src/rules/no-redundant-parentheses.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  if (tokenBefore !== null && tokenAfter !== null && isParenthesesPair(tokenBefore, tokenAfter)) {
+    let previousToken = sourceCode.getTokenBefore(tokenBefore);
+    let followingToken = sourceCode.getTokenAfter(tokenAfter);
+
+    const parent = getParent(context);
+    let skipFirst = !!parent && isParentStatementWithParentheses(node, parent);
+
+    while (
+      previousToken !== null &&
```
vilchik-elena: can't we move these nullability checks inside `isParenthesesPair`? Also I would use `!!previousToken` to make sure it's not `undefined`

eslint-bridge/src/rules/no-redundant-parentheses.ts
```diff
@@ -0,0 +1,110 @@
...
+    return {
+      "*": function(node: estree.Node) {
+        checkRedundantParenthesis(context.getSourceCode(), node, context);
+      },
+    };
+  },
+};
+
+function isParentStatementWithParentheses(node: estree.Node, parent: estree.Node): boolean {
+  return (
```
vilchik-elena: this is unreadable, how about at least split into 2 variables or functions? one checking condition another call expression.

eslint-bridge/src/rules/no-redundant-parentheses.ts
```diff
@@ -0,0 +1,110 @@
...
+  const tokenAfter = sourceCode.getTokenAfter(node);
+
+  if (tokenBefore !== null && tokenAfter !== null && isParenthesesPair(tokenBefore, tokenAfter)) {
+    let previousToken = sourceCode.getTokenBefore(tokenBefore);
+    let followingToken = sourceCode.getTokenAfter(tokenAfter);
+
+    const parent = getParent(context);
+    let skipFirst = !!parent && isParentStatementWithParentheses(node, parent);
+
+    while (
```
vilchik-elena: I don't get why we need a loop here. It will be executed twice max, right?

eslint-bridge/src/rules/no-redundant-parentheses.ts
christophe-zurn-sonarsource: If we move the null check inside the `isParenthesesPair` function, typescript compiler complains that that `previousToken` (and the others) may be null inside the `if` body, so it forces to use `!` everywhere (ex: `sourceCode.getTokenBefore(tokenBefore!);`), hence why I kept the null check inside the `if` condition.
Good point for `!!` instead of simply null check

eslint-bridge/src/rules/no-redundant-parentheses.ts
christophe-zurn-sonarsource: It will be executed once for each redundant pair of parentheses found. So if there are more than 1 additional pair, it will raise one issue for each pair that is redundant.
For instance, `(((((a)))))` will loop 4 times inside the while loop

Issue #871, opened 2017-12-10
[feature request] Native ESLint support


AMorgaut:
**I want to request a feature.**

I'd want a "Native" ESLint support from SonarJS with:

* integrated eslint launch
* integrated Rules support (with list & enable/disable from sonarqube UI)
* bonus: support of eslintrc / eslintignore

### Context

There is plenty or JS linters out there like at the very least:

- [JSLint](http://jslint.com): the original JS Linter from Douglas Crockford
- [JSHint](http://jshint.com/): the first fork due to license & rules disagreements
- [JavaScriptLint](http://www.javascriptlint.com): a more isolated mono-developer project 
- [ESLint](https://eslint.org/): The most active & extensible Linter

To me, Sonar is not only a linter but also a *Continuous Integration* platform with its integrated Lint flavor

At best, I'd love a CI tool supporting any linter and handling for them:

- direct feedback (SonarLint)
- the PASSED/FAILED validation (quality gate), 
- dashboards, 
- and history (not yet very usable in Sonarqube, but that would be another ticket for sonarqube...)

### Technical Context

Usually, Lint is done in 2 steps:

- code parsing (generates an AST: Abstract Syntax Tree)
- looks for rules conformity in this AST

Luckily, the AST format is quite standardized (cf  [ESTree spec](https://github.com/estree/estree)), even if there is few flavors

Main parsers:

- [acorn](https://github.com/ternjs/acorn) from ternjs
- [babylon](https://github.com/babel/babel/tree/master/packages/babylon) from babel, based on acorn
- [espree](https://github.com/eslint/espree) variant used by ESLint, initially based on Esprima (compatible), use acorn
- [esprima](http://esprima.org/) used by istanbul and so many others... but istanbul team moved to the [nyc](https://github.com/istanbuljs/nyc) project with its internal [babel-plugin-istanbul](https://github.com/istanbuljs/babel-plugin-istanbul) so babylon must not be far...
- [cherow](https://github.com/cherow/cherow)

Note also that most of them also support JSX and/or TypeScript syntax

What would be sad is to parse the code for each linter we want to support, but luckily again, it turns out that tools like ESLint (or Babel) provide an API that accept already generated AST

```javascript
const { SourceCode, Lint } = require("eslint");
const linter = new Linter();
const messages = linter.verify(
    new SourceCode(sourceContent, ast), 
    config, // rules, 
    filename
);
```
> https://eslint.org/docs/developer-guide/nodejs-api

[...]

PR #1628, opened 2019-10-07
Add rule S4782 ('prefer-optional' in tslint) for TS

### LINKED ISSUE #1530: Add rule S4782 ('prefer-optional') for TS

dvorsky:
Fixes #1530 

Hi! I wanted to contribute and try translating tslint rules into eslint. I wasn't sure how to approach SonarQube plugin so if you're open to merging this I would appreciate help with that!


vilchik-elena:
Hi @dvorsky, 
First, thanks for the contribution, we value that a lot!

Concerning the rule, in fact it's a controversial rule (see #871). Some users a claiming (and I believe they are right) that replacing `x: Foo | undefined` with `x? : Foo` is not equivalent. First syntax requires the property in the object, while another allows to omit it. While this difference is pretty subtle, we believe this rule does not make sense for most of people.

We decided to limit scope of the rule only with redundant syntax like this:
```
interface Person {
  pet?: Animal | undefined; // Noncompliant, "undefined" or "?" is redundant, remove one of it
}
```  

In rule description we will speak about which syntax to prefer in which situation. 

For now you can either just close this PR or adjust to report only on redundant cases. In the latter case it's probably better to wait until we will update [RSPEC 
](https://jira.sonarsource.com/browse/RSPEC-4782) with new rule scope (hopefully will be done this week)

saberduck:
Replaced by #1706 , thanks for you contribution @dvorsky, we have used the idea, but we improved the implementation to be more correct.

PR #1630, opened 2019-10-08
Add rule S4622 ('max-union-size' in tslint) for TS

### LINKED ISSUE #1526: Add rule S4622 ('max-union-size') for TS

yassin-kammoun-sonarsource:
Fixes #1526


its/ruling/src/test/expected/ts/TypeScript/typescript-S4622.json
```diff
@@ -1,7 +1,4 @@
 {
-'TypeScript:scripts/buildProtocol.ts':[
-81,
```
yassin-kammoun-sonarsource: We fail to parse this file because it uses JSX and the file extension is not `.tsx`.

eslint-bridge/tests/rules/max-union-size.test.ts
```diff
@@ -0,0 +1,139 @@
...
+      }`,
+      options: [DEFAULT_THRESHOLD],
+      errors: [
+        {
+          message: `Refactor this union type to have less than ${DEFAULT_THRESHOLD} elements.`,
+        },
+      ],
+    },
+    {
+      code: `type U = (A | B | C | D) & E;`,
```
christophe-zurn-sonarsource: I don't think we should raise an issue here, as it is explicitly mentioned in the RSPEC `This rule ignores union types part of type statement:`. This is related to the comment about `TSIntersectionType` above

eslint-bridge/src/rules/max-union-size.ts
```diff
@@ -0,0 +1,45 @@
...
+            node,
+          });
+        }
+      },
+    };
+  },
+};
+
+function isFromTypeStatement(node: TSESTree.TSUnionType): boolean {
+  return node.parent !== undefined && node.parent.type === "TSTypeAliasDeclaration";
```
christophe-zurn-sonarsource: We should recursively look for the parent type when the current parent is an `TSIntersectionType` (See comment below on unit test)

eslint-bridge/tests/parser.test.ts
```diff
@@ -107,11 +107,7 @@ describe("parseTypeScriptSourceFile", () => {
   it("should parse typescript syntax", () => {
     const file = __dirname + "/fixtures/ts-project/sample.lint.ts";
     const sourceCode = parseTypeScriptSourceFile(
-      `if (b == 0) { // Noncompliant  
-      doOneMoreThing();
-    } else {
-      doOneMoreThing();
-    }
+      `function nokFn(a: A | (B | C | D)) {}
```
christophe-zurn-sonarsource: Is this change wanted?
I understand it was not testing actual TS code before, but I think the change should be on a separate commit, as it is not directly linked to the the ticket. Side note: the name `nokFn`, does not make sense in this context, as the function can be parsed correctly here, and is OK

its/ruling/src/test/expected/ts/TypeScript/typescript-S4622.json
christophe-zurn-sonarsource: In this case, the file is actually parsed correctly, otherwise it would appear in a `typescript-ParsingError.json` file, as we have a rule tracking parsing error of the analysis.
The syntax used here is the old (1st ?) way of doing type assertion (`<MyType> val`, which is now `as MyType`). I don't know why the `TSUnionType` doesn't appear in the program AST in that case, but I'm not sure it is worth investing time to support such an old syntax

eslint-bridge/tests/parser.test.ts
yassin-kammoun-sonarsource: Sorry, this change wasn't wanted.

eslint-bridge/src/rules/max-union-size.ts
yassin-kammoun-sonarsource: We don't because SonarTS didn't (see [SonarTS's implementation](https://github.com/SonarSource/SonarTS/blob/62c6a2de708ecd4963430edff5adb5f6d15869fc/sonarts-core/src/rules/maxUnionSizeRule.ts)). We could definitely improve the rule to accommodate that (and other degenerated cases like `let x: A | (B | C | (D)))`), but I was suggested to maintain the original behaviour of the rule.

eslint-bridge/tests/rules/max-union-size.test.ts
yassin-kammoun-sonarsource: I should have mentioned that, but all test cases come from SonarTS. This particular example used to be accepted by SonarTS (see [this](https://github.com/SonarSource/SonarTS/blob/62c6a2de708ecd4963430edff5adb5f6d15869fc/sonarts-core/tests/rules/maxUnionSizeRule/maxUnionSizeRule.lint.ts#L25)). However, the RSPEC may need to clarify that point.

eslint-bridge/src/rules/max-union-size.ts
christophe-zurn-sonarsource: If it was decided not to change behavior, then ok. I guess these are corner cases anyways

PR #1631, opened 2019-10-08
Add rules S1751 and S3626 for TS/JS (S3626 new rule for JS)

### LINKED ISSUE #1494: Add rule S1751 (based on 'no-one-iteration-loop') for TS/JS

### LINKED ISSUE #1510: Add rule S3626 ('no-redundant-jump') for TS/JS (new for JS)

vilchik-elena:
Fixes #1510 and fixes #1494 

S3626:
For TS some issues requiring CFG are missing, still there are many new (probably our CFG impl was limited)

S1751:
For JS issue location changed (reported on unconditional jump before, now on the loop)


eslint-bridge/tests/rules/no-redundant-jump.test.ts
```diff
@@ -0,0 +1,136 @@
...
+        }
+        if (b) {
+          console.log("b");
+          return;
+        }
+        console.log("useful");
+      }
+
+      while (x == 1) {
+        continue; // Ok, we ignore when 1 statement
```
quentin-jaquier-sonarsource: Why this exception? I did not find any documentation related to this.

eslint-bridge/tests/rules/no-redundant-jump.test.ts
```diff
@@ -0,0 +1,136 @@
+import { RuleTester } from "eslint";
+
+const ruleTester = new RuleTester({ parserOptions: { ecmaVersion: 2018 } });
+import { rule } from "../../src/rules/no-redundant-jump";
+
+function invalid(code: string) {
```
quentin-jaquier-sonarsource: 💯

eslint-bridge/src/rules/no-redundant-jump.ts
```diff
@@ -0,0 +1,77 @@
...
+const message = "Remove this redundant jump.";
+const loops = "WhileStatement, ForStatement, DoWhileStatement, ForInStatement, ForOfStatement";
+
+export const rule: Rule.RuleModule = {
+  create(context: Rule.RuleContext) {
+    function reportIfLastStatement(node: estree.ContinueStatement | estree.ReturnStatement) {
+      const withArgument = node.type === "ContinueStatement" ? !!node.label : !!node.argument;
+      if (!withArgument) {
+        const ancestors = context.getAncestors();
+        const block = ancestors[ancestors.length - 1] as estree.BlockStatement;
```
christophe-zurn-sonarsource: You could use `getParent` from "eslint-plugin-sonarjs/lib/utils/nodes" package here for simplicity

eslint-bridge/src/rules/no-redundant-jump.ts
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    function reportIfLastStatementInsideIf(
+      node: estree.ContinueStatement | estree.ReturnStatement,
+    ) {
+      const ancestors = context.getAncestors();
+      const ifStatement = ancestors[ancestors.length - 2];
+      const upperBlock = ancestors[ancestors.length - 3] as estree.BlockStatement;
+      if (upperBlock.body[upperBlock.body.length - 1] === ifStatement) {
+        reportIfLastStatement(node as estree.ContinueStatement);
```
christophe-zurn-sonarsource: the type assertion `node as estree.ContinueStatement` is not needed here, as `node` already has correct type

eslint-bridge/tests/rules/no-redundant-jump.test.ts
vilchik-elena: Indeed I didn't document this exception, will add it to RSPEC now.
Idea is that between empty block and block containing single `continue`, second option is more readable.

eslint-bridge/src/rules/no-redundant-jump.ts
vilchik-elena: I wonder why SQ didn't report that..

PR #1633, opened 2019-10-09
Add rule S1533 ('no-primitive-wrappers') for TS/JS

### LINKED ISSUE #1426: Add rule S1533 ('use-primitive-type' in tslint) for TS/JS

christophe-zurn-sonarsource:
Fixes #1426


christophe-zurn-sonarsource:
The exception to the rule was removed in JS.

its/ruling/src/test/expected/js/js-project/javascript-S1533.json
```diff
@@ -1,4 +1,9 @@
 {
+'js-project:google-closure-library/testing/asserts.js':[
+89,
+92,
+95,
+],
```
christophe-zurn-sonarsource: These new issues in JS come from the fact the exception to the rule was removed

PR #1637, opened 2019-10-10
Add rule S4324 ('no-return-type-any') for TS

### LINKED ISSUE #1521: Add rule S4324 ('no-return-type-any' in tslint) for TS

quentin-jaquier-sonarsource:
Fixes #1521


eslint-bridge/src/rules/no-return-type-any.ts
```diff
@@ -0,0 +1,94 @@
...
+  create(context: Rule.RuleContext) {
+    const services = context.parserServices;
+    try {
+      const ts = require("typescript");
+
+      if (isRequiredParserServices(services)) {
+        let returnedExpressions: returnsArrayType = [];
+        const checker = services.program.getTypeChecker();
+
+        function allReturnTypesEqual(returns: returnsArrayType): boolean {
```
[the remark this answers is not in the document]
quentin-jaquier-sonarsource: I wanted these functions to be in the same scope as `const ts = require("typescript");`, to avoid passing it as argument with `any` type, but it's not too bad after all.

eslint-bridge/tests/RuleTesterTs.ts
```diff
@@ -58,6 +62,14 @@ class RuleTesterTs extends RuleTester {
     });
 
     super.run(name, rule, tests);
+
+    tests.invalid.map(test => {
```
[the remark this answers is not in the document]
quentin-jaquier-sonarsource: It isn't used, I forgot to remove it

eslint-bridge/src/rules/no-return-type-any.ts
```diff
@@ -0,0 +1,93 @@
...
+export const rule: Rule.RuleModule = {
+  create(context: Rule.RuleContext) {
+    const services = context.parserServices;
+
+    if (isRequiredParserServices(services)) {
+      const ts = require("typescript");
+      let returnedExpressions: ReturnedExpression[] = [];
+      return {
+        ReturnStatement(node: estree.Node) {
+          returnedExpressions.push((node as estree.ReturnStatement).argument);
```
yassin-kammoun-sonarsource: This array of returned expressions is shared across nested functions which can prevent valid issues from being raised.

Consider the following:

```typescript
function outer(): number {
  if (false) return -1;
  function inner(): any {
    return "";
  } // returnedExpressions === [-1, ""] @inner:exit
  return 0;
}
```

An issue should be raised for `inner` since its return type is `any` and its only returned expression has type `string`.

No issue is raised here because every returned expression between a nesting function declaration and a nested function declaration are visible to the latter, visible in the sense of being included in `returnedExpressions`.

quentin-jaquier-sonarsource: Nice catch! It was also reporting FP in some cases for the outer function:

```
            // OK, returns different primitive types
            function outer(x: any): any {
              if (x) return "";
              function inner(): number {}
              return 1;
            }
```

I updated the implementation to support these two cases, let me known what you think!

eslint-bridge/tests/rules/no-return-type-any.test.ts
```diff
@@ -152,5 +160,16 @@ ruleTester.run("Primitive return types should be used.", rule, {
         }`,
       errors: 1,
     },
+    {
+      code: `//Nested functions
+            function outer(): number {
+              if (false) return -1;
+              function inner(): any {
+                return ""; //Nomcompliant
+              } // returnedExpressions === [-1, ""] @inner:exit
```
yassin-kammoun-sonarsource: We can probably drop this comment since it's not the case anymore.

eslint-bridge/src/rules/no-return-type-any.ts
yassin-kammoun-sonarsource: It's perfect!

PR #1639, opened 2019-10-10
Add rule S1451 ('file-header') for TS/JS (new parameters for TS)

### LINKED ISSUE #1489: Add rule S1451 for TS/JS (`file-header` in tslint)
Old implementation of this rule for TS is relying on tslint rule `file-header`, that's why formats of input are different for JS and TS:
* for TS single parameter taking format **without** comment characters and it's considered as regular expression
* for JS there are 2 parameters:
  * first parameter is taking format **with** comment characters
  * second one specifies if first one should be considered as regular expression (not by default)

To be consistent between 2 languages both languages will have parameters as for JS. This means that for projects having this rule activated, parameter value(s) should be updated, otherwise there will be an issue on each TS file

vilchik-elena:
Fixes #1489 

❗️ For TS there is a incompatible change in parameter


eslint-bridge/src/rules/file-header.ts
```diff
@@ -0,0 +1,90 @@
...
+export const rule: Rule.RuleModule = {
+  create(context: Rule.RuleContext) {
+    const [{ headerFormat, isRegularExpression }] = context.options;
+    const expectedLines = headerFormat.split(/(?:\r)?\n|\r/);
+    let searchPattern: RegExp;
+    if (isRegularExpression) {
+      try {
+        searchPattern = new RegExp(headerFormat, "s");
+      } catch (e) {
+        console.error("Failed to compile regular expression for rule S1451");
```
saberduck: I think we should fail the analysis in such case

vilchik-elena: it's not possible, we are catching everything:) but warning in UI might be an option

PR #1648, opened 2019-10-11
Rely on TypeScript to get list of files for tsconfig

### LINKED ISSUE #1608: Support trailling comma in tsconfig
Currently, we parse tsconfig using Java Gson parser, which doesn't support trailing commas. As a consequence we will be not able to assign files to projects using such tsconfig.

We should use TypeScript API to parse tsconfig. We will need new endpoint to sent the tsconfig file to eslint-bridge and receive list of files belonging to the project.

vilchik-elena:
Fixes #1608


eslint-bridge/src/tsconfig.ts
```diff
@@ -0,0 +1,54 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+// this import will be removed after transpilation as it's used only for types
+import * as tsForTypes from "typescript";
```
saberduck: this seems a bit convoluted, we can probably just omit the type annotation on `parseConfigHost` and let the type inference do its magic :)

eslint-bridge/src/tsconfig.ts
```diff
@@ -0,0 +1,54 @@
...
+import * as path from "path";
+
+export function getFilesForTsConfig(tsConfig: string): string[] {
+  const ts = require("typescript");
+
+  const parseConfigHost: tsForTypes.ParseConfigHost = {
+    fileExists: fs.existsSync,
+    readDirectory: ts.sys.readDirectory,
+    // 'readFile' is unused for our purposes
+    readFile: file => /* istanbul ignore next */ fs.readFileSync(file, "utf8"),
```
saberduck: you will have to type the parameter like `(file: string) => ...` , if you remove the type

vilchik-elena: in fact I just dropped this property as it's not used anyway

PR #1653, opened 2019-10-11
Add rule S1541 ('cyclomatic-complexity') for TS/JS

### LINKED ISSUE #1493: Add rule S1541 ('cyclomatic-complexity') for TS/JS

yassin-kammoun-sonarsource:
Fixes #1493

### Implementation reference

SonarJS

- [FunctionComplexityVisitor](https://github.com/SonarSource/SonarJS/blob/daf09a4acf719a49a37b4bfbeaede20887d4346f/javascript-checks/src/main/java/org/sonar/javascript/checks/FunctionComplexityCheck.java)
  - depends on [ComplexityVisitor](https://github.com/SonarSource/SonarJS/blob/868e772f75bf5a5fca4d2588c06b8755bfd1e9b6/javascript-frontend/src/main/java/org/sonar/javascript/metrics/ComplexityVisitor.java)
  - depends on [FunctionDefiningModuleVisitor](https://github.com/SonarSource/SonarJS/blob/daf09a4acf719a49a37b4bfbeaede20887d4346f/javascript-frontend/src/main/java/org/sonar/javascript/metrics/FunctionDefiningModuleVisitor.java)

SonarTS

- [mccabeComplexityRule](https://github.com/SonarSource/SonarTS/blob/e84fe9e1f17984559600539b61573bfcc88a2414/sonarts-core/src/rules/mccabeComplexityRule.ts)
  - depends on [cyclomaticComplexity](https://github.com/SonarSource/SonarTS/blob/e84fe9e1f17984559600539b61573bfcc88a2414/sonarts-core/src/utils/cyclomaticComplexity.ts)

### Changes in the rule

SonarJS used to ignore function expressions passed to method calls on Angular modules:

```javascript
var moduleX = angular.module("moduleX");

moduleX.controller("Name", function() { // ignored
  var a = true && false && true;
});
```

For this to be possible, SonarJS used to annotate expressions like `angular.module(...)` with `Type.Kind.ANGULAR_MODULE`. 

We no longer ignore those function expressions because we don't have particular type information for Angular module anymore.

### Notes

#1659 was created to update the implementation of file-level cyclomatic complexity to take into account the corner cases of SonarJS's original implementation, namely AMD pattern and immediately invoked functions.


its/ruling/src/test/expected/ts/vscode/typescript-S1541.json
```diff
@@ -76,7 +76,6 @@
 ],
 'vscode:src/vs/base/browser/dom.ts':[
 64,
-271,
```
yassin-kammoun-sonarsource: Used to be a false positive.

its/ruling/src/test/expected/ts/TypeScript/typescript-S1541.json
```diff
@@ -449,9 +435,6 @@
 64,
 259,
 ],
-'TypeScript:src/server/builder.ts':[
```
yassin-kammoun-sonarsource: Not included in local tsconfig.json.

its/ruling/src/test/expected/ts/TypeScript/typescript-S1541.json
```diff
@@ -266,10 +256,6 @@
 3958,
 4118,
 ],
-'TypeScript:src/compiler/moduleNameResolver.ts':[
```
yassin-kammoun-sonarsource: Not included in local tsconfig.

eslint-bridge/src/rules/mccabe-complexity.ts
```diff
@@ -0,0 +1,293 @@
...
+
+  getComplexityTokens() {
+    return this.tokens;
+  }
+
+  private tokenOf(node: FunctionLike): ComplexityToken | undefined | null {
+    if (node.type === "FunctionExpression") {
+      const ancestors = this.context.getAncestors();
+      if (ancestors.length > 0) {
+        const parent = ancestors[ancestors.length - 1];
```
vilchik-elena: you can rely on `import { getParent } from "eslint-plugin-sonarjs/lib/utils/nodes";`

eslint-bridge/src/rules/mccabe-complexity.ts
```diff
@@ -0,0 +1,293 @@
...
+    };
+
+    visitNode(this.root);
+  }
+
+  getComplexityTokens() {
+    return this.tokens;
+  }
+
+  private tokenOf(node: FunctionLike): ComplexityToken | undefined | null {
```
vilchik-elena: there is function https://github.com/SonarSource/eslint-plugin-sonarjs/blob/master/src/utils/locations.ts#L43 which you can reuse here

javascript-checks/src/main/java/org/sonar/javascript/checks/McCabeComplexityCheck.java
```diff
@@ -0,0 +1,54 @@
...
+@JavaScriptRule
+@TypeScriptRule
+@DeprecatedRuleKey(ruleKey = "FunctionComplexity")
+@Rule(key = "S1541")
+public class McCabeComplexityCheck extends EslintBasedCheck {
+
+  private static final int DEFAULT_THRESHOLD = 10;
+
+  @RuleProperty(
+    key = "Threshold",
```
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: The [RSPEC](https://jira.sonarsource.com/browse/RSPEC-1541) mentions `Threshold`, not `maximumFunctionComplexityThreshold`. Should I still undo the change?

eslint-bridge/src/rules/mccabe-complexity.ts
```diff
@@ -0,0 +1,293 @@
...
+import * as estree from "estree";
+import { EncodedMessage, IssueLocation } from "eslint-plugin-sonarjs/lib/utils/locations";
+import visit, { childrenOf } from "../utils/visitor";
+
+type FunctionLike =
+  | estree.FunctionDeclaration
+  | estree.FunctionExpression
+  | estree.ArrowFunctionExpression;
+
+const FunctionLikeType = "FunctionDeclaration, FunctionExpression, ArrowFunctionExpression";
```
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: They are implicitly visited. Methods are represented with nodes of type `MethodDefinition` which happens to have a `value` property of type `FunctionExpression`. This essentially means that a method definition wraps an anonymous function under the hood.

eslint-bridge/tests/rules/mccabe-complexity.test.ts
```diff
@@ -0,0 +1,675 @@
...
+           //^^
+        else {}        // +0
+      }
+      `,
+      options: [DEFAULT_THRESHOLD],
+      errors: [
+        error(3, DEFAULT_THRESHOLD, [
+          sndloc(2, 6, 2, 14),
+          sndloc(4, 8, 4, 10),
+          sndloc(6, 13, 6, 15),
```
quentin-jaquier-sonarsource: I just had a quick first look: maybe you could generate these secondary locations thanks to the `//^` comments?

eslint-bridge/src/rules/mccabe-complexity.ts
```diff
@@ -0,0 +1,293 @@
...
+  constructor(private context: Rule.RuleContext) {}
+
+  visit() {
+    const sourceCode = this.context.getSourceCode();
+    visit(sourceCode, node => {
+      if (node.type === "CallExpression") {
+        this.checkForImmediatelyInvokedFunction(node.callee);
+      }
+
+      if (node.type === "NewExpression") {
```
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: I got confused because I though that `new function(){}` was like creating an instance of an anonymous class and didn't denote a regular function call.

eslint-bridge/tests/rules/mccabe-complexity.test.ts
yassin-kammoun-sonarsource: I though about it, and it would definitely make the tests more readable. However, it would be too much effort from what I observed. There are several test cases that expect more than one issue and it's kind of difficult to associate a secondary location to a particular issue.

javascript-checks/src/main/java/org/sonar/javascript/checks/McCabeComplexityCheck.java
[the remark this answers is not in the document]
vilchik-elena: in fact it's not so easy. TS rule is using `Threshold` and JS rules is using `maximumFunctionComplexityThreshold`. I think the following trick should work: create 2 java rules both relying on cyclomatic-complexity. One will be for JS, another for TS.

eslint-bridge/tests/rules/cyclomatic-complexity.test.ts
```diff
@@ -0,0 +1,623 @@
...
+    },
+    {
+      code: `
+      define([], "module name", function(){  // AMD PATTERN - OK
+        var a = true && false && true;
+      });
+      `,
+      options: [DEFAULT_THRESHOLD],
+    },
+    // TODO not supported yet
```
vilchik-elena: Could you create ticket?

eslint-bridge/tests/rules/cyclomatic-complexity.test.ts
```diff
@@ -0,0 +1,623 @@
...
+): string {
+  const encodedMessage: EncodedMessage = {
+    message: `Function has a complexity of ${complexity} which is greater than ${threshold} authorized.`,
+    cost: complexity - threshold,
+    secondaryLocations,
+  };
+  return JSON.stringify(encodedMessage);
+}
+
+function sndloc(line: number, column: number, endLine: number, endColumn: number): IssueLocation {
```
vilchik-elena: IMO it's not such a trouble to parse secondary locations. Indeed it's harder when code sample contains more than 1 issue, but I think you can just split it.  Take inspiration here https://github.com/SonarSource/SonarJS/blob/master/eslint-bridge/tests/rules/no-redundant-jump.test.ts#L6

eslint-bridge/src/rules/cyclomatic-complexity.ts
```diff
@@ -0,0 +1,202 @@
...
+        } else {
+          token = { loc: getMainFunctionTokenLocation(node, this.parent, this.context) };
+        }
+      } else {
+        switch (node.type) {
+          case "ConditionalExpression":
+            token = this.context
+              .getSourceCode()
+              .getFirstTokensBetween(node.test, node.consequent)
+              .find(token => token.value === "?");
```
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: Well spotted !

javascript-checks/src/main/java/org/sonar/javascript/checks/CyclomaticComplexityCheck.java
```diff
@@ -24,11 +24,9 @@
 
 import org.sonar.check.Rule;
 import org.sonar.check.RuleProperty;
-import org.sonar.javascript.checks.annotations.JavaScriptRule;
 import org.sonar.javascript.checks.annotations.TypeScriptRule;
 import org.sonarsource.analyzer.commons.annotations.DeprecatedRuleKey;
 
-@JavaScriptRule
```
vilchik-elena: we don't need deprecated rule key here

javascript-checks/src/main/java/org/sonar/javascript/checks/FunctionComplexityCheck.java
```diff
@@ -0,0 +1,52 @@
...
+
+import org.sonar.check.Rule;
+import org.sonar.check.RuleProperty;
+import org.sonar.javascript.checks.annotations.JavaScriptRule;
+import org.sonarsource.analyzer.commons.annotations.DeprecatedRuleKey;
+
+@JavaScriptRule
+@DeprecatedRuleKey(ruleKey = "FunctionComplexity")
+@Rule(key = "S1541")
+public class FunctionComplexityCheck extends EslintBasedCheck {
```
vilchik-elena: I think keeping names more explicit (like `CyclomaticComplexityJavaScriptCheck` and `CyclomaticComplexityTypeScriptCheck`) is better

javascript-checks/src/main/java/org/sonar/javascript/checks/FunctionComplexityCheck.java
```diff
@@ -0,0 +1,52 @@
...
+
+import java.util.Collections;
+import java.util.List;
+
+import org.sonar.check.Rule;
+import org.sonar.check.RuleProperty;
+import org.sonar.javascript.checks.annotations.JavaScriptRule;
+import org.sonarsource.analyzer.commons.annotations.DeprecatedRuleKey;
+
+@JavaScriptRule
```
vilchik-elena: add comment explaining why we need 2 rules

eslint-bridge/tests/rules/cyclomatic-complexity.test.ts
```diff
@@ -178,433 +178,310 @@ ruleTester.run("Functions should not be too complex", rule, {
...
+      return 32;
+    }
+    `,
+      10,
+    ),
   ],
 });
 
-function err(complexity: number, threshold: number, secondaryLocations: IssueLocation[]) {
+function invalid(code: string, threshold = THRESHOLD) {
```
vilchik-elena: btw if you renamed `threshold` to `THRESHOLD` to not have `threshold = threshold` I think it should be ok

vilchik-elena: you are not writing a generic library used for every rule test, you are doing it here only for your rule. So I think you could do it in a more simplistic manner. For example parse `//----` for a primary location (you know it contributes to complexity `+1`) and `//^^` for secondary (it's also always `+1`), there is even no need to support several secondaries in the same line and no need to have several issues in the same code sample.
When you write a code imagine yourself in 1h needing to fix a bug there, I think it will be pretty painful.

yassin-kammoun-sonarsource: The compiler does not allow it.

PR #1654, opened 2019-10-11
Prepare bugfix vesion 6.0.1

saberduck:


saberduck:
QA is broken because checkout of submodules in QA is  wrong , I changed the version in the branch itself

PR #1660, opened 2019-10-14
Add rule S4323 ('use-type-alias') for TS

### LINKED ISSUE #1520: Add rule S4323 ('use-type-alias' in tslint) for TS

yassin-kammoun-sonarsource:
Fixes #1520


its/ruling/src/test/expected/ts/TypeScript/typescript-S4323.json
```diff
@@ -8,7 +8,6 @@
 ],
 'TypeScript:src/compiler/factory.ts':[
 844,
-3752,
```
yassin-kammoun-sonarsource: Was a false positive; had only two occurrences.

its/ruling/src/test/expected/ts/ag-grid/typescript-S4323.json
```diff
@@ -1,12 +1,28 @@
 {
+'ag-grid:src/ts/baseFrameworkFactory.ts':[
+15,
+],
 'ag-grid:src/ts/columnController/columnController.ts':[
 46,
-59,
```
yassin-kammoun-sonarsource: Was a false positive; had only two occurrences.

its/ruling/src/test/expected/ts/rxjs/typescript-S4323.json
```diff
@@ -1,5 +1,9 @@
 {
 'rxjs:src/operator/merge.ts':[
+12,
```
yassin-kammoun-sonarsource: These are true positive.

eslint-bridge/src/rules/use-type-alias.ts
```diff
@@ -0,0 +1,69 @@
...
+        }),
+      "TSUnionType, TSIntersectionType": (node: estree.Node) => {
+        const ancestors = context.getAncestors();
+        const declaration = ancestors.find(
+          ancestor => (ancestor as TSESTree.Node).type === "TSTypeAliasDeclaration",
+        );
+        if (!declaration) {
+          const composite = (node as unknown) as TSESTree.TSUnionType | TSESTree.TSIntersectionType;
+          if (composite.types.length > TYPE_THRESHOLD) {
+            const text = context.getSourceCode().getText(node);
```
christophe-zurn-sonarsource: Using the whole node text as key, we will miss the following cases:
- same union/intersection type with different parameter order: `string | null | number;` vs `string | number | null;`
- same union/intersection type with different spacing: `string | null | number;` vs `string|null|number;`

It might be better to create a key based on the children types, for instance:
`const text = composite.types.map(typeNode => context.getSourceCode().getText(typeNode)).sort((a, b) => a.localeCompare(b)).join("|")` (example, you may use something else :) )

eslint-bridge/src/rules/use-type-alias.ts
```diff
@@ -0,0 +1,69 @@
...
+    return {
+      Program: () => (usage = new Map<string, TSESTree.Node[]>()),
+      "Program:exit": () =>
+        usage.forEach(nodes => {
+          if (nodes.length > USAGE_THRESHOLD) {
+            const [node, ...rest] = nodes;
+            const kind = node.type === "TSUnionType" ? "union" : "intersection";
+            const message = toEncodedMessage(
+              `Replace this ${kind} type with a type alias.`,
+              rest.slice(0, USAGE_THRESHOLD),
```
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: This is the behaviour of SonarTS implementation of the rule ([reference](https://github.com/SonarSource/SonarTS/blob/e84fe9e1f17984559600539b61573bfcc88a2414/sonarts-core/src/rules/useTypeAliasRule.ts#L66)). My guess is that one wanted to avoid having too many secondary locations. For example, if `A | B | C` appears twenty times in a given file, then there will be one primary location and nineteen secondary ones.

Should we drop this limit ?

eslint-bridge/tests/rules/use-type-alias.test.ts
```diff
@@ -0,0 +1,97 @@
...
+    },
+    {
+      code: `
+      let x: number | string;
+      let y: number | string;
+      let z: number | string; // this fine because only 2 types in the union`,
+      options: [],
+    },
+    {
+      code: `let a: any[] | null; // this is not a union when strict is false`,
```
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: Indeed. It was actually taken as it is from SonarTS. Not sure why it was there given the threshold used...

eslint-bridge/src/rules/use-type-alias.ts
yassin-kammoun-sonarsource: Since a fully working fix was generously given, I will use it as it is ;).

eslint-bridge/src/rules/use-type-alias.ts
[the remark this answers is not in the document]
christophe-zurn-sonarsource: To me, the way the limit/threshold was used in the SonarTS reference implementation looks like a bug, because the issue was only raised when there were __exactly__ 3 repetitions of the type. (one of the reason we have more true positives in this new implementation).
Also, there was no secondary locations in SonarTS, the message was containing the lines of duplicated types. In my opinion, we can drop this limit, as the way the secondary locations are displayed is not too invasive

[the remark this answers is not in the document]
yassin-kammoun-sonarsource: There is no bug actually; the original implementation is indeed confusing, but it is correct though. What essentially happens is that it keeps track of union/intersection types usage and tests on the fly if the limit was reached. The equality test will be true at some point and it is at that particular moment that the issue will be raised. In any case, I will drop the limit (which is not even mentioned in the spec).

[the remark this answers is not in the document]
christophe-zurn-sonarsource: Oh ok. Thanks for the explanation. That was not obvious when having a quick look. Yep, I think it's OK to do it like this. Thanks

its/ruling/src/test/expected/ts/TypeScript/typescript-S4323.json
```diff
@@ -8,6 +8,7 @@
 ],
 'TypeScript:src/compiler/factory.ts':[
 844,
+3752,
```
yassin-kammoun-sonarsource: Not a false positive after all because of different ordering.

its/ruling/src/test/expected/ts/ag-grid/typescript-S4323.json
```diff
@@ -4,8 +4,7 @@
 ],
 'ag-grid:src/ts/columnController/columnController.ts':[
 46,
-73,
-1210,
+59,
```
yassin-kammoun-sonarsource: Issue `59` not only was not a false positive after all, but it also replaces issues `73` and `1210` because of order- and spacing-insensitivity.

PR #1663, opened 2019-10-14
Fix handling of encodings other than UTF-8

### LINKED ISSUE #1657: Fix handling of encodings other then utf8
Currently, we don't send the content of the file between Java and Node.js process, only the file path to improve the speed of the analysis. When file content is missing, Node.js process will read the file from the filesystem. However, it will not take into account the encoding provided by the scanner, it always loads the file with `utf8` encoding.

When we detect that file is using different encoding then `utf8` we should also send the content of the file to the Node.js process, to avoid handling of different file encondigs on Node.js side.

saberduck:
Fixes #1657


saberduck:
Yes, you are right, I changed to UTF-16 which should fail with previous implementation

PR #1666, opened 2019-10-14
Add rule S4322 ('prefer-type-guard') for TS

### LINKED ISSUE #1519: Add rule S4322 ('prefer-type-guard' in tslint) for TS

quentin-jaquier-sonarsource:
Fixes #1519


quentin-jaquier-sonarsource:
In #1656, there is logic related to function that we could re-use, I propose to wait until it's merge on master to simplify the implementation of this one.

eslint-bridge/tests/rules/prefer-type-guard.test.ts
```diff
@@ -0,0 +1,233 @@
...
+        message: `Declare this function return type using type predicate "animal is Fish".`,
+        line,
+        endLine: line,
+      },
+    ],
+  };
+}
+
+ruleTester.run("Type guards should be used", rule, {
+  /*
```
yassin-kammoun-sonarsource: Do we still need those ?

eslint-bridge/src/rules/prefer-type-guard.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  const body = functionDeclaration.body;
+  const returnedExpression = getReturnedExpression(body);
+  if (!returnedExpression) {
+    return;
+  }
+
+  if (isInequalityBinaryExpression(returnedExpression)) {
+    const { left, right } = returnedExpression;
+    if (isUndefined(right)) {
```
christophe-zurn-sonarsource: Why are we not checking the other way around too ? if `left` is `undefined`, why don't we check `right` expression whether it's a `checkCastType` ?

quentin-jaquier-sonarsource: Good idea, it's a nice simple improvement to the rule!

PR #1668, opened 2019-10-16
Add rule S2870 ('no-array-delete') for TS/JS

### LINKED ISSUE #1506: Add rule S2870 ('no-array-delete' in tslint) for TS/JS

yassin-kammoun-sonarsource:
Fixes #1506


yassin-kammoun-sonarsource:
> In my opinion, there are too many FP in JS, knowing that this rule is in Sonarway, I don't think the current situation is acceptable.

As agreed, we will keep the Java-based implementation for JavaScript for now until we attempt first to parse JavaScript code with `@typescript-eslint/parser`.

PR #1678, opened 2019-10-17
Add rule S1226 for TS/JS (`no-ignored-initial-value`)

### LINKED ISSUE #1491: Add rule S1226 for TS/JS (`no-ignored-initial-value` in tslint)

christophe-zurn-sonarsource:
Fixes #1491


christophe-zurn-sonarsource:
Changes in the rule:

Implementation:
SonarTS and SonarJS implementations were a bit different. In the end, to avoid FP as much as possible, the implementation chosen is one close to the original SonarTS one. This implies the following:
- variables written and read on the same line are not raising issue in JS anymore. Ex: the following is now valid: `function(p1, p2) { foo(p1++); p2 /= 2; }`, this reduces a lot of FP in JS
- issues are reported at all levels, not only top level of function (how it previously was for TS). This makes some issues appear in TS, which can arguably be FP(?). This can be discussed, whether to put this exception back in the rule

Other:
- Message used is the one from the RSPEC, which is a bit different for JS from what it used to be.
- Highlighting location chosen is the one described in RSPEC and used in SonarJS, issues for SonarTS have therefore changed location slighlty.

eslint-bridge/src/rules/no-parameter-reassignment.ts
```diff
@@ -0,0 +1,270 @@
...
+        ) {
+          invalidReassignmentReferences.push(currentReference);
+        }
+        markAsRead(variableUsageContext, variableName);
+      }
+    }
+
+    function isUsedInWriteExpression(variableName: string, writeExpr: estree.Node | null) {
+      return (
+        !!writeExpr &&
```
saberduck: not sure `!!` is needed

eslint-bridge/src/rules/no-parameter-reassignment.ts
```diff
@@ -0,0 +1,270 @@
...
+        !!context.getSourceCode().getFirstToken(writeExpr, token => token.value === variableName)
+      );
+    }
+
+    function raiseIssue(reference: Scope.Reference) {
+      let locationHolder: { node: estree.Node } | { loc: AST.SourceLocation } = {
+        node: reference.identifier,
+      };
+      const location = getPreciseLocation(reference, reference.identifier.loc);
+      if (!!location) {
```
saberduck: do we need `!!` ?

its/ruling/src/test/expected/ts/rxjs/typescript-S1226.json
```diff
@@ -0,0 +1,14 @@
+{
```
quentin-jaquier-sonarsource: All new issues on TS (and in JS) looks like:

```
function foo(x, y) {
  if(x) {
    y = somethingElse();
  }
//Do something with y
}
```
One can argue that the initial value is not ignored here, and I would agree...
Therefore, I'm in favor of adding the exception about top level of function, we can discuss it further if you want!

eslint-bridge/tests/rules/no-parameter-reassignment.test.ts
```diff
@@ -0,0 +1,239 @@
...
+        for (let {prop1, prop2} in obj) {
+          prop1 = foo(); // Noncompliant
+        }
+        
+        for (let x of obj) {
+          x = foo(); // Noncompliant
+        }
+        
+        var y = 1;
+        y = 42;
```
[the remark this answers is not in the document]
christophe-zurn-sonarsource: I updated how invalid tests work for better readability

its/ruling/src/test/expected/ts/rxjs/typescript-S1226.json
christophe-zurn-sonarsource: Good point!
After experimenting a bit, I found the best compromise was to not raise issues inside top-level if-statements, as they are indeed often used to shift a function parameters, or to set default values depending on a global/field state.

christophe-zurn-sonarsource:
@quentin-jaquier-sonarsource After updating the implementation to ignore top-level if statements, it revealed the implementation shortcomings as it was easier to detect the other FP. I fixed the implementation in the last 2 commits. (Fix for the "arguments" variable, read variables propagation)

its/ruling/src/test/expected/ts/rxjs/typescript-S1226.json
quentin-jaquier-sonarsource: 👍 , don't forget to update the "changes in the rule" message when this is merged.

eslint-bridge/src/rules/no-parameter-reassignment.ts
```diff
@@ -0,0 +1,332 @@
...
+}
+
+function getPreciseLocationHolder(
+  reference: Scope.Reference,
+): { node: estree.Node } | { loc: AST.SourceLocation } {
+  const identifierLoc = reference.identifier.loc;
+  if (identifierLoc && reference.writeExpr && reference.writeExpr.loc) {
+    return { loc: { start: identifierLoc.start, end: reference.writeExpr.loc.end } };
+  }
+  return { node: reference.identifier };
```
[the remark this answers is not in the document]
christophe-zurn-sonarsource: Yes, it should actually never happen that `writeExpr` is undefined, since we only raise issues on reassignments. I still put the check here just in case (for defensive programming) as well as to avoid code smell/bug being raised.

PR #1681, opened 2019-10-18
Use bundledDependencies for packaging of eslint-bridge

saberduck:
Uses `bundledDependencies` in `package.json` to create the bundle with `npm pack`. Changes xz compression to gzip, which makes bundle bigger -> 7.5mb to 9.3mb

Also uses project references for TypeScript compilation, see here for details
https://www.typescriptlang.org/docs/handbook/project-references.html


eslint-bridge/jest.config.js
```diff
@@ -2,7 +2,7 @@ module.exports = {
   collectCoverageFrom: ["src/**/*.ts"],
   globals: {
     "ts-jest": {
-      skipBabel: false
+      tsConfig: "tests/tsconfig.json"
```
christophe-zurn-sonarsource: I guess this change come from the fact that `ts-jest` version has been updated in `package.json` (`"ts-jest": "23.1.4"` => `"ts-jest": "24.1.0"`). It seems this update was necessary (otherwise there is a build error), I just don't understand why it is the case...

saberduck: I believe it's related to the fact that we use project references with TypeScript (`tsc -b`) , but I didn't investigate further.

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/BundleImplTest.java
```diff
@@ -17,35 +17,37 @@
...
+import org.sonar.api.utils.internal.JUnitTempFolder;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
-public class BundleUtilsTest {
+public class BundleImplTest {
 
-  @Rule
-  public TemporaryFolder temporaryFolder = new TemporaryFolder();
+  @org.junit.Rule
```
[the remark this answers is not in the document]
saberduck: no, this was copy&paste

PR #1683, opened 2019-10-18
Add rule S4030 ('no-unused-array') for TS/JS

### LINKED ISSUE #1514: Add rule S4030 ('no-unused-array' in tslint) for TS/JS

quentin-jaquier-sonarsource:
Fixes #1514


its/ruling/src/test/expected/ts/TypeScript/typescript-S4030.json
```diff
@@ -1,5 +0,0 @@
-{
-'TypeScript:Gulpfile.ts':[
-665,
```
quentin-jaquier-sonarsource: This issue is correctly reported, but the file is not present in any tsconfig.

eslint-bridge/tests/rules/no-unused-collection.test.ts
```diff
@@ -0,0 +1,232 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+import { RuleTester } from "eslint";
+import * as path from "path";
+
+const ruleTester = new RuleTester({
+  parser: path.resolve(`${__dirname}/../../node_modules/@typescript-eslint/parser`),
```
saberduck: I think this can be changed to use `require.resolve` , I will fix all occurrences

eslint-bridge/src/rules/no-unused-collection.ts
```diff
@@ -0,0 +1,199 @@
...
+
+export const rule: Rule.RuleModule = {
+  create(context: Rule.RuleContext) {
+    return {
+      "Program:exit": function() {
+        const unusedArrays: Scope.Variable[] = [];
+        collectUnusedCollections(context.getScope(), unusedArrays);
+
+        unusedArrays.forEach(unusedArray => {
+          if (unusedArray.defs.length > 0) {
```
christophe-zurn-sonarsource: Could you please add a test related to this condition.
Currently, if I remove this `if`, the tests pass without any problem

eslint-bridge/src/rules/no-unused-collection.ts
```diff
@@ -0,0 +1,199 @@
...
+  });
+}
+
+function isUnusedCollection(variable: Scope.Variable) {
+  if (variable.references.length <= 1) {
+    return false;
+  }
+  let assignCollection = false;
+
+  for (let ref of variable.references) {
```
christophe-zurn-sonarsource: SonarQube is raising a code smell here

eslint-bridge/src/rules/no-unused-collection.ts
```diff
@@ -0,0 +1,199 @@
...
+
+  //All the write statement that we search are part of ExpressionStatement, if there is none, it's a read
+  return true;
+}
+
+function findFirstMatchingAncestor(
+  node: TSESTree.Node,
+  predicate: (node: TSESTree.Node) => boolean,
+) {
+  let currentNode = node.parent;
```
christophe-zurn-sonarsource: This `node.parent` property is coming from the `TSESTree.Node` definition. As we're casting it from `estree.Node`, is it possible that in some cases (depending on parser used) this property won't be available?

eslint-bridge/src/rules/no-unused-collection.ts
quentin-jaquier-sonarsource: In fact, if we reached this point, we went though at least one definition or reference, we can safely remove this check.

eslint-bridge/src/rules/no-unused-collection.ts
quentin-jaquier-sonarsource: I tried with both parser that we use to parse JS code, and it worked with both. I am not sure to understand why this property is not exposed on estree.Node, but since we are ourselves choosing the parser, I think it's fine this way!

PR #1686, opened 2019-10-21
S1541: Enable support of secondary locations

yassin-kammoun-sonarsource:
The rule metadata for secondary locations were missing.


yassin-kammoun-sonarsource:
> LGTM.
> However, this change makes me wonder whether we should have directly in unit test files something checking expected secondary locations/messages correctly depending on this parameter to avoid having the same problem in the future

Any idea how to do proceed ?

christophe-zurn-sonarsource:
> > LGTM.
> > However, this change makes me wonder whether we should have directly in unit test files something checking expected secondary locations/messages correctly depending on this parameter to avoid having the same problem in the future
> 
> Any idea how to do proceed ?

You can merge the PR already, once it's been rebased on master. I'll bring the subject up during standup

PR #1689, opened 2019-10-21
Add rule S3516 ('no-invariant-returns') for TS

### LINKED ISSUE #1508: Add rule S3516 ('no-invariant-return' in tslint) for TS

christophe-zurn-sonarsource:
Fixes #1508


christophe-zurn-sonarsource:
Chnages in ruling results are new true positives coming from slight change of implementation: not based on SE anymore, but simplistic check on identifiers with only single (first) write operation.

its/ruling/src/test/expected/js/js-project/javascript-S3516.json
```diff
@@ -1,5 +1,54 @@
...
+],
+'js-project:google-closure-library/editor/plugins/enterhandler.js':[
+265,
+319,
+],
+'js-project:google-closure-library/fx/draglistgroup.js':[
+610,
+],
 'js-project:mootools-core-1.4.5-full-nocompat.js':[
+1029,
```
yassin-kammoun-sonarsource: With the point made about function calls, this is a FP since `window.execScript(text)` at line 1032 could modify `text`, but should disappear if we decide to handle this case.

PR #1690, opened 2019-10-22
Add rule S4335 ('no-useless-intersection') for TS

### LINKED ISSUE #1522: Add rule S4335 ('no-useless-intersection' in tslint) for TS

yassin-kammoun-sonarsource:
Fixes #1522


eslint-bridge/src/rules/no-useless-intersection.ts
```diff
@@ -0,0 +1,103 @@
...
+
+function isTypeWithoutMembers(tp: any, ts: any): boolean {
+  const isNullLike = isNullType(tp, ts) || isUndefinedType(tp, ts) || isVoidType(tp, ts);
+  const isEmptyInterface = Boolean(
+    tp.symbol &&
+      tp.symbol.members &&
+      tp.symbol.members.size === 0 &&
+      isStandaloneInterface(tp.symbol, ts),
+  );
+  return isNullLike || isEmptyInterface;
```
quentin-jaquier-sonarsource: I think we can refactor the body of this method.
- create one function `isNullLike` instead of 3 different.
- and one for `isEmptyInterface`

The function becomes: 
```
return isNullLike(...) || (isEmptyInterface(...) && isStandaloneInterface(...))
```
In addition, this will enable short circuit to do it's magic.

PR #1691, opened 2019-10-22
Add rule S2871 ('no-alphabetical-sort') for TS

### LINKED ISSUE #1507: Add rule S2871 ('no-alphabetical-sort' in tslint) for TS

yassin-kammoun-sonarsource:
Fixes #1507


christophe-zurn-sonarsource:
Have these changes in the rule been discussed at some point? This seem to be quite a big change, and prone to raise FP.

For instance, I don't see why
```
let Strings = [ new String("a"), new String("b"), new String("c") ];
Strings.sort();
```
would raise an issue. This seem totally valid to me.

The same for the issue raised in the IT:
```
 let tsConfigFiles: NormalizedPath[];
// ...
tsConfigFiles.sort();

// with
type NormalizedPath = string & { __normalizedPathTag: any };
```

And any class redefining the `toString` method could potentially rely on lexicographical order for sorting without it posing a problem.
I think we should discuss this decision

yassin-kammoun-sonarsource:
I replaced the implementation by the original version of SonarTS, i.e. rising issues only on number arrays. Also, I created #1696 suggesting a few cases to improve the rule.

eslint-bridge/src/rules/no-alphabetical-sort.ts
```diff
@@ -0,0 +1,66 @@
...
+    if (isRequiredParserServices(services)) {
+      const ts = require("typescript");
+      return {
+        CallExpression: (node: estree.Node) => {
+          const call = node as TSESTree.CallExpression;
+          const callee = call.callee;
+          if (call.arguments.length === 0 && callee.type === "MemberExpression") {
+            const { object, property } = callee;
+            const text = context.getSourceCode().getText(property as estree.Node);
+            if (["sort", '"sort"', "'sort'"].includes(text)) {
```
christophe-zurn-sonarsource: Nitpick: you could extract this array `["sort", '"sort"', "'sort'"]` into a global constant to avoid having to create it each time the function is called

PR #1692, opened 2019-10-22
Add rule S4634 ('prefer-promise-shorthand') for TS

### LINKED ISSUE #1529: Add rule S4634 ('prefer-promise-shorthand' in tslint) for TS

yassin-kammoun-sonarsource:
Fixes #1529


eslint-bridge/src/rules/prefer-promise-shorthand.ts
```diff
@@ -0,0 +1,116 @@
...
+  rejectParameterName: string | undefined,
+): string | undefined {
+  switch (callee) {
+    case resolveParameterName:
+      return "resolve";
+    case rejectParameterName:
+      return "reject";
+    default:
+      return undefined;
+  }
```
quentin-jaquier-sonarsource: I feel that this is not fully tested, you may want to add a test where names are not `resolve` and `reject`, something like this:
```
let rejectedPromise = new Promise((p1, p2) => p2(new Error('fail')));
// message: `Replace this trivial promise with "Promise.reject(new Error('fail'))".`
```

PR #1700, opened 2019-10-25
Fix import of unit test report

vilchik-elena:


vilchik-elena:
I'm not sure I will be able to verify PR before merging it

pom.xml
```diff
@@ -66,6 +66,8 @@
   </ciManagement>
 
   <properties>
+    <sonar.testExecutionReportPaths>eslint-bridge/test-report.xml</sonar.testExecutionReportPaths>
```
saberduck: maybe we can add a comment that this is necessary because of drop of modules

vilchik-elena:
Tested locally manually, all is good (~700 more unit tests)

PR #1701, opened 2019-10-25
Add rule S2681 ('no-unenclosed-multiline-block') for TS/JS

### LINKED ISSUE #1504: Add rule S2681 ('no-unenclosed-multiline-block') for TS/JS

yassin-kammoun-sonarsource:
Fixes #1504

### Changes in the rule
The rule implementation is taken from what was done in SonarTS.; more cases are now covered for JavaScript, like the following:

```javascript
if (condition) action1();
  action2();

if (condition)
   action1(); action2();
```


eslint-bridge/src/rules/no-unenclosed-multiline-block.ts
```diff
@@ -0,0 +1,139 @@
...
+  nestingStatement: NestingStatement,
+  others: Statement[],
+  context: Rule.RuleContext,
+) {
+  const [primaryMessage, secondaryMessage] =
+    nestingStatement.type === "IfStatement"
+      ? [IF_PRIMARY_MESSAGE, IF_SECONDARY_MESSAGE]
+      : [LOOP_PRIMARY_MESSAGE, LOOP_SECONDARY_MESSAGE];
+
+  let firstStatementInPseudoBlock = nestedStatement(nestingStatement);
```
christophe-zurn-sonarsource: Why not pass the nested statement directly, as it is already computed on line 68 before calling the `raiseIssue` function?

eslint-bridge/src/rules/no-unenclosed-multiline-block.ts
```diff
@@ -0,0 +1,139 @@
...
+  let lastStatementInPseudoBlock = statement;
+
+  const secondaryLocations: Statement[] = [];
+  for (const other of others) {
+    if (column(other) !== column(statement)) {
```
christophe-zurn-sonarsource: you could extract the value of `column(statement)` outside the loop, to avoid recomputing it each time

eslint-bridge/src/rules/no-unenclosed-multiline-block.ts
```diff
@@ -0,0 +1,139 @@
...
+}
+
+function isNestingStatement(node: estree.Node): node is NestingStatement {
+  return [
+    "IfStatement",
+    "ForStatement",
+    "ForInStatement",
+    "ForOfStatement",
+    "WhileStatement",
+  ].includes(node.type);
```
christophe-zurn-sonarsource: You could extract this array creation inside a global variable

eslint-bridge/tests/rules/no-unenclosed-multiline-block.test.ts
```diff
@@ -0,0 +1,196 @@
...
+    },
+    {
+      code: `
+      if (condition)
+        action1();
+        action2();
+        action3();
+      action4();
+      `,
+      errors: 1,
```
christophe-zurn-sonarsource: I think it would be interesting to check the message here, as we want to make sure that the line with `action4();` is not part of the secondary lcoation

eslint-bridge/tests/rules/no-unenclosed-multiline-block.test.ts
```diff
@@ -0,0 +1,196 @@
...
+    },
+    {
+      code: `
+        if (condition)
+      action1();
+      action2();
+      `,
+    },
+    {
+      code: `if (condition) action1(); action2();`,
```
christophe-zurn-sonarsource: This example, as well as the following snippets:
```
if (condition) action1();
  action2();

if (condition)
   action1(); action2();
```
were raising an issue in SonarTS. Is the change in the rule behaviour intended ? If yes, could you document it in the description of the PR so that it is clear it was done on purpose.

eslint-bridge/src/rules/no-unenclosed-multiline-block.ts
```diff
@@ -0,0 +1,139 @@
...
+
+function nestedStatement(node: NestingStatement) {
+  if (node.type === "IfStatement") {
+    return node.consequent;
+  }
+  return node.body;
+}
+
+function column(node: estree.Node) {
+  return node.loc!.start.column;
```
christophe-zurn-sonarsource: This line and line 138 are raising code smells in SQ. You could either add a check for defensive programing or use `as TSESTree.Node` type cast

eslint-bridge/tests/rules/no-unenclosed-multiline-block.test.ts
yassin-kammoun-sonarsource: Ah, I forgot to consider what was done in SonarTS. It's probably better to rely on SonarTS's implementation of the rule since it covers more cases.

yassin-kammoun-sonarsource: Does it make sense ?

its/ruling/src/test/expected/js/js-project/javascript-S2681.json
```diff
@@ -0,0 +1,87 @@
+{
+'js-project:angular-1.0.3.js':[
+13253,
+],
+'js-project:ecmascript6/di.js/visualize/lib/angular.js':[
+18850,
+],
+'js-project:mootools-core-1.4.5-full-nocompat.js':[
+1884,
```
christophe-zurn-sonarsource: This looks like a FP, as the catch clause should not be considered as a separate statement

eslint-bridge/src/rules/no-unenclosed-multiline-block.ts
```diff
@@ -0,0 +1,218 @@
...
+      return result;
+    }, new Array<{ prev: NestingStatement; next: Statement }>())
+    .map(pair => {
+      return new ChainedStatements(pair.prev, extractLastBody(pair.prev), pair.next);
+    });
+}
+
+function extractLastBody(statement: NestingStatement) {
+  if (statement.type === "IfStatement") {
+    if (!statement.alternate) {
```
christophe-zurn-sonarsource: This condition could be reversed to slightly improve readability

its/ruling/src/test/expected/js/js-project/javascript-S2681.json
yassin-kammoun-sonarsource: This looks indeed like a FP, but it's not actually. The issue is raised because of the semicolon after the catch statement; the semicolon is considered as an empty statement. 

This example is not really different from:

```javascript
if (condition) try {} catch {}
  ;
```

christophe-zurn-sonarsource: Oh ok. Thanks for the precision!

christophe-zurn-sonarsource:
LGTM

PR #1702, opened 2019-10-25
Update typescript-eslint to 2.5.0

vilchik-elena:


vilchik-elena:
We failed to update to latest 2.5.0 version as unit tests on Windows are failing. It's unclear due to which reason, we are postponing this investigation.

PR #1703, opened 2019-10-28
Add rule S3981 ('no-collection-size-mischeck') for TS/JS

### LINKED ISSUE #1512: Add rule S3981 ('no-collection-size-mischeck') for TS/JS
We can make rule rely on types for TS

yassin-kammoun-sonarsource:
Fixes #1512


eslint-bridge/src/rules/no-collection-size-mischeck.ts
```diff
@@ -0,0 +1,77 @@
...
+} from "../utils/isRequiredParserServices";
+import { TSESTree } from "@typescript-eslint/experimental-utils";
+
+const CollectionLike = ["Array", "Map", "Set"];
+const CollectionSizeLike = ["length", "size"];
+
+export const rule: Rule.RuleModule = {
+  create(context: Rule.RuleContext) {
+    const services = context.parserServices;
+    if (isRequiredParserServices(services)) {
```
saberduck: we can invert this condition and use early return, so the indentation of the whole function is decreased

eslint-bridge/src/rules/no-collection-size-mischeck.ts
```diff
@@ -0,0 +1,77 @@
...
+
+import { Rule } from "eslint";
+import * as estree from "estree";
+import {
+  isRequiredParserServices,
+  RequiredParserServices,
+} from "../utils/isRequiredParserServices";
+import { TSESTree } from "@typescript-eslint/experimental-utils";
+
+const CollectionLike = ["Array", "Map", "Set"];
```
saberduck: can we add also `WeakSet` and `WeakMap` ?

yassin-kammoun-sonarsource: Yes, we can.

eslint-bridge/src/rules/no-collection-size-mischeck.ts
yassin-kammoun-sonarsource: Won't need to do that because the logic changed a bit.

PR #1704, opened 2019-10-28
Add rule S4822 ('no-try-promise') for TS

### LINKED ISSUE #1533: Add rule S4822 ('no-try-promise') for TS

yassin-kammoun-sonarsource:
Fixes #1533


its/plugin/projects/ts-rule-project/S4822.ts
```diff
@@ -0,0 +1,129 @@
+function returningPromise() {
```
vilchik-elena: I think we should not have entire unit test the rule here. The point of this test is just to make sure rule is executed normally.
Could you keep only couple main cases?

PR #1706, opened 2019-10-29
Add rule S4782 ('no-redundant-optional') for TS

### LINKED ISSUE #1530: Add rule S4782 ('prefer-optional') for TS

christophe-zurn-sonarsource:
Fixes #1530 

**Changes in rule**
As mentioned [here](https://github.com/SonarSource/SonarJS/pull/1628#issuecomment-539517621), [RSPEC-4782](https://jira.sonarsource.com/browse/RSPEC-4782) was updated to only report issues when both the optional `?` token as well as `undefined` type are present for a property.

This change made many issues disappear in the ruling ITs.


javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S4782.json
```diff
@@ -0,0 +1,19 @@
...
+  ],
+  "status": "ready",
+  "remediation": {
+    "func": "Constant\/Issue",
+    "constantCost": "1min"
+  },
+  "tags": [
+    "redundant"
+  ],
+  "defaultSeverity": "Major",
```
saberduck: we can maybe change it to minor

PR #1707, opened 2019-10-30
Add S1121 for TS/JS ('no-nested-assignment')

### LINKED ISSUE #1408: Add S1121 for TS/JS ('no-nested-assignment')
We allow assignment in loop condition, in eslint it's possible to wrap in parentheses to ignore assignment in condition. I think we should go for this option enabled and w/o ability to configure it in SQ.

yassin-kammoun-sonarsource:
Fixes #1408

### Changes in the rule
The rule now raises issues for assignments in any kind of expressions except assignment expression statements, i.e. `a = 0;`.


its/ruling/src/test/expected/ts/Joust/typescript-S1121.json
```diff
@@ -2,4 +2,16 @@
 'Joust:ts/components/EventLog.tsx':[
 51,
 ],
+'Joust:ts/components/GameWidget.tsx':[
```
yassin-kammoun-sonarsource: Should we exclude this pattern, i.e. lambda whose body is assignment expression ?

eslint-bridge/src/rules/no-conditional-assignment.ts
```diff
@@ -0,0 +1,63 @@
...
+  );
+  const text = sourceCode.getText(node.left as estree.Node);
+  context.report({
+    message: `Extract the assignment of \"${text}\" from this expression.`,
+    loc: operator!.loc,
+  });
+}
+
+function parentOf(node: TSESTree.Node): TSESTree.Node | undefined {
+  const parent = node.parent;
```
vilchik-elena: hm, I didn't know about `node.parent`, that's something internal for estree API. Could you use classic `context.getParent()` to make sure it's always there?

eslint-bridge/tests/rules/no-conditional-assignment.test.ts
```diff
@@ -0,0 +1,156 @@
...
+      code: `for (i = j = 0;;) {}`,
+    },
+    {
+      code: `for (i, j = 0;;) {}`,
+    },
+    {
+      code: `for (;; i = 0) {}`,
+    },
+  ],
+  invalid: [
```
vilchik-elena: you forgot to test issue location and message

eslint-bridge/tests/rules/no-conditional-assignment.test.ts
```diff
@@ -0,0 +1,156 @@
...
+    {
+      code: `(a = 0) ? b : c;`,
+      errors: 1,
+    },
+    {
+      code: `a ? b = 0 : c;`,
+      errors: 1,
+    },
+    {
+      code: `while (a = 0) {}`,
```
vilchik-elena: we should now remove exception from rule description

eslint-bridge/tests/rules/no-conditional-assignment.test.ts
```diff
@@ -0,0 +1,174 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+import { RuleTester } from "eslint";
+
+const ruleTester = new RuleTester({ parserOptions: { ecmaVersion: 2018 } });
+import { rule } from "../../src/rules/no-conditional-assignment";
```
vilchik-elena: I just realized that rule key should be different now as this rule has nothing to do this conditions (how about `no-nested-assignment`?)

eslint-bridge/src/rules/no-conditional-assignment.ts
```diff
@@ -0,0 +1,72 @@
...
+import * as estree from "estree";
+import { getParent } from "eslint-plugin-sonarjs/lib/utils/nodes";
+
+export const rule: Rule.RuleModule = {
+  create(context: Rule.RuleContext) {
+    return {
+      AssignmentExpression: (node: estree.Node) => {
+        const assignment = node as estree.AssignmentExpression;
+        const ancestor = ancestorOf(assignment, context.getAncestors());
+        const parent = getParent(context);
```
vilchik-elena: tbh this logic with parents and ancestors is fuzzy. How about extracting logic checking for arrow function into a separate function to hide `parent` variable there?

yassin-kammoun-sonarsource: I agree, it's not nice the way it is right now.

PR #1710, opened 2019-10-30
Add rule S2068 ('no-hardcoded-credentials') for TS/JS

### LINKED ISSUE #1498: Add rule S2068 ('no-hardcoded-credentials') for TS/JS (new rule for JS)

yassin-kammoun-sonarsource:
Fixes #1498


eslint-bridge/src/rules/no-hardcoded-credentials.ts
```diff
@@ -0,0 +1,91 @@
...
+      node: literal,
+    });
+  }
+}
+
+function isStringLiteral(node: estree.Node): node is estree.Literal {
+  return (
+    node.type === "Literal" &&
+    node.raw !== undefined &&
+    (node.raw.startsWith(`\"`) || node.raw.startsWith(`'`))
```
christophe-zurn-sonarsource: I think it would be cleaner to check for `typeof node.value === "string"` instead of checking the raw value for quotes

javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S2068.html
```diff
@@ -0,0 +1,31 @@
...
+      <li> API </li>
+      <li> Service </li>
+    </ul> </li>
+  <li> Application re-distribution is required before updating the credentials. </li>
+</ul>
+<p>You are at risk, if you answered yes to any of these questions.</p>
+<h2>Recommended Secure Coding Practices</h2>
+<ul>
+  <li> Credentials should be stored outside of the code in a strongly-protected encrypted configuration file or database. </li>
+</ul>
```
christophe-zurn-sonarsource: The Compliant/Noncompliant examples are missing because the RSPEC subtask only exists for Typescript at the moment.

PR #1713, opened 2019-10-31
Add rule S4275 ('no-accessor-field-mismatch') for TS/JS (new rule for JS)

### LINKED ISSUE #1518: Add rule S4275 ('no-accessor-field-mismatch') for TS/JS (new rule for JS)

christophe-zurn-sonarsource:
Fixes #1518


eslint-bridge/src/rules/no-accessor-field-mismatch.ts
```diff
@@ -0,0 +1,253 @@
...
+    if (!isRequiredParserServices(services)) {
+      return {};
+    }
+
+    function checkAccessor(accessor: TSESTree.Property | TSESTree.MethodDefinition) {
+      const accessorIsPublic =
+        accessor.type !== "MethodDefinition" || accessor.accessibility === "public";
+      const accessorInfo = getAccessorInfo(accessor);
+      const fields = getFields(accessor.parent as TSESTree.Node);
+      if (!accessorInfo || !fields || !accessorIsPublic) {
```
vilchik-elena: `!fields` is always true

eslint-bridge/src/rules/no-accessor-field-mismatch.ts
```diff
@@ -0,0 +1,253 @@
...
+    const services = context.parserServices;
+    if (!isRequiredParserServices(services)) {
+      return {};
+    }
+
+    function checkAccessor(accessor: TSESTree.Property | TSESTree.MethodDefinition) {
+      const accessorIsPublic =
+        accessor.type !== "MethodDefinition" || accessor.accessibility === "public";
+      const accessorInfo = getAccessorInfo(accessor);
+      const fields = getFields(accessor.parent as TSESTree.Node);
```
vilchik-elena: you would do this multiple times for every class. Should we cache it or visit class directly?

eslint-bridge/src/rules/no-accessor-field-mismatch.ts
```diff
@@ -0,0 +1,253 @@
...
+  containingClass: TSESTree.ClassBody,
+): TSESTree.FunctionExpression | TSESTree.TSEmptyBodyFunctionExpression | undefined {
+  for (const classElement of containingClass.body) {
+    if (classElement.type === "MethodDefinition" && getName(classElement.key) === "constructor") {
+      return classElement.value;
+    }
+  }
+}
+
+function fieldNameMatch(field: Field, name: string) {
```
vilchik-elena: it will be more consistent to have here as input 2 strings or 2 objects

eslint-bridge/src/rules/no-accessor-field-mismatch.ts
```diff
@@ -0,0 +1,253 @@
...
+
+    function checkAccessor(accessor: TSESTree.Property | TSESTree.MethodDefinition) {
+      const accessorIsPublic =
+        accessor.type !== "MethodDefinition" || accessor.accessibility === "public";
+      const accessorInfo = getAccessorInfo(accessor);
+      const fields = getFields(accessor.parent as TSESTree.Node);
+      if (!accessorInfo || !fields || !accessorIsPublic) {
+        return;
+      }
+      const matchingFields = fields.filter(field => fieldNameMatch(field, accessorInfo.name));
```
vilchik-elena: you can compute it later

eslint-bridge/src/rules/no-accessor-field-mismatch.ts
```diff
@@ -0,0 +1,253 @@
...
+      const statements = getFunctionBody(accessor.value);
+      if (!statements || statements.length > 1) {
+        return;
+      }
+
+      if (
+        matchingFields.length > 0 &&
+        (statements.length === 0 ||
+          isNotUsingAccessorFieldInBody(statements[0], accessorInfo, matchingFields))
```
vilchik-elena: it's better to not use negation in func name

eslint-bridge/tests/rules/no-accessor-field-mismatch.test.ts
```diff
@@ -0,0 +1,347 @@
...
+          }
+          
+          public getY(someParam: number) { // Compliant, not a zero-parameters getter
+            return 3;
+          }
+        }`,
+    },
+    {
+      code: `
+        class Exceptions {
```
vilchik-elena: why most methods here duplicate the one from line 65?

eslint-bridge/tests/rules/no-accessor-field-mismatch.test.ts
```diff
@@ -0,0 +1,347 @@
...
+          line: 8,
+          column: 20,
+          endLine: 8,
+          endColumn: 21,
+        },
+      ],
+    },
+    {
+      code: `
+      class NOK_Issue476 {
```
vilchik-elena: what is `476`?

eslint-bridge/src/rules/no-accessor-field-mismatch.ts
christophe-zurn-sonarsource: As you suggested, I will investigate to use `TypeChecker` to retrieve fields instead, so this part will surely change a bit

eslint-bridge/tests/rules/no-accessor-field-mismatch.test.ts
christophe-zurn-sonarsource: Copy-paste error, after which I added some code for coverage in both the examples. I removed the duplication

eslint-bridge/tests/rules/no-accessor-field-mismatch.test.ts
christophe-zurn-sonarsource: It's referring to the issue: https://github.com/SonarSource/SonarTS/issues/476, I can remove it as it does not make sense anymore now that it is not the same repo

eslint-bridge/src/rules/no-accessor-field-mismatch.ts
```diff
@@ -0,0 +1,250 @@
...
+    schema: [
+      {
+        // internal parameter for rules having secondary locations
+        enum: ["sonar-runtime"],
+      },
+    ],
+  },
+
+  create(context: Rule.RuleContext) {
+    const services = context.parserServices;
```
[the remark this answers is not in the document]
christophe-zurn-sonarsource: It was required to only run the rule when it was implemented for TS only. Following your suggestion, the scope of the rule was increased to also cover JS, so this part has been removed.

eslint-bridge/src/rules/no-accessor-field-mismatch.ts
```diff
@@ -0,0 +1,250 @@
...
+
+        context.report({
+          message: toEncodedMessage(primaryMessage, secondaryLocations, secondaryMessages),
+          loc: accessor.key.loc,
+        });
+      }
+    }
+
+    return {
+      Property: (node: estree.Node) => checkAccessor(node as TSESTree.Property),
```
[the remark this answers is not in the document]
christophe-zurn-sonarsource: Indeed, it makes more sense to do it like this. I updated the implementation to compute only once the fields for the current class/object declaration

christophe-zurn-sonarsource:
@saberduck Good point. We agreed that it made sense to also enable this rule for JS

eslint-bridge/tests/rules/no-accessor-field-mismatch.test.ts
```diff
@@ -164,6 +168,18 @@ ruleTester.run("Getters and setters should access the expected fields", rule, {
             this.z = x;
           },
           
+          set w(x: number) {
+            this.w_ = x;
+          },
+          
+          get myVal() {
+            return this.myVal_;
```
vilchik-elena: will we report if names don't match?

christophe-zurn-sonarsource: Yes. I updated the unit tests accordingly

PR #1715, opened 2019-11-01
Add rule S1874 ('deprecation') for TS

### LINKED ISSUE #1432: Add rule S1874 ('deprecation') for TS
This is a tricky one, investigate rule implementation before applying it (does it work cross-file?)

vilchik-elena:
Fixes #1432


eslint-bridge/src/rules/deprecation.ts
```diff
@@ -0,0 +1,221 @@
...
+
+function isShorthandPropertyAssignment(
+  node: tsTypes.Node,
+  ts: any,
+): node is tsTypes.ShorthandPropertyAssignment {
+  return node.kind === ts.SyntaxKind.ShorthandPropertyAssignment;
+}
+
+function isShortHandProperty(parent: estree.Node | undefined): parent is estree.Property {
+  return !!parent && parent.type === "Property" && parent.shorthand;
```
saberduck: you can remove `!!`

eslint-bridge/src/rules/deprecation.ts
```diff
@@ -0,0 +1,221 @@
...
+}
+
+function getSignatureDeprecation(signature?: tsTypes.Signature): Deprecation | undefined {
+  if (!signature) {
+    return undefined;
+  }
+  return getJsDocDeprecation(signature.getJsDocTags());
+}
+
+function getJsDocDeprecation(tags: tsTypes.JSDocTagInfo[]): Deprecation | undefined {
```
saberduck: I would drop the return type, it can be inferred (also in other functions)

its/ruling/src/test/expected/ts/angular/typescript-S1874.json
```diff
@@ -0,0 +1,429 @@
+{
```
saberduck: we don't need this file, angular was removed from ruling

its/ruling/src/test/expected/ts/vscode/typescript-S1874.json
```diff
@@ -0,0 +1,172 @@
+{
```
saberduck: vscode project is not part of the ruling anymore, there should be no expectations

saberduck:
few minor comments, otherwise lgtm

eslint-bridge/src/rules/deprecation.ts
vilchik-elena: no, I can't. Function should return boolean

PR #1716, opened 2019-11-01
Fix TypeScriptRuleTest

vilchik-elena:


its/plugin/projects/ts-rule-project/S4158.ts
```diff
@@ -160,7 +160,7 @@ function arrayInitializedByFunctionCall(init: () => number[]) {
 
 function arrayNotInitialized() {
   let notInitializedArray!: number[];
-  foo(notInitializedArray[0]); // Noncompliant
+  foo(notInitializedArray[0]);  // Not reported
```
vilchik-elena: it's not reported in unit test as well

its/plugin/projects/ts-rule-project/S2871.ts
```diff
@@ -3,7 +3,7 @@ arrayOfNumbers.sort(); // Noncompliant
 
 arrayOfNumbers.sort((n, m) => n - m);
 
-var emptyArrayOfNumbers: number[] = [];
+var emptyArrayOfNumbers: number[] = [1, 2, 3];
```
vilchik-elena: to not have issue for another rule

christophe-zurn-sonarsource: I would either remove this test or rename the variable, as it is confusing to have emptyArrayOfNumbers`` for a non-empty array

vilchik-elena: indeed I remove it

PR #1717, opened 2019-11-01
Add rule S4328 ('no-implicit-dependencies') for TS

### LINKED ISSUE #1476: Add rule S4328 ('no-implicit-dependencies') for TS
Check if applicable to JS

yassin-kammoun-sonarsource:
Fixes #1476


eslint-bridge/tests/rules/no-implicit-dependencies.test.ts
```diff
@@ -0,0 +1,144 @@
...
+    },
+    {
+      code: `import "foo/bar";`,
+      filename,
+      options,
+      errors: 1,
+    },
+    {
+      code: `import "foo";`,
+      filename: path.join(__dirname, "../fixtures/empty-package-json-project/file.js"),
```
vilchik-elena: what's the point of this test with empty package?

eslint-bridge/tests/rules/no-implicit-dependencies.test.ts
```diff
@@ -0,0 +1,144 @@
...
+    },
+    {
+      code: `import "foo";`,
+      filename: path.join(__dirname, "../fixtures/package-json-project/dir/subdir/file.js"),
+      options,
+      errors: 1,
+    },
+    {
+      code: `import "foo";`,
+      filename: "/",
```
vilchik-elena: what's that for?

javascript-checks/src/main/java/org/sonar/javascript/checks/ImplicitDependenciesCheck.java
```diff
@@ -0,0 +1,50 @@
...
+
+  @RuleProperty(
+    key = "whitelist",
+    description = "Comma separated list of modules to ignore while checking in package.json.",
+    defaultValue = "" + DEFAULT)
+  public String whitelist = DEFAULT;
+
+  @Override
+  public List<Object> configurations() {
+    return Arrays.asList((Object[])whitelist.split("\\s*,\\s*"));
```
vilchik-elena: let's do split and trim, without regular expressions (first regular expressions are often error prone, second they might introduce security problems)

eslint-bridge/tests/fixtures/package-json-project/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "test",
+  "version": "1.0.0",
+  "dependencies": {
+    "foobar": "latest",
+    "@angular/core": "4.4.5"
+  },
+  "devDependencies": {
+    "jest": "latest"
```
vilchik-elena: it's better to use names showing that you want to test, e.g.

```
  "dependencies": {
    "dependency": "latest",
    "@namespaced/dependency": "4.4.5"
  },
  "devDependencies": {
    "devDependency": "latest"
  }
```

eslint-bridge/src/rules/no-implicit-dependencies.ts
```diff
@@ -0,0 +1,163 @@
...
+  const cached = cache.get(dirname);
+  if (cached) {
+    return cached;
+  }
+
+  const result = new Set<string>();
+  const packageJsonPath = findPackageJson(path.resolve(path.dirname(fileName)));
+  if (packageJsonPath !== undefined) {
+    try {
+      // don't use require here to avoid caching
```
vilchik-elena: `require` what?

eslint-bridge/src/rules/no-implicit-dependencies.ts
```diff
@@ -0,0 +1,163 @@
...
+      // remove BOM from file content before parsing
+      const content = JSON.parse(
+        fs.readFileSync(packageJsonPath, "utf8").replace(/^\uFEFF/, ""),
+      ) as PackageJson;
```
vilchik-elena: I think we can drop `PackageJson` and `Dependencies` types, they don't bring a lot of clarity to the code

eslint-bridge/src/rules/no-implicit-dependencies.ts
```diff
@@ -0,0 +1,163 @@
...
+  if (cached) {
+    return cached;
+  }
+
+  const result = new Set<string>();
+  const packageJsonPath = findPackageJson(path.resolve(path.dirname(fileName)));
+  if (packageJsonPath !== undefined) {
+    try {
+      // don't use require here to avoid caching
+      // remove BOM from file content before parsing
```
vilchik-elena: do we have a test for file with bom?

eslint-bridge/src/rules/no-implicit-dependencies.ts
```diff
@@ -0,0 +1,163 @@
...
+function getDependencies(fileName: string) {
+  const dirname = path.dirname(fileName);
+
+  const cached = cache.get(dirname);
+  if (cached) {
+    return cached;
+  }
+
+  const result = new Set<string>();
+  const packageJsonPath = findPackageJson(path.resolve(path.dirname(fileName)));
```
vilchik-elena: `path.dirname(fileName)` is computed already

eslint-bridge/src/rules/no-implicit-dependencies.ts
```diff
@@ -0,0 +1,163 @@
...
+        const moduleName = (node as estree.ImportDeclaration).source.value;
+        const importToken = context.getSourceCode().getFirstToken(node);
+        raiseOnImplicitImport(moduleName, importToken!.loc, dependencies, whitelist, context);
+      },
+    };
+  },
+};
+
+function raiseOnImplicitImport(
+  moduleName: string | number | boolean | RegExp | undefined | null,
```
vilchik-elena: why not passing `estree.Literal` instead to avoid this union?

eslint-bridge/src/rules/no-implicit-dependencies.ts
```diff
@@ -0,0 +1,163 @@
...
+    !dependencies.has(packageName)
+  ) {
+    context.report({
+      message: "Either remove this import or add it as a dependency.",
+      loc,
+    });
+  }
+}
+
+function getPackageName(name: string, whitelist: string[]) {
```
vilchik-elena: I don't get what's going on in this function. Could you explain?

eslint-bridge/tests/rules/no-implicit-dependencies.test.ts
yassin-kammoun-sonarsource: Essentially branch coverage in `getDependencies` when there is a package.json without any kind of dependencies.

eslint-bridge/tests/rules/no-implicit-dependencies.test.ts
yassin-kammoun-sonarsource: This case tests missing package.json. `findPackageJson` looks for a package.json in the directory of the source code. If there is no such file, it then looks at the parent directory and continues doing so until it finds one or loops (loops in the sense that the parent directory of the current directory is the current directory itself, e.g. `/`).

eslint-bridge/src/rules/no-implicit-dependencies.ts
yassin-kammoun-sonarsource: One can do `require('./package.json')`, but this is not what we want. I think it's better if I drop this comment.

eslint-bridge/src/rules/no-implicit-dependencies.ts
yassin-kammoun-sonarsource: We don't need it actually.

eslint-bridge/src/rules/no-implicit-dependencies.ts
```diff
@@ -44,28 +44,29 @@ export const rule: Rule.RuleModule = {
         ) {
           const [argument] = call.arguments;
           if (argument.type === "Literal") {
-            const moduleName = argument.value;
+            const module = argument;
```
vilchik-elena: why just reassigning it?

eslint-bridge/tests/fixtures/package-json-project/package.json
```diff
@@ -1,14 +1,12 @@
...
-  "version": "1.0.0",
   "dependencies": {
-    "foobar": "latest",
-    "@angular/core": "4.4.5"
+    "dependency": "latest",
+    "@namespaced/dependency": "latest"
   },
   "devDependencies": {
-    "jest": "latest"
+    "devDependencies": "latest"
```
vilchik-elena: use singular `devDependency` to not confuse with option itself

eslint-bridge/tests/fixtures/package-json-project/package.json
```diff
@@ -1,14 +1,12 @@
...
   },
   "peerDependencies": {
-    "typescript": "3.5.3"
+    "peerDependencies": "latest"
```
vilchik-elena: same here

PR #1718, opened 2019-11-01
Add rule S1854 ('no-dead-store') for TS/JS

### LINKED ISSUE #1497: Add rule S1854 ('no-dead-store' in tslint) for TS/JS

saberduck:
Fixes #1497


eslint-bridge/src/rules/no-deadstore.ts
```diff
@@ -0,0 +1,503 @@
...
+      const ref = scope.references.find(r => r.identifier === node);
+      if (ref) {
+        return { ref, variable: ref.resolved };
+      } else {
+        // if it's not a reference, it can be just declaration without initializer
+        const variable = scope.variables.find(v => v.defs.find(def => def.name === node));
+        if (variable) {
+          return { ref: null, variable };
+        }
+        return resolveReferenceRecursively(node, scope.upper);
```
vilchik-elena: as far as I understood this code only required for switch expression, and for that recursion strictly speaking is not required, only looking at upper scope should be enough

eslint-bridge/src/rules/no-deadstore.ts
```diff
@@ -0,0 +1,503 @@
...
+      }
+    }
+  },
+};
+
+class CodePathContext {
```
vilchik-elena: wdyt about moving these classes to another file? This one is too huge

eslint-bridge/src/rules/no-deadstore.ts
```diff
@@ -0,0 +1,503 @@
...
+  lhs = new Set<ReferenceLike>();
+  rhs = new Set<ReferenceLike>();
+
+  isRhs(node: TSESTree.Node) {
+    return isAssignmentExpression(this.node) ? this.node.right === node : this.node.init === node;
+  }
+
+  isLhs(node: TSESTree.Node) {
+    return isAssignmentExpression(this.node) ? this.node.left === node : this.node.id === node;
+  }
```
yassin-kammoun-sonarsource: I can't find an example where there is more than one ref in `lhs` or `rhs`. Can it happen at all?

eslint-bridge/src/rules/no-deadstore.ts
```diff
@@ -0,0 +1,503 @@
...
+      peek(codePathStack).assignmentStack.push(new AssignmentContext(node));
+    }
+
+    function popAssignmentContext() {
+      const assignment = peek(codePathStack).assignmentStack.pop()!;
+      assignment.rhs.forEach(r => processReference(r));
+      assignment.lhs.forEach(r => processReference(r));
+    }
+
+    function lva(liveVariablesMap: Map<string, LiveVariables>) {
```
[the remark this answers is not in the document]
saberduck: I prefer to keep the parameter, but I moved the method to the `LiveVariables` class

eslint-bridge/src/rules/no-deadstore.ts
saberduck: I moved the lva algorithm with `LiveVariables` class to separate file

eslint-bridge/src/rules/no-deadstore.ts
saberduck: yes, it seems that this is required only for the `switch`. However, I would prefer to keep the recursion for safety and also I find it a bit more readable in this case. I added a small comment explaining this.

eslint-bridge/src/rules/no-deadstore.ts
saberduck: I am not sure I understand the comment, but this test 
```
function f(a,b) {
  var d = (d = a - b) * d + (d = a - b) * d;
  foo(d);
}
```
the assignment will have many references on rhs. On lhs it is more rare that there are multiple references, but you can imagine something like
`a[foo(x,y)] = 42` , and btw this is the case where current algoritm is not entirely correct, because `foo(x,y)`  should be evaluated before rhs, but we will consider it to be evaluated after.

PR #1730, opened 2019-11-07
Rely on eslint 6.6.0

### LINKED ISSUE #1729: S1440 is reporting on entire line (eslint eqeqeq does not provide endColumn)

vilchik-elena:
Fixes #1729


vilchik-elena:
Still 2 times slower...

PR #1734, opened 2019-11-07
Analyze only files which are part of the tsconfig.json

saberduck:


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/TypeScriptSensor.java
```diff
@@ -102,11 +101,7 @@ static FilePredicate filePredicate(FileSystem fileSystem) {
   @Override
   void analyzeFiles(List<InputFile> inputFiles) throws IOException {
     List<String> tsConfigs = tsConfigs();
-    if (tsConfigs.size() > PER_TSCONFIG_ANALYSIS_THRESHOLD) {
-      splitAnalysisByTsConfig(inputFiles, tsConfigs);
-    } else {
-      analyzeFilesWithTsConfigs(inputFiles, tsConfigs);
-    }
+    splitAnalysisByTsConfig(inputFiles, tsConfigs);
```
vilchik-elena: `analyzeFilesWithTsConfigs` does not need to take list of tsconfigs. also I think you can inline `splitAnalysisByTsConfig`

saberduck: I agree with inlining, but parameter is needed (I removed the field to avoid state in the sensor)

PR #1743, opened 2019-11-11
Rely on run-node for MacOS

### LINKED ISSUE #1740: Fix node execution from SonarLint Intellij on MacOS
By default on MacOS we currently launch NodeJS with `bin/sh -c node`. This was done to work properly in SonarLint. Still even it works for VSCode it does not in Intellij (and may be also Eclipse)  as environment is not available and `node` can't be located

To fix that we will rely on [`run-node`](https://www.npmjs.com/package/run-node) package discovering `node` even when it's not part of `PATH`.

vilchik-elena:
Fixes #1740

### DIFF
--- a/nodejs-utils/src/test/java/org/sonarsource/nodejs/NodeCommandTest.java
+++ b/nodejs-utils/src/test/java/org/sonarsource/nodejs/NodeCommandTest.java
@@ -29,7 +29,6 @@ import java.nio.file.Paths;
 import java.util.List;
 import java.util.Map;
 import org.junit.Before;
-import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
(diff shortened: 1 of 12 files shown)


nodejs-utils/src/test/java/org/sonarsource/nodejs/NodeCommandTest.java
```diff
@@ -138,8 +143,9 @@ public void test_version_check() {
     assertThat(NodeCommandBuilderImpl.checkVersion("Invalid version", 6)).isFalse();
   }
 
+  @Ignore
   @Test
-  public void test_max_old_space_size_setting() {
+  public void test_max_old_space_size_setting() throws IOException {
```
vilchik-elena: @saberduck could you have a look why this test is failing?

nodejs-utils/src/test/resources/package/node_modules/run-node/run-node
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env bash
+node "$@"
```
saberduck: nice trick!

PR #1764, opened 2019-12-02
Fix code explanation comment in WithStatement rule

KrzysztofKarol:


vilchik-elena:
Hi, 

Thanks for noticing this typo! I fixed in our rule base (https://jira.sonarsource.com/browse/RSPEC-1321), the change will be applied with the next release.

PR #1777, opened 2019-12-10
[DO-NOT-MERGE] Drop SE

quentin-jaquier-sonarsource:


quentin-jaquier-sonarsource:
Part of an Innovation sprint to Use TypeScript compiler on JavaScript code.

PR #1813, opened 2019-12-16
Add rule S1135 (TODO tag) for JS/TS (new rule for TS)

### LINKED ISSUE #1787: Add rule S1135 (TODO tag) for JS/TS

quentin-jaquier-sonarsource:
Fixes #1787 

### Changes in the rule

We are now reporting only the "TODO" and not the whole line.


quentin-jaquier-sonarsource:
Differences are due to the fact that we are now reporting the whole comment containing the tag.

This change the reported location, especially for multi-line comments, and might remove issues when multiple tags are contained in the same comment.

PR #1816, opened 2019-12-17
Add rule S1067 (Expression Complexity) for JS/TS (new rule for TS)

### LINKED ISSUE #1795: Add rule S1067 (Expression Complexity) for JS/TS

yassin-kammoun-sonarsource:
Fixes #1795


eslint-bridge/src/rules/expression-complexity.ts
```diff
@@ -0,0 +1,145 @@
...
+function reportIssue(
+  node: TSESTree.Node,
+  operators: AST.Token[],
+  max: number,
+  context: Rule.RuleContext,
+) {
+  const complexity = operators.length;
+  const message = `Reduce the number of conditional operators (${complexity}) used in the expression (maximum allowed ${max}).`;
+  const secondaryLocationsHolder = operators;
+  const secondaryMessages = Array(complexity).fill("+1");
```
saberduck: we don't need `new` for `Array()`?

yassin-kammoun-sonarsource: If we omit `new`, `Array()` behaves like a function call and returns a newly instantiated Array object.

PR #1821, opened 2019-12-18
Add rule S1219 (NonCaseLabelInSwitch) for JS/TS (new rule for TS)

### LINKED ISSUE #1799: Add rule S1219 (NonCaseLabelInSwitch) for JS/TS

yassin-kammoun-sonarsource:
Fixes #1799


eslint-bridge/src/rules/no-case-label-in-switch.ts
```diff
@@ -0,0 +1,74 @@
...
+      stack.push(stack.pop()! + 1);
+    }
+    function leaveCase() {
+      stack.push(stack.pop()! - 1);
+    }
+    function inCase() {
+      return stack[stack.length - 1] > 0;
+    }
+    return {
+      "*": (node: estree.Node) => {
```
quentin-jaquier-sonarsource: Matching on `"*"` and then filtering seems weird, are you sure you can not filter further, or match on specific nodes?

yassin-kammoun-sonarsource: Indeed. I don't remember why I did that this way.

its/plugin/projects/ts-rule-project/S1439.ts
```diff
@@ -6,19 +6,16 @@ function compliant(x: number) {
...
+        break label1;
+      } else {
+        break label2;
       }
-    default:
-      return 0;
-  }
+      return 3;
+    }
+  } while (true);
```
yassin-kammoun-sonarsource: This had to change otherwise S1219 would have raised an issue.

its/plugin/projects/ts-rule-project/S1439.ts
```diff
@@ -6,19 +6,16 @@ function compliant(x: number) {
...
-        } else {
-          break label2;
-        }
-        return 3;
+  label1: do {
+    label2: for (let i = 0; i < 3; i++) {
+      if (x % i == 1) {
+        break label1;
+      } else {
+        break label2;
```
yassin-kammoun-sonarsource: This had to change otherwise S1219 would have raised an issue.

eslint-bridge/src/rules/no-case-label-in-switch.ts
```diff
@@ -0,0 +1,66 @@
...
+      },
+      "FunctionExpression, FunctionDeclaration, SwitchCase, LabeledStatement :exit": (
+        node: estree.Node,
+      ) => {
+        if (node.type === "SwitchCase") {
+          leaveCase();
+        } else if (node.type !== "LabeledStatement") {
+          stack.pop();
+        }
+      },
```
quentin-jaquier-sonarsource: I'm still not convinced by this code, why not something like that:

```js
    SwitchCase: () => {
        enterCase();
      },
      LabeledStatement: (node: estree.Node) => {
        if (inCase()) {
          const label = (node as estree.LabeledStatement).label;
          context.report({
            message: `Remove this misleading "${label.name}" label.`,
            node: label,
          });
        }
      },
      "FunctionExpression, FunctionDeclaration": () => {
        stack.push(0);
      },
      "SwitchCase:exit": () => {
        leaveCase();
      },
      "FunctionExpression, FunctionDeclaration :exit": () => {
        stack.pop();
      },
```

yassin-kammoun-sonarsource: Definitely better, thanks !

PR #1823, opened 2019-12-18
Add rule S1128 (Unused Import) for JS/TS (new rule for TS)

### LINKED ISSUE #1796: Add rule S1128 (Unused Import) for JS/TS

yassin-kammoun-sonarsource:
Fixes #1796 

### Changes in the rule
- We non longer raise issues for import duplications since those produce parsing errors with the JavaScript parsers we use.


its/ruling/src/test/expected/js/jshint/javascript-S1128.json
```diff
@@ -1,11 +0,0 @@
-{
-'jshint:tests/unit/fixtures/es6-import-export.js':[
```
yassin-kammoun-sonarsource: Parsing this file fails because of some default export duplication.

eslint-bridge/src/parser.ts
```diff
@@ -56,9 +56,9 @@ export type Parse = (
 ) => SourceCode | ParsingError;
 
 export function parseJavaScriptSourceFile(fileContent: string): SourceCode | ParsingError {
-  let parseFunctions = [espree.parse, babel.parse];
+  let parseFunctions = [espree.parse, babel.parseForESLint];
```
vilchik-elena: I think we should use `parseForESLint` for all parsers, that way you will not need condition on line 152

javascript-checks/src/main/java/org/sonar/javascript/checks/UnusedImportCheck.java
```diff
@@ -19,45 +19,17 @@
...
-import org.sonar.plugins.javascript.api.symbols.Symbol;
-import org.sonar.plugins.javascript.api.symbols.SymbolModel;
-import org.sonar.plugins.javascript.api.symbols.Usage;
-import org.sonar.plugins.javascript.api.tree.ModuleTree;
-import org.sonar.plugins.javascript.api.tree.expression.IdentifierTree;
-import org.sonar.plugins.javascript.api.visitors.DoubleDispatchVisitorCheck;
+import org.sonar.javascript.checks.annotations.TypeScriptRule;
 
 @JavaScriptRule
+@TypeScriptRule
```
vilchik-elena: add "new rule for TS" in title

its/ruling/src/test/expected/js/javascript-test-sources/javascript-S1128.json
```diff
@@ -52,6 +53,9 @@
 5,
 5,
 ],
+'javascript-test-sources:src/ecmascript6/http/test/Serialize.spec.js':[
+1,
```
vilchik-elena: why do we have these new issues?

eslint-bridge/src/rules/unused-import.ts
```diff
@@ -0,0 +1,74 @@
...
+import * as estree from "estree";
+import { TSESTree } from "@typescript-eslint/experimental-utils";
+
+const EXCLUDED_IMPORTS = ["React"];
+
+export const rule: Rule.RuleModule = {
+  create(context: Rule.RuleContext) {
+    const unusedImports: estree.Identifier[] = [];
+    const tsTypeIdentifiers: Set<string> = new Set();
+    const saveTypeIdentifier = (node: estree.Node) =>
```
vilchik-elena: I think it's safer to pass identifier type as argument instead of casting inside

its/ruling/src/test/expected/js/javascript-test-sources/javascript-S1128.json
yassin-kammoun-sonarsource: The Java implementation of the rule used to visit only imports within modules, not scripts.

yassin-kammoun-sonarsource: These new issues are from files that are scripts.

eslint-bridge/src/parser.ts
quentin-jaquier-sonarsource: It seems to me that `espree` does not provide a `parseForESLint` function, I don't think we can avoid the condition on line 152 nicely...

its/ruling/src/test/expected/js/javascript-test-sources/javascript-S1128.json
vilchik-elena: This does not make a lot of sense. File can't be script if it contains `import`

yassin-kammoun-sonarsource: Well, it depends on what basis the Java frontend considers a file to be a script or a module. I tried to analyze one of those files using directly the Java implementation of the rule, and the callbacks - one for module, one for import - were never invoked. I then added a callback for script, and it worked. I didn't investigate further the matter, but my guess is that the frontend considers a file as a module if it has a (default?) export statement.

PR #1824, opened 2019-12-18
Add rule S1134 (FIXME tag) for JS/TS (new rule for TS)

### LINKED ISSUE #1786: Add rule S1134 (FIXME tag) for JS/TS

quentin-jaquier-sonarsource:
Fixes #1786


eslint-bridge/src/rules/todo-tag.ts
```diff
@@ -22,53 +22,65 @@
...
+  context: Rule.RuleContext,
+  pattern: string,
+  message: string,
+) {
+  const sourceCode = context.getSourceCode();
+  (sourceCode.getAllComments() as TSESTree.Comment[]).forEach(comment => {
+    const rawText = comment.value.toLowerCase();
+
+    if (rawText.includes(pattern)) {
+      const lines = rawText.split(/\r\n?|\n/);
```
saberduck: maybe use `/\r?\n/` for regex, it's a bit simpler. `\r` should not be used alone afaik

quentin-jaquier-sonarsource: I believe that `\r` alone is still (rarely) used, and this regex is also used in other plugins (php, java), I think we can keep it as it is.

PR #1829, opened 2019-12-23
Add rule S1472 (FunctionCallArgumentsOnNewLine) for JS/TS (new rule for TS)

### LINKED ISSUE #1800: Add rule S1472 (FunctionCallArgumentsOnNewLine) for JS/TS

yassin-kammoun-sonarsource:
Fixes #1800


its/ruling/src/test/expected/js/p5.js/javascript-S1472.json
```diff
@@ -0,0 +1,6 @@
+{
+'p5.js:test/js/sinon.js':[
+65,
```
vilchik-elena: I believe these 2 are FPs
Looks like when callee is parenthesised we don't correctly find its last line and thus report such FPs.

Test to reproduce:
```
(foo
)(bar);
```

eslint-bridge/src/rules/call-argument-line.ts
```diff
@@ -0,0 +1,45 @@
...
+export const rule: Rule.RuleModule = {
+  create(context: Rule.RuleContext) {
+    return {
+      CallExpression: (node: estree.Node) => {
+        const call = node as estree.CallExpression;
+        if (call.callee.type !== "CallExpression" && call.arguments.length === 1) {
+          const sourceCode = context.getSourceCode();
+          const calleeLastLine = sourceCode.getLastToken(call.callee)!.loc.end.line;
+          const { start } = sourceCode.getTokenAfter(call.callee)!.loc;
+          const { end } = sourceCode.getLastToken(call)!.loc;
```
vilchik-elena: computation of `end` can be moved inside of condition

PR #1830, opened 2019-12-26
Add rule S1539 (StrictMode) for JS/TS (new rule for TS)

### LINKED ISSUE #1805: Add rule S1539 (StrictMode) for JS/TS

quentin-jaquier-sonarsource:
Fixes #1805


its/ruling/src/test/expected/js/p5.js/javascript-S1539.json
```diff
@@ -5,17 +5,8 @@
 4625,
 ],
 'p5.js:lib/addons/p5.sound.js':[
-77,
```
quentin-jaquier-sonarsource: In fact, these missing issues are not due to the single quotes, but to the fact that the 'use strict' is not the first statement in the function, therefore this is not a correct 'use strict'.

its/ruling/src/test/expected/js/jshint/javascript-S1539.json
```diff
@@ -98,16 +98,8 @@
 'jshint:tests/unit/fixtures/latedefundef.js':[
 16,
 ],
-'jshint:tests/unit/fixtures/reserved.js':[
```
vilchik-elena: do you know why these issues are removed?

vilchik-elena: parsing is failing, that's why

quentin-jaquier-sonarsource: Yes, I confirm that reserved.js, strings.js and with.js parsing is failing. For strict_incorrect.js, it is due to the fact that the literal is not in first position.

PR #1833, opened 2019-12-27
Add rule S1515 (FunctionDefinitionInsideLoop) for JS/TS (new rule for TS)

### LINKED ISSUE #1802: Add rule S1515 (FunctionDefinitionInsideLoop) for JS/TS

quentin-jaquier-sonarsource:
Fixes #1802


quentin-jaquier-sonarsource:
I inspired myself from both the old Java implementation and Eslint core rule [no-loop-func](https://eslint.org/docs/rules/no-loop-func) implementation.

its/ruling/src/test/expected/js/qunit/javascript-S1515.json
```diff
@@ -1,5 +0,0 @@
-{
-'qunit:src/cli/run.js':[
-59,
```
quentin-jaquier-sonarsource: The code looks good to me, and is not reported either by [no-loop-func](https://eslint.org/docs/rules/no-loop-func), I think it's fine to not report it.

eslint-bridge/tests/rules/function-inside-loop.test.ts
```diff
@@ -0,0 +1,264 @@
...
+          
+          for (var i = 0; i < 10; i++) {
+              funs[i] = (function * () {       // Noncompliant
+                  return function() {           // OK
+                      return i;
+                  };
+              }(i));
+          }
+      `,
+      errors: [
```
vilchik-elena: why  4 `// Noncompliant` in the code and just 2 issues?

eslint-bridge/tests/rules/function-inside-loop.test.ts
```diff
@@ -0,0 +1,264 @@
...
+    {
+      code: `
+      while (true) {
+        new Promise((resolve) => setTimeout(resolve, 5000))
+        await ping()
+      }
+      `,
+    },
+  ],
+  invalid: [],
```
vilchik-elena: can we add invalid test for TS?

eslint-bridge/tests/rules/function-inside-loop.test.ts
```diff
@@ -0,0 +1,264 @@
...
+const ruleTesterTS = new RuleTester({
+  parserOptions: { ecmaVersion: 2018, sourceType: "module" },
+  parser: require.resolve("@typescript-eslint/parser"),
+});
+
+ruleTesterTS.run("Functions should not be defined inside loops", rule, {
+  valid: [
+    {
+      code: `
+      while (true) {
```
vilchik-elena: why this test for TS?

eslint-bridge/src/rules/function-inside-loop.ts
```diff
@@ -0,0 +1,151 @@
...
+  "every",
+  "some",
+  "reduce",
+  "reduceRight",
+  "sort",
+];
+
+export const rule: Rule.RuleModule = {
+  create(context: Rule.RuleContext) {
+    const functionAndLoopScopes: estree.Node[] = [];
```
vilchik-elena: why storing this structure? can't we rely on `parents` of context?

eslint-bridge/src/rules/function-inside-loop.ts
```diff
@@ -0,0 +1,151 @@
...
+          const mainLocation = getMainFunctionTokenLocation(
+            node as estree.Function,
+            getParent(context),
+            context,
+          );
+
+          const loopNode = peek(functionAndLoopScopes);
+
+          if (
+            mainLocation &&
```
vilchik-elena: why this condition? It seems to me that mainLocation is always available. And we can move its computation inside this `if`

eslint-bridge/tests/rules/function-inside-loop.test.ts
quentin-jaquier-sonarsource: This code can not be parsed by JS parser, but in fact it's due to `await ping()`.
Sine this part of the test code is not critical, I removed it and used only one RuleTester.

eslint-bridge/tests/rules/function-inside-loop.test.ts
quentin-jaquier-sonarsource: I removed RuleTester for TS, it is not needed anymore.

eslint-bridge/src/rules/function-inside-loop.ts
quentin-jaquier-sonarsource: Correct, I took this code from another rule, but it seems indeed that mainLocation is always available, I refactored the code.

eslint-bridge/src/rules/function-inside-loop.ts
quentin-jaquier-sonarsource: Good point

quentin-jaquier-sonarsource:
@vilchik-elena In addition to your comments, I tried to search for enclosing loop outside of a local scope (see last commit), I think the results are good enough, let me know what you think.

PR #1834, opened 2019-12-27
Fix eval-or-arguments-use undefined foreach

quentin-jaquier-sonarsource:

### DIFF
--- a/eslint-bridge/tests/rules/eval-or-arguments-use.test.ts
+++ b/eslint-bridge/tests/rules/eval-or-arguments-use.test.ts
@@ -162,3 +162,18 @@ ruleTester.run('"eval" and "arguments" should not be bound or assigned', rule, {
     },
   ],
 });
+
+const ruleTesterBabel = new RuleTester({
+  parser: require.resolve("babel-eslint"),
+});
+
+ruleTesterBabel.run('"eval" and "arguments" should not be bound or assigned', rule, {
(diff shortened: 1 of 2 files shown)


quentin-jaquier-sonarsource:
**We should really use `":function"` with care in the future.**

It's in fact defined as `.startsWith("function")`, since we are using different parser, it doesn't mean that the matching node will be of type `estree.Function`.
In this specific case, when parsing with babel-eslint, a function can be FunctionTypeParam, with params property possibly undefined, causing the exception.

I double check uses of `":function"` in current code, the only place we are using it is [here](https://github.com/SonarSource/SonarJS/blob/master/eslint-bridge/src/rules/no-parameter-reassignment.ts#L219), and it seems fine to me.

eslint-bridge/tests/rules/eval-or-arguments-use.test.ts
```diff
@@ -129,8 +129,8 @@ ruleTester.run('"eval" and "arguments" should not be bound or assigned', rule, {
         function fun(arguments, ...a) {}
         var f = function(eval) {}
         const arguments = eval;
-        `,
-      errors: 7,
+        var arrowFunction = (arguments) => {}`,
```
vilchik-elena: I believe we should add test which was failing before, with FunctionParameterType

PR #1880, opened 2020-01-08
Add rule S1994 (ForLoopConditionAndUpdateCheck) for JS/TS (new rule for TS)

### LINKED ISSUE #1808: Add rule S1994 (ForLoopConditionAndUpdate) for JS/TS (new rule for TS)

vilchik-elena:
Fixes #1808


eslint-bridge/src/rules/misplaced-loop-counter.ts
```diff
@@ -0,0 +1,167 @@
...
+      },
+
+      CallExpression: (node: estree.Node) => {
+        const callee = getCallee(node as estree.CallExpression);
+        if (isInsideUpdate(node) && callee != null) {
+          peekFor().updatedExpressions.push(callee);
+        }
+      },
+
+      Identifier: (node: estree.Node) => {
```
quentin-jaquier-sonarsource: Maybe we can use selectors `"ForStatement Identifier"` to avoid computing `isInsideTest(node)` for every identifier, if we already know that it's not in a `for`.
The same can apply for others nodes aswell.

_Ideally, we would want to compute `isInsideTest(node)` directly in the selector, something like `"ForStatement.test Identifier"`, but I'm not sure it's possible._

eslint-bridge/src/rules/misplaced-loop-counter.ts
```diff
@@ -0,0 +1,167 @@
...
+  while (callee.type === "MemberExpression") {
+    callee = callee.object;
+  }
+  if (callee.type === "Identifier" && callee !== node.callee) {
+    return callee;
+  }
+  return null;
+}
+
+function computeAssignedExpressions(node: estree.Node, assigned: estree.Node[]) {
```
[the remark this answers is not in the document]
vilchik-elena: I prefer to keep mine. In the case like this `{ a: a[i] } = obj` mine implementation will return `a[i]`, while impl from `utils` will return nothing

eslint-bridge/src/rules/misplaced-loop-counter.ts
vilchik-elena: Yeah, I tried to have something like `"ForStatement.test Identifier"` but didn't manage. Still having "ForStatement Identifier" is a good idea

PR #1884, opened 2020-01-10
Fix conflicting keys in different plugins

quentin-jaquier-sonarsource:


its/plugin/projects/eslint_based_rules_with_ts/main.js
```diff
@@ -0,0 +1,9 @@
+/*eslint no-magic-numbers: "error"*/
```
vilchik-elena: could we remove first 2 lines to not test same thing twice (and not test 2 things at a time :D)

PR #1885, opened 2020-01-10
Add rule S2310 (CounterUpdatedInLoopCheck) for JS/TS (new rule for TS)

### LINKED ISSUE #1855: Add rule S2310 (CounterUpdatedInLoopCheck) for JS/TS (new rule for TS)

vilchik-elena:
Fixes #1855


eslint-bridge/tests/rules/updated-loop-counter.test.ts
```diff
@@ -0,0 +1,250 @@
...
+      code: `
+      let fl = false;
+
+      for (; i < m && !fl; i++) {
+        fl = true;   // Compliant
+        m = 10;      // Compliant
+      }
+      `,
+      errors: [],
+    },
```
quentin-jaquier-sonarsource: I think you can move this test to the valid list.

eslint-bridge/src/rules/updated-loop-counter.ts
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    return {
+      "ForStatement > BlockStatement": (node: estree.Node) => {
+        const forLoop = getParent(context) as estree.ForStatement;
+        if (forLoop.update) {
+          checkLoop(forLoop.update, collectCountersFor, node);
+        }
+      },
+      ":matches(ForInStatement > BlockStatement, ForOfStatement > BlockStatement)": (
```
quentin-jaquier-sonarsource: I don't think the `:matches` is needed here.

eslint-bridge/src/rules/updated-loop-counter.ts
```diff
@@ -0,0 +1,115 @@
...
+  } else if (updateExpression.type === "SequenceExpression") {
+    updateExpression.expressions.forEach(e => collectCountersFor(e, counters));
+  }
+
+  if (counter && counter.type === "Identifier") {
+    counters.push(counter);
+  }
+}
+
+function isUsedInsideBody(id: estree.Identifier, loopBody: estree.Node) {
```
quentin-jaquier-sonarsource: We have the same logic [here](https://github.com/SonarSource/SonarJS/blob/master/eslint-bridge/src/rules/no-equals-in-for-termination.ts#L178), you may want to re-use it

eslint-bridge/tests/rules/updated-loop-counter.test.ts
```diff
@@ -0,0 +1,250 @@
...
+          i = 5;      // Noncompliant 
+          j = 5;      // compliant, not in update section
+        }
+      `,
+      errors: [
+        {
+          line: 3,
+          column: 11,
+          endColumn: 12,
+          message: `{"message":"Remove this assignment of \\"i\\".","secondaryLocations":[{"message":"Counter variable update","column":38,"line":2,"endColumn":39,"endLine":2}]}`,
```
quentin-jaquier-sonarsource: We only test concretely when the message reports variable `i`, we should change the name for one test.

eslint-bridge/src/rules/updated-loop-counter.ts
vilchik-elena: it's just one-liner, I think it's ok to keep this duplication

PR #1888, opened 2020-01-10
Add rule S3513 (ArgumentsUsageCheck) for JS/TS (new rule for TS)

### LINKED ISSUE #1865: Add rule S3513 (ArgumentsUsageCheck) for JS/TS (new rule for TS)

alban-auzeill:
Fixes #1865


eslint-bridge/src/rules/arguments-usage.ts
```diff
@@ -0,0 +1,64 @@
...
+import { Rule, Scope } from "eslint";
+import { TSESTree } from "@typescript-eslint/experimental-utils";
+import { toEncodedMessage } from "./utils";
+
+const message = "Use the rest syntax to declare this function's arguments.";
+
+export const rule: Rule.RuleModule = {
+  create(context: Rule.RuleContext) {
+    return {
+      "Program:exit": () => checkArgumentsUsageInScopeRecursively(context, context.getScope()),
```
quentin-jaquier-sonarsource: Problematic `arguments` can only appear in functions right? I would prefer to match on functions to avoid checking the global scope every time.

eslint-bridge/src/rules/arguments-usage.ts
```diff
@@ -0,0 +1,64 @@
...
+// https://jira.sonarsource.com/browse/RSPEC-3513
+
+import { Rule, Scope } from "eslint";
+import { TSESTree } from "@typescript-eslint/experimental-utils";
+import { toEncodedMessage } from "./utils";
+
+const message = "Use the rest syntax to declare this function's arguments.";
+
+export const rule: Rule.RuleModule = {
+  create(context: Rule.RuleContext) {
```
quentin-jaquier-sonarsource: You have to add this parameters for rules having secondary location, otherwise, it will not work in Sonarqube!
```
  meta: {
    schema: [
      {
        // internal parameter for rules having secondary locations
        enum: ["sonar-runtime"],
      },
    ],
  },
```

See [this rule](https://github.com/SonarSource/SonarJS/blob/master/eslint-bridge/src/rules/no-variable-usage-before-declaration.ts) for example.

eslint-bridge/src/rules/arguments-usage.ts
alban-auzeill: If we register on FunctionDeclaration and FunctionExpression, we would have some duplicated issues in case of nested functions, because we look into scopes recursively. Instead, we could just ignore the root scope, it's less complex to implement.

PR #1893, opened 2020-01-14
Add rule S3973 (ConditionalIndentationCheck) for JS/TS (new rule for TS)

### LINKED ISSUE #1873: Add rule S3973 (ConditionalIndentationCheck) for JS/TS (new rule for TS)

quentin-jaquier-sonarsource:
Fixes #1873


eslint-bridge/src/rules/conditional-indentation.ts
```diff
@@ -0,0 +1,90 @@
...
+          const alternate = ifStatement.alternate;
+          if (alternate.type === "IfStatement") {
+            //case with "else if", we have to check the consequent of the next if
+            checkIndentation(elseToken, alternate.consequent, context);
+          } else {
+            checkIndentation(elseToken, alternate, context);
+          }
+        }
+      },
+      "WhileStatement, ForStatement, ForInStatement, ForOfStatement": (node: estree.Node) => {
```
saberduck: what about `DoWhileStatement` ?

quentin-jaquier-sonarsource: It was not reported in the Java implementation, and I believe the reason is that code like this:
```js
do
i++;
while(i < 10);
```

can't be ambiguous as we can't have more than one line between `do` and `while`. 

I agree that it's bad coding style, but reporting it with Critical severity seems harsh. 
Should we report it anyway?

saberduck: makes sense, we can keep as is

PR #1894, opened 2020-01-14
Add rule S134 (NestedControlFlowDepthCheck) for JS/TS (new for TS)

### LINKED ISSUE #1874: Add rule S134 (NestedControlFlowDepthCheck) for JS/TS (new for TS)

saberduck:
Fixes #1874


eslint-bridge/src/rules/nested-control-flow.ts
```diff
@@ -0,0 +1,108 @@
...
+      "ForInStatement",
+      "ForOfStatement",
+      "WhileStatement",
+      "DoWhileStatement",
+      "IfStatement",
+      "TryStatement",
+      "SwitchStatement",
+    ].join(",");
+    return {
+      [`:matches(${controlFlowNodes})`]: (node: estree.Node) => {
```
quentin-jaquier-sonarsource: We don't need `:matches` here.
Neither at line 78.

eslint-bridge/src/rules/nested-control-flow.ts
```diff
@@ -0,0 +1,108 @@
...
+    function push(n: AST.Token) {
+      nodeStack.push(n);
+    }
+    function pop() {
+      return nodeStack.pop();
+    }
+    function check(node: estree.Node) {
+      if (nodeStack.length === threshold) {
+        context.report({
+          message: encodeMessage(threshold, nodeStack.map(n => toSecondaryLocation(n))),
```
quentin-jaquier-sonarsource: You can use `toEncodedMessage` from `./utils` instead.

eslint-bridge/src/rules/nested-control-flow.ts
```diff
@@ -0,0 +1,108 @@
...
+      "ForOfStatement",
+      "WhileStatement",
+      "DoWhileStatement",
+      "IfStatement",
+      "TryStatement",
+      "SwitchStatement",
+    ].join(",");
+    return {
+      [`:matches(${controlFlowNodes})`]: (node: estree.Node) => {
+        if (!isElseIf(node)) {
```
quentin-jaquier-sonarsource:
```js
      if (true) {

      } else if (false) {
        for (const i of arr) {
        }
      }
```

This code will report the first `if` as secondary location, and not else if one. I believe you can pop/push in the else of this condition to improve the location.

PR #1898, opened 2020-01-15
Add rule S1527 (FutureReservedWords) for JS/TS

### LINKED ISSUE #1803: Add rule S1527 (FutureReservedWords) for JS/TS

vilchik-elena:
Fixes #1803


its/ruling/src/test/expected/js/jshint/javascript-FutureReservedWords.json
```diff
@@ -1,6 +0,0 @@
-{
-'jshint:tests/unit/fixtures/reserved.js':[
```
vilchik-elena: this file is not parsed

eslint-bridge/src/rules/future-reserved-words.ts
```diff
@@ -0,0 +1,70 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+// https://jira.sonarsource.com/browse/RSPEC-1523
```
quentin-jaquier-sonarsource: 152**7**

eslint-bridge/src/rules/future-reserved-words.ts
```diff
@@ -0,0 +1,70 @@
...
+  "static",
+  "yield",
+  "await",
+];
+
+export const rule: Rule.RuleModule = {
+  create(context: Rule.RuleContext) {
+    let reported: Scope.Variable[] = [];
+    return {
+      Program: (_node: estree.Node) => {
```
quentin-jaquier-sonarsource: Isn't it better to remove the parameter instead of using the prefix `_`?

eslint-bridge/src/rules/future-reserved-words.ts
```diff
@@ -0,0 +1,70 @@
...
+        reported = [];
+      },
+      Identifier: (node: estree.Node) => {
```
quentin-jaquier-sonarsource: I have the feeling that it is not ideal to match on Identifier, we will potentially do a lot of unnecessary work, and if the project is using reserved keyword, we will reach line 59 for **every** reserved identifier use, even though we are reporting only the declaration!

What about an approach similar to [no-empty-collection, line 88](https://github.com/SonarSource/SonarJS/blob/master/eslint-bridge/src/rules/no-empty-collection.ts#L88), where we could go though every Variable (_Symbol_) only once?

vilchik-elena:
One condition is not covered (`variable.defs.length > 0`) but I prefer to keep it to be more defensive

PR #1900, opened 2020-01-15
S4798: Do not report on function with no body

### LINKED ISSUE #1768: FP S4798: function without body
Functions without body should be excluded as it is not possible to add a parameter initializer.

**Reproducer** 
- Function overloads
```typescript
function foo(b?: boolean); // FP, it's not possible to add a parameter initializer here
function foo(b) {
  // ...
} 
```
- Abstract methods

```ts
abstract class A{
    abstract foo(p?: boolean): number; // FP
}
```

- Interface methods
```ts
interface I {
  foo: (b?:  boolean) => void; // FP
}
```

- Function Type
```ts
type Foo = (p?: boolean) => void; //A parameter initializer is only allowed in a function or constructor implementation
```

quentin-jaquier-sonarsource:
Fixes #1768 
This issue was also reported in SonarTS:
Fixes https://github.com/SonarSource/SonarTS/issues/823


eslint-bridge/src/rules/bool-param-default.ts
```diff
@@ -27,15 +27,12 @@ type FunctionLike =
...
-  | TSESTree.TSMethodSignature
-  | TSESTree.TSFunctionType
-  | TSESTree.TSConstructSignatureDeclaration;
+  | TSESTree.TSFunctionType;
 
 export const rule: Rule.RuleModule = {
   create(context: Rule.RuleContext) {
     return {
-      "FunctionDeclaration, FunctionExpression, ArrowFunctionExpression, TSEmptyBodyFunctionExpression, TSMethodSignature, TSFunctionType, TSConstructSignatureDeclaration": (
+      "FunctionDeclaration, FunctionExpression, ArrowFunctionExpression, TSFunctionType": (
```
vilchik-elena: I checked `TSFunctionType`, TS compiler also does not allow initializer there:
```
type Foo = (p: boolean = true) => void; //A parameter initializer is only allowed in a function or constructor implementation
```

eslint-bridge/src/rules/bool-param-default.ts
```diff
@@ -27,15 +27,12 @@ type FunctionLike =
   | TSESTree.FunctionDeclaration
   | TSESTree.FunctionExpression
   | TSESTree.ArrowFunctionExpression
-  | TSESTree.TSEmptyBodyFunctionExpression
-  | TSESTree.TSMethodSignature
-  | TSESTree.TSFunctionType
-  | TSESTree.TSConstructSignatureDeclaration;
```
vilchik-elena: Could you add test for construct signature?

PR #1901, opened 2020-01-16
Add rule RSPEC-1529 (BitwiseOperatorsCheck) for JS/TS (new rule for TS)

### LINKED ISSUE #1850: Add rule RSPEC-1529 (BitwiseOperatorsCheck) for JS/TS (new rule for TS)

alban-auzeill:
Fixes #1850


eslint-bridge/src/rules/bitwise-operators.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    function isNumericType(type: tsTypes.Type): boolean {
+      return (
+        (type.getFlags() & (ts.TypeFlags.NumberLike | ts.TypeFlags.BigIntLike)) !== 0 ||
+        (type.isUnionOrIntersection() && !!type.types.find(isNumericType))
+      );
+    }
+
+    return (node: estree.Node) =>
+      isNumericType(checker.getTypeAtLocation(nodeMap.get(node as TSESTree.Node)));
```
quentin-jaquier-sonarsource: You could use `getTypeFromTreeNode` from `./utils` here.

eslint-bridge/src/rules/bitwise-operators.ts
```diff
@@ -0,0 +1,126 @@
...
+              message: `Review this use of bitwise "${op}" operator; conditional "${op}${op}" might have been intended.`,
+            });
+          }
+        }
+      },
+    };
+  },
+};
+
+function insideCondition(node: estree.Node, ancestors: estree.Node[]) {
```
quentin-jaquier-sonarsource: You could use `findFirstMatchingAncestor` from `./utils` here, to avoid storing the ancestors.

alban-auzeill: `findFirstMatchingAncestor` could be used to find the first ancestor of type: If, For, While, ... But it's not possible to check directly that the descended is inside the `test` property. So I prefer not using `findFirstMatchingAncestor`.

eslint-bridge/src/rules/bitwise-operators.ts
```diff
@@ -0,0 +1,123 @@
...
+      return parent.test === child;
+    }
+    child = parent;
+  }
+  return false;
+}
+
+type NumericTypeChecker = (node: estree.Node) => boolean;
+
+function getNumericTypeChecker(context: Rule.RuleContext): NumericTypeChecker {
```
quentin-jaquier-sonarsource: I like this code, I will probably re-use the idea in the future!

PR #1907, opened 2020-01-16
Add own implementation of S138 to fix issue location

### LINKED ISSUE #1750: S138 TooManyLinesInFunctionCheck: report on function signature, not entire function
As we are relying on Eslint for this rule, it's reporting on the entire function, which looks ugly in SonarLint and SonarQube UI. We should probably implement our own rule reporting on function signature (name) only.

quentin-jaquier-sonarsource:
Fixes #1750


eslint-bridge/src/rules/sonar-max-lines-per-function.ts
```diff
@@ -0,0 +1,116 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+// Greatly inspired by https://github.com/eslint/eslint/blob/master/lib/rules/max-lines-per-function.js
```
vilchik-elena: it's better to put permalink

vilchik-elena: also mention that we had to fork implementation to have proper reporting

PR #1912, opened 2020-01-17
Add debug logs when importing external issues report

### LINKED ISSUE #1840: Add debug logs when importing external issues report
Currently we don't log anything when importing external issues report (only report path), which makes it hard to investigate if something goes wrong. I suggest we log a debug message for each issue in the report (file, line and rule key might be useful in this message).

It will help in future to investigate cases like [this one](https://community.sonarsource.com/t/dont-see-tslint-report-into-sonarcloud/17385/11).

alban-auzeill:
Fixes #1840

### DIFF
--- a/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/external/EslintReportSensor.java
+++ b/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/external/EslintReportSensor.java
@@ -26,10 +26,13 @@ import java.io.IOException;
 import java.io.InputStreamReader;
 import java.nio.charset.StandardCharsets;
 import org.sonar.api.batch.fs.InputFile;
+import org.sonar.api.batch.fs.TextPointer;
 import org.sonar.api.batch.fs.TextRange;
+import org.sonar.api.batch.rule.Severity;
 import org.sonar.api.batch.sensor.SensorContext;
 import org.sonar.api.batch.sensor.issue.NewExternalIssue;
 import org.sonar.api.batch.sensor.issue.NewIssueLocation;
+import org.sonar.api.rules.RuleType;
 import org.sonar.api.utils.log.Logger;
 import org.sonar.api.utils.log.Loggers;
 import org.sonar.plugins.javascript.rules.EslintRulesDefinition;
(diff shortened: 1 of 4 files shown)


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/external/TslintReportSensor.java
```diff
@@ -70,18 +71,24 @@ private void saveTslintError(SensorContext context, TslintError tslintError) {
     if (inputFile == null) {
       return;
     }
+    TextRange location = getLocation(tslintError, inputFile);
+    RuleType ruleType = TslintRulesDefinition.ruleType(tslintKey);
+
+    LOG.debug("External issue, file:\"{}\", id:{}, message:\"{}\", location:{}, type: {}",
+      tslintError.name, tslintKey, tslintError.failure, location, ruleType);
```
vilchik-elena: I think just line for the location should be enough. Also relying on `TextRange#toString` is not very reliable, SQ may change that in the future

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/external/TslintReportSensor.java
```diff
@@ -71,11 +72,13 @@ private void saveTslintError(SensorContext context, TslintError tslintError) {
...
       return;
     }
+
     TextRange location = getLocation(tslintError, inputFile);
+    TextPointer start = location.start();
     RuleType ruleType = TslintRulesDefinition.ruleType(tslintKey);
 
-    LOG.debug("External issue, file:\"{}\", id:{}, message:\"{}\", location:{}, type: {}",
-      tslintError.name, tslintKey, tslintError.failure, location, ruleType);
+    LOG.debug("External Tslint issue { file:\"{}\", id:{}, message:\"{}\", line:{}, offset:{}, type: {} }",
```
vilchik-elena: -> `Saving external TSLint issue` (add verb and fix casing of tslint)

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/external/EslintReportSensor.java
```diff
@@ -72,29 +75,37 @@ void importReport(File report, SensorContext context) {
...
     }
 
+    TextRange location = getLocation(eslintError, inputFile);
+    TextPointer start = location.start();
+    ExternalRuleLoader ruleLoader = EslintRulesDefinition.loader(eslintKey);
+    RuleType ruleType = ruleLoader.ruleType(eslintKey);
+    Severity severity = ruleLoader.ruleSeverity(eslintKey);
+    Long effortInMinutes = ruleLoader.ruleConstantDebtMinutes(eslintKey);
+
+    LOG.debug("External Eslint issue { file:\"{}\", id:{}, message:\"{}\", line:{}, offset:{}, type: {}, severity:{}, remediation:{} }",
```
vilchik-elena: same here:
-> `Saving external ESLint issue`

PR #1914, opened 2020-01-17
Improve error message when Node.js is unresponsive

### LINKED ISSUE #1847: Rephrase timeout error to be more user friendly

saberduck:
Fixes #1847


sonar-javascript-plugin/src/main/resources/static/documentation.md
```diff
@@ -28,6 +28,19 @@ There are 2 built-in rule profiles for JavaScript: `Sonar way` (default) and `So
 * `Sonar way` profile is activated by default. It defines a trimmed list of high-value/low-noise rules useful in almost any JS development context.
 * `Sonar way Recommended` contains all rules from `Sonar way`, plus more rules that mandate high code readability and long-term project evolution.
 
+## Troubleshooting
+
+### Slow or unresponsive analysis
+
+On a big project it can happen that analyzer needs to allocate a lot of memory to analyze the project. This would manifest by analysis getting stuck and following stacktrace might appear in the logs
```
vilchik-elena: should not it be `This would be manifested` in passive?

vilchik-elena: `a lot` -> `more`

vilchik-elena: **the** following

sonar-javascript-plugin/src/main/resources/static/documentation.md
```diff
@@ -28,6 +28,19 @@ There are 2 built-in rule profiles for JavaScript: `Sonar way` (default) and `So
...
+
+```
+ERROR: Failed to get response while analyzing [file].ts
+java.io.InterruptedIOException: timeout
+```   
+You can use `sonar.javascript.node.maxspace` property to allow the analyzer to use more memory. Set this property to `4096` or `8192` for big projects. This property should be set in `sonar-project.properties` file or on command line for scanner (with `-Dsonar.javascript.node.maxspace=4096`).
```
vilchik-elena: Are you sure that the only 2 ways to specify properties? Should not we use some more generic phasing?

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
```diff
@@ -192,6 +193,13 @@ private String request(String json, String endpoint) throws IOException {
     try (Response response = client.newCall(request).execute()) {
       // in this case response.body() is never null (according to docs)
       return response.body().string();
+    } catch (InterruptedIOException e) {
+      if ("timeout".equals(e.getMessage())) {
+        String msg = "eslint-bridge Node.js process is unresponsive. This is most likely caused by process running out of memory." +
+          " Consider setting sonar.javascript.node.maxspace to higher value (e.g. 4096).";
+        LOG.error(msg);
```
vilchik-elena: can we test this log?

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImplTest.java
```diff
@@ -154,10 +154,9 @@ public void should_throw_if_failed_to_start() throws Exception {
     eslintBridgeServer = createEslintBridgeServer("throw.js");
     eslintBridgeServer.deploy();
 
-    thrown.expect(NodeCommandException.class);
-    thrown.expectMessage("Failed to start server (" + TEST_TIMEOUT_SECONDS + "s timeout)");
-
-    eslintBridgeServer.startServer(context);
+    assertThatThrownBy(() -> eslintBridgeServer.startServer(context))
```
vilchik-elena: is it just refactoring?

sonar-javascript-plugin/src/main/resources/static/documentation.md
saberduck: I prefer to be concrete, if we are generic people will not understand and set the property on the server side. This should work with all scanners afaik

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImplTest.java
saberduck: no, for some reason this assertion doesn't work on windows anymore. Assertj is a bit better

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
saberduck: added test

PR #1922, opened 2020-01-20
Add rule S1119 (LabelledStatementCheck) for JS/TS (new for TS)

### LINKED ISSUE #1904: Add rule S1119 (LabelledStatementCheck) for JS/TS (new for TS)
Try https://eslint.org/docs/rules/no-labels

saberduck:
Fixes #1904


javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/Sonar_way_profile.json
```diff
@@ -13,7 +13,6 @@
     "S1110",
     "S1116",
     "S1117",
-    "S1119",
```
quentin-jaquier-sonarsource: As discussed, we should add S1439 in Sonar-way instead of this one.

PR #1970, opened 2020-03-16
Prefer direct TypeScript dependency (#1928)

### LINKED ISSUE #1928: Prefer direct project TypeScript dependency to a transitive one
Project might have a direct TypeScript dependency (located in top level `node_modules`) and a transitive one (located in nested `node_modules`). In this case when searching for a TypeScript dependency we might first find the transitive dependency which has an old version and analysis will not be executed.

See https://github.com/SonarSource/SonarJS/blob/fdc6425a982d78ef9dd62feccdad884ea52a4497/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java#L327-L333

https://community.sonarsource.com/t/error-about-unsupported-ts-version-while-project-is-using-supported-version/15776/7

sandro-h:
Adjust getTypeScriptLocation to pick the top-most
(in terms of directory hierarchy) TypeScript dependency it
can find.
This prevents it from finding a nested transitive
TypeScript dependency that might not match the version
requirements.

The average search performance is a bit worse now, since
it's forced to do a full directory search. Before, it
might've found a match earlier, though worst-case performance
was still more or less a full search.

Fixes #1928

### DIFF
--- a/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
+++ b/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
@@ -28,6 +28,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.time.Duration;
+import java.util.Comparator;
 import java.util.Optional;
 import java.util.stream.Stream;
 import okhttp3.HttpUrl;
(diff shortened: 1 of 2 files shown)


vilchik-elena:
@sandro-h nice fix, but I will not merge it yet. We have some other problems related to TS search, and may be we will go for a more global solution (e.g. packaging TS inside analyzer). 
In case we will decide to do several patches, your PR might be useful.

sandro-h:
👍 works for me! Thanks for the heads-up.

vilchik-elena:
closing PR as we decided to bundle TS dependency (#2010)

PR #1971, opened 2020-03-17
Always look for TypeScript in SonarLint context

### LINKED ISSUE #1820: TypeScript location in SonarLint is not detected correctly in project mixing JS and TS
In SonarLint when project contains both JS and TS files, if the first analyzed file is JS file, we will not attempt to detect TypeScript location, and this detection will not be tried on any other analysis. If user subsequently opens TS file, he will receive the error about missing TypeScript dependency.

TypeScript location is detected on the start of eslint-bridge server, and it is not reattempted after the start. In SonarLint context we should retry detecting TypeScript on every analysis.

jblievremont:
Fixes #1820

### DIFF
--- a/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
+++ b/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
@@ -36,6 +36,7 @@ import okhttp3.OkHttpClient;
 import okhttp3.Request;
 import okhttp3.RequestBody;
 import okhttp3.Response;
+import org.sonar.api.SonarProduct;
 import org.sonar.api.batch.fs.FileSystem;
 import org.sonar.api.batch.sensor.SensorContext;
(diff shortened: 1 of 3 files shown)


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
```diff
@@ -135,7 +136,7 @@ private void initNodeCommand(SensorContext context, File scriptFile) throws IOEx
       .getInt(MAX_OLD_SPACE_SIZE_PROPERTY)
       .ifPresent(nodeCommandBuilder::maxOldSpaceSize);
 
-    if (shouldDetectTypeScript(context.fileSystem())) {
+    if (context.runtime().getProduct() == SonarProduct.SONARLINT || shouldDetectTypeScript(context.fileSystem())) {
```
saberduck: I think we could move the condition to `shouldDetectTypescript` method and change the parameter to be `context`

PR #1984, opened 2020-04-03
Prepare for next development iteration (6.3.0)

vilchik-elena:


sameerinfodb:
SonarJS 6.3.0 is compatible with which version of Sonarqube?

PR #2001, opened 2020-06-17
S3524 Fix FPs for arrow-function-convention: 'async' keyword, generics

### LINKED ISSUE #1999: FP S3524 (arrow-function-convention): 'async' keyword
```
let myFunc = async (x) =>  { console.log(x); };  // FP  
```

andrey-tyukin-sonarsource:
Fixes #1999

### DIFF
--- a/eslint-bridge/src/rules/arrow-function-convention.ts
+++ b/eslint-bridge/src/rules/arrow-function-convention.ts
@@ -68,22 +68,20 @@ function checkParameters(
   if (arrowFunction.params.length !== 1) {
     return;
   }
-  const firstParameter = arrowFunction.params[0];
-  const firstTokenOffset = arrowFunction.async ? 1 : 0;
-  const firstGenericsOrParamlListToken = context
-    .getSourceCode()
-    .getFirstToken(arrowFunction, firstTokenOffset);
-  const lastParamListToken = context.getSourceCode().getTokenAfter(firstParameter);
-  const hasGenerics =
-    firstGenericsOrParamlListToken && firstGenericsOrParamlListToken.value === '<';
+  const parameter = arrowFunction.params[0];
   // Looking at the closing parenthesis after the parameter to avoid problems with cases like
   // `functionTakingCallbacks(x => {...})` where the opening parenthesis before `x` isn't part
   // of the function literal
-  const hasParameterParentheses = lastParamListToken && lastParamListToken.value === ')';
+  const tokenAfterParameter = context.getSourceCode().getTokenAfter(parameter);
+  const hasParameterParentheses = tokenAfterParameter && tokenAfterParameter.value === ')';
 
   if (requireParameterParentheses && !hasParameterParentheses) {
-    context.report({ node: firstParameter, message: MESSAGE_ADD_PARAMETER });
-  } else if (!requireParameterParentheses && !hasGenerics && hasParameterParentheses) {
+    context.report({ node: parameter, message: MESSAGE_ADD_PARAMETER });
+  } else if (
+    !requireParameterParentheses &&
+    !hasGeneric(context, arrowFunction) &&
+    hasParameterParentheses
+  ) {
     const arrowFunctionComments = context.getSourceCode().getCommentsInside(arrowFunction);
     const arrowFunctionBodyComments = context.getSourceCode().getCommentsInside(arrowFunction.body);
     // parameters comments inside parentheses are not available, so use the following subtraction:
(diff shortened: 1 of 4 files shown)


eslint-bridge/tests/rules/arrow-function-convention.test.ts
```diff
@@ -94,6 +94,19 @@ var foo = (a, b) => {   // OK, ignore multiline return
 `,
       options: [{ requireParameterParentheses: true, requireBodyBraces: false }],
     },
+
+    {
+      code: `// valid mandatory parentheses on parameters of async arrow
+var f = async (x) => x
```
vilchik-elena, line 100: I would simply put these cases inside the cases above, we test compliant with each combination of parameters here.

eslint-bridge/tests/rules/arrow-function-convention.test.ts
```diff
@@ -94,6 +94,19 @@ var foo = (a, b) => {   // OK, ignore multiline return
 `,
       options: [{ requireParameterParentheses: true, requireBodyBraces: false }],
     },
+
+    {
+      code: `// valid mandatory parentheses on parameters of async arrow
```
vilchik-elena, line 99: IMHO these comments in front of each new test case don't make things much clearer but just redundant and create some noise which can be avoided easily

eslint-bridge/tests/rules/arrow-function-convention.test.ts
```diff
@@ -94,6 +94,19 @@ var foo = (a, b) => {   // OK, ignore multiline return
...
+var f = async (x) => x
+      `,
+      options: [{ requireParameterParentheses: true, requireBodyBraces: false }],
+    },
+    {
+      code: `// valid non-mandatory parentheses on parameters of async arrow
+var f = async x => x
```
vilchik-elena, line 106: let's add test with multiple parameters and with typed parameter (`async (x: string) => x`)

eslint-bridge/src/rules/arrow-function-convention.ts
```diff
@@ -70,8 +70,10 @@ function checkParameters(
     return;
   }
   const firstParameter = arrowFunction.params[0];
-  const firstToken = context.getSourceCode().getFirstToken(arrowFunction);
-  const hasParameterParentheses = firstToken && firstToken.value === '(';
+  const firstParamListToken = context
+    .getSourceCode()
+    .getFirstToken(arrowFunction, arrowFunction.async ? 1 : 0);
```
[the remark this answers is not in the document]
andrey-tyukin-sonarsource: Indeed, in combination with the name `getFirstToken`, it's pretty confusing.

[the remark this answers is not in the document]
andrey-tyukin-sonarsource: Same for TypeScript: `<T extends {}>(x: T) => x` (which can also appear in combination with `async`). Instead of looking for first token after the `async`-modifier, it's now looking at the last token before the first parameter.

eslint-bridge/tests/rules/arrow-function-convention.test.ts
andrey-tyukin-sonarsource: Doesn't matter once tests are merged.

eslint-bridge/tests/rules/arrow-function-convention.test.ts
andrey-tyukin-sonarsource: Added, but for the case that requires parens: if type annotations are merged, then TS cannot parse it without parens at all.

eslint-bridge/src/rules/arrow-function-convention.ts
```diff
@@ -70,8 +70,9 @@ function checkParameters(
     return;
   }
   const firstParameter = arrowFunction.params[0];
-  const firstToken = context.getSourceCode().getFirstToken(arrowFunction);
-  const hasParameterParentheses = firstToken && firstToken.value === '(';
+  const lastParameter = arrowFunction.params[arrowFunction.params.length - 1];
+  const lastParamListToken = context.getSourceCode().getTokenAfter(lastParameter);
```
vilchik-elena, line 74: let's add comment explaining why we are checking token after parameter and not before (as that seems more intuitive)

eslint-bridge/tests/rules/arrow-function-convention.test.ts
```diff
@@ -34,6 +34,13 @@ var foo = (a /*some comment*/) => { foo(); }
 var foo = (a, b) => { foo(); }
 var foo = () => { foo(); }
 var foo = (a = 1) => { foo(); }
+var foo = async (x) => x
+var foo = <T extends unknown>(x: T) => x
+var foo = async <T extends unknown>(x: T) => x
+var foo = async (x1: number, x2: number = 42) => (x1 + x2)
+var foo = (...xs) => xs
+var foo = async (...xs) => xs
+var foo = async (/* deprecated x, */ y) => y
```
vilchik-elena, line 43: let's also have a test with a comment after the parameter

eslint-bridge/tests/rules/arrow-function-convention.test.ts
```diff
@@ -94,11 +103,16 @@ var foo = (a, b) => {   // OK, ignore multiline return
...
+    {
+      code: `x => x // parameter as the very first token`,
+      options: [{ requireParameterParentheses: false, requireBodyBraces: false }],
+    },
   ],
   invalid: [
     {
       code: `// invalid mandatory parentheses 
 var foo = a => { foo(); }  // Noncompliant
+var foo = async x => x
```
vilchik-elena, line 115: can we have a test with generic?

PR #2005, opened 2020-06-19
Reduce noise of SCA reports

### LINKED ISSUE #2004: Reduce the noise of SCA reporting
The bundled library `commons-compress` suffers from the potential vulnerability described in https://cve.mitre.org/cgi-bin/cvename.cgi?name=2019-12402. The JS plugin is not vulnerable but it generates a false-positive in Software Composition Analysis reports. Upgrading the dependency kills the noise.

simonbrandhof:
Fixes #2004


vilchik-elena:
Thanks @simonbrandhof!

PR #2007, opened 2020-06-24
Fix FP S4158: array overridden inside nested function can be nonempty

### LINKED ISSUE #1974: FP S4158: array written inside function
```js
function foo() {
  let array = [];
  f();
  array.indexOf('x'); // FP

  function f() {
    array = h();
    // array[1] = 42; // note that when the line above is replaced by this one, FP issue disappears
  }
}
```

Also FP disappears if function `f` is declared above `array.indexOf('x')`

andrey-tyukin-sonarsource:
Fixes #1974


andrey-tyukin-sonarsource:
The rule now completely ignores the order of usages.
The boolean state that was both mutated and read during the traversal of usages has been eliminated.
Now it reports if and only if it non-vacuously holds that for all writing usages the write must leave the collection empty, i.e.:

  * There is at least one write that definitely makes the collection empty.
  * There are no writes that might make the collection non-empty.

Note that it is not necessary to check the state of `emptyAssignmentExists` during the iteration, i.e. what ends up in the list of reported issues no longer depends on the order of usages.

eslint-bridge/src/rules/no-empty-collection.ts
```diff
@@ -102,30 +102,42 @@ function reportEmptyCollectionUsage(variable: Scope.Variable, context: Rule.Rule
...
     if (ref.isWriteOnly()) {
-      isEmptyCollection = isReferenceAssigningEmptyCollection(ref);
-    } else {
-      if (isReadCollectionPattern(ref)) {
-        if (isEmptyCollection) {
-          usedReferences.push(ref);
-        }
-      } else {
-        // One references is a write
+      const isEmptyAssignment = isReferenceAssigningEmptyCollection(ref);
```
vilchik-elena, line 116: I would inline this variable

eslint-bridge/src/rules/no-empty-collection.ts
```diff
@@ -102,30 +102,42 @@ function reportEmptyCollectionUsage(variable: Scope.Variable, context: Rule.Rule
...
+      if (!isEmptyAssignment) {
```
vilchik-elena, line 117: remove negation and swap the branches

eslint-bridge/src/rules/no-empty-collection.ts
```diff
@@ -102,30 +102,42 @@ function reportEmptyCollectionUsage(variable: Scope.Variable, context: Rule.Rule
...
+        // At least one usage of the collection is nonempty.
```
vilchik-elena, line 118: -> `// collection is written to a non-empty array`

eslint-bridge/src/rules/no-empty-collection.ts
```diff
@@ -102,30 +102,42 @@ function reportEmptyCollectionUsage(variable: Scope.Variable, context: Rule.Rule
     return;
   }
 
+  if (variable.defs.some(d => d.type === 'Parameter' || d.type === 'ImportBinding')) {
+    // Bound value initialized elsewhere, could be non-empty.
+    return;
+  }
+
+  // All reading usages that are meaningful only if the accessed collection can be nonempty.
   const usedReferences = [];
```
vilchik-elena, line 111: let's rename to `readReferences`(then may be comment can be dropped)

eslint-bridge/src/rules/no-empty-collection.ts
```diff
@@ -102,30 +102,42 @@ function reportEmptyCollectionUsage(variable: Scope.Variable, context: Rule.Rule
...
+      return;
     }
   }
 
-  usedReferences.forEach(ref => {
-    context.report({
-      message: `Review this usage of "${ref.identifier.name}" as it can only be empty here.`,
-      node: ref.identifier,
+  if (emptyAssignmentExists) {
+    // All assignments are empty, and there exists at least one such assignment.
```
vilchik-elena, line 133: IMHO comments are required when logic can't be easily inferred from the code. Here I believe it's better without comment

eslint-bridge/src/rules/no-empty-collection.ts
```diff
@@ -102,30 +102,42 @@ function reportEmptyCollectionUsage(variable: Scope.Variable, context: Rule.Rule
...
+        // At least one usage of the collection is nonempty.
+        // We ignore the order of usages, and therefore consider all reads to be safe.
         return;
+      } else {
+        emptyAssignmentExists = true;
       }
+    } else if (isReadCollectionPattern(ref)) {
+      usedReferences.push(ref);
+    } else {
+      // Some not explicitly handled construct that might make the collection non-empty.
```
vilchik-elena, line 127: same about comment, write it properly or better not write at all (pretty valid option :) try to make code logic and variable names as expressing as possible so that comment is not required)

if you want to keep the comment I would phase it this way: `// some unknown operation on the collection, to not produce FPs we assume it can make collection non-empty`

eslint-bridge/src/rules/no-empty-collection.ts
```diff
@@ -159,6 +171,10 @@ function isEmptyCollectionType(node: estree.Node) {
   return false;
 }
 
+/**
+ * Checks whether this usage would be invalid if the collection turns out
```
vilchik-elena, line 175: better rename method if you believe the name is not expressive enough than adding this comment

eslint-bridge/tests/rules/no-empty-collection.test.ts
```diff
@@ -168,6 +170,52 @@ ruleTester.run('Empty collections should not be accessed or iterated', rule, {
...
+        innerFunction();
+        a.indexOf('x');
+      
+        function innerFunction() {
+          a[0] = 42;
+        }
+      }`,
+    },
+    {
+      code: `function overridingInInnerFunction() {
```
vilchik-elena, line 185: `overwriting` may be better?

eslint-bridge/tests/rules/no-empty-collection.test.ts
```diff
@@ -168,6 +170,52 @@ ruleTester.run('Empty collections should not be accessed or iterated', rule, {
...
+        a.indexOf('x');
+      
+        function innerFunction() {
+          a = unknownFunction();
+        }
+      }`,
+    },
+    {
+      code: `
+      function parametersAreIgnoreUnlessReAsigned(parameterArray: number[]) {
```
vilchik-elena, line 197: I believe parameters are always ignored, so name could be just `parametersAreIgnored`

eslint-bridge/tests/rules/no-empty-collection.test.ts
```diff
@@ -168,6 +170,52 @@ ruleTester.run('Empty collections should not be accessed or iterated', rule, {
...
+        foo(parameterArray[1]);
+        parameterArray = [];
+        foo(parameterArray[1]); // FN, traded for FP in issue-1974
```
vilchik-elena, line 200: `// FN, to avoid FPs (like in issue-1974) we ignore the order of references and so completely ignore parameters`

eslint-bridge/tests/rules/no-empty-collection.test.ts
```diff
@@ -168,6 +170,52 @@ ruleTester.run('Empty collections should not be accessed or iterated', rule, {
...
+    },
+    {
+      code: `
+      function argumentsAreNonempty() {
+        console.log(arguments[0]);
+      }`,
+    },
+    {
+      code: `
+      function minErr() {
```
vilchik-elena, line 211: what are we testing here?

andrey-tyukin-sonarsource: It's a somewhat reduced code snippet that somewhat unexpectedly failed the integration tests. Removed it from unit tests - if the problem reoccurs, it will be seen in the integration tests.

eslint-bridge/tests/rules/no-empty-collection.test.ts
andrey-tyukin-sonarsource: Added something along this lines.

eslint-bridge/tests/rules/no-empty-collection.test.ts
andrey-tyukin-sonarsource: Very true. https://stackoverflow.com/questions/8651562/overwrite-or-override/8651622#comment18314356_8651622

eslint-bridge/src/rules/no-empty-collection.ts
andrey-tyukin-sonarsource: Renamed. Retained the comment anyway, because there is no way to cram the precise formulation with all the modalities into a function name - it would be too long.

eslint-bridge/src/rules/no-empty-collection.ts
andrey-tyukin-sonarsource: In combination with the comment in line 127, this one indeed becomes unnecessary.

eslint-bridge/src/rules/no-empty-collection.ts
andrey-tyukin-sonarsource: I renamed the method into `isReadingCollectionUsage` and this one into `readingUsages` to keep it somewhat consistent.

eslint-bridge/src/rules/no-empty-collection.ts
andrey-tyukin-sonarsource: Fair enough, I guess [early return](https://softwareengineering.stackexchange.com/a/18459/267303) doesn't really apply inside of loops.

eslint-bridge/src/rules/no-empty-collection.ts
vilchik-elena: @andrey-tyukin-sonarsource not handled

eslint-bridge/src/rules/no-empty-collection.ts
```diff
@@ -102,30 +102,40 @@ function reportEmptyCollectionUsage(variable: Scope.Variable, context: Rule.Rule
...
 
-  const usedReferences = [];
-  let isEmptyCollection = false;
+  if (variable.defs.some(d => d.type === 'Parameter' || d.type === 'ImportBinding')) {
+    // Bound value initialized elsewhere, could be non-empty.
+    return;
+  }
+
+  const readingUsages = [];
+  let assignmentOfEmptyCollectionExists = false;
```
vilchik-elena, line 111: imho it's better to start boolean var with `is` or `has`

eslint-bridge/src/rules/no-empty-collection.ts
andrey-tyukin-sonarsource: I assume you meant "collection is overwritten by a non-empty array". I've added a comment that is arguably more precise than what was there previously, I think it should suffice.

eslint-bridge/src/rules/no-empty-collection.ts
```diff
@@ -102,30 +102,42 @@ function reportEmptyCollectionUsage(variable: Scope.Variable, context: Rule.Rule
     return;
   }
 
+  if (variable.defs.some(d => d.type === 'Parameter' || d.type === 'ImportBinding')) {
```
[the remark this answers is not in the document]
andrey-tyukin-sonarsource: Added a test with an import from a non-existent package `nonemptyCollections` (just a hint so it's easier to search)

eslint-bridge/src/rules/no-empty-collection.ts
vilchik-elena: @andrey-tyukin-sonarsource what do you mean? I still see the same comment. And you are right I was wrong, it should be `// something which is not empty array is assigned to the collection`
I would not use "overwrite" here, as it could be first and the only "write", so nothing to "overwrite"

eslint-bridge/src/rules/no-empty-collection.ts
andrey-tyukin-sonarsource: Ok. I think, one can rule out the "is" in this case. The prefix `is` is almost exclusively used for predicates. For example, the predicate "mortal(Socrates)" would be expressed as "Socrates is mortal" in prose, or as `socrates.isMortal()` in code. But this here does not check a predicate, it computes a truth value of a proposition with existential quantifier `exists(collection -> isReferenceAssigningEmptyCollection(collection))`, which would be expressed in prose as "There exists an assignment of an empty collection". There is no "is" in that either. 

Regarding the `has` - I'm not sure, what would be the subject in the "x has y" relation? The code? I guess, if one agrees that one implicitly means the code here, then one could change it to `has`.

eslint-bridge/src/rules/no-empty-collection.ts
andrey-tyukin-sonarsource: Ok, yes. It's doesn't have to be an overrite. It also doesn't necessarily have to be an array. I'd leave it at "there is at least one operation that might make the collection non-empty", if there are no objections.

PR #2012, opened 2020-06-30
Fix FP S4328: add support for path mappings in tsconfig

### LINKED ISSUE #1851: Typescript S4328 does not support aliases
**I want to report a bug.**

**SonarJS version:** 6.1.0.11503<br>

**Rule key:** typescript:S4328

**Reproducer** 

If I use aliases with import of internal modules the S4328 rule is raised for those imports.

**Example**

tsconfig.json
```
{
  "compileOnSave": false,
  "compilerOptions": {
    "baseUrl": "./",
    "paths": {
      "$core/*": ["src/app/core/*"],
...
```
in sometypescript.ts
```
import { ConfigurationService } from '$core/services/configuration.service';
```

**Expected behavior**

It should not raise S4328 rule for that import since it is not an external module.

---

Reading that ROADMAP here: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/ROADMAP.md, it seems that no-implicit-dependencies from tslint has been implemented in the eslint-plugin-import: https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-extraneous-dependencies.md

It seems that aliases are supported in that rule. Does it make sense to use that plugin instead for that kind of rule?

andrey-tyukin-sonarsource:
Fixes #1851 

Docs https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping


eslint-bridge/src/rules/no-implicit-dependencies.ts
```diff
@@ -154,3 +179,44 @@ function findPackageJson(current: string): string | undefined {
...
+  const paths = (compilerOptions && compilerOptions.paths) || [];
+  const pathMappings: PathMapping[] = [];
+  for (const p in paths) {
+    if (p === '*') {
+      // All paths can be mapped, possibly into a directory with completely synthetic code.
+      // Skip the entire check to avoid false positives.
+      return 'skip';
+    } else if (p.endsWith('/*')) {
+      pathMappings.push(new PathMappingPrefix(p.substring(0, p.length - 1)));
+    } else if (p.indexOf('*') < 0) {
```
vilchik-elena, line 214: -> `includes`

eslint-bridge/src/rules/no-implicit-dependencies.ts
```diff
@@ -154,3 +179,44 @@ function findPackageJson(current: string): string | undefined {
...
+      pathMappings.push(new PathMappingExact(p));
+    } else {
+      // Intentionally left blank.
```
vilchik-elena, line 217: so we can have FPs in this case?

eslint-bridge/src/rules/no-implicit-dependencies.ts
```diff
@@ -154,3 +179,44 @@ function findPackageJson(current: string): string | undefined {
   }
   return undefined;
 }
+
+interface PathMapping {
+  isApplicableTo(name: string): boolean;
+}
+
+/** A path mapping that matches exactly one name. */
```
vilchik-elena, line 187: The names of the classes are pretty explicit, no need for these comments

eslint-bridge/tests/fixtures/ts-project-with-path-aliases/main.ts
```diff
@@ -0,0 +1 @@
+import { f } from '$core/services/configuration.service';
```
vilchik-elena, line 1: can we simplify test? I don't think using example from the ticket is reasonable here, we better simplify it

eslint-bridge/tests/rules/no-implicit-dependencies.test.ts
```diff
@@ -153,3 +153,131 @@ ruleTester.run('Dependencies should be explicit', rule, {
     },
   ],
 });
+
+// Refer to
+// https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/parser/README.md#configuration
```
vilchik-elena, line 158: use permalinks

eslint-bridge/tests/rules/no-implicit-dependencies.test.ts
```diff
@@ -153,3 +153,131 @@ ruleTester.run('Dependencies should be explicit', rule, {
...
+// especially for the description of `tsconfigRootDir` + `project` and the `createDefaultProgram`-option.
+const ruleTesterForPathMappings = new RuleTester({
```
vilchik-elena, line 160: On my machine this test is now executed 3-4 times slower... can we do something about that?

eslint-bridge/tests/rules/no-implicit-dependencies.test.ts
```diff
@@ -153,3 +153,131 @@ ruleTester.run('Dependencies should be explicit', rule, {
...
+    createDefaultProgram: true,
+  },
+});
+
+const filenameForFileWithPathMappings = path.join(
+  __dirname,
+  '../fixtures/ts-project-with-path-aliases/file.ts',
+);
+
+ruleTesterForPathMappings.run('Path aliases should be exempt.', rule, {
```
vilchik-elena, line 176: remove dot, we usually not putting dots in such context

eslint-bridge/tests/rules/no-implicit-dependencies.test.ts
```diff
@@ -153,3 +153,131 @@ ruleTester.run('Dependencies should be explicit', rule, {
...
+  },
+});
+
+const filenameWildcardExample = path.join(
+  __dirname,
+  '../fixtures/ts-project-with-wildcard-path-alias/file.ts',
+);
+
+ruleTesterForWildcardExample.run(
+  'Avoid FPs for projects with an unconstrained wildcard mapping',
```
vilchik-elena, line 263: -> `'Do not report when wildcard mapping is used ("*")'`

eslint-bridge/tests/rules/no-implicit-dependencies.test.ts
```diff
@@ -153,3 +153,131 @@ ruleTester.run('Dependencies should be explicit', rule, {
...
+    },
+  ],
+});
+
+const ruleTesterForWildcardExample = new RuleTester({
+  parser: require.resolve('@typescript-eslint/parser'),
+  parserOptions: {
+    ecmaVersion: 2018,
+    sourceType: 'module',
+    tsconfigRootDir: './tests/fixtures/ts-project-with-wildcard-path-alias',
```
vilchik-elena, line 251: let's put all projects used to test this rule under `no-implicit-dependencies` directory (there are 3 old projects and 2 new you've added)

eslint-bridge/tests/fixtures/ts-project-with-path-aliases/main.ts
vilchik-elena, line 1: do we need any ts files at all? all the rule needs is `tsconfig` as far as I can tell

eslint-bridge/tests/fixtures/ts-project-with-wildcard-path-alias/tsconfig.json
```diff
@@ -0,0 +1,17 @@
+{
+  "compilerOptions": { 
+    "baseUrl": "./",
+    "paths": {
+      "is/this/*/even/valid": ["coverage/only"],
```
vilchik-elena, line 5: can we check if such syntax is supported?

eslint-bridge/tests/fixtures/ts-project-with-wildcard-path-alias/tsconfig.json
```diff
@@ -0,0 +1,17 @@
...
+      "*": ["*", "generated/*"]
+    },
+    "rootDir": "./",
```
vilchik-elena, line 8: drop all compiler options not relevant for the test

eslint-bridge/tests/rules/no-implicit-dependencies.test.ts
```diff
@@ -153,3 +153,131 @@ ruleTester.run('Dependencies should be explicit', rule, {
     },
   ],
 });
+
+// Refer to
```
vilchik-elena, line 157: instead of referencing the doc I would shortly explain what additional parser options stand for. And tbh `tsconfigRootDir` and `project` are pretty obvious, so only `createDefaultProgram` requires comment (and btw from docs I didn't get what it does, only experimenting with removing it explained me what it does)

eslint-bridge/tests/rules/no-implicit-dependencies.test.ts
```diff
@@ -153,3 +153,131 @@ ruleTester.run('Dependencies should be explicit', rule, {
...
+
+const filenameForFileWithPathMappings = path.join(
+  __dirname,
+  '../fixtures/ts-project-with-path-aliases/file.ts',
+);
+
+ruleTesterForPathMappings.run('Path aliases should be exempt.', rule, {
+  valid: [
+    {
+      code: `import { f } from '$core/services/configuration.service';`,
```
vilchik-elena, line 179: using same word (`core`) makes it hard to follow what the point of each test. How about `$dollarPrefix`, `atPrefix`, `noSymbolPrefix`?

eslint-bridge/tests/rules/no-implicit-dependencies.test.ts
```diff
@@ -153,3 +153,131 @@ ruleTester.run('Dependencies should be explicit', rule, {
...
+    {
+      code: `import { f } from '@core/services/configuration.service';`,
+      filename: filenameForFileWithPathMappings,
+    },
+    {
+      code: `import { f } from 'core/services/configuration.service';`,
+      filename: filenameForFileWithPathMappings,
+    },
+    {
+      code: `import { f } from 'foo/bar/services/configuration.service';`,
```
vilchik-elena, line 191: `foo/bar` -> `nested/directory`

eslint-bridge/tests/rules/no-implicit-dependencies.test.ts
```diff
@@ -153,3 +153,131 @@ ruleTester.run('Dependencies should be explicit', rule, {
...
+      options,
+      errors: 1,
+    },
+    {
+      code: `let f = require("this/doesnt/exist").f;`,
+      filename: filenameForFileWithPathMappings,
+      options,
+      errors: 1,
+    },
+  ],
```
vilchik-elena, line 243: let's also test the mix of tsconfig/package.json data? dependency is not matching any `path` and [not] found in `package.json`

eslint-bridge/tests/rules/no-implicit-dependencies.test.ts
```diff
@@ -153,3 +153,131 @@ ruleTester.run('Dependencies should be explicit', rule, {
...
+      filename: filenameForFileWithPathMappings,
+    },
+    {
+      code: `import { f } from 'concrete';`,
+      filename: filenameForFileWithPathMappings,
+    },
+    {
+      code: `let f = require("foo/bar/services/configuration.service").f;`,
+      filename: filenameForFileWithPathMappings,
+      options,
```
vilchik-elena, line 201: do we need `options`? it seems to work without them

eslint-bridge/src/rules/no-implicit-dependencies.ts
```diff
@@ -154,3 +179,44 @@ function findPackageJson(current: string): string | undefined {
...
+
+function extractPathMappings(parserServices: RequiredParserServices): PathMapping[] | 'skip' {
+  const compilerOptions = parserServices.program && parserServices.program.getCompilerOptions();
+  const paths = (compilerOptions && compilerOptions.paths) || [];
+  const pathMappings: PathMapping[] = [];
+  for (const p in paths) {
+    if (p === '*') {
+      // All paths can be mapped, possibly into a directory with completely synthetic code.
+      // Skip the entire check to avoid false positives.
+      return 'skip';
```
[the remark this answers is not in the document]
andrey-tyukin-sonarsource: Not sure how to interpret that. It's exactly because `*` behaves quite differently (essentially deactivating all the other checks), that there is a separate case for that. One could replace it by something like an empty prefix pattern, then it would match all paths, but it would still iterate over all the other patterns that came before it, instead of exiting early.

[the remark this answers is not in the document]
vilchik-elena: @andrey-tyukin-sonarsource I'm talking only about using different literal here

eslint-bridge/src/rules/no-implicit-dependencies.ts
andrey-tyukin-sonarsource: I've modified it so it can cope with `prefix/*/suffix`-patterns (tsc indeed does this), and checked that `tsc` emits error if there is more than one asterisk. So, the remaining case should now never occur.

eslint-bridge/src/rules/no-implicit-dependencies.ts
andrey-tyukin-sonarsource: I've made them even more explicit, and deleted the comments.

eslint-bridge/tests/fixtures/ts-project-with-path-aliases/main.ts
andrey-tyukin-sonarsource: Replaced everything by `a/b/c/d.e` (kept something with a `.` in the middle of directory name, though)

eslint-bridge/tests/rules/no-implicit-dependencies.test.ts
andrey-tyukin-sonarsource: I've eliminated the `createDefaultProgram` entirely, apparently, it's deprecated, and could also have negative impact on test performance.

eslint-bridge/tests/rules/no-implicit-dependencies.test.ts
andrey-tyukin-sonarsource: Since now it's all just `a/b/c`, there is no need to duplicate the `$` as `dollarPrefix` or `@` as `atPrefix` - those are the very first characters in 5-8 character long strings, really difficult to miss.

eslint-bridge/tests/rules/no-implicit-dependencies.test.ts
andrey-tyukin-sonarsource: Well... The directories were already nested previously. I just needed some names that are different from `a`, `b`, `c`, and `foo` and `bar` seemed as meaningful as anything else.

eslint-bridge/tests/rules/no-implicit-dependencies.test.ts
andrey-tyukin-sonarsource: Added `dependency-in-package-json`.

eslint-bridge/tests/rules/no-implicit-dependencies.test.ts
andrey-tyukin-sonarsource: indeed, works without `options`. Not sure what they are for in all the previously existing test cases.

eslint-bridge/tests/rules/no-implicit-dependencies.test.ts
andrey-tyukin-sonarsource: Weird `createDefaultApp`-option is eliminated, this line is no longer needed.

eslint-bridge/tests/rules/no-implicit-dependencies.test.ts
andrey-tyukin-sonarsource: Accelerated it by factor 3 (roughly), now it takes around 3.3 seconds on my machine. Previously, it was around 1.9. Seems like an acceptable cost for having three times as many different `tsc`-configurations.

eslint-bridge/tests/fixtures/ts-project-with-path-aliases/main.ts
andrey-tyukin-sonarsource: Rule - yes; But this way, one can actually go into the directory and try out what an actual `tsc` invocation is doing. Since it seems to be mostly implementation-defined - i.e. there is no external specification that tells what it should be doing - the cheapest way for getting some feeling for what it's actually doing is to set up a little project and try it out, and that's exactly what has been done here. I've removed the files from `ts-project-with-catch-all-path-alias`, though, since they indeed weren't needed in order to keep the project valid.

eslint-bridge/tests/fixtures/ts-project-with-wildcard-path-alias/tsconfig.json
andrey-tyukin-sonarsource: Strangely enough, it really is supported. Changed implementation accordingly (now the patterns can also have a suffix).

ct24p:
@vilchik-elena: could you please let us know from which version of sonarqube, S4328 mapped path issues for typescript is fixed? We are using Sonarqube 7.9.5 and still having this issue, and how to fix for it without update sonarqube? thanks

vilchik-elena:
@ct24p There is no way to do that without updating sonarqube. Anyway new LTS version 8.9 is released and 7.9.x is not supported anymore, you should upgrade. 
Issue should be fixed in SQ 8.9

alkemist:
@vilchik-elena  We are using Sonarqube 8.9 and still having this issue :/

vilchik-elena:
@alkemist then welcome to https://community.sonarsource.com/ to properly report the problem you are experiencing :)

PR #2014, opened 2020-06-30
Format/bump prettier version 2x

andrey-tyukin-sonarsource:
Bumps `prettier` version to latest version 2.0.5 (older 1.14.2 version was no longer capable of parsing new syntactical constructs from TypeScript >=3.7 like optional chaining).

Pins as many options as possible in `package.json` to avoid future changes in case that the convention is changed in newer versions of prettier. The options are chosen such that the formatting changes are as small as possible, i.e. are mostly consistent with what we've already been using in the project anyway. In particular:

  * We do not enforce `lf` line endings (new default in 2.0.5), we kept `auto`-setting; Enforcing `lf` everywhere would require big changes in 16 more files.
  * The `arrowParens` setting is set to `avoid` (`always` is the new default).

Some lengthy string literals have been fixed manually (breaking in the middle of `${ ... }` is now avoided).


eslint-bridge/package.json
```diff
@@ -71,7 +71,9 @@
   "prettier": {
     "printWidth": 100,
     "trailingComma": "all",
-    "singleQuote": true
+    "singleQuote": true,
+    "arrowParens": "avoid",
+    "endOfLine": "auto"
```
vilchik-elena, line 76: I would go for default here (lf). @saberduck wdyt?

andrey-tyukin-sonarsource: Changed the option to `lf`. It modified >2k lines of code that I'd otherwise not have touched, but it didn't seem to break anything.

PR #2021, opened 2020-07-02
Use babel-eslint when parsing VueJS

### LINKED ISSUE #1978: Support 'import()' in VueJS files
`import()` syntax is a proposal at stage 4. It's supported by babel eslint parser, but not by eslint default parser. As for VueJS we use https://www.npmjs.com/package/vue-eslint-parser without specifying underlying parser, default eslint parser is used. We should specify `babel-eslint` in the parser options.
```
"parserOptions": {
    "parser": "babel-eslint",
    ...
}
```

https://github.com/tc39/proposal-dynamic-import

vilchik-elena:
Fixes #1978

### DIFF
--- a/eslint-bridge/src/parser.ts
+++ b/eslint-bridge/src/parser.ts
@@ -142,7 +142,9 @@ export function unloadTypeScriptEslint() {
 
 export function parseVueSourceFile(fileContent: string): SourceCode | ParsingError {
   let exceptionToReport: ParseException | null = null;
-  for (const config of [PARSER_CONFIG_MODULE, PARSER_CONFIG_SCRIPT]) {
+  // setting parser to be able to parse more code (by default `espree` is used by vue parser)
+  const vueModuleConfig = { ...PARSER_CONFIG_MODULE, parser: 'babel-eslint' };
+  for (const config of [vueModuleConfig, PARSER_CONFIG_SCRIPT]) {
     try {
       const result = VueJS.parseForESLint(fileContent, config);
       return new SourceCode(fileContent, result.ast as any);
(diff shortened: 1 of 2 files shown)


kakaru1331:
Hello! First I really appreciate your works. Thank you.
BTW can you tell me when will this be applied to sonar-scanner?

IGx89:
Looks like this was released over a month ago in v6.3.0, but the Marketplace hasn't been updated with that version yet (latest there is still 6.2.1, while latest here is 6.4.1). You could manually install it -- I'm running into this same issue and will have to do that myself if the Marketplace isn't updated soon.

kakaru1331:
@IGx89 Oh you saved my life. I'll try it!
thank you :D

0m4r:
> Looks like this was released over a month ago in v6.3.0, but the Marketplace hasn't been updated with that version yet (latest there is still 6.2.1, while latest here is 6.4.1). You could manually install it -- I'm running into this same issue and will have to do that myself if the Marketplace isn't updated soon.

how can I install manually the version that solves this issue?
I'd like to get this running with the sonarLint extention for VS Code

IGx89:
Can't answer about SonarLint, don't use it myself. However, my guess is that for whatever reason they're holding all these new versions back for the next version of SonarQube, 8.5. Last I heard they were targeting to release that end of Sept. and the online docs already say 8.5, so it should be releasing shortly and then hopefully this whole confusing situation will be in the past...

IGx89:
I can confirm SQ 8.5 contains the latest release of SonarJS (6.5)! Download links on sonarqube.org are still for 8.4, but I found a direct link elsewhere: https://binaries.sonarsource.com/Distribution/sonarqube/sonarqube-8.5.0.37579.zip

PR #2022, opened 2020-07-03
Fix S905 FP with optional chaining

### LINKED ISSUE #1930: S905: FP with optional chaining
Reproducer:

```ts
x?.somefunction("abc")
``` 

For S905, we don't have our own implementation, we are relying on "no-unused-expressions". The version of typescript-eslint we are currently using (we updated versions in #1849) already include the fix for this FP, if we use the implementation from typescript-eslint.
However, we are in fact [using a chai friendly](https://github.com/SonarSource/SonarJS/blob/fdc6425a982d78ef9dd62feccdad884ea52a4497/eslint-bridge/src/linter.ts#L58) implementation, who is based on eslint (therefore not supporting correctly TS...).

All in all, we are interested in the improvement of both implementations, but we are not supporting correctly optional chaining in the current situation.

andrey-tyukin-sonarsource:
Fixes #1930

### DIFF
--- a/eslint-bridge/src/linter.ts
+++ b/eslint-bridge/src/linter.ts
@@ -58,7 +58,6 @@ try {
   console.log(typescriptEslintRules);
   linter.defineRules(typescriptEslintRules);
 
-  // Decorate the no-unused-expressions rule with additional chai-specific conditions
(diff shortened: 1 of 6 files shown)


vilchik-elena:
@andrey-tyukin-sonarsource Please give a proper title to this PR

PR #2028, opened 2020-07-08
Merge JavaScript and TypeScript settings in UI

saberduck:
Fixes #1938

PR #2029, opened 2020-07-08
Improve UX around language-specific exclusion property

### LINKED ISSUE #2023: Improve UX around language-specific exclusion property
* move the parameter from setting UI to the troubleshooting section of docs
* Document that the parameters should only need to be changed when the user wants to analyze code in node_modules or bower_components
* Maybe have a warning in the documentation to be careful because this exclusion applies across all project files?
* Make the debug log message a bit more explicit so that any problems can be tracked down more easily. Maybe something like:
```
13:18:14.085 DEBUG: 'src/cobol/Custmnt2.cbl' excluded by sonar.typescript.exclusions filter
```

saberduck:
Fixes #2023 

needs #2028 to be merged before

### DIFF
--- a/sonar-javascript-plugin/src/main/resources/static/documentation.md
+++ b/sonar-javascript-plugin/src/main/resources/static/documentation.md
@@ -55,7 +55,7 @@ You can use `sonar.javascript.node.maxspace` property to allow the analyzer to u
 ### Default exclusions
 
 By default, analyzer will exclude files from dependencies in `node_modules` and `bower_components`. If for some reason analysis of files in these directories
-is desired, it can be configured by setting `sonar.javascript.exclusions` property. 
+is desired, it can be configured by setting `sonar.javascript.exclusions` property to empty value, i.e. `sonar.javascript.exclusions=""`, or to comma separated list of paths to be excluded. This property will exclude the files also for other plugins, similar to `sonar.exclusions` property, however `sonar.exclusions` property should be preferred to configure general exclusions for the project.
 
 <!-- sonarqube -->
 ## Custom rules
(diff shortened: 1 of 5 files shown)


sonar-javascript-plugin/src/main/resources/static/documentation.md
```diff
@@ -52,6 +52,11 @@ java.io.InterruptedIOException: timeout
 You can use `sonar.javascript.node.maxspace` property to allow the analyzer to use more memory. Set this property to `4096` or `8192` for big projects. This property should be set in `sonar-project.properties` file or on command line for scanner (with `-Dsonar.javascript.node.maxspace=4096`).
 
 
+### Default exclusions
+
+By default, analyzer will exclude files from dependencies in `node_modules` and `bower_components`. If for some reason analysis of files in these directories
+is desired, it can be configured by setting `sonar.javascript.exclusions` property.
```
vilchik-elena, line 58: setting to what?

saberduck: I improved the wording here, please have a look again.

sonar-javascript-plugin/src/main/resources/static/documentation.md
```diff
@@ -52,6 +52,11 @@ java.io.InterruptedIOException: timeout
 You can use `sonar.javascript.node.maxspace` property to allow the analyzer to use more memory. Set this property to `4096` or `8192` for big projects. This property should be set in `sonar-project.properties` file or on command line for scanner (with `-Dsonar.javascript.node.maxspace=4096`).
 
 
+### Default exclusions
+
+By default, analyzer will exclude files from dependencies in `node_modules` and `bower_components`. If for some reason analysis of files in these directories
+is desired, it can be configured by setting `sonar.javascript.exclusions` property to empty value, i.e. `sonar.javascript.exclusions=""`, or to comma separated list of paths to be excluded. This property will exclude the files also for other plugins, similar to `sonar.exclusions` property, however `sonar.exclusions` property should be preferred to configure general exclusions for the project.
```
vilchik-elena, line 58: `for other plugins` -> `for other languages` may be better?

PR #2045, opened 2020-07-29
Initialize Linter only once

### LINKED ISSUE #2041: Linter should be initialized lazily
Currently we call `linter.defineRule` and `linter.getRules()` functions everytime we analyze JS or TS file. These functions are costly, we should initialize linter once, and then reuse `linterConfig` for every subsequent invocation of `analyze` function.

saberduck:
Fixes #2041

### DIFF
--- a/eslint-bridge/src/analyzer.ts
+++ b/eslint-bridge/src/analyzer.ts
@@ -62,7 +62,6 @@ export const COGNITIVE_COMPLEXITY_RULE: AdditionalRule = {
 export interface AnalysisInput {
   filePath: string;
   fileContent: string | undefined;
-  rules?: Rule[];
   ignoreHeaderComments?: boolean;
   tsConfigs?: string[];
 }
@@ -142,8 +141,8 @@ function analyze(input: AnalysisInput, parse: Parse): AnalysisResponse {
   if (!fileContent) {
     fileContent = getFileContent(input.filePath);
   }
-  if (input.rules) {
-    initLinter(input.rules);
+  if (!linter) {
+    throw new Error('Linter is undefined. Did you call /init-linter?');
   }
   const result = parse(fileContent, input.filePath, input.tsConfigs);
   if (result instanceof SourceCode) {
--- a/eslint-bridge/src/analyzer.ts
+++ b/eslint-bridge/src/analyzer.ts
@@ -27,7 +27,6 @@ import {
 import getHighlighting, { Highlight } from './runner/highlighter';
 import getMetrics, { EMPTY_METRICS, Metrics } from './runner/metrics';
 import getCpdTokens, { CpdToken } from './runner/cpd';
-import * as linter from './linter';
 import { SourceCode } from 'eslint';
 import {
   HighlightedSymbol,
@@ -36,6 +35,7 @@ import {
 } from './runner/symbol-highlighter';
 import * as fs from 'fs';
 import { rules as sonarjsRules } from 'eslint-plugin-sonarjs';
+import { LinterWrapper, AdditionalRule } from './linter';
 
 const COGNITIVE_COMPLEXITY_RULE_ID = 'internal-cognitive-complexity';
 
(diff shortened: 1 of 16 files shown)


eslint-bridge/src/analyzer.ts
```diff
@@ -131,11 +131,20 @@ function stripBom(s: string) {
...
+export function initLinter(rules: Rule[]) {
+  linter = new LinterWrapper(rules, SYMBOL_HIGHLIGHTING_RULE, COGNITIVE_COMPLEXITY_RULE);
+}
+
 function analyze(input: AnalysisInput, parse: Parse): AnalysisResponse {
   let fileContent = input.fileContent;
   if (!fileContent) {
     fileContent = getFileContent(input.filePath);
   }
+  if (input.rules) {
```
vilchik-elena, line 145: let's drop this logic

eslint-bridge/src/analyzer.ts
```diff
@@ -131,11 +131,20 @@ function stripBom(s: string) {
...
 }
 
+let linter: LinterWrapper;
+
+export function initLinter(rules: Rule[]) {
+  linter = new LinterWrapper(rules, SYMBOL_HIGHLIGHTING_RULE, COGNITIVE_COMPLEXITY_RULE);
+}
+
 function analyze(input: AnalysisInput, parse: Parse): AnalysisResponse {
   let fileContent = input.fileContent;
```
vilchik-elena, line 141: how about throwing when `linter` is `undefined`?

eslint-bridge/tests/analyzer.test.ts
```diff
@@ -73,6 +74,20 @@ describe('#analyzeJavaScript', () => {
     expect(issues).toContainEqual(noDuplicateStringIssue);
   });
 
+  it('should report issue using initLinter', () => {
```
vilchik-elena, line 77: we can probably drop these tests when others are updated to not provide `rules`

eslint-bridge/tests/server.test.ts
```diff
@@ -189,11 +189,33 @@ describe('server', () => {
...
+    expect(JSON.parse(response)).toEqual(expectedResponse);
+  });
+
+  it('should respond to JavaScript analysis request using init', async () => {
+    expect.assertions(2);
+    expect(server.listening).toEqual(true);
+
+    await post(
+      JSON.stringify([{ key: 'no-all-duplicated-branches', configurations: [] }]),
+      '/init',
```
vilchik-elena, line 213: should not it be `init-linter`? how is it working with the wrong url?

eslint-bridge/tests/server.test.ts
```diff
@@ -208,11 +230,38 @@ describe('server', () => {
...
+  it('should respond to TypeScript analysis request using init', async () => {
+    const filePath = join(__dirname, './fixtures/ts-project/sample.lint.ts');
+    const tsConfig = join(__dirname, './fixtures/ts-project/tsconfig.json');
+
+    expect.assertions(2);
+    expect(server.listening).toEqual(true);
+
+    await post(
+      JSON.stringify([{ key: 'no-all-duplicated-branches', configurations: [] }]),
+      '/init',
```
vilchik-elena, line 259: same here

sonar-javascript-plugin/pom.xml
```diff
@@ -143,7 +143,7 @@
             <configuration>
               <rules>
                 <requireFilesSize>
-                  <minsize>10000000</minsize>
+                  <minsize>9000000</minsize>
```
vilchik-elena, line 146: do you know why side decreased?

saberduck: looks like eslint is smaller

eslint-bridge/tests/server.test.ts
saberduck: it works because tests are executed in order and it's sufficient for one of them to init the linter. However these tests were redundant, because they assumed that both with init and without init protocol should work. I removed them

sonar-javascript-plugin/pom.xml
vilchik-elena: but eslint is the same in this branch

PR #2038, opened 2020-07-16
Upgrade ESLint to 7.5.0 and typescript-eslint to 3.7.1

saberduck:
Bump of dependencies
* eslint to 7.5.0
* typescript-eslint to 3.7.1
  * this change makes us to support different range of TypeScript: >=3.3.1 <4.0.0 (technically we test against 3.10.0 to exclude 4.0.0 see https://github.com/typescript-eslint/typescript-eslint/blob/5eedbff01178ea33b98ab22e556df4c1a195f839/packages/typescript-estree/src/parser.ts#L23 )

Changes in ruling are related to improvements in eslint
* https://github.com/eslint/eslint/pull/12279 for S1656
* https://github.com/eslint/eslint/pull/11509 for S2814
* https://github.com/eslint/eslint/pull/12947 for S121

PR #2046, opened 2020-07-29
linter init eslint7

saberduck:


saberduck:
this was just a test. work is done in #2038

PR #2049, opened 2020-07-31
Normalize path from bundle to fix theoretical vulnerability

saberduck:
Vulnerability is theoretical, because we control the bundle which is being extracted.


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/BundleImpl.java
```diff
@@ -109,7 +109,7 @@ private static void extractFromClasspath(InputStream resource, Path targetPath)
   }
 
   private static Path entryPath(Path targetPath, ArchiveEntry entry) {
-    Path entryPath = targetPath.resolve(entry.getName());
+    Path entryPath = targetPath.resolve(entry.getName()).normalize();
```
vilchik-elena, line 112: tbh I don't get how normalization can make code safer here

PR #2053, opened 2020-08-05
Improve S5122: raise only for permissive access (*) and report on Express.js API

### LINKED ISSUE #1837: Improve S5122: raise only for permissive access (*)
### Raise only for permissive access (*)
[S5122](https://jira.sonarsource.com/browse/RSPEC-5122) should be reworked to raise only when permissive CORS policy is obvious (`'Access-Control-Allow-Origin': '*'`)

Following code should not raise an issue
```
const srv2 = http.createServer((req, res) => {
	res.setHeader('Access-Control-Allow-Origin', 'http://mytrustedorigin'); // FP
```

### Add support for Express.js API

Code examples: https://github.com/SonarSource/security-expected-issues/tree/master/javascript/rules/hotspots/RSPEC-5122

For the `cors` middleware raise an issue when the `origin` property of the options is set to `*`:
```
let corsOptions = {
  origin: '*' // Sensitive (main location)
};

let app = express();
app.use(cors(corsOptions));  // Sensitive (secondary location on the cors() call)
```
Without options passed in the call (ie: `cors()`) the code is sensitive (equals to `origin` set to `*`)

Related issue:
https://github.com/SonarSource/SonarJS/issues/1145

### LINKED ISSUE #2058: Improve S5122: Support cors middleware
### Add support for Express.js cors middleware 

https://expressjs.com/en/resources/middleware/cors.html

Code examples: https://github.com/SonarSource/security-expected-issues/tree/master/javascript/rules/hotspots/RSPEC-5122

For the `cors` middleware raise an issue when the `origin` property of the options is set to `*`:
```
let corsOptions = {
  origin: '*' // Sensitive (main location)
};

let app = express();
app.use(cors(corsOptions));  // Sensitive (secondary location on the cors() call)
```
Without options passed in the call (ie: `cors()`) the code is sensitive (equals to `origin` set to `*`)

Related issue:
https://github.com/SonarSource/SonarJS/issues/1145

saberduck:
Fixes #1837 
Fixes #2058


eslint-bridge/src/rules/cors.ts
```diff
@@ -64,14 +96,48 @@ export const rule: Rule.RuleModule = {
...
+
+function isCorsHeader(node: estree.Node) {
+  return isLiteral(node) && node.value === CORS_HEADER;
+}
+
+function isAnyDomain(node: estree.Node) {
+  return isLiteral(node) && node.value === '*';
+}
+
+function isSensitiveCorsOptions(node: estree.Node | undefined | null): estree.Property | undefined {
```
vilchik-elena, line 132: rename method to not start it with `is` (`getSensitiveCorsProperty`)

eslint-bridge/src/rules/cors.ts
```diff
@@ -21,35 +21,67 @@
...
 
-      Literal(node: estree.Node) {
-        const { value } = node as estree.Literal;
-        if (String(value).includes(CORS_HEADER_PREFIX)) {
-          context.report({ message: MESSAGE, node });
+      VariableDeclarator(node: estree.Node) {
+        const decl = node as estree.VariableDeclarator;
+        const corsOptions = isSensitiveCorsOptions(decl.init);
+        if (decl.id.type === 'Identifier' && corsOptions) {
+          sensitiveCorsOptions.set(decl.id.name, corsOptions);
```
vilchik-elena, line 84: I think my approach is more precise as it's based on variable object, not its name only

eslint-bridge/src/rules/cors.ts
```diff
@@ -64,14 +96,48 @@ export const rule: Rule.RuleModule = {
...
-            context.report({
-              message: MESSAGE,
-              node,
-            });
+          if (moduleName?.value === 'cors') {
+            if (call.arguments.length === 0 || isSensitiveCorsOptions(call.arguments[0])) {
+              report(node);
+            }
+            if (call.arguments[0]?.type === 'Identifier') {
+              corsCallArguments.push(call.arguments[0]);
```
vilchik-elena, line 104: can't we directly check here if this variable set to sensitive options instead of doing it at the end in 'Program:exit'?

eslint-bridge/tests/rules/cors.test.ts
```diff
@@ -18,9 +18,41 @@
...
 import { RuleTester } from 'eslint';
+import { rule } from '../../src/rules/cors';
+import { RuleTesterTs } from '../RuleTesterTs';
 
 const ruleTester = new RuleTester({ parserOptions: { ecmaVersion: 2018, sourceType: 'module' } });
-import { rule } from '../../src/rules/cors';
+
+const ruleTesterTs = new RuleTesterTs();
+
+ruleTesterTs.run('CORS', rule, {
```
vilchik-elena, line 28: I don't get why we need this test with types, is it some leftover after debug?

eslint-bridge/tests/rules/cors.test.ts
```diff
@@ -33,17 +65,44 @@ ruleTester.run('Enabling Cross-Origin Resource Sharing is security-sensitive', r
     {
       code: `res.writeHead(200, { 'Content-Type': 'text/html' });`,
     },
+    {
+      code: `
+      const srv2 = http.createServer((req, res) => {
+        res.setHeader('Access-Control-Allow-Origin', 'http://mytrustedorigin'); // FP
```
vilchik-elena, line 71: remove `FP`

eslint-bridge/tests/rules/cors.test.ts
```diff
@@ -33,17 +65,44 @@ ruleTester.run('Enabling Cross-Origin Resource Sharing is security-sensitive', r
...
+      });
+      `,
+    },
+    {
+      code: `
+        const express = require('express');
+        const cors = require('cors');
```
vilchik-elena, line 78: unused, can we drop them?

eslint-bridge/src/rules/cors.ts
```diff
@@ -64,14 +96,48 @@ export const rule: Rule.RuleModule = {
 
         if (usingExpressFramework && callee.type === 'Identifier') {
```
vilchik-elena, line 97: tbh I don't understand why we need to check that `express` is used here. Does not `cors` guarantee that?

eslint-bridge/tests/rules/cors.test.ts
```diff
@@ -18,9 +18,41 @@
...
+                    }
+
+                    return waitForXHR(xhr);
+                }`,
+      errors: 1,
+    },
+  ],
+});
+
+const EXPECTED_MESSAGE = JSON.stringify({
```
vilchik-elena, line 52: how about renaming `EXPECTED_MESSAGE_NO_SECONDARY`?

eslint-bridge/tests/rules/cors.test.ts
```diff
@@ -54,35 +113,110 @@ ruleTester.run('Enabling Cross-Origin Resource Sharing is security-sensitive', r
...
+    },
+    {
+      code: `
+        res.header('Access-Control-Allow-Origin', '*' ); // Sensitive
+    `,
+      errors: 1,
+    },
+    {
+      code: `
+        res.set('Access-Control-Allow-Origin', '*' ); // Sensitive
```
vilchik-elena, line 145: in fact any callee will work, right? how about testing `foo('Access-Control-Allow-Origin', '*' );`

eslint-bridge/src/rules/cors.ts
```diff
@@ -64,14 +96,48 @@ export const rule: Rule.RuleModule = {
...
+              report(node);
+            }
+            if (call.arguments[0]?.type === 'Identifier') {
+              corsCallArguments.push(call.arguments[0]);
+            }
           }
         }
+
+        if (isSettingCorsHeader(call)) {
+          report(call.callee);
```
vilchik-elena, line 110: I would prefer to report on arguments or entire call here (and btw location is not tested in this case)

eslint-bridge/src/rules/cors.ts
saberduck: how do you guarantee that order of visit is the same as order of execution?

saberduck: reworked using `getUniqueWriteUsage`

eslint-bridge/tests/rules/cors.test.ts
saberduck: removed

eslint-bridge/src/rules/cors.ts
saberduck: I removed the check for express

eslint-bridge/tests/rules/cors.test.ts
saberduck: sure, but why not testing with real API?

eslint-bridge/src/rules/cors.ts
saberduck: ok, also added test for location

eslint-bridge/tests/rules/cors.test.ts
vilchik-elena: I think having just such tests will make somebody think that we are checking API while we don't

eslint-bridge/src/rules/cors.ts
```diff
@@ -125,15 +97,32 @@ function isCorsHeader(node: estree.Node) {
   return isLiteral(node) && node.value === CORS_HEADER;
 }
 
-function isAnyDomain(node: estree.Node) {
+function isAnyDomain(node: estree.Node | undefined) {
   return isLiteral(node) && node.value === '*';
 }
 
-function isSensitiveCorsOptions(node: estree.Node | undefined | null): estree.Property | undefined {
+function getObjectExpressionProperty(
```
vilchik-elena, line 104: I had similar pattern, I did this https://github.com/SonarSource/SonarJS/pull/2059/files#diff-e3d2168698afd38bfed3dd8a3c64992eR152-R169
That way you don't need to duplicate the check for `isAnyPattern` after calling this function

saberduck: I prefer to keep this method to have just this single purpose, not adding the test to it

eslint-bridge/src/rules/cors.ts
```diff
@@ -45,77 +43,51 @@ export const rule: Rule.RuleModule = {
...
-              report(node);
-            }
-            if (call.arguments[0]?.type === 'Identifier') {
-              corsCallArguments.push(call.arguments[0]);
+        if (isCorsCall(callee)) {
+          if (call.arguments.length === 0) {
+            report(call);
+            return;
+          }
+          const [arg] = call.arguments;
```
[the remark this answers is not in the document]
saberduck: I tried to rework it this way, but it get's tricky - reporting is not the same in the two cases, because with identifier I report secondary location for the call. I think avoiding duplication is not worth it in this case.

eslint-bridge/src/rules/cors.ts
```diff
@@ -21,57 +21,112 @@
...
 
-      Literal(node: estree.Node) {
-        const { value } = node as estree.Literal;
-        if (String(value).includes(CORS_HEADER_PREFIX)) {
-          context.report({ message: MESSAGE, node });
-        }
-      },
+    function isCorsCall(callee: estree.Node) {
+      return (
+        callee.type === 'Identifier' && getModuleNameOfIdentifier(callee, context)?.value === 'cors'
```
andrey-tyukin-sonarsource, line 53: This doesn't seem to allow inlining `require`s:
```
{
      code: `
      const express = require('express');
      // const cors = require('cors');
      app.use(require('cors')({
        origin: '*', // Sensitive
        optionsSuccessStatus: 200
      }));
      `,
      errors: 1,
    },
```

vilchik-elena: there are many things we don't support, the point is to support the most common cases. Yours seems to me pretty fantastic.

andrey-tyukin-sonarsource: It's not as fantastic as it might seem on first glance. The `app.use(require('some-middleware')())` is not uncommon.
Here is an [example from the main file of the standard vulnerable node application DVNA](https://github.com/appsecco/dvna/blob/c637437d6515bd4c732e91c58e62d38e88260d3c/server.js#L35): 

```
// Initialize express-flash
app.use(require('express-flash')());
```

I'm also reasonably sure that I've seen it more than once.

PR #2059, opened 2020-08-07
Rule S2598: support 'multer'

### LINKED ISSUE #1814: Rule S2598 File uploads should be restricted
Implements [RSPEC-2598](https://jira.sonarsource.com/browse/RSPEC-2598)

For both JS/TS

Code examples: https://github.com/SonarSource/security-expected-issues/tree/master/javascript/rules/vulnerabilities/RSPEC-2598

Message:
The message is dynamic and depends on the noncompliant property detected.
Restrict [the extension|folder destination] of uploaded files.
if both are non compliant:
Restrict the extension and folder destination of uploaded files.

Detection behavior:

**Formidable**
For all kind of Formidable object instantiations:
```
const formidable = require('formidable');
const form = formidable(options);
// or
const { formidable } = require('formidable');
const form = formidable(options);
// or
const { IncomingForm } = require('formidable');
const form = new IncomingForm(options);
// or
const { Formidable } = require('formidable');
const form = new Formidable(options);
```

We should check the form options. They can be provided in 2 ways: object passed in the constructor (1st argument) or by property assignment after the instantiation.
```
let options = {
uploadDir: "./uploads/",
keepExtensions: false
};
const form = new Formidable(options); // Compliant
// or
const form = new Formidable(); // Compliant because
form.uploadDir = "./uploads/";    // file destination folder is specified (and not empty string)
form.keepExtensions = false;      // and extensions are not kept (report if value is set to 'true')
```

**Multer**
When `multer` is called without `storage` property, it is compliant (a default safe storage is used):
```
let diskupload1 = multer({ // Compliant
  //storage: storage1,
  limits: { 
    fileSize: 2000000 
  }
})
```

if storage is defined look at where the storage object is instantiated and if the destination property of the first object argument exists. If not, issue should be reported on `storage` property with secondary location on `multer.diskStorage({})`
```
let storage1 = multer.diskStorage({ // Noncompliant (secondary location): no destination specified
  /*
  destination: (req, file, cb) => {
    console.log("here 1");
    error = null;
    cb(error, './uploads/')
  },*/
  filename: (req, file, cb) => {
    console.log("here 2");
    const buf = crypto.randomBytes(20);
    cb(null, buf.toString('hex'))
  }
}); 

let diskupload1 = multer({
  storage: storage1, // Noncompliant (main location)
  limits: { 
    fileSize: 2000000 
  }
}) 
```

vilchik-elena:
Fixes #1814

### DIFF
--- a/eslint-bridge/tests/rules/file-uploads.test.ts
+++ b/eslint-bridge/tests/rules/file-uploads.test.ts
@@ -20,7 +20,7 @@
 import { RuleTester } from 'eslint';
 
 const ruleTester = new RuleTester({ parserOptions: { ecmaVersion: 2018, sourceType: 'module' } });
-import { rule } from '../../src/rules/file-uploads';
+import { rule } from 'rules/file-uploads';
 
 function encodedMessage(
   message: string,
--- a/eslint-bridge/tests/rules/file-uploads.test.ts
+++ b/eslint-bridge/tests/rules/file-uploads.test.ts
@@ -22,12 +22,40 @@ import { RuleTester } from 'eslint';
 const ruleTester = new RuleTester({ parserOptions: { ecmaVersion: 2018, sourceType: 'module' } });
 import { rule } from '../../src/rules/file-uploads';
 
+function encodedMessage(
+  message: string,
+  secondary?: { line: number; column: number; endColumn: number; endLine: number },
+) {
+  let secondaryLocations = [];
+  if (secondary) {
+    secondaryLocations = [
+      {
+        message: 'no destination specified',
+        column: secondary.column,
+        line: secondary.line,
+        endColumn: secondary.endColumn,
+        endLine: secondary.endLine,
+      },
+    ];
+  }
+  return JSON.stringify({
+    message,
+    secondaryLocations,
+  });
+}
+
 ruleTester.run('File uploads should be restricted', rule, {
   valid: [
     {
       code: `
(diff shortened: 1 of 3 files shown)


eslint-bridge/tests/rules/file-uploads.test.ts
```diff
@@ -22,6 +22,28 @@ import { RuleTester } from 'eslint';
 const ruleTester = new RuleTester({ parserOptions: { ecmaVersion: 2018, sourceType: 'module' } });
 import { rule } from '../../src/rules/file-uploads';
```
saberduck, line 23: you can remove `../../src`

eslint-bridge/tests/rules/file-uploads.test.ts
```diff
@@ -22,6 +22,28 @@ import { RuleTester } from 'eslint';
...
+function encodedMessage(
+  message: string,
+  secondary?: { line: number; column: number; endColumn: number; endLine: number },
+) {
+  let secondaryLocations = [];
+  if (secondary) {
+    secondaryLocations = [
+      {
+        message: 'no destination specified',
+        column: secondary.column,
```
saberduck, line 34: you can maybe use spread operator here `...secondary` ?

vilchik-elena: I can't :( order of properties in json is different

PR #2063, opened 2020-08-13
Rule S5691: Statically serving hidden files is security-sensitive

### LINKED ISSUE #1842: Rule S5691: Statically serving hidden files is security-sensitive
Implements [RSPEC-5691](https://jira.sonarsource.com/browse/RSPEC-5691)

For both JS/TS

Code examples: https://github.com/SonarSource/security-expected-issues/tree/master/javascript/rules/hotspots/RSPEC-5691

Raise when `dotfiles` property of the 2nd argument of serveStatic() call is set to `allow`.
```
let serveStaticMiddleware = serveStatic('public', 
   { 'index': false, 'dotfiles': 'allow'});
```

vilchik-elena:
Fixes #1842


andrey-tyukin-sonarsource:
Another not entirely uncommon way to import things

```
import serveStatic from 'serve-static'
```

(Here is an example usage: https://nuxtjs.org/api/configuration-servermiddleware/)

doesn't work, but it will be fixed shortly in a separate commit, no action required here.

vilchik-elena:
@andrey-tyukin-sonarsource not sure what is in this example, but I tested it manually, and it is not working
```
import serveStatic from 'serve-static';
console.log(serveStatic); // prints undefined
```

PR #2064, opened 2020-08-13
Support default imports and inlined requires in 'utils'

andrey-tyukin-sonarsource:
Two small enhancements in utils.ts, specifically of the methods that search for module names of expressions:

  1. Inlined requires are now supported:
        ```
        app.use(require('middleware')());
        ```
  now works without requiring intermediate variables. This is a common enough way to instantiate middleware ([example from DVNA](https://github.com/appsecco/dvna/blob/c637437d6515bd4c732e91c58e62d38e88260d3c/server.js#L35))

  2. Default export aliases are supported in import clauses (no need for `* as foo` - `foo` is enough):

        ```
        import serveStatic from 'serve-static'
        ```

  This is important, because middleware-modules usually don't have any other useful functions in the namespace: it's often just the middleware, so the `* as middlewareName` doesn't make much sense. [Example usage](https://nuxtjs.org/api/configuration-servermiddleware/).

### DIFF
--- a/eslint-bridge/src/rules/utils.ts
+++ b/eslint-bridge/src/rules/utils.ts
@@ -61,11 +61,11 @@ export const FUNCTION_NODES = [
 export function getModuleNameOfIdentifier(
   identifier: estree.Identifier,
   context: Rule.RuleContext,
-) {
+): estree.Literal | undefined {
   const { name } = identifier;
   // check if importing using `import * as X from 'module_name'`
-  const importDeclaration = getImportDeclarations(context).find(importDecl =>
-    isNamespaceSpecifier(importDecl, name),
+  const importDeclaration = getImportDeclarations(context).find(
+    importDecl => isNamespaceSpecifier(importDecl, name) || isDefaultSpecifier(importDecl, name),
   );
   if (importDeclaration) {
     return importDeclaration.source;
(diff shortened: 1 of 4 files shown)


andrey-tyukin-sonarsource:
Also, relevant not only for express, new hotspot from `src/ecmascript6/reddit-mobile/src/config.es6.js`, because of:
```
import crypto from 'crypto'
```

eslint-bridge/src/rules/utils.ts
```diff
@@ -78,6 +78,31 @@ export function getModuleNameOfIdentifier(
   return undefined;
 }
 
+/**
+ * Returns the module name of a module-valued expression in following three cases:
```
vilchik-elena, line 82: what is `module-valued`?

eslint-bridge/src/rules/utils.ts
```diff
@@ -78,6 +78,31 @@ export function getModuleNameOfIdentifier(
...
+ *  1. If the node `n` is a `require('m')` call;
+ *  2. If the node `n` is an identifier `i` bound by an import, as in `import i from 'm'`;
+ *  3. If the node `n` is an identifier `i`, and there is a single assignment with a `require`
+ *     on the right hand side, i.e. `var i = require('m')`;
+ *
+ * then, in all three cases, the returned value will be the name of the module `'m'`.
+ *
+ * @param node the expression that is expected to evaluate to a module
+ * @param context the rule context
+ * @return name of the module or `undefined`.
```
vilchik-elena, line 93: with this wording I would expect to get `string` type, not `estree.Literal`

eslint-bridge/src/rules/utils.ts
```diff
@@ -61,11 +61,11 @@ export const FUNCTION_NODES = [
...
   identifier: estree.Identifier,
   context: Rule.RuleContext,
-) {
+): estree.Literal | undefined {
   const { name } = identifier;
   // check if importing using `import * as X from 'module_name'`
-  const importDeclaration = getImportDeclarations(context).find(importDecl =>
-    isNamespaceSpecifier(importDecl, name),
+  const importDeclaration = getImportDeclarations(context).find(
+    importDecl => isNamespaceSpecifier(importDecl, name) || isDefaultSpecifier(importDecl, name),
```
vilchik-elena, line 68: with `isDefaultSpecifier(importDecl, name)` the first phrase in the doc `Returns the module name, when an identifier represents a namespace for that module.` is not true anymore

eslint-bridge/src/rules/utils.ts
```diff
@@ -78,6 +78,31 @@ export function getModuleNameOfIdentifier(
   return undefined;
 }
 
+/**
+ * Returns the module name of a module-valued expression in following three cases:
+ *
+ *  1. If the node `n` is a `require('m')` call;
```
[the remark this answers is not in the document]
andrey-tyukin-sonarsource: Indeed, shouldn't unnecessarily multiply entities here. Removed `n`.

eslint-bridge/src/rules/utils.ts
andrey-tyukin-sonarsource: I meant any expression which, when evaluated, results in some kind of module, with `require('module-name')` being the most prominent example of such expressions. Rephrased.

andrey-tyukin-sonarsource:
I didn't re-request a review, but I've addressed all the mentioned issues in a PR that will follow immediately thereafter.

PR #2066, opened 2020-08-17
Rule S5693 Allowing requests with excessive content length is security-sensitive

### LINKED ISSUE #1882: Rule S5693 Allowing requests with excessive content length is security-sensitive
Implements [RSPEC-5693](https://jira.sonarsource.com/browse/RSPEC-5693)

For both JS/TS

Code examples: https://github.com/SonarSource/security-expected-issues/tree/master/javascript/rules/hotspots/RSPEC-5693

#### Formidable
Check #1814 for the possible ways to use `formidable`
We should check `maxFileSize` against max value defined by parameter for upload requests

#### multer
We should check options passed to `multer` (`limits.fileSize`) against max value defined by parameter for upload requests
```
let diskUpload = multer({
  storage: diskStorage,
  limits: { 
    fileSize: 10000000; // Sensitive: 10MB is more than the recommended limit of 8MB
  }
});

let diskUploadUnlimited = multer({ // Sensitive: the default value is no limit
  storage: diskStorage, 
});
```

#### body-parser
For [body-parser](https://github.com/expressjs/body-parser) all kind of parsers should be handled:
```
bodyParser.json([options])
bodyParser.raw([options])
bodyParser.text([options])
bodyParser.urlencoded([options])
```
`options` has a `limit` property that should be checked against max value defined by parameter for regular requests
Value can be number (than it's bytes) or string (then use https://www.npmjs.com/package/bytes to parse)

vilchik-elena:
Fixes #1882


eslint-bridge/src/rules/content-length.ts
```diff
@@ -0,0 +1,247 @@
...
+    const maxFileSizeValue = getSizeValue(context, property.value);
+    if (maxFileSizeValue) {
+      report(context, property, maxFileSizeValue, useStandardSizeLimit);
+    }
+  } else {
+    report(context, callExpr, defaultLimit, useStandardSizeLimit);
+  }
+}
+
+function getProperty(options: estree.ObjectExpression, key: string) {
```
saberduck, line 189: you can use `getObjectExpressionProperty` from `utils` to replace this function

PR #2067, opened 2020-08-17
Issue 2037: Exclude huge files from analysis

### LINKED ISSUE #2037: Exclude huge files from analysis
Currently we attempt to detect minified files by computing average line length. This is not always sufficient, as some bundles can have standard line size, but they are large overall. This can make `eslint-bridge` unresponsive and block the analysis. We should also exclude very big files from the analysis (threshold to be defined, but I think anything bigger than 512kb is probably generated).

One example offender is this 7MB file in https://github.com/plotly/plotly.py/tree/master/packages/python/plotly/plotlywidget/static

andrey-tyukin-sonarsource:
Fixes #2037


saberduck (approved):
few minor comments, otherwise lgtm!

PR #2072, opened 2020-08-20
Issue 2070: Remove dynamic `require('typescript')`.

### LINKED ISSUE #2070: Remove dynamic require for TypeScript
Since we embed typescript now, there is no need to require it dynamically anymore. We should remove dynamic require

andrey-tyukin-sonarsource:
Fixes #2070


andrey-tyukin-sonarsource:
I've just noticed that we also have

```
const tsParser = require('@typescript-eslint/parser');
```
in a few places.

saberduck (approved):

PR #2081, opened 2020-08-25
Apply the maximum size filter only to JS/TS files

### LINKED ISSUE #2079: Only JS/TS files should be excluded from analysis
In #2067 we started to exclude files bigger than 1MB from analysis. However, we are not testing the language of the file. This might exclude unrelated files from analysis (e.g. COBOL). We should test the language property of `InputFile`.

andrey-tyukin-sonarsource:
Fixes #2079


saberduck (approved):
lgtm!

PR #2088, opened 2020-09-02
Migrate S1321 to eslint implementation (no-with)

guillaume-dequenne-sonarsource:
Removed expected issues on jshint due to parse error in strict mode (https://eslint.org/docs/rules/no-with).


its/ruling/src/test/expected/js/jshint/javascript-S1321.json
```diff
@@ -1,6 +0,0 @@
-{
```
andrea-guarino-sonarsource, line 1: Since there are no issues, I think we should write ourselves some javascript code to assert this( in javascript-test-sources project)

javascript-checks/src/main/java/org/sonar/javascript/checks/WithStatementCheck.java
```diff
@@ -21,21 +21,15 @@
 
 import org.sonar.check.Rule;
 import org.sonar.javascript.checks.annotations.JavaScriptRule;
-import org.sonar.plugins.javascript.api.tree.statement.WithStatementTree;
-import org.sonar.plugins.javascript.api.visitors.DoubleDispatchVisitorCheck;
 import org.sonarsource.analyzer.commons.annotations.DeprecatedRuleKey;
 
 @JavaScriptRule
```
andrea-guarino-sonarsource, line 26: Shouldn't we add it also for TypeScript?

andrea-guarino-sonarsource (changes requested):

its/ruling/src/test/expected/js/jshint/javascript-S1321.json
guillaume-dequenne-sonarsource: Actually, there are issues. This file is only for `jshint` project but we do have issues in `knockout` or `paper` projects for instance.

andrea-guarino-sonarsource: ah sorry indeed, I forgot issues are separated by projects. Nevermind then

javascript-checks/src/main/java/org/sonar/javascript/checks/WithStatementCheck.java
guillaume-dequenne-sonarsource: Indeed, done in 292c3e107fde8e4be40727f571355a4edd3d53d9.

guillaume-dequenne-sonarsource: ...and I actually removed it, as apparently `with` statements are not allowed in TypeScript.

andrea-guarino-sonarsource (approved):

PR #2092, opened 2020-09-02
Migrate S1534 to eslint implementation (no-dupe-keys)

### LINKED ISSUE #2091: Migrate S1534 to eslint implementation (no-dupe-keys) for JS

yassin-kammoun-sonarsource:
Fixes #2091 

--
The issues for the jshint project are no longer raised because the involved Javascript file includes non-parsable code in strict mode, e.g. getters with parameters generate a parsing error.


guillaume-dequenne-sonarsource (changes requested):
With now only 1 case reported on the ruling, I'm afraid our testing is a bit light. Also, the behaviour of the rule is not tested for TS code.
Could you please add a `file-for-rules` folder in `typescript-test-sources` and maybe add some code that would raise an issue (maybe inspired from our previous test cases) ?

yassin-kammoun-sonarsource:
Actually, the rule does not apply to TypeScript as the compiler would detect duplicate properties and raise a compilation error.

guillaume-dequenne-sonarsource:
Fair enough! Could you still please add some of our test cases to the ruling sources for JavaScript so that we can keep some tests for the behaviour of this rule?

guillaume-dequenne-sonarsource (approved):

guillaume-dequenne-sonarsource (approved):
lgtm

PR #2096, opened 2020-09-03
Migrate S3358 to eslint implementation (no-nested-ternary)

### LINKED ISSUE #2094: Migrate S3358 to eslint implementation (no-nested-ternary) for JS/TS (new rule for TS)

yassin-kammoun-sonarsource:
Fixes #2094 

--
We finally don't rely on [no-nested-ternary](https://eslint.org/docs/rules/no-delete-var) because of the following:
- Issues are reported on the nesting ternary expression, not the nested one.
- There are some corner cases in the Java implementation of the rule.
- Only direct children are considered.


eslint-bridge/src/rules/no-nested-conditional.ts
```diff
@@ -0,0 +1,58 @@
...
+          });
+        }
+      },
+    };
+  },
+};
+
+function isNestingBroken(ancestors: estree.Node[]) {
+  let parent = ancestors.pop()!;
+  while (parent.type !== 'ConditionalExpression') {
```
andrea-guarino-sonarsource, line 42: Shouldn't we check that parent is not undefined? or it's not necessary because of the 'ConditionalExpression ConditionalExpression' pattern?

andrea-guarino-sonarsource (approved):
I left a comment to be sure I understood correctly. Otherwise LGTM

eslint-bridge/src/rules/no-nested-conditional.ts
yassin-kammoun-sonarsource: Exactly! We don't need to because we know for sure that the nested expression of type `ConditionalExpression` has a parent of type `ConditionalExpression`.

PR #2098, opened 2020-09-03
Migrate S3500 to eslint implementation (no-const-assign)

### LINKED ISSUE #2097: Migrate S3500 to eslint implementation (no-const-assign) for JS

yassin-kammoun-sonarsource:
Fixes #2097


guillaume-dequenne-sonarsource (approved):
Given how small the rule implementation is, I think I'd prefer to reimplement it, as I feel our message is more informative and the secondary location on the declaration is a pretty nice thing that we would lose. Not sure if we have the API (listing symbol usages) to do it as easily on eslint frontend though, so I let you judge that by yourself (if it would be too long to do I think it's ok as it is).

yassin-kammoun-sonarsource:
As you emphasised, having secondary locations is too valuable to be ignored. As a result, I adapted ESLint's original implementation.

guillaume-dequenne-sonarsource (changes requested):
I think you need to define the rule in `main.ts` as I see on the ruling it's still using the eslint message and location. Besides, I'd suggest renaming the name of the check to avoid any confusion with eslint implementation (maybe to something like `updated-const-variable` to keep the link with the name used for the java implem).

guillaume-dequenne-sonarsource (changes requested):
I think you also need to set the metadata `sonar-runtime` so that the secondary location appears correctly in SonarQube (see `argument-order.ts` for instance).

guillaume-dequenne-sonarsource (approved):
LGTM

PR #2099, opened 2020-09-04
Migrate S3531 to custom eslint-based implementation (generator-without-yield)

guillaume-dequenne-sonarsource:
Adapted from eslint implementation, with more readable issue reporting: https://github.com/eslint/eslint/blob/master/lib/rules/require-yield.js


eslint-bridge/src/rules/generator-without-yield.ts
```diff
@@ -0,0 +1,64 @@
...
+          loc: getMainFunctionTokenLocation(functionNode, getParent(context), context),
+        });
+      }
+    }
+
+    return {
+      FunctionDeclaration: enterFunction,
+      'FunctionDeclaration:exit': exitFunction,
+      FunctionExpression: enterFunction,
+      'FunctionExpression:exit': exitFunction,
```
yassin-kammoun-sonarsource, line 56: This is not mandatory, but just that you know, you can actually simplify these callback subscriptions with:

```typescript
':function[generator=true]': enterFunction,
':function[generator=true]:exit': exitFunction,
```

By doing so, you can get rid of the above if statements testing whether a function-like node is a generator.

yassin-kammoun-sonarsource (approved):
An optional suggestion for improvement, otherwise looks good to me.

eslint-bridge/src/rules/generator-without-yield.ts
guillaume-dequenne-sonarsource: Thanks for the suggestion! I didn't know that! Will do :-)

PR #2102, opened 2020-09-07
Migrate S2703 to eslint implementation (no-implicit-global)

andrea-guarino-sonarsource:
Note that this rule doesn't apply to typescript because it already reports an error.


eslint-bridge/src/rules/no-implicit-global.ts
```diff
@@ -0,0 +1,52 @@
...
+  `Add the "let", "const" or "var" keyword to this declaration of "${variable}" to make it explicit.`;
+
+const excludedNames = flatMap(Object.values(globalsByLibraries), globals => globals);
+
+export const rule: Rule.RuleModule = {
+  create(context: Rule.RuleContext) {
+    return {
+      'Program:exit'() {
+        const globalScope = context.getScope();
+        const alreadyReported: string[] = [];
```
guillaume-dequenne-sonarsource, line 36: Why not use some kind of set here? Genuine question, as I'm not sure what the best practices/implementations are in TS.

guillaume-dequenne-sonarsource (approved):
lgtm, 1 small question.

eslint-bridge/src/rules/no-implicit-global.ts
andrea-guarino-sonarsource: That's a good point. As far as I can remeber, Sets don't gurantee uniqueness with objects (i.e. there is no way to express equality between objects besides relying on reference equality). However sets of strings are guaranteed to preserve uniqueness, so indeed it makes sense to use it. Thanks! :)

PR #2107, opened 2020-09-08
Migrate S1528 to eslint implementation (no-array-constructor)

### LINKED ISSUE #2113: Migrate S1528 to eslint frontend (new rule for TS)

guillaume-dequenne-sonarsource:
Fixes #2113


eslint-bridge/src/rules/array-constructor.ts
```diff
@@ -17,17 +17,28 @@
...
+    function checkNewExpression(node: estree.Node) {
+      const newExpression = node as estree.NewExpression;
+      if (newExpression.callee.type === 'Identifier' && newExpression.callee.name === 'Array') {
+        let message = 'Use either a literal or "Array.from()" instead of the "Array" constructor.';
+        if (
+          newExpression.arguments.length === 1 &&
+          newExpression.arguments[0].type === 'Literal' &&
+          typeof newExpression.arguments[0].value === 'number'
+        ) {
+          message = 'Use "Array.from()" instead of the "Array" constructor.';
```
andrea-guarino-sonarsource, line 34: Why did you change the message? If I'm not wrong Array.from has been introduced starting from ES6, so that may be invalid for people writing code in older JS environment

guillaume-dequenne-sonarsource: From my understanding, it's the only alternative if you want to create an array with a given initial size. 
Since calling the `Array` constructor is so error-prone, we decided to keep raising an issue for those cases nevertheless. But then, the message instructing to use a literal is not valid.

andrea-guarino-sonarsource (approved):

PR #2116, opened 2020-09-09
Migrate S2424 to eslint implementation (no-redeclare)

### LINKED ISSUE #2115: Migrate S2424 to eslint frontend (new rule for TS)

yassin-kammoun-sonarsource:
Fixes #2115 

--
This PR introduces [ESLint es6 environment](https://eslint.org/docs/user-guide/configuring#specifying-environments) during analysis which essentially exposes additional built-in global objects.


eslint-bridge/src/rules/no-built-in-override.ts
```diff
@@ -0,0 +1,104 @@
...
+  'ParallelArray',
+  'StopIteration',
+];
+
+export const rule: Rule.RuleModule = {
+  create(context: Rule.RuleContext) {
+    const overriden: Set<estree.Identifier> = new Set();
+
+    function isBuiltIn(variable: Scope.Variable) {
+      return BUILT_IN_OBJECTS.includes(variable.name);
```
andrea-guarino-sonarsource, line 75: we have as well a list of builtin inside `utils/globals.ts` , can you please check if we can reuse that one?

eslint-bridge/src/rules/no-built-in-override.ts
```diff
@@ -0,0 +1,104 @@
...
+  'Intl',
+  'Generator',
+  'Iterator',
+  'ParallelArray',
+  'StopIteration',
+];
+
+export const rule: Rule.RuleModule = {
+  create(context: Rule.RuleContext) {
+    const overriden: Set<estree.Identifier> = new Set();
```
andrea-guarino-sonarsource, line 72: is this code called at each file? I fear it may not be reinitialised each time, hence we may have a memory leak across files.

eslint-bridge/src/linter.ts
```diff
@@ -90,6 +90,7 @@ export class LinterWrapper {
     const ruleConfig: Linter.Config = {
       rules: {},
       parserOptions: { sourceType: 'module', ecmaVersion: 2018 },
+      env: { es6: true },
```
andrea-guarino-sonarsource, line 93: why do we need to introduce this `env` property?

eslint-bridge/src/rules/no-built-in-override.ts
```diff
@@ -0,0 +1,104 @@
...
+      if (isBuiltIn(variable)) {
+        variable.defs.forEach(def => overriden.add(def.name));
+        variable.references
+          .filter(ref => ref.isWrite())
+          .forEach(ref => overriden.add(ref.identifier));
+      }
+    }
+
+    return {
+      '*': (node: estree.Node) => {
```
andrea-guarino-sonarsource, line 88: can we be more precise? I think this code will be called many times (at each node of the AST). Maybe another approach would be to
- either visit all the scopes, starting from the root
- or having a handler for all AST nodes that bind a variable (assignment statement, assignment expression, for initiliazer, ...)

eslint-bridge/src/rules/no-built-in-override.ts
yassin-kammoun-sonarsource: It is called at each file since the source code we get from the context corresponds to a single source file. Therefore, the set is definitely reinitialised at each time.

eslint-bridge/src/rules/no-built-in-override.ts
yassin-kammoun-sonarsource: We can be more precise, and we do now. I went with the scope-based solution you suggested.

eslint-bridge/src/linter.ts
yassin-kammoun-sonarsource: As mentioned above, this property exposed additional built-in global objects. Furthermore, it allows the detection of global redefinitions like this one:

```
Set = ...;
```

For some reason, the scope does not include references of assigned variables that are undeclared without enabling this property.

eslint-bridge/src/rules/no-built-in-override.ts
andrea-guarino-sonarsource: As discussed, this might not be true if one day we decide to pass more than a single files. Let's play safe and reinitilize this set

eslint-bridge/src/linter.ts
andrea-guarino-sonarsource: This might cause FP on js code older than ES6. However I think we can accept it for the time being and we can check what are the results on peach.

andrea-guarino-sonarsource (approved):
LGTM

PR #2135, opened 2020-09-15
Downgrade eslint @typescript-eslint/parser dependencies to allow Node.js v8

### LINKED ISSUE #2156: Revert Node.js 10 requirement change
We need to revert to eslint 6.7 and typescript-eslint 2.24, to allow analysis on Node.js 8

We should add additional axes on CI to test with Node.js 8, 10, 12, 14

saberduck:
Fix #2156 

It was necessary to revert S2432 back to Java implementation because ESLint rule is available only from v7.5.0

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -142,7 +142,9 @@ ruling_task:
     - build
   <<: *ONLY_SONARSOURCE_QA
   gke_container:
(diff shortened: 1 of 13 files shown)


andrey-tyukin-sonarsource (approved):
lgtm

PR #2136, opened 2020-09-15
Migrate S3533 to eslint-based implementation

### LINKED ISSUE #1869: Add rule S3533 (NonStandardImportCheck) for JS/TS
We already use the [implementation from typescript eslint](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/src/rules/no-require-imports.ts) for TS, however, the current implementation based on Java for JS behaves differently.
With the help of types (see [MMF-1772](https://jira.sonarsource.com/browse/MMF-1772)), we can improve the rule for both for JS and TS.

guillaume-dequenne-sonarsource:
Fixes #1869. I did not rely on type inference (used to detect if the argument of `require` or `define` is actually a string - otherwise it might be the wrong function) as I did not see a single false positive case due to that on the ruling. This also helps align the behaviors of the JS and TS implementations (TS had a high amount of FPs while JS was under-reporting). 

After implementation of MMF-1772 we could further improve this rule if we deem it valuable.


eslint-bridge/src/rules/no-require-or-define.ts
```diff
@@ -0,0 +1,67 @@
...
+  identifier: estree.Identifier,
+): boolean {
+  if (identifier.name !== 'require' && identifier.name !== 'define') {
+    return false;
+  }
+  if (callExpression.arguments.length !== 2 && callExpression.arguments.length !== 3) {
+    return false;
+  }
+  return (
+    callExpression.arguments[callExpression.arguments.length - 1].type === 'FunctionExpression'
```
yassin-kammoun-sonarsource, line 65: I think we can also consider `'ArrowFunctionExpression'` here.

yassin-kammoun-sonarsource (approved):
A minor suggestion, but LGTM!

PR #2149, opened 2020-09-17
Migrate S2376 to eslint implementation (accessor-pairs)

### LINKED ISSUE #2148: Migrate S2376 to eslint implementation (new rule for TS)

yassin-kammoun-sonarsource:
Fixes #2148


guillaume-dequenne-sonarsource (commented):
lgtm, but I wonder why the rule is specified to report on getters without setters as well. 

I'm not sure how the practice translates to JS, but in Java I'd certainly see value to reporting setters without getters but not the other way around, which can be used to create immutable objects.

Now that we can use `eslint` config, I'd advocate for setting `getWithoutSet` to `false` by default and allow the user to configure it as well if they want to.

Let's discuss this on Monday before merging.

javascript-checks/src/main/java/org/sonar/javascript/checks/GetterSetterCheck.java
```diff
@@ -32,17 +33,21 @@
 @Rule(key = "S2376")
 public class GetterSetterCheck implements EslintBasedCheck {
 
+  private static final boolean DEFAULT_GET_WITHOUT_SET = false;
+
+  @RuleProperty(
+    key = "getWithoutSet",
+    description = "s",
```
guillaume-dequenne-sonarsource, line 40: I  think you forgot to update this field.

guillaume-dequenne-sonarsource (approved):
lgtm, don't forget to update property description before merging.

PR #2154, opened 2020-09-18
Migrate S139 (TrailingCommentCheck) to eslint implementation (new rule for TS)

### LINKED ISSUE #2155: Migrate S139 to eslint implementation (new rule for TS)

guillaume-dequenne-sonarsource:
Fixes #2155
A couple notes:
- We have some issues disappearing due to default `eslint` exclusions, which I kept as I found them valuable.
- Some issues disappear due to parsing errors (as always)
- A weird FP appears when comments follow block comments like: `/* some comment*/ // this will trigger FP`, I did not try to get rid of them as I don't find this pattern very readable anyway.
- The regex provided as custom pattern must now comply to JS regex flavor, in particular, that means possessive quantifiers are no longer accepted. Analysis will fail (no file will be analyzed) if such incorrect regex is used.


javascript-checks/src/test/java/org/sonar/javascript/checks/TrailingCommentCheckTest.java
```diff
@@ -19,27 +19,18 @@
...
+  public void configurations() {
+    TrailingCommentCheck check = new TrailingCommentCheck();
 
-    JavaScriptCheckVerifier.issues(check, new File("src/test/resources/checks/trailingComment.js"))
-      .next().atLine(2).withMessage("Move this trailing comment on the previous empty line.")
-      .next().atLine(9)
-      .noMore();
+    // default configuration
+    String defaultConfigAsString = new Gson().toJson(check.configurations());
+    assertThat(defaultConfigAsString).isEqualTo("[{\"ignorePattern\":\"^\\\\s*[^\\\\s]+$\"}]");
```
yassin-kammoun-sonarsource, line 34: You can add another assertion for custom configuration (see [this example](https://github.com/SonarSource/SonarJS/blob/97842545c973481f868b287fddeeb09d8fa3a1e5/javascript-checks/src/test/java/org/sonar/javascript/checks/ClassNameCheckTest.java#L37)).

yassin-kammoun-sonarsource (approved):
Looks good to me. You can just add another assertion checking that rule customisation works as expected.

PR #2161, opened 2020-09-21
Fix SSF-122

### LINKED ISSUE #2123: Fix SSF-122

yassin-kammoun-sonarsource:
Fixes #2123

### DIFF
--- a/eslint-bridge/tests/server.test.ts
+++ b/eslint-bridge/tests/server.test.ts
@@ -289,7 +289,7 @@ describe('server', () => {
 
 describe('support custom rules', () => {
   it('should load custom rule bundles', async () => {
-    const server: http.Server = await start(0, ['custom-rule-bundle']);
+    const server: http.Server = await start(0, '127.0.0.1', ['custom-rule-bundle']);
     expect.assertions(2);
     expect(server.listening).toEqual(true);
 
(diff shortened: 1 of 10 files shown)


eslint-bridge/tests/server.test.ts
```diff
@@ -338,7 +338,7 @@ describe('should send error when failing', () => {
   let close;
 
   beforeEach(async () => {
-    server = await startServer(0, failAnalysis, failAnalysis);
+    server = await startServer(0, '127.0.0.1', failAnalysis, failAnalysis);
```
andrea-guarino-sonarsource, line 341: can we add a test where the loopback address is not '127.0.0.1'?

andrea-guarino-sonarsource (commented):
I would like to have an (integration?) test asserting that the node server is only listening on 127.0.0.1 and not on any other ip addresses. 
However I'm not sure this is very easy to test: if you think it's too complicated, I'm ok with trusting that `app.listen` will behave as expected.

yassin-kammoun-sonarsource:
> * I would like to have an (integration?) test asserting that the node server is only listening on 127.0.0.1 and not on any other ip addresses.

As discussed during standup, it seems quite trick to write such a test. Let's trust Express.js on that one.

> * why is the coverage at 87.5 %? I couldn't understand from SQ UI

It was due to parameters with default values. If default values are not used, the test framework considers that branches are not covered.

eslint-bridge/tests/server.test.ts
yassin-kammoun-sonarsource: I don't think we can. At least, for the particular case of macOS, it does not work unless you manually create some kind of alias (see [here](https://superuser.com/questions/458875/how-do-you-get-loopback-addresses-other-than-127-0-0-1-to-work-on-os-x)).

andrea-guarino-sonarsource (approved):
LGTM

PR #2162, opened 2020-09-21
Fix FP on S1515 when a constant is defined inside a loop

### LINKED ISSUE #1998: FP S1515 ('function-inside-loop'): constant defined inside loop
When variable declared with `const` inside the loop (or in the signature of for-of, for-in loop) using them inside the function is safe. We have FPs in that case:

```
const object = {a: 1, b: 2, c: 3};
for (const property in object) {
   funs[i] = function() {              // FP
        return property;
   };
}

const tasks = [];
while(collection.length > 0) {
  const batch = collection.splice(0, batchSize);
  tasks.push(new Task(() => doSomethingWith(batch))); // FP
}
```

Reported in https://community.sonarsource.com/t/can-s1515-no-functions-declared-inside-a-loop-be-smarter/25258

guillaume-dequenne-sonarsource:
Fixes #1998


yassin-kammoun-sonarsource (approved):
Looks good.

PR #2163, opened 2020-09-22
Fix FP on S1128 for "jsx" import used in "Emotion" pragma

### LINKED ISSUE #1273: FP S1128 for "jsx" import used in "Emotion" pragma
```
/** @jsx jsx */
import { jsx } from '@emotion/core'
```

See docs https://emotion.sh/docs/css-prop

Reported in [community thread](https://community.sonarsource.com/t/jsx-pragma-does-not-prevent-unnecessary-import-warning/7790).

yassin-kammoun-sonarsource:
Fixes #1273

### DIFF
--- a/eslint-bridge/src/rules/unused-import.ts
+++ b/eslint-bridge/src/rules/unused-import.ts
@@ -27,14 +27,26 @@ const EXCLUDED_IMPORTS = ['React'];
 
 export const rule: Rule.RuleModule = {
   create(context: Rule.RuleContext) {
+    const isJsxPragmaSet =
+      context
+        .getSourceCode()
+        .getAllComments()
+        .findIndex(comment => comment.value.includes('@jsx jsx')) > -1;
     const unusedImports: estree.Identifier[] = [];
     const tsTypeIdentifiers: Set<string> = new Set();
     const saveTypeIdentifier = (node: estree.Identifier) => tsTypeIdentifiers.add(node.name);
     return {
       ImportDeclaration: (node: estree.Node) => {
+        const { source } = node as estree.ImportDeclaration;
         const variables = context.getDeclaredVariables(node);
         for (const variable of variables) {
-          if (!EXCLUDED_IMPORTS.includes(variable.name) && variable.references.length === 0) {
+          if (
+            !EXCLUDED_IMPORTS.includes(variable.name) &&
+            variable.references.length === 0 &&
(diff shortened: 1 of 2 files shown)


eslint-bridge/src/rules/unused-import.ts
```diff
@@ -27,14 +27,26 @@ const EXCLUDED_IMPORTS = ['React'];
...
+        const { source } = node as estree.ImportDeclaration;
         const variables = context.getDeclaredVariables(node);
         for (const variable of variables) {
-          if (!EXCLUDED_IMPORTS.includes(variable.name) && variable.references.length === 0) {
+          if (
+            !EXCLUDED_IMPORTS.includes(variable.name) &&
+            variable.references.length === 0 &&
+            variable.name !== 'jsx' &&
+            source.value !== '@emotion/core' &&
+            !isJsxPragmaSet
```
guillaume-dequenne-sonarsource, line 48: To me it's a bit unclear what we are trying to exclude.
Do we really want to exclude all import of `jsx` names and all imports from `@emotion/core` specifically?

From my understanding of the [documentation](https://emotion.sh/docs/css-prop#jsx-pragma), it's really the pragma comment that makes us want to ignore this pattern (and it seems like it being absent would cause problems with the Emotion library). 

If I'm mistaken and there are other patterns (maybe in another version of this framework or something) that we actually want to exclude (maybe just to be safe...), I think it'd be worth it to have explicit tests for them.

guillaume-dequenne-sonarsource (changes requested):
Looks good but please have a look at my comment: I think we can either make the condition a bit more strict or be more explicit about the pattern we exclude.

eslint-bridge/src/rules/unused-import.ts
yassin-kammoun-sonarsource: > To me it's a bit unclear what we are trying to exclude.
> Do we really want to exclude all import of `jsx` names and all imports from `@emotion/core` specifically?

You are right! It's not specific to `@emotion/core`. I just relaxed that condition.

> If I'm mistaken and there are other patterns (maybe in another version of this framework or something) that we actually want to exclude (maybe just to be safe...), I think it'd be worth it to have explicit tests for them.

I added a test case for "any" imported source.

guillaume-dequenne-sonarsource (approved):
lgtm

PR #2167, opened 2020-09-22
Fix FP on S128 when 'process.exit' is called

### LINKED ISSUE #2042: FP S128 (sonar-no-fallthrough): 'process.exit' call
```
switch (x) {
  case 42: // FP
    process.exit(1);
}
```

guillaume-dequenne-sonarsource:
Fixes #2042

### DIFF
--- a/eslint-bridge/src/rules/sonar-no-fallthrough.ts
+++ b/eslint-bridge/src/rules/sonar-no-fallthrough.ts
@@ -28,8 +28,8 @@ export const rule: Rule.RuleModule = {
   create(context: Rule.RuleContext) {
     let currentCodePath: Rule.CodePath | null = null;
     let currentCodeSegment: Rule.CodePathSegment | null = null;
-    const segmentsWithExit = new Set();
-    const initialSegmentsBySwitchCase = new Map();
+    const segmentsWithExit: Set<string> = new Set();
+    const initialSegmentsBySwitchCase: Map<estree.SwitchCase, Rule.CodePathSegment> = new Map();
     const segmentsInSwitchCase: Map<estree.Node, Set<string>> = new Map();
 
     function noComment(node: estree.Node) {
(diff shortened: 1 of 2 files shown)


eslint-bridge/src/rules/sonar-no-fallthrough.ts
```diff
@@ -27,27 +27,88 @@ import { getParent } from 'eslint-plugin-sonarjs/lib/utils/nodes';
...
+        }
+      }
+      return true;
+    }
+
     return {
       onCodePathStart(codePath: Rule.CodePath) {
         currentCodePath = codePath;
       },
       onCodePathEnd() {
```
saberduck, line 61: I think this can be deleted (I know it's not your change, but to simplify logic)

eslint-bridge/src/rules/sonar-no-fallthrough.ts
```diff
@@ -27,27 +27,79 @@ import { getParent } from 'eslint-plugin-sonarjs/lib/utils/nodes';
...
+        const initialSegment: Rule.CodePathSegment = initialSegmentsBySwitchCase.get(switchCase)!;
+        const segmentsToConsider = segmentsInSwitchCase.get(switchCase);
+        let allPathsCallExit = false;
+        if (!!segmentsToConsider && segmentsWithExit.size !== 0) {
+          segmentsToConsider.add(initialSegment.id);
+          allPathsCallExit = allSegmentsReachProcessExitCall(initialSegment, segmentsToConsider);
+        }
         const { cases } = getParent(context) as estree.SwitchStatement;
         if (
           currentCodePath!.currentSegments.some(s => s.reachable) &&
```
saberduck, line 98: we can probably extend this condition that all reachable blocks here, which have a previous segment with `process.exit` call are also unreachable for the purpose of this rule. This should help to avoid the need of keeping track of current segments. However, we still need to be able to stop looking beyond the initial segment.

eslint-bridge/src/rules/sonar-no-fallthrough.ts
guillaume-dequenne-sonarsource: Actually it's still required to handle nested functions in the switch cases.

saberduck (approved):
nice work!

PR #2170, opened 2020-09-24
Add warning when using EOL version of Node.js

### LINKED ISSUE #2159: Add deprecation warning when Node.js 8 is used
We want to update to ESLint v7 and TS 4, those dependencies require Node.js 10 as a minimum. Node.js 8 is EOL since 2019-12-31

The recommended version is LTS 12.x or 14.x 

We should use `AnalysisWarning` API in the sensor, and add a log message on `WARN` level when Node.js 8 is used.

`AnalysisWarning` API should only be used on SonarQube. For SonarCloud we have better mechanism in place which can also add message to the PR decoration.

saberduck:
Fixes #2159

### DIFF
--- a/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/NodeDeprecationWarning.java
+++ b/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/NodeDeprecationWarning.java
@@ -49,8 +49,9 @@ public class NodeDeprecationWarning {
 
   void logNodeDeprecation(int actualNodeVersion) {
     if (actualNodeVersion < MIN_RECOMMENDED_NODE_VERSION) {
-      String msg = "You are using Node.js version {}, which reached end-of-life. " +
-        "Support for this version will be dropped in future release, please upgrade Node.js to more recent version.";
+      String msg = String.format("You are using Node.js version %d, which reached end-of-life. " +
+        "Support for this version will be dropped in future release, please upgrade Node.js to more recent version.",
+        actualNodeVersion);
       LOG.warn(msg);
       if (isSonarQube() && analysisWarnings != null) {
         analysisWarnings.addUnique(msg);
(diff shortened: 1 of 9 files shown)


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/NodeDeprecationWarning.java
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  public NodeDeprecationWarning(SonarRuntime sonarRuntime, @Nullable AnalysisWarnings analysisWarnings) {
+    this.sonarRuntime = sonarRuntime;
+    this.analysisWarnings = analysisWarnings;
+  }
+
+  void logNodeDeprecation(int actualNodeVersion) {
+    if (actualNodeVersion < MIN_RECOMMENDED_NODE_VERSION) {
+      String msg = "You are using Node.js version {}, which reached end-of-life. " +
```
yassin-kammoun-sonarsource, line 52: The actual node version is missing in the message.

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/NodeDeprecationWarningTest.java
```diff
@@ -0,0 +1,86 @@
...
+import org.sonar.api.internal.SonarRuntimeImpl;
+import org.sonar.api.utils.Version;
+import org.sonar.api.utils.log.LogTester;
+import org.sonar.api.utils.log.LoggerLevel;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class NodeDeprecationWarningTest {
+
+  private static final String MSG = "You are using Node.js version {}, which reached end-of-life. Support for this version will be dropped in future release, please upgrade Node.js to more recent version.";
```
yassin-kammoun-sonarsource, line 38: Probably needs to change as well.

yassin-kammoun-sonarsource (changes requested):
The node version is missing in the deprecation warning.

yassin-kammoun-sonarsource (approved):
Looks good!

PR #2176, opened 2020-09-28
Parse JavaScript code with TypeScript ESLint parser

yassin-kammoun-sonarsource:
Fixes #

--

A default TSConfig file is now generated and passed to eslint-bridge for each JavaScript file analysis. The configuration file sets the following compiler options:
- `allowJs: true`, to support parsing of JavaScript-specific syntax
- `noImplicitAny: true`, to make TypeScript compiler "better infer types" (see [innovation sprint's feedback](https://docs.google.com/document/d/1Hw9WI_hZLgTy9GvgAMulWs57nKKerQ-IdSVX0JubC2w/))

As a side note, the ruling expectations for TypeScript were updated. The changes only affect [S125](https://jira.sonarsource.com/browse/RSPEC-125), and are due to TypeScript ESLint parser that is now used first by the rule to parsing commented code.


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/JavaScriptEslintBasedSensor.java
```diff
@@ -42,43 +45,53 @@
...
+  private final TempFolder tempFolder;
 
   /**
    * Required for SonarLint
    */
   public JavaScriptEslintBasedSensor(JavaScriptChecks checks, NoSonarFilter noSonarFilter,
                                      FileLinesContextFactory fileLinesContextFactory,
-                                     EslintBridgeServer eslintBridgeServer) {
-    this(checks, noSonarFilter, fileLinesContextFactory, eslintBridgeServer, null);
+                                     EslintBridgeServer eslintBridgeServer, TempFolder folder) {
```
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: Unless there is some other way I am not aware of, I think we do need the temp folder so that we can write, save the `tsconfig` file in the file system, and pass the file path to eslint-bridge. That's actually what the TypeScript sensor does when a TypeScript project is missing such a configuration file, and a default one is provided.

its/ruling/src/test/expected/js/es5-shim/javascript-S3981.json
```diff
@@ -1,5 +0,0 @@
-{
-'es5-shim:es5-shim.js':[
-1084,
```
andrea-guarino-sonarsource, line 3: This one looks like a regression. What do you think?

yassin-kammoun-sonarsource: No, on the contrary. The rule raises an issue for `value.length >= 0` if the target object happens to be an array. If we look at the function code as a whole, there is nothing that suggests that `value` is an array. So I would argue that it was a false positive that is no longer raised thanks to TypeScript's type inference.

yassin-kammoun-sonarsource: What do you think ?

eslint-bridge/src/parser.ts
```diff
@@ -59,7 +59,17 @@ export type Parse = (
...
+export function parseJavaScriptSourceFile(
+  fileContent: string,
+  filePath: string,
+  tsConfigs?: string[],
+): SourceCode | ParsingError {
+  const parsed = parseTypeScriptSourceFile(fileContent, filePath, tsConfigs);
+  if (parsed instanceof SourceCode) {
+    return parsed;
+  }
+  console.log(`Failed to parse ${filePath} with TypeScript compiler: ${parsed.message}`);
```
yassin-kammoun-sonarsource, line 71: I have mixed feelings about this log. If you look at TypeScript ruling on Cirrus, you will see it quite a lot. Those come from the rule that try to parse and analyse commented code. Should we keep logging ?

its/ruling/src/test/expected/js/es5-shim/javascript-S3981.json
andrea-guarino-sonarsource: Indeed, that makes sense. Thanks for clarifying!

eslint-bridge/src/parser.ts
andrea-guarino-sonarsource: Indeed, that's quite noise.I think we could be:
- either log it only at DEBUG level.
- or raise an exception and let the caller decide if logging or not.

I would suggest that we merge this PR first btw, as it's blocking other tickets. We can then change this log in a separate PR.

yassin-kammoun-sonarsource: I agree.

andrea-guarino-sonarsource (approved):
LGTM

PR #2177, opened 2020-09-28
Wait for the node process to terminate properly

henryju:
When stopping the ESLint bridge, we should wait for the node process to be terminated
before continuing, in order to avoid issues like file locking on Windows.
Using destroy does not garantee that the process is terminated.


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
```diff
@@ -286,7 +283,7 @@ public void clean() {
       } catch (IOException e) {
         LOG.warn("Failed to close server", e);
       }
-      nodeCommand.destroy();
+      nodeCommand.waitFor();
```
henryju, line 286: This is the important change!

nodejs-utils/src/main/java/org/sonarsource/nodejs/NodeCommand.java
```diff
@@ -161,7 +154,14 @@ public Process start(List<String> commandLine, Map<String, String> env) throws I
 
     @Override
     public int waitFor(Process process) throws InterruptedException {
-      return process.waitFor();
+      boolean success = process.waitFor(1, TimeUnit.MINUTES);
```
henryju, line 157: This change is not mandatory, but I think it is better in general to avoid waiting forever.

saberduck: I think you are right about the change, but I think this logic should be implemented in `NodeCommand` rather than `ProcessWrapper` interface, goal of `ProcessWrapper` interface was to abstract from JDK API to make testing easier - so it should map 1:1 to `java.lang.Process` in most cases.

henryju: Make sense

saberduck (approved):
lgtm, thanks!

PR #2187, opened 2020-09-29
Rule: A compare function should be provided when using Array.prototype.sort()

### LINKED ISSUE #706: Rule: A compare function should be provided when using Array.prototype.sort()
Implements [RSPEC-2871](https://jira.sonarsource.com/browse/RSPEC-2871)

We tried to implement this rule but it produces a lot of FP: raises issues for non-numeric array. 
So we need information about array's elements types to solve this problem.

yassin-kammoun-sonarsource:
Fixes #706


andrea-guarino-sonarsource (approved):
LGTM

PR #2188, opened 2020-09-29
Rule S5659: JWT should be signed and verified with strong cipher algo…

### LINKED ISSUE #1776: Rule S5659 JWT should be signed and verified with strong cipher algorithms
Implements [RSPEC-5659](https://jira.sonarsource.com/browse/RSPEC-5659)

For both JS/TS

Code examples: https://github.com/SonarSource/security-expected-issues/tree/master/javascript/vulnerabilities/RSPEC-5659

* Report on `jsonwebtoken.sign()` if third argument contains `algorithm: 'none'` property:
```js
const jwt = require('jsonwebtoken');
let token = jwt.sign({ foo: 'bar' }, key, { algorithm: 'none' }); 
```

* Report on `jsonwebtoken.verify()` if third argument contains `algorithms` property with array containing `none` or if `algorithms` property is missing:
```js
const jwt = require('jsonwebtoken');
const NONE = 'none';
jwt.verify(forgedtoken, key, { expiresIn: 360000 * 5, algorithms: ['RS256', NONE] }, callbackcheck); // Noncompliant
jwt.verify(forgedtoken, key, { expiresIn: 360000 * 5, algorithms: ['RS256', 'none'] }, callbackcheck); // Noncompliant
jwt.verify(forgedtoken, key, { expiresIn: 360000 * 5 }, callbackcheck); // Noncompliant: by default signature algorithm is : none

```

guillaume-dequenne-sonarsource:
…rithm

Fixes #1776


eslint-bridge/src/rules/insecure-jwt-token.ts
```diff
@@ -0,0 +1,162 @@
...
+    function checkCallToSign(
+      callExpression: estree.CallExpression,
+      thirdArgumentValue: estree.ObjectExpression,
+      secondaryLocations: estree.Node[],
+    ) {
+      const algorithmProperty = getObjectExpressionProperty(thirdArgumentValue, 'algorithm');
+      if (!algorithmProperty) {
+        return;
+      }
+      const algorithmValue =
```
saberduck, line 50: I am not entirely sure what was the goal, but it seems to me that ternary is doing the work which is the goal of `getValueOfExpression` utility function. Imo this can be replaced with following

`const algorithmValue = getValueOfExpression<estree.Literal>(context, algorithmProperty.value, 'Literal');`

eslint-bridge/src/rules/insecure-jwt-token.ts
```diff
@@ -0,0 +1,162 @@
...
+            return;
+          }
+          const isCallToSign = isPropertyNamed(callee, 'sign');
+          const isCallToVerify = isPropertyNamed(callee, 'verify');
+          if (callExpression.arguments.length < 3) {
+            // algorithm(s) property is contained in third argument of "sign" and "verify" calls
+            return;
+          }
+          const thirdArgument = callExpression.arguments[2];
+          const thirdArgumentValue =
```
saberduck, line 133: same as above, you don't need to check the type of the node, that's what `getValueOfExpression` is doing for you

eslint-bridge/src/rules/insecure-jwt-token.ts
```diff
@@ -0,0 +1,162 @@
...
+          if (isCallToVerify) {
+            checkCallToVerify(callExpression, thirdArgumentValue, secondaryLocations);
+          }
+        }
+      },
+    };
+  },
+};
+
+function isPropertyNamed(memberExpression: estree.MemberExpression, name: string) {
```
saberduck, line 160: see `isIdentifier` function in `utils`, I think you can replace this one with it.

eslint-bridge/src/rules/insecure-jwt-token.ts
```diff
@@ -0,0 +1,162 @@
...
+              context,
+              algorithmsProperty.value,
+              'ArrayExpression',
+            ) as estree.ArrayExpression);
+      if (!algorithmsValue) {
+        return;
+      }
+      const algorithmsContainNone = algorithmsValue.elements.some(e => {
+        const value =
+          e.type === 'Literal'
```
saberduck, line 97: same as above, you don't need to check the type of the node, that's what `getValueOfExpression` is doing for you

eslint-bridge/src/rules/insecure-jwt-token.ts
```diff
@@ -0,0 +1,162 @@
...
+          node: callExpression.callee,
+          message: toEncodedMessage(
+            'Use only strong cipher algorithms when verifying the signature of this JWT.',
+            secondaryLocations,
+          ),
+        });
+        return;
+      }
+      const algorithmsValue =
+        algorithmsProperty.value.type === 'ArrayExpression'
```
saberduck, line 85: same as above, you don't need to check the type of the node, that's what `getValueOfExpression` is doing for you

eslint-bridge/src/rules/insecure-jwt-token.ts
```diff
@@ -0,0 +1,162 @@
...
+        return value === 'none';
+      });
+      if (algorithmsContainNone) {
+        if (algorithmsProperty.value !== algorithmsValue) {
+          secondaryLocations.push(algorithmsValue);
+        }
+        context.report({
+          node: callExpression.callee,
+          message: toEncodedMessage(
+            'Use only strong cipher algorithms when verifying the signature of this JWT.',
```
saberduck, line 109: maybe we can extract the message to a context or common function, so it's not duplicated

saberduck (changes requested):

saberduck (approved):

PR #2189, opened 2020-09-29
Rule: "for in" should not be used with iterables

### LINKED ISSUE #627: Rule: "for in" should not be used with iterables
Implements [RSPEC-4139](https://jira.sonarsource.com/browse/RSPEC-4139)

yassin-kammoun-sonarsource:
Fixes #627 

--
The rule was reimplemented since tslint rule only considers arrays while we say to cover any kind of iterables.


its/ruling/src/test/expected/js/jshint/javascript-S4139.json
```diff
@@ -0,0 +1,8 @@
+{
+'jshint:src/jshint.js':[
+3504,
+4353,
```
guillaume-dequenne-sonarsource, line 4: The code is not quite readable but I wonder why TS compiler is certain `tokens` is an array here.
 That's the same for the other issues but especially on this one, since a couple lines above I can see:

```
var names = [];
if (_.includes(["{", "["], state.tokens.next.value)) {
    tokens = destructuringPattern(context);
    lone = false;
} else {
    tokens = [ { id: identifier(context), token: state.tokens.curr } ];
    lone = true;
}
```

So I guess `tokens` could be a proper object and the `for...in` could be warranted here?
Maybe there is a configuration on the compiler that could be tuned to avoid inferring a type for sure in such case?
Just wondering here...

guillaume-dequenne-sonarsource (approved):
LGTM
I left a couple questions on ruling issues as I'm not sure I understand exactly how TypeScript type inference works, but I'm not sure there is anything to do about it.

its/ruling/src/test/expected/js/paper.js/javascript-S4139.json
```diff
@@ -0,0 +1,5 @@
+{
+'paper.js:src/item/Item.js':[
+2175,
```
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: I had a look at TypeScript compiler's implementation, and it turns out that there is a compilation phase that transforms the AST by replacing JSDoc annotations with TypeScript-like type annotations (see [annotateWithTypeFromJSDoc](https://github.com/microsoft/TypeScript/blob/eb3645f16b66ed05110ea957c77773cea739bae8/src/services/codefixes/annotateWithTypeFromJSDoc.ts)). This essentially means that type inference is not triggered for `name` since the following:

```javascript
// @param {String} param
function foo(param) {
  /* whatever */
}
```

would be rewritten to:

```typescript
function foo(param: string) {
  /* whatever */
}
```

and type information for `param` would be already available before type checking.

In the end, even tough the code from the highlighted issue claims that name could either be a string, an object, or a function, the TypeScript compilers only considers it to be a string because of the provided JSDoc annotation. As you said, the case is fair to report here.

its/ruling/src/test/expected/js/jshint/javascript-S4139.json
yassin-kammoun-sonarsource: Indeed, the code is not readable, but I would argue that `tokens` is always an array, and the case is also fair to report here. Type inference here takes the union of typing the right-hand sides of both assignments:
- `[ { id: identifier(context), token: state.tokens.curr } ]` is definitely of type `Array` (maybe you missed the square brackets).
- `destructuringPattern(context)` requires typing the function definition of `destructuringPattern` which happens to return an array as well from what I saw.

guillaume-dequenne-sonarsource: Indeed I actually missed the `[]`, fair to report then :)

PR #2192, opened 2020-09-30
Use loopback host address also for requests

### LINKED ISSUE #2191: Use InetAddress.getLoopback also for requests
We use `InetAddress.getLoopbackAddress().getHostAddress()` to configure the host of eslint-bridge. We should use the same host also to make request to the bridge.

saberduck:
Fixes #2191

### DIFF
--- a/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/NetUtils.java
+++ b/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/NetUtils.java
@@ -20,10 +20,7 @@
 package org.sonar.plugins.javascript.eslint;
 
 import java.io.IOException;
-import java.net.InetSocketAddress;
 import java.net.ServerSocket;
-import java.net.Socket;
-import java.net.SocketAddress;
 
 public class NetUtils {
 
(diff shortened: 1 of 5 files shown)


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/NetUtils.java
```diff
@@ -36,29 +33,4 @@ public static int findOpenPort() throws IOException {
...
-  }
-
-  private static boolean serverListening(SocketAddress address) {
-    try (Socket s = new Socket()) {
-      s.connect(address, 1);
-      return true;
-    } catch (Exception e) {
-      return false;
-    }
-  }
```
yassin-kammoun-sonarsource, line 63: How are those changes related to the loopback address issue ?

yassin-kammoun-sonarsource (approved):
Looks good to me. However, I am curious about some changes that are not relate to the loopback address issue unless I'm wrong.

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/NetUtils.java
saberduck: this function was taking `host` as a parameter. I could keep it and update the argument to use the new field, however, this function is legacy of the older version when we didn't have `isAlive` method, which is functionally equivalent. Instead of updating the argument, I took the opportunity to remove it altogether.

yassin-kammoun-sonarsource: Makes sense. Thanks for clarifying.

PR #2195, opened 2020-09-30
Rule S2092: Creating cookies without the 'secure' flag is security-se…

### LINKED ISSUE #1789: Rule S2092 Creating cookies without the "secure" flag is security-sensitive
Implements [RSPEC-2092](https://jira.sonarsource.com/browse/RSPEC-2092)

For both JS/TS

Code examples: https://github.com/SonarSource/security-expected-issues/tree/master/javascript/hotspots/RSPEC-2092

guillaume-dequenne-sonarsource:
…nsitive

Fixes #1789


eslint-bridge/src/rules/insecure-cookie.ts
```diff
@@ -0,0 +1,212 @@
...
+          checkCsurfAndExpressSession(callExpression, true);
+          return;
+        }
+        if (moduleName?.value === 'express-session') {
+          checkCsurfAndExpressSession(callExpression, false);
+          return;
+        }
+        const callee = callExpression.callee;
+        if (callee.type === 'MemberExpression') {
+          const objectValue = getValueOfExpression(
```
saberduck, line 197: should be `getValueOfExpression<estree.NewExpression>` instead of cast

eslint-bridge/src/rules/insecure-cookie.ts
```diff
@@ -0,0 +1,212 @@
...
+          });
+        }
+      }
+    }
+
+    function checkCookiesMethodCall(callExpression: estree.CallExpression) {
+      if (!isIdentifier((callExpression.callee as estree.MemberExpression).property, 'set')) {
+        return;
+      }
+      if (callExpression.arguments.length < 3) {
```
saberduck, line 119: the logic from here seems to be the same as in `checkCookieSession` function, only the argument index is different. Can be factor it out to separate function?

eslint-bridge/src/rules/insecure-cookie.ts
```diff
@@ -0,0 +1,212 @@
...
+      const thirdArgument = callExpression.arguments[2];
+      const cookieObjectExpression = getValueOfExpression<estree.ObjectExpression>(
+        context,
+        thirdArgument,
+        'ObjectExpression',
+      );
+      if (!cookieObjectExpression) {
+        return;
+      }
+      const secureProperty = getObjectExpressionProperty(cookieObjectExpression, 'secure');
```
saberduck, line 132: the code here looks the same function `checkSecurreFlagOnCookieExpression` below. Can we reuse that function?

eslint-bridge/src/rules/insecure-cookie.ts
```diff
@@ -0,0 +1,212 @@
...
+        }
+        if (moduleName?.value === 'csurf') {
+          checkCsurfAndExpressSession(callExpression, true);
+          return;
+        }
+        if (moduleName?.value === 'express-session') {
+          checkCsurfAndExpressSession(callExpression, false);
+          return;
+        }
+        const callee = callExpression.callee;
```
saberduck, line 195: just a small tip, you can use object destructuring to extract and declare variable with the same name from object, like this `const { callee } = callExpression;` Also you can reuse this variable on line 182

eslint-bridge/src/rules/insecure-cookie.ts
```diff
@@ -0,0 +1,212 @@
...
+        }
+        context.report({
+          node: callExpression.callee,
+          message: toEncodedMessage(MESSAGE, secondaryLocations),
+        });
+      }
+    }
+
+    // Common method to check csurf and expression-session libraries, as both set cookies in a similar way
+    function checkCsurfAndExpressSession(callExpression: estree.CallExpression, isCsurf: boolean) {
```
saberduck, line 70: it is a bit strange to have function `checkXandY` with boolean parameter `isY`. It somehow suggests that there is both shared and different logic for X and Y. I think it would be nicer to have a separate function for shared logic which is then used by specific functions for Csurf and Express.

saberduck (changes requested):
there is an opportunity for a few refactorings which I believe would make code a bit more clear

saberduck (approved):

PR #2198, opened 2020-10-01
Migrate S4043 to eslint frontend for JS

### LINKED ISSUE #2183: Migrate S4043 to eslint frontend for JS

yassin-kammoun-sonarsource:
Fixes #2183 

--
Given the newly raised issues, I would argue that "misleading" mutating calls used as arguments should be considered false positive, and therefore not reported. Since the scope of this PR is only about rule migration, I suggest we create a ticket for it, if my claim makes sense.


its/ruling/src/test/expected/js/backbone/javascript-S4043.json
```diff
@@ -1,6 +0,0 @@
-{
-'backbone:test/vendor/qunit.js':[
-1809,
```
andrea-guarino-sonarsource, line 3: why are those removed?

yassin-kammoun-sonarsource: The implementation of the rule in eslint-bridge defines an exception to ignore mutating calls in return statements, i.e. `return array.sort()`. However, as it is written, the exception is not accurate enough, and thus skips cases like the one you highlighted, which includes such a call with ancestors: call expression < binary expression < return statement.

andrea-guarino-sonarsource: Thanks for explanation!

andrea-guarino-sonarsource (approved):

PR #2214, opened 2020-10-01
Rule S3330: Creating cookies without the "HttpOnly" flag is security-sensitive

### LINKED ISSUE #1788: Rule S3330 Creating cookies without the "HttpOnly" flag is security-sensitive
Implements [RSPEC-3330](https://jira.sonarsource.com/browse/RSPEC-3330)

For both JS/TS

Code examples: https://github.com/SonarSource/security-expected-issues/tree/master/javascript/hotspots/RSPEC-3330

guillaume-dequenne-sonarsource:
Fixes #1788


eslint-bridge/src/rules/abstract-cookie-flag-check.ts
```diff
@@ -0,0 +1,196 @@
...
+  isIdentifier,
+  toEncodedMessage,
+} from './utils';
+
+export class AbstractCookieFlagCheck {
+  context: Rule.RuleContext;
+  flag: string;
+  issueMessage: string;
+
+  constructor(context: Rule.RuleContext, flag: string) {
```
saberduck, line 35: I think we can use literal types for flag like `flag: 'http-only'|'secure'` see https://www.typescriptlang.org/docs/handbook/literal-types.html

eslint-bridge/src/rules/abstract-cookie-flag-check.ts
```diff
@@ -0,0 +1,196 @@
...
+import * as estree from 'estree';
+import {
+  getModuleNameOfNode,
+  getObjectExpressionProperty,
+  getValueOfExpression,
+  isIdentifier,
+  toEncodedMessage,
+} from './utils';
+
+export class AbstractCookieFlagCheck {
```
saberduck, line 30: this class is not really abstract in the usual meaning of that word. I would omit it from name.

eslint-bridge/src/rules/abstract-cookie-flag-check.ts
saberduck, line 35: you can use parameter properties https://www.typescriptlang.org/docs/handbook/classes.html#parameter-properties , to avoid the need to declare and assign the field. Just add `private` before the parameter and it will create and assign the field for you.

saberduck (approved):
Some minor comments, otherwise lgtm!

PR #2217, opened 2020-10-02
Rule S5542 Encryption algorithms should be used with secure mode and padding scheme

### LINKED ISSUE #1782: Rule S5542: Encryption algorithms should be used with secure mode and padding scheme
Implements [RSPEC-5542](https://jira.sonarsource.com/browse/RSPEC-5542)

For both JS/TS

Code examples: https://github.com/SonarSource/security-expected-issues/tree/master/javascript/vulnerabilities/RSPEC-5542

* Should raise when substring "CBC" or "ECB" are found in the first argument of crypto.createCipheriv() AND for some alias like AES192 (alias of AES-192-CBC)
* Case insensitive
* The complete list of incorrect algorithms could be found in security-expected-issues

guillaume-dequenne-sonarsource:
Fixes #1782


eslint-bridge/src/rules/encryption-secure-mode.ts
```diff
@@ -0,0 +1,88 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+import { Rule } from 'eslint';
+import * as estree from 'estree';
+import { getModuleNameOfNode, getValueOfExpression } from './utils';
+
+export const rule: Rule.RuleModule = {
+  create(context: Rule.RuleContext) {
+    const aliases: string[] = [
```
saberduck, line 26: it's probably better to declare constants on module level than on the rule level - we don't need this constant array to be allocated on every rule instantiation (although this is a theoretical concern, the rule is probably instantiated only once).

eslint-bridge/src/rules/encryption-secure-mode.ts
```diff
@@ -0,0 +1,88 @@
...
+        const callExpression = node as estree.CallExpression;
+        const { callee } = callExpression;
+        if (callee.type !== 'MemberExpression') {
+          return;
+        }
+        const moduleName = getModuleNameOfNode(context, callee.object);
+        if (moduleName?.value !== 'crypto') {
+          return;
+        }
+        if (callee.property.type !== 'Identifier' || callee.property.name !== 'createCipheriv') {
```
saberduck, line 60: I think that this method name is so specific that we can ignore testing for `crypto` module, let's just test for method name itself. Also, you can use `isIdentifier` function from `utils`.

saberduck (approved):
You can keep testing for `crypto` module if you prefer, I think that method name is so specific, that we could skip it. Otherwise lgtm!

PR #2218, opened 2020-10-02
Remove noisy log from isAlive()

saberduck:
In case of failure proper exception is raised by caller.


guillaume-dequenne-sonarsource (approved):
lgtm

PR #2219, opened 2020-10-05
Use slack_channel argument instead of slack action for release notifi…

tomverin:
…cations

Fixes #


saberduck (approved):

saberduck:
thanks @tomverin !

PR #2221, opened 2020-10-05
Log failures from TypeScript compiler without analysis abort

yassin-kammoun-sonarsource:
Fixes #

--
Failures from TypeScript compiler raised in eslint-bridge are communicated to ESLint sensor as "parsing" errors. The sensor then logs these errors, but does not abort the analysis.

### DIFF
--- a/eslint-bridge/src/analyzer.ts
+++ b/eslint-bridge/src/analyzer.ts
@@ -170,9 +170,21 @@ function analyze(input: AnalysisInput, parse: Parse): AnalysisResponse {
 }
 
 function analyzeFile(sourceCode: SourceCode, input: AnalysisInput) {
-  const issues = linter.analyze(sourceCode, input.filePath).issues;
+  let issues: Issue[] = [];
+  let parsingError: ParsingError | undefined = undefined;
+  try {
+    issues = linter.analyze(sourceCode, input.filePath).issues;
+  } catch (e) {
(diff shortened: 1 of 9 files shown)


eslint-bridge/src/analyzer.ts
```diff
@@ -170,9 +170,21 @@ function analyze(input: AnalysisInput, parse: Parse): AnalysisResponse {
...
 function analyzeFile(sourceCode: SourceCode, input: AnalysisInput) {
-  const issues = linter.analyze(sourceCode, input.filePath).issues;
+  let issues: Issue[] = [];
+  let parsingError: ParsingError | undefined = undefined;
+  try {
+    issues = linter.analyze(sourceCode, input.filePath).issues;
+  } catch (e) {
+    // turns exceptions from TypeScript compiler into "parsing" errors
+    if (e.stack.indexOf('typescript.js:') > -1) {
+      parsingError = { message: e.message, code: ParseExceptionCode.FailingTypeScript };
```
andrea-guarino-sonarsource, line 180: I'm not completely sure about considering those as parsing exceptions. @saberduck wdyt?

saberduck: I think it's OK, it would be cumbersome to create new type of error here.

saberduck (approved):

PR #2223, opened 2020-10-05
Rule S4426: Cryptographic keys should be robust

### LINKED ISSUE #1784: Rule S4426: Cryptographic keys should be robust
Implements [RSPEC-4426](https://jira.sonarsource.com/browse/RSPEC-4426)

For both JS/TS

Code examples: https://github.com/SonarSource/security-expected-issues/tree/master/javascript/rules/vulnerabilities/RSPEC-4426

* Look for crypto.generateKeyPair, crypto.generateKeyPairSync functions
* If first argument of these functions is 'rsa', 'dsa' or 'ec'
* Then look for 2nd argument of these functions, an object
* For rsa if object.modulusLength < 2048 raise an issue
* For dsa if object.modulusLength < 2048 or object.divisorLength < 224  raise an issue
* For ec look for the hardcoded string/curve in object.namedCurve (complete list of noncompliant curves could be found in security-expected-issues code examples)

saberduck:
Fixes #1784


eslint-bridge/src/rules/main.ts
```diff
@@ -239,6 +240,7 @@ ruleModules['no-useless-increment'] = noUselessIncrement;
 ruleModules['no-useless-intersection'] = noUselessIntersection;
 ruleModules['no-variable-usage-before-declaration'] = noVariableUsageBeforeDeclaration;
 ruleModules['no-wildcard-import'] = noWildcardImport;
+ruleModules['no-weak-crypto'] = noWeakCrypto;
```
guillaume-dequenne-sonarsource, line 243: Bit of a nitpick but, as I understand we have one rule reporting on weak cipher algorithms and another reporting on weak keys, maybe it would make sense to be a bit more explicit on the naming of each rule?

eslint-bridge/src/rules/no-weak-crypto.ts
```diff
@@ -0,0 +1,130 @@
...
+      const namedCurveProperty = getObjectExpressionProperty(options, 'namedCurve');
+      const namedCurve = getValueOfExpression<estree.Literal>(
+        context,
+        namedCurveProperty?.value,
+        'Literal',
+      )?.value?.toString();
+      if (namedCurveProperty && namedCurve && WEAK_CURVES.includes(namedCurve)) {
+        context.report({
+          node: namedCurveProperty,
+          message: `${namedCurve} doesn't provide enough security. Use stronger curve.`,
```
guillaume-dequenne-sonarsource, line 94: _nitpick_: It feels to me the second part of the message is missing an `a`: `Use a stronger curve.`

eslint-bridge/src/rules/no-weak-crypto.ts
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    function checkRsaAndDsaOptions(algorithm: string, options: estree.Node) {
+      const modulusProperty = getObjectExpressionProperty(options, 'modulusLength');
+      const modulusLength = getNumericValue(modulusProperty?.value);
+      if (modulusProperty && modulusLength && modulusLength < MINIMAL_MODULUS_LENGTH) {
+        context.report({
+          node: modulusProperty,
+          message: `Use a modulus length of at least ${MINIMAL_MODULUS_LENGTH} bits for ${algorithm} cipher algorithm.`,
+        });
```
guillaume-dequenne-sonarsource, line 72: Do you think it'd make sense to add a secondary location when the `modulusProperty` is an identifier assigned somewhere else (as well as for `divisorLength` and `namedCurve`) ?
By default I've been doing it for other rules as I guess it could be helpful in cases it's assigned somewhere else, but I genuinely wonder if it's worth it.

guillaume-dequenne-sonarsource (approved):
lgtm, I just left minor nitpicks/questions.

eslint-bridge/src/rules/no-weak-crypto.ts
saberduck: I think it's not worth it for this rule. I don't expect assignments to be "far" from the call.

PR #2228, opened 2020-10-07
Rule S4830 Server certificates should be verified during SSL⁄TLS connections

### LINKED ISSUE #1781: Rule S4830: Server certificates should be verified during SSL/TLS connections
Implements [RSPEC-4830](https://jira.sonarsource.com/browse/RSPEC-4830)

For both JS/TS

Code examples: https://github.com/SonarSource/security-expected-issues/tree/master/javascript/vulnerabilities/RSPEC-4830

* Should raise when rejectUnauthorized is set to false

guillaume-dequenne-sonarsource:
Fixes #1781


eslint-bridge/src/rules/utils.ts
```diff
@@ -524,6 +524,43 @@ export function getObjectExpressionProperty(
...
+  functionName: string,
+) {
+  const { callee } = callExpression;
+  if (callee.type !== 'MemberExpression') {
+    return false;
+  }
+  const module = getModuleNameOfNode(context, callee.object);
+  return (
+    module?.value === moduleName &&
+    isIdentifier((callExpression.callee as estree.MemberExpression).property, functionName)
```
saberduck, line 560: we can use `callee` variable and so we shouldn't need cast

saberduck (approved):
lgtm! one minor comment

eslint-bridge/src/rules/utils.ts
saberduck: I wonder how feasible it would be to have a rule for this, probably a bit complicated

PR #2230, opened 2020-10-08
Migrate S3403 to eslint frontend for JS

### LINKED ISSUE #2203: Migrate S3403 to eslint frontend for JS

yassin-kammoun-sonarsource:
Fixes #2203 

--
Even though the `===` operator takes into account the type of its operands, the rule takes exception for:
- any operand of type `any` to avoid reporting potential FPs when TypeScript's type inference "fails" to infer type.
- any operand of type `undefined` or `null` to avoid raising issues on undefined/null-check patterns.


its/ruling/src/test/expected/js/ocanvas/javascript-S3403.json
```diff
@@ -0,0 +1,18 @@
+{
```
guillaume-dequenne-sonarsource, line 1: I understand why we raise this and it makes sense, but to me it feels like this is where the type inference of TS falls a bit short for JS code as those variables are clearly used as unions of `string` and `number` and the operations make sense (still I believe they deserve an issue for that, just not this one 😄 )

its/ruling/src/test/expected/js/jquery/javascript-S3403.json
```diff
@@ -1,5 +1,18 @@
...
+'jquery:src/selector.js':[
+1470,
+],
 'jquery:test/data/jquery-1.9.1.js':[
-9429,
+5302,
+5556,
+],
+'jquery:test/integration/data/gh-1764-fullscreen.js':[
+5,
```
guillaume-dequenne-sonarsource, line 10: I think this one is a FP as both objects are `Window`, although using `getTypeAsString`, I get types `Window` and `Window and typeof globalThis` (and no common flags between the two). Maybe it's just a special case that we have to exclude manually, or maybe there is something else we can do? Not sure...

its/ruling/src/test/expected/js/p5.js/javascript-S3403.json
```diff
@@ -1,5 +1,25 @@
...
 ],
+'p5.js:src/events/acceleration.js':[
+643,
+643,
+645,
+],
+'p5.js:src/io/files.js':[
+1854,
+],
+'p5.js:src/webgl/p5.Matrix.js':[
```
guillaume-dequenne-sonarsource, line 18: In this file I think we also have a problem with `this` keyword. Maybe we should ignore it for this rule?

guillaume-dequenne-sonarsource (changes requested):
Implementation looks good, although you forgot to add the `sonar-runtime` for secondary locations to render correctly.

Ruling issues are a bit disappointing. I left comments on the issues which looked fishy to me, although I don't expect all of them to be actionable, but some might... We can discuss it if you want.

eslint-bridge/tests/rules/different-types-comparison.test.ts
```diff
@@ -0,0 +1,164 @@
...
+    },
+  ],
+  invalid: [
+    {
+      code: `
+        let str = 'str', num = 5;
+        str === num;`,
+      errors: [
+        {
+          message: `{\"message\":\"Remove this \\\"===\\\" check; it will always be false. Did you mean to use \\\"==\\\"?\",\"secondaryLocations\":[{\"column\":8,\"line\":3,\"endColumn\":11,\"endLine\":3},{\"column\":16,\"line\":3,\"endColumn\":19,\"endLine\":3}]}`,
```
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: That's a great idea! Thank you.

its/ruling/src/test/expected/js/underscore/javascript-S3403.json
```diff
@@ -0,0 +1,5 @@
+{
+'underscore:underscore.js':[
+246,
```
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: This case is interesting; the returned type of `keyFinder` is the union `string | number`, which the current implementation does not take into account. I think we need to consider both intersection and union.

its/ruling/src/test/expected/js/jquery/javascript-S3403.json
yassin-kammoun-sonarsource: Same as above. It is definitely a false positive because we don't consider intersections.

its/ruling/src/test/expected/js/p5.js/javascript-S3403.json
```diff
@@ -1,5 +1,25 @@
 {
 'p5.js:lib/addons/p5.sound.js':[
```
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: Same as above, i.e., missing handling of unions (and intersections). Type inference based on JSDoc is correct here, and actually helps.

its/ruling/src/test/expected/js/angular.js/javascript-S3403.json
```diff
@@ -1,4 +1,12 @@
 {
+'angular.js:src/ng/directive/ngModel.js':[
+714,
+],
+'angular.js:src/ng/rootScope.js':[
+274,
```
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: That's right, and I know why those were raised; it's because the implementation didn't consider the underlying symbol of the types, here `Scope`. This can be generalized to functions and classes.

its/ruling/src/test/expected/js/p5.js/javascript-S3403.json
yassin-kammoun-sonarsource: I added an exception for `this` that can't be resolved.

eslint-bridge/tests/rules/different-types-comparison.test.ts
```diff
@@ -114,6 +114,22 @@ ruleTesterTs.run(`Strict equality operators should not be used with dissimilar t
         let str = 'str', undefinedd = undefed;
         str !== undefinedd;`,
     },
+    {
+      code: `
```
guillaume-dequenne-sonarsource, line 118: I think those don't actually test what you intend to test, as types of `union` and `str` are inferred to be `any`. I believe the `TS` parser used for tests doesn't use JSDoc comments to infer types.

guillaume-dequenne-sonarsource (approved):
Nice improvement of the rule!
There are a couple tests that don't work as intended, not sure if you can replicate the behavior from JSDoc type inference to have something that works correctly though.

eslint-bridge/tests/rules/different-types-comparison.test.ts
yassin-kammoun-sonarsource: Didn't manage to understand why JSDoc were not taken into account. As a consequence, those were replaced with TypeScript annotations.

PR #2236, opened 2020-10-12
Generate TSConfig file per analyzed JavaScript file

yassin-kammoun-sonarsource:
A TSConfig file is now generated per analysed JavaScript source file. Furthermore, each JavaScript analysis requests clearing the cache of typescript-eslint parser. And even though the impact is minimal here, the changes in ruling show a positive impact for rules relying on type information. Type inference was incorrect for these cases leading to FPs, which now are no longer reported.

### DIFF
--- a/its/ruling/src/test/java/org/sonar/javascript/it/JavaScriptRulingTest.java
+++ b/its/ruling/src/test/java/org/sonar/javascript/it/JavaScriptRulingTest.java
@@ -85,9 +85,9 @@ public class JavaScriptRulingTest {
   @Parameters(name = "{0}")
   public static Object[][] projects() {
     return new Object[][]{
-      jsProject("amplify", "external/**"),
+      //jsProject("amplify", "external/**"),
       jsProject("angular.js", "src/ngLocale/**", "i18n/**"),
-      jsProject("backbone"),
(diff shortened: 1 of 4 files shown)


its/ruling/src/test/java/org/sonar/javascript/it/JavaScriptRulingTest.java
```diff
@@ -85,9 +85,9 @@ public JavaScriptRulingTest(String project, String language, String sourceDir, L
   @Parameters(name = "{0}")
   public static Object[][] projects() {
     return new Object[][]{
-      jsProject("amplify", "external/**"),
+      //jsProject("amplify", "external/**"),
```
saberduck, line 88: should we uncomment all projects now?

saberduck (changes requested):
looks like not all test are executed?

its/ruling/src/test/java/org/sonar/javascript/it/JavaScriptRulingTest.java
yassin-kammoun-sonarsource: Yes, we should.

saberduck (approved):

PR #2241, opened 2020-10-12
Enhance S3533 with type information

### LINKED ISSUE #2240: Enhance S3533 with type information
Rely on semantic information instead of syntactical analysis.

yassin-kammoun-sonarsource:
Fixes #2240

### DIFF
--- a/eslint-bridge/src/rules/no-require-or-define.ts
+++ b/eslint-bridge/src/rules/no-require-or-define.ts
@@ -20,9 +20,18 @@
 // https://jira.sonarsource.com/browse/RSPEC-3533
 import { Rule } from 'eslint';
 import * as estree from 'estree';
+import {
+  isRequiredParserServices,
+  RequiredParserServices,
+} from '../utils/isRequiredParserServices';
+import { isFunction, isString } from './utils';
 
 export const rule: Rule.RuleModule = {
   create(context: Rule.RuleContext) {
+    const services = context.parserServices;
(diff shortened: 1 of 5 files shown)


eslint-bridge/src/rules/no-require-or-define.ts
```diff
@@ -47,22 +56,25 @@ export const rule: Rule.RuleModule = {
...
     return false;
   }
   if (callExpression.arguments.length !== 2 && callExpression.arguments.length !== 3) {
     return false;
   }
-  const lastArgumentType = callExpression.arguments[callExpression.arguments.length - 1].type;
-  return (
-    lastArgumentType === 'FunctionExpression' || lastArgumentType === 'ArrowFunctionExpression'
-  );
+  return isFunction(callExpression.arguments[callExpression.arguments.length - 1], services);
```
guillaume-dequenne-sonarsource, line 79: I guess by using this new method, you can detect more cases than the previous approach (such as if the argument is a function reference I assume). As we don't have explicit unit tests for `utils` functions, I think it would be quite valuable to have a test case for this.

guillaume-dequenne-sonarsource (approved):
One minor suggestion to add a test case but otherwise LGTM.

eslint-bridge/src/rules/no-require-or-define.ts
yassin-kammoun-sonarsource: I agree. We are indeed missing a test case with inferred functions.

PR #2242, opened 2020-10-12
Add property to disable parsing JavaScript files with TS compiler

andrea-guarino-sonarsource:
Fixes #

### DIFF
--- a/eslint-bridge/src/context.ts
+++ b/eslint-bridge/src/context.ts
@@ -20,7 +20,7 @@
 
 interface Context {
   workDir: string;
-  shouldUseTypescriptParserForJS: boolean;
+  shouldUseTypeScriptParserForJS: boolean;
 }
 
 let context: Context;
--- a/eslint-bridge/src/context.ts
+++ b/eslint-bridge/src/context.ts
@@ -20,6 +20,7 @@
 
 interface Context {
   workDir: string;
+  shouldUseTypescriptParserForJS: boolean;
 }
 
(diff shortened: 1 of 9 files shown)


eslint-bridge/src/parser.ts
```diff
@@ -64,12 +65,15 @@ export function parseJavaScriptSourceFile(
   filePath: string,
   tsConfigs?: string[],
 ): SourceCode | ParsingError {
-  const parsed = parseTypeScriptSourceFile(fileContent, filePath, tsConfigs);
-  if (parsed instanceof SourceCode) {
-    return parsed;
+  const context = getContext();
+  const shouldUseTypescriptParserForJS = context ? context.shouldUseTypescriptParserForJS : true;
```
yassin-kammoun-sonarsource, line 69: Why can't we directly use `context.shouldUseTypescriptParserForJS` instead of this conditional expression? Can the context be undefined?

eslint-bridge/src/context.ts
```diff
@@ -20,6 +20,7 @@
 
 interface Context {
   workDir: string;
+  shouldUseTypescriptParserForJS: boolean;
```
yassin-kammoun-sonarsource, line 23: Could you capitalise the `s` in `Typescript` so that we use a single naming within the project?

yassin-kammoun-sonarsource (approved):
Looks good to me. I left a few comments that need to be addressed before merging.

PR #2245, opened 2020-10-12
Rule S4423: Weak SSL/TLS protocols should not be used

### LINKED ISSUE #1779: Rule S4423: Weak SSL/TLS protocols should not be used
Implements [RSPEC-4423](https://jira.sonarsource.com/browse/RSPEC-4423)

For both JS/TS

Code examples: https://github.com/SonarSource/security-expected-issues/tree/master/javascript/rules/vulnerabilities/RSPEC-4423

- Should look at (`minVersion`/`maxVersion`, `secureProtocol`, `secureOptions`) options

To be compliant the developer should use TLS 1.2 or TLS 1.3, to do that he can use:

- `minVersion`/`maxVersion` properties, `minVersion` should not be lower than `TLSv1.2`, example (see exhaustive cases on security-expected-issues):
```
var options = {
  hostname: 'www.example.com',
  port: 443,
  path: '/',
  method: 'GET',
  minVersion: 'TLSv1.2', 
  maxVersion: 'TLSv1.3'
};

var req = https.request(options, (res) => { // Compliant
  res.on('data', (d) => {
    process.stdout.write(d);
  });
});
```

- or `secureProtocol` property which is a string that should select TLSv1.2 or TLSv1.3, example (see all the noncompliant exhaustive cases on security-expected-issues):
```

var options = {
  hostname: 'www.example.com',
  port: 443,
  path: '/',
  method: 'GET',
  secureProtocol: 'TLSv1_2_method'
};

var req = https.request(options, (res) => {
  res.on('data', (d) => {
    process.stdout.write(d);
  });
});  // Compliant
```

- or `secureOptions` property which is a numeric bitmask value that should remove weak protocols: SSLv2, SSLv3 (already disabled by default on node normally) and TLSv1.0 or TLSv1.1, example (see all the compliant exhaustive cases on security-expected-issues):
```
var options = {
  hostname: 'www.example.com',
  port: 443,
  path: '/',
  method: 'GET',
  secureOptions: constants.SSL_OP_NO_TLSv1 | constants.SSL_OP_NO_TLSv1_1
};

var req = https.request(options, (res) => { // Compliant
  res.on('data', (d) => {
    process.stdout.write(d);
  });
});
```

- If all of these properties (`minVersion`/ `maxVersion`, `secureProtocol`, `secureOptions`) are not defined the code is compliant (see https://nodejs.org/api/tls.html#tls_tls_default_min_version)
- if some of them are defined, only one compliant solution of them (`minVersion`/`maxVersion` or `secureProtocol`or `secureOptions`) is enough to be compliant

If the value of these properties cannot be retrieved (because it's not a literal for instance) the code should be considered as compliant:

```
var options = {
  hostname: 'www.example.com',
  port: 443,
  path: '/',
  method: 'GET',
  secureOptions: getSecureOptionsComplexFunction()
};

var req = https.request(options, (res) => { // Compliant
  res.on('data', (d) => {
    process.stdout.write(d);
  });
});
```

saberduck:
Fixes #1779


eslint-bridge/src/rules/weak-ssl.ts
```diff
@@ -0,0 +1,137 @@
...
+    ) {
+      const unsafeProperty = getObjectExpressionProperty(objectExpression, propertyName);
+      if (unsafeProperty) {
+        return getValueOfExpression(context, unsafeProperty.value, 'Literal');
+      }
+      return undefined;
+    }
+
+    function checkMinMaxVersion(propertyName: string, property: estree.Literal | undefined) {
+      if (property && property.value !== 'TLSv1.2' && property.value !== 'TLSv1.3') {
```
guillaume-dequenne-sonarsource, line 54: Would it make sense to report if the value is equal to `TLSv1` or `TLSv1.1` rather than if it's different from `TLSv1.2 `and `TLSv1.3` ? Just so we don't have to change the implementation in case some new version comes up.
I'm not sure what happens in case an invalid string is passed though, but default min version seems to be `1.2` anyway (https://nodejs.org/api/tls.html#tls_tls_default_min_version)

eslint-bridge/tests/rules/weak-ssl.test.ts
```diff
@@ -0,0 +1,301 @@
...
+      };
+      
+      var req = https.request(options, (res) => {
+        console.log('statusCode:', res.statusCode);
+        console.log('headers:', res.headers);
+      
+        res.on('data', (d) => {
+          process.stdout.write(d);
+        });
+      }); // Noncompliant
```
guillaume-dequenne-sonarsource, line 247: You can remove the `Noncompliant` comment here.

guillaume-dequenne-sonarsource (approved):
Some minor comments but LGTM. Also don't forget to rebase before merging.

eslint-bridge/src/rules/weak-ssl.ts
saberduck: yeah, I was wondering about the same.

PR #2253, opened 2020-10-15
Migrate S3800 to eslint frontend for JS

### LINKED ISSUE #2207: Migrate S3800 to eslint frontend for JS

yassin-kammoun-sonarsource:
Fixes #2207

--
Here are some remarks about the rule implementation:
- The rule ignores functions having multiple return types (i.e. unions) if the function documentation includes the `@return` JSDoc, and considers that the developer is aware of that.
- `ts.Type#isUnion` is not enough to determine whether a type is really a union. There are cases where TypeScript compiler types an expression as being a union while it's not, e.g. the return type of a function including `if (condition) return true; else return false;` is `false | true`.
- Because TypeScript has a dedicated type for object literals, e.g., `{ foo: 'hello' }` has type `{ foo: string }`, the rule disregards the underlying structure of the object types and considers those as `object`. This relaxation in the rule allows, for example, return types of the form `{ foo: string } | {}`.


its/ruling/src/test/expected/js/p5.js/javascript-S3800.json
```diff
@@ -1,15 +1,31 @@
...
-'p5.js:src/utilities/conversion.js':[
-72,
-106,
-134,
-173,
-206,
-237,
-271,
+'p5.js:test/manual-test-examples/tutorials/Integrating-other-libraries/0/buzz.js':[
+184,
```
guillaume-dequenne-sonarsource, line 16: In this file we highlight the `return null`. I believe we should not as it's not the reason we raise an issue.

its/ruling/src/test/expected/js/p5.js/javascript-S3800.json
```diff
@@ -1,15 +1,31 @@
...
+467,
+],
+'p5.js:lib/addons/p5.sound.js':[
+237,
+],
+'p5.js:src/core/error_helpers.js':[
+487,
+],
 'p5.js:src/typography/p5.Font.js':[
+626,
```
guillaume-dequenne-sonarsource, line 12: I think secondary location is a bit hard to read as it enumerates all the properties of the object. I think a simple `return object` would be enough in this kind of cases.

guillaume-dequenne-sonarsource (changes requested):
Implementation looks good to me.
As discussed, there are just some minor changes to issue reporting that I believe would make the readability a bit better.

its/ruling/src/test/expected/js/p5.js/javascript-S3800.json
```diff
@@ -1,15 +1,31 @@
 {
+'p5.js:docs/yuidoc-p5-theme/assets/js/reference.js':[
+467,
+],
+'p5.js:lib/addons/p5.sound.js':[
+237,
+],
+'p5.js:src/core/error_helpers.js':[
+487,
```
guillaume-dequenne-sonarsource, line 9: Here we still have some reporting problems for arrays of objects, I think it would be nice if we could display `object[]` instead.

its/ruling/src/test/expected/js/javascript-test-sources/javascript-S3800.json
```diff
@@ -20,8 +17,11 @@
 'javascript-test-sources:src/ecmascript6/Ghost/core/server/permissions/effective.js':[
 39,
 ],
-'javascript-test-sources:src/ecmascript6/http/test/TestUtils.js':[
-2,
+'javascript-test-sources:src/ecmascript6/expressionist.js/test/watch_parser.spec.js':[
+252,
```
guillaume-dequenne-sonarsource, line 21: Sorry I missed this previously, but I believe we should not report if the returned elements are always functions that differ only by signature.
Also, when we report on function, I believe we can simply display `function` instead of its signature, as it could get hard to read.

guillaume-dequenne-sonarsource (changes requested):
A couple additional comments on reporting for some edge cases.

its/ruling/src/test/expected/js/p5.js/javascript-S3800.json
yassin-kammoun-sonarsource: As I feared, something goes wrong on TypeScript side when resolving the element type of arrays. As a consequence, I resolved their string representation manually for a few cases.

its/ruling/src/test/expected/js/knockout/javascript-S3800.json
```diff
@@ -1,8 +1,8 @@
 {
-'knockout:src/binding/expressionRewriting.js':[
-9,
+'knockout:spec/lib/innershiv.js':[
+12,
 ],
-'knockout:src/subscribables/mappingHelpers.js':[
-23,
+'knockout:src/subscribables/observableArray.js':[
+67,
```
guillaume-dequenne-sonarsource, line 6: Here we now report with secondary only on of of the returns statement with type `any[]`. Is skipping some highlighting on return statements with unresolvable types intended?

its/ruling/src/test/expected/js/angular.js/javascript-S3800.json
```diff
@@ -27,13 +21,12 @@
 1542,
 1548,
 ],
-'angular.js:src/ng/sce.js':[
-51,
-],
-'angular.js:src/shallowCopy.js':[
-10,
+'angular.js:src/ngAnimate/animateCssDriver.js':[
+54,
```
guillaume-dequenne-sonarsource, line 25: Here we now only report on 1 return statement that returns `object`. Is this intended?

guillaume-dequenne-sonarsource (commented):
There are still 2 cases on ruling that appear a bit weird to me. Let's discuss them tomorrow.

its/ruling/src/test/expected/js/knockout/javascript-S3800.json
yassin-kammoun-sonarsource: This is a corner case with return type of the form `void | something` which is not handled correctly. Will fix that.

its/ruling/src/test/expected/js/angular.js/javascript-S3800.json
yassin-kammoun-sonarsource: Nope. Same problem as above.

its/ruling/src/test/expected/js/knockout/javascript-S3800.json
yassin-kammoun-sonarsource: Even though [dd4b5d9](https://github.com/SonarSource/SonarJS/pull/2253/commits/dd4b5d908c49ebb3d2c12e6d13c8ea61be04d836) fixes some cases, it didn't get rid of this wrongly raised issue. After having another look, it turned out to by TypeScript's fault, which incorrectly infer the return type of the function.

Not sure we can do anything about that.

guillaume-dequenne-sonarsource (approved):
LGTM

PR #2254, opened 2020-10-16
Fix accidentally skipped tests

yassin-kammoun-sonarsource:
`it.only` was accidentally merged in a previous PR which led to skipping all tests but the one decorated with `only`.


guillaume-dequenne-sonarsource (approved):
lgtm

PR #2260, opened 2020-10-19
Fix FP on S2999 when instantiating 'this'

### LINKED ISSUE #2261: Fix FP on S2999 when instantiating 'this'
When TypeScript inference mechanism fails to infer `this()` as being instantiable, we wrongly raise issues for instantiations of the form:

```javascript
class MyClass {
  static getInstance() {
    return new this();
  }
}
```

yassin-kammoun-sonarsource:
Fixes #2261


saberduck (approved):
lgtm, I understand that we couldn't reproduce the case in test

PR #2262, opened 2020-10-20
Migrate S3782 to eslint frontend for JS

### LINKED ISSUE #2205: Migrate S3782 to eslint frontend for JS

saberduck:
Fixes #2205 

We lost some TP on parameters of type Function or generics, because I didn't find the way how to test compatibility between actual and generic declared type. However, I think they are acceptable.
There is also one new FP for the case where variable is first assign as string and later used as number. I think this code pattern is fishy in the first place,, so FP is acceptable.


eslint-bridge/src/rules/argument-type.ts
```diff
@@ -0,0 +1,121 @@
...
+      // and no relevant TP, e.g. RegExp, Function
+      return ['String', 'Math', 'Array', 'Number', 'Date'].includes(fqn);
+    }
+
+    function isVarArg(param: ts.ParameterDeclaration) {
+      return !!param.dotDotDotToken;
+    }
+
+    function isFunctionType(declaredType: ts.Type) {
+      return declaredType.getCallSignatures().length > 0;
```
yassin-kammoun-sonarsource, line 57: No need to change anything, but FYI, there is also `ts#isFunctionTypeNode()` that can be used for that purpose.

eslint-bridge/src/rules/argument-type.ts
```diff
@@ -0,0 +1,121 @@
...
+    function isTypeParameter(type: ts.Type) {
+      return type.getFlags() & ts.TypeFlags.TypeParameter;
+    }
+
+    function declarationMatch(
+      declaration: ts.SignatureDeclaration,
+      callExpression: estree.CallExpression,
+    ) {
+      const parameters = declaration.parameters;
+      for (let i = 0; i < parameters.length; i++) {
```
yassin-kammoun-sonarsource, line 69: Just as a preventive measure, we could replace the loop condition with:

```javascript
Math.min(parameters.length, callExpression.arguments.length)
```

eslint-bridge/src/rules/argument-type.ts
```diff
@@ -0,0 +1,121 @@
...
+        const symbol = tc.getSymbolAtLocation(tsCallExpr);
+
+        if (symbol && isBuiltInMethod(symbol)) {
+          let result: {
+            actualType: ts.Type;
+            declaredType: ts.Type;
+            node: estree.Node;
+          } | null = null;
+          for (const declaration of symbol.declarations) {
+            result = declarationMatch(declaration as ts.SignatureDeclaration, callExpression);
```
yassin-kammoun-sonarsource, line 104: What do you think about renaming `result` to `mismatch` and `declarationMatch` to `declarationMismatch` ?

yassin-kammoun-sonarsource (changes requested):

eslint-bridge/src/rules/argument-type.ts
```diff
@@ -0,0 +1,121 @@
...
+            if (!result) {
+              return;
+            }
+          }
+          if (result) {
+            context.report({
+              node: result.node,
+              message: `Verify that argument is of correct type: '${tc.typeToString(
```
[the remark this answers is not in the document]
saberduck: I updated the RSPEC, I think new message is better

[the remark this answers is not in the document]
saberduck: not directly, because I don't have `TypeNode`, but I used same mechanism

yassin-kammoun-sonarsource (approved):
Looks good!

PR #2265, opened 2020-10-20
S2201: Fix FP on "replace" with callbacks and avoid reporting on Array "some" and "every" methods

guillaume-dequenne-sonarsource:


saberduck (approved):
lgtm

PR #2266, opened 2020-10-20
Migrate S3796 to eslint frontend for JS

### LINKED ISSUE #2209: Migrate S3796 to eslint frontend for JS

andrea-guarino-sonarsource:
Fixes #2209


eslint-bridge/src/rules/array-callback-without-return.ts
```diff
@@ -0,0 +1,111 @@
...
+export const rule: Rule.RuleModule = {
+  create(context: Rule.RuleContext) {
+    const services = context.parserServices;
+
+    if (!isRequiredParserServices(services)) {
+      return {};
+    }
+
+    return {
+      'CallExpression[callee.type="MemberExpression"]'(node: estree.Node) {
```
guillaume-dequenne-sonarsource, line 67: That's quite a nitpicky edge case, but I think it might be worth to restrict `MemberExpression` to be only dot member expression (non-computed). Otherwise, you could theoretically have the following FP:

```
function someFunc(callback) {
    callback();
}
const arr = [someFunc];
const some = 0;
arr[some](function() {console.log("hello there!")});
```
Which would report because the property is indeed an identifier named `some`, but is not actually the method we're trying to detect.

guillaume-dequenne-sonarsource (approved):
One minor suggestion for edge case handling, but lgtm otherwise.

eslint-bridge/src/rules/array-callback-without-return.ts
andrea-guarino-sonarsource: Good point! thanks! Fixed in 0d5d86712a5a38d3007e08ed34b624f4351186dd

PR #2294, opened 2020-11-02
Rule S5247: Disabling auto-escaping features is security-sensitive

### LINKED ISSUE #1947: RSPEC-5247: Disabling auto-escaping features is security-sensitive
Implements [RSPEC-5247](https://jira.sonarsource.com/browse/RSPEC-5247)

For both JS/TS

Code examples: https://github.com/SonarSource/security-expected-issues/tree/master/javascript/rules/hotspots/RSPEC-5247

### Mustache
Report on `Mustache.escape` when a user redefines the escape function to the behaviour returning the only parameter without changes:
```
const Mustache = require("mustache");
Mustache.escape = function(text) {return text;}; // Sensitive
// or
Mustache.escape = a => a; // Sensitive
```

### kramed
Report on the property `sanitize` when set to `false`:
```js
let kramed = require('kramed');

var options = {
  renderer: new kramed.Renderer({
    sanitize: false // Sensitive
  })
};
``` 

For all other libraries check RSPEC as pattern to report is straightforward. Report on property disabling the auto-escape (e.g. `noEscape: true`) as primary location. Secondary location should be on method callee using these options
```js
const Handlebars = require('handlebars');
let source = "<p>attack {{name}}</p>";
let template = Handlebars.compile(source, { noEscape: true }); // Sensitive
    secondary   ^^^^^^^^^^^^^^^^^^^^                  ^^^^^^^^^^^^^^^ primary
```

guillaume-dequenne-sonarsource:
Fixes #1947


eslint-bridge/src/rules/disabled-auto-escaping.ts
```diff
@@ -0,0 +1,187 @@
...
+        onlyStatement.argument &&
+        isIdentifier(onlyStatement.argument, firstParamName)
+      ) {
+        return true;
+      }
+      return false;
+    }
+
+    function isInvalidSanitizerFunction(node: estree.Node) {
+      type AssignedFunction =
```
andrea-guarino-sonarsource, line 113: why do we need this type alias? I think it's used only at line 118 right?

andrea-guarino-sonarsource (approved):
Overall LGTM. Because there are no issues in ruling, can you please add one file manually?

eslint-bridge/src/rules/disabled-auto-escaping.ts
guillaume-dequenne-sonarsource: Some code quality tool from our pipeline told me to...:smile:
https://jira.sonarsource.com/browse/RSPEC-4622

To be honest I didn't really know what to do with it, so I just went with the suggestion and made the union explicit. The issue was that if I didn't specify a type explicitly, I had an compile error at line 123 when assigning a `FunctionDeclaration`, because the type checker didn't expect this type at declaration.

If you have a suggestion for a better solution, I'd sure love to make this part a bit cleaner.

andrea-guarino-sonarsource: ok I understand..I think it's fine, we can leave it like that! thanks!

PR #2302, opened 2020-11-03
RSPEC-5742 Disabling Certificate Transparency monitoring is security-sensitive

### LINKED ISSUE #1958: RSPEC-5742 Disabling Certificate Transparency monitoring is security-sensitive
Implements [RSPEC-5742](https://jira.sonarsource.com/browse/RSPEC-5742) MMF-1898

For both JS/TS

Code examples: https://github.com/SonarSource/security-expected-issues/tree/master/javascript/express.js/RSPEC-5742

Should raise on `expectCt` property of an argument of `helmet()` call is set to `false`.
```
app.use(
    helmet({
      expectCt: false // Sensitive
    })
  );
```

We don't care about the values of the expectCt directives (`enforce`, `maxage`) for the moment when expectCt middleware is used (always compliant).

yassin-kammoun-sonarsource:
Fixes #1958

--
This PR introduces an abstraction `Express.SensitiveMiddlewarePropertyRule` that can be reused for all the rules of MMF-1898. One basically needs to provide the following to implement one of such rules:
- a sensitive property finder, that is, a function that returns a property introducing some sensitive setting when using some middleware along with Express.js,
- a message that serves as the reported message in the presence of an issue, i.e., an occurrence of a sensitive property found by the sensitive property finder.


eslint-bridge/src/rules/utils.ts
```diff
@@ -654,4 +654,62 @@ export namespace Express {
...
+    message: string,
+  ): Rule.RuleModule {
+    return {
+      create(context: Rule.RuleContext) {
+        let instantiatedApp: estree.Identifier | null;
+        let sensitiveProperty: estree.Property | undefined;
+        let isSafe: boolean;
+
+        function isExposing(node: estree.Node): boolean {
+          return Boolean((sensitiveProperty = findSensitiveProperty(node)));
```
andrey-tyukin-sonarsource, line 672: This, indeed, seems to warrant using two separate lines. Also, one pair of parentheses seems too much.

eslint-bridge/tests/rules/certificate-transparency.test.ts
```diff
@@ -0,0 +1,98 @@
...
+  ],
+  invalid: [
+    {
+      code: `
+        const helmet = require('helmet');
+        const express = require('express');
+        const app = express(); // Noncompliant
+        app.use(
+          helmet({
+            expectCt: false,
```
andrey-tyukin-sonarsource, line 84: The `Noncompliant` annotation seems to be not where the `errors` position specify it to be.

eslint-bridge/tests/rules/certificate-transparency.test.ts
```diff
@@ -0,0 +1,98 @@
...
+          message: `Make sure disabling Certificate Transparency monitoring is safe here.`,
+          line: 7,
+          endLine: 7,
+          column: 13,
+          endColumn: 28,
+        },
+      ],
+    },
+  ],
+});
```
andrey-tyukin-sonarsource, line 98: Does it also work when `helmet`-call is extracted into a variable? Could we maybe add an example with

```
var h = helmet({ expectCt: false })
app.use(... h ...)
```
?

eslint-bridge/src/rules/utils.ts
```diff
@@ -654,4 +654,62 @@ export namespace Express {
...
+        }
+
+        function findSensitiveProperty(node: estree.Node): estree.Property | undefined {
```
andrey-tyukin-sonarsource, line 675: This `node` parameter corresponds to a single `middleware` passed to `use`, right? I think we should rename this entire chain of `node`-`node`-`node` parameters to something else. maybe `middlewareNode`, meaning that the entire syntactical node should eventually evaluate to a middleware?

eslint-bridge/src/rules/utils.ts
```diff
@@ -654,4 +654,62 @@ export namespace Express {
...
+    ) => estree.Property | undefined,
+    message: string,
+  ): Rule.RuleModule {
+    return {
+      create(context: Rule.RuleContext) {
+        let instantiatedApp: estree.Identifier | null;
+        let sensitiveProperty: estree.Property | undefined;
+        let isSafe: boolean;
+
+        function isExposing(node: estree.Node): boolean {
```
andrey-tyukin-sonarsource, line 671: This `node` parameter should be renamed to reflect more clearly what part of the AST it's checking. I think it should be the node corresponding to a single middleware, but currently, it could be mistaken for the entire `app.use(..)` call, or for the config object.

eslint-bridge/src/rules/utils.ts
```diff
@@ -654,4 +654,62 @@ export namespace Express {
     }
     return false;
   }
+
+  export function SensitiveMiddlewarePropertyRule(
+    sensitivePropertyFinder: (
+      context: Rule.RuleContext,
+      node: estree.CallExpression,
```
andrey-tyukin-sonarsource, line 661: I believe that this parameter could benefit from a more telling name.

andrey-tyukin-sonarsource (changes requested):
That looks like a nice abstraction over an entire bunch of rules!

Just a few minor suggestions:

  * I think there are too many methods taking `node`-parameters. Can we change the name so it reflects better what part of the AST is checked, and what node it corresponds to?
  * Does it follow the references? Would it work for `var h = helmet(badConfig); app.use(h);`? The `isUsingMiddleware` and `flattenArgs` seem to do this, but I think we should add an `invalid` test that shows it.

eslint-bridge/tests/rules/certificate-transparency.test.ts
yassin-kammoun-sonarsource: Forgot to update that after deciding to raise on property instead of instantiated app.

eslint-bridge/src/rules/utils.ts
yassin-kammoun-sonarsource: The extra pair of parentheses is automatically added by the prettier.

andrey-tyukin-sonarsource (approved):
Looks good!

Can be merged unless something comes up in the quality gate.

PR #2303, opened 2020-11-04
RSPEC-5734 Allowing browsers to sniff MIME types is security-sensitive

### LINKED ISSUE #1953: RSPEC-5734 Allowing browsers to sniff MIME types is security-sensitive
Implements [RSPEC-5734](https://jira.sonarsource.com/browse/RSPEC-5734) MMF-1898

For both JS/TS

Code examples: https://github.com/SonarSource/security-expected-issues/tree/master/javascript/express.js/RSPEC-5734

Should raise on the `noSniff` property of the argument of an `helmet()` call when set to `false`:

```
  app.use(
    helmet({
      noSniff: false, // Sensitive
    })
  );
```

yassin-kammoun-sonarsource:
Fixes #1953

--
In addition to implement RSPEC-5734, this PR introduces two changes to `Express.SensitiveMiddlewarePropertyRule`:
- raised issues now include secondary locations: primary on `app.use`, secondary on sensitive property,
- injected `app` as parameters of exported functions are now also considered along with instantiated `app`.


its/ruling/src/test/expected/js/javascript-test-sources/javascript-S105.json
```diff
@@ -23,4 +23,7 @@
 'javascript-test-sources:src/ecmascript6/router/src/util.js':[
 2,
 ],
+'javascript-test-sources:src/file-for-rules/S5743.js':[
```
andrey-tyukin-sonarsource, line 26: I'm probably confusing something.

Could you please briefly comment why there are so many `S5743` mentions in the updated expected issues?

Please double-check that it's not some kind of confusion between

```
S5743
S5734
```

It seems that it suddenly found something about allowing DNS prefetching, but I somehow don't quite understand why it did that now.

javascript-checks/src/main/java/org/sonar/javascript/checks/NoSniffCheck.java
```diff
@@ -0,0 +1,36 @@
...
+package org.sonar.javascript.checks;
+
+import org.sonar.check.Rule;
+import org.sonar.plugins.javascript.api.EslintBasedCheck;
+import org.sonar.plugins.javascript.api.JavaScriptRule;
+import org.sonar.plugins.javascript.api.TypeScriptRule;
+
+@JavaScriptRule
+@TypeScriptRule
+@Rule(key = "S5734")
```
andrey-tyukin-sonarsource, line 29: This one is the mime-sniff, ending with 34. This looks fine (no questions here).

andrey-tyukin-sonarsource (changes requested):
Looks very good!

I think it might be advantageous that both the `findPropertyWithNameXValueY` as well as the batch of tests can be parametrized and kept in one place. The unit-tests in particular seem somewhat repetitive.

Also, it was not entirely clear to me why the expected issues for an example of S57**4**3 were updated, whereas there is no mention of S57**3**4; I'd appreciate a comment here.

its/ruling/src/test/expected/js/javascript-test-sources/javascript-S105.json
yassin-kammoun-sonarsource: Yes, this was particularly annoying to deal with. It is actually related to people working MMF-1974. They are actually adding files in [javascript-test-sources](https://github.com/SonarSource/javascript-test-sources) that are used in integration tests. As long as the updated expected issues from their respective PRs are not merged to the master branch, other PRs can never be green at the same time.

It was probably a bad idea to take ahead and include those changes in this PR.

yassin-kammoun-sonarsource: I will drop b0b62c7.

eslint-bridge/tests/rules/no-sniff.test.ts
```diff
@@ -0,0 +1,258 @@
...
+            helmet({
+              noSniff: false,
+            })
+          );
+        }`,
+    },
+    {
+      code: `
+        const helmet = require('helmet');
+        module.exports = function (foo) { // Not app
```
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: Sorry, it's not different. I guess I duplicated the test by accident.

javascript-checks/src/main/java/org/sonar/javascript/checks/NoSniffCheck.java
yassin-kammoun-sonarsource: I should probably rename the rule to `no-mime-sniff` to emphasise that it's about MIME sniffing.

eslint-bridge/src/rules/no-sniff.ts
```diff
@@ -0,0 +1,53 @@
...
+export const rule: Rule.RuleModule = Express.SensitiveMiddlewarePropertyRule(
+  findFalseNoSniffPropertyFromHelmet,
+  `Make sure allowing browsers to sniff MIME types is safe here.`,
+);
+
+/**
+ * Looks for property `noSniff: false` in node looking
+ * somewhat similar to `helmet(<options>?)`, and returns it.
+ */
+function findFalseNoSniffPropertyFromHelmet(
```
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: Indeed, there are several rules that basically check some property's value when invoking a middleware constructor, like `helmet` in that particular case. However, sensible settings can be passed not only on constructor invocations but also on method calls with the very same middleware, i.e., either `helmet(settings)` or `helmet.method(settings)`. And these settings may not necessarily be a pair of property and value, but a more elaborate object expression. Therefore, I think we should avoid extracting anything here.

eslint-bridge/tests/rules/no-sniff.test.ts
```diff
@@ -0,0 +1,258 @@
...
+            helmet({
+              noSniff: true,
+            })
+          );
+        }`,
+    },
+    {
+      code: `
+        const helmet = require('helmet');
+        module.exports = function (foo) {
```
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: The situation is somehow similar to the one above. One could definitely abstract further and provide parameterizable test suites for this MMF. But again, this is possible only if all rules assert a pair of property and value, which is not always the case. There are indeed rules that require the assertion of object expressions.

andrey-tyukin-sonarsource (approved):
LGTM

PR #2308, opened 2020-11-05
RSPEC-5730 Allowing mixed-content is security-sensitive

### LINKED ISSUE #1950: RSPEC-5730 Allowing mixed-content is security-sensitive
Implements [RSPEC-5732](https://jira.sonarsource.com/browse/RSPEC-5730) MMF-1898

For both JS/TS

Should raise on the csp()/helmet() call (depending on the `helmet-csp`/`helmet` module used) if the `blockAllMixedContent` ("block-all-mixed-content" is an alias) directive is missing:

```
var app = express()

  app.use(csp({
    directives: {
      defaultSrc: ["'self'", 'example.com', 'code.jquery.com'],
    } // Sensitive: blockAllMixedContent is not set
  }));
```
Code examples: https://github.com/SonarSource/security-expected-issues/tree/master/javascript/express.js/RSPEC-5730

when `contentSecurityPolicy` property when set to `false`  NO NEED TO RAISE, because [rule RSPEC-5728](https://github.com/SonarSource/SonarJS/issues/1949) will be already triggered::

```
app.use(
    helmet({
      contentSecurityPolicy: false, // sensitive for rule S5728 only
    })
  );
```

yassin-kammoun-sonarsource:
Fixes #1950


guillaume-dequenne-sonarsource (approved):
Looks good to me, left a minor suggestion.

Small unrelated remark: it seems to me the checks on `utils-express` to guarantee that we indeed have an Express app are not necessary (the creation of the unsafe `helmet` config is enough to report issues, as discussed). Still, as it's already merged, I don't think changing that is worth the trouble.

Before merging, can you add a small test on the ruling test files so we have an issue there as well?

yassin-kammoun-sonarsource:
> Before merging, can you add a small test on the ruling test files so we have an issue there as well?

If it is okay with you, I think I will create one single PR with integration tests for all rules as I forgot to do so for previously merged rules.

PR #2311, opened 2020-11-05
Rule S5725: Disabling resource integrity features is security-sensitive

### LINKED ISSUE #1944: Rule S5725 Disabling resource integrity features is security-sensitive
Implements [RSPEC-5725](https://jira.sonarsource.com/browse/RSPEC-5725)

For both JS/TS

Code examples: https://github.com/SonarSource/security-expected-issues/tree/master/javascript/rules/hotspots/RSPEC-5725

```js
var script = document.createElement("script"); // Sensitive
script.src = "https://cdnexample.com/script-latest.js";
script.crossOrigin = "anonymous";
document.head.appendChild(script);
```

Report on `document.createElement("script")` when for resulting element following conditions are met:
* `script.src` is set to a value starting with "http" or "//"
* `script.integrity` is not defined

guillaume-dequenne-sonarsource:
Fixes #1944
As there aren't issues on ruling, I'll add one test file once https://github.com/SonarSource/SonarJS/pull/2317 is merged.


eslint-bridge/src/rules/disabled-resource-integrity.ts
```diff
@@ -0,0 +1,116 @@
...
+        const { callee } = callExpression;
+        if (left.type !== 'Identifier') {
+          return;
+        }
+        if (callee.type !== 'MemberExpression') {
+          return;
+        }
+        if (
+          !isIdentifier(callee.object, 'document') ||
+          !isIdentifier(callee.property, 'createElement')
```
andrea-guarino-sonarsource, line 98: I think we should check that we're creating a `script` element.
```javascript
var foo = document.createElement("foo");
foo.src = "https://code.jquery.com/jquery-3.4.1.min.js"; // FP
foo.crossOrigin = "anonymous";
document.head.appendChild(foo)
```

eslint-bridge/src/rules/disabled-resource-integrity.ts
```diff
@@ -0,0 +1,116 @@
...
+        const left = variableDeclarator.id;
+        const { callee } = callExpression;
+        if (left.type !== 'Identifier') {
+          return;
+        }
+        if (callee.type !== 'MemberExpression') {
+          return;
+        }
+        if (
+          !isIdentifier(callee.object, 'document') ||
```
andrea-guarino-sonarsource, line 97: Another approach we could also take is using types.
```
const script =document.createElement("script"); // type of script is 'HTMLScriptElement'
```

eslint-bridge/src/rules/disabled-resource-integrity.ts
```diff
@@ -0,0 +1,116 @@
...
+        return false;
+      }
+      return true;
+    }
+
+    function isUnsafeSrcAssignment(memberExpression: TSESTree.Node): boolean {
+      if (!isSrcAssignment(memberExpression)) {
+        return false;
+      }
+      const right = (memberExpression.parent as estree.AssignmentExpression).right;
```
andrea-guarino-sonarsource, line 77: is this cast always safe?

eslint-bridge/src/rules/disabled-resource-integrity.ts
```diff
@@ -0,0 +1,116 @@
...
+      if (right.type !== 'Literal') {
+        return false;
+      }
+      return !!right.raw && !!right.raw.match('^"http');
```
andrea-guarino-sonarsource, line 81: In the ticket, it says we should consider it unsafe also when it starts with `//`

andrea-guarino-sonarsource (changes requested):

eslint-bridge/src/rules/disabled-resource-integrity.ts
guillaume-dequenne-sonarsource: `isSrcAssignment` will indeed return true only if the parent is an `AssignmentExpression`.

eslint-bridge/src/rules/disabled-resource-integrity.ts
guillaume-dequenne-sonarsource: Thanks for the suggestion. I decided to keep both (check that the script has correct type and that it's created through `document.createElement`) as I figure a conservative approach is better, knowing we are afraid of potential FPs with this rule.

andrea-guarino-sonarsource (approved):
LGTM, I think you need to rebase before merging

PR #2315, opened 2020-11-06
Fix S5739

yassin-kammoun-sonarsource:


andrey-tyukin-sonarsource (approved):
Is fast-forward & green, looks good.

PR #2337, opened 2020-11-11
Migrate S3523 to eslint implementation

### LINKED ISSUE #2329: Migrate S3523 to eslint implementation

guillaume-dequenne-sonarsource:
Fixes #2329
The rule was actually deprecated and could probably have been dropped, but as the existing eslint implementation behaves almost identically, I just kept it instead.


saberduck (approved):
lgtm

PR #2338, opened 2020-11-11
Detect missing secondary location support for rules with secondary locations

### LINKED ISSUE #2334: Detect missing secondary location support for rules with secondary locations
The idea is to parse all the rule files in `src/rules` with `typescript-eslint/parser` and analyse each source code with an internal rule. This internal rule will essentially go through the AST, check whether `toEncodeMessage` from `src/rules/utils.ts` is called, and detect whether the expected object expression enabling the internal parameter for having secondary locations is missing or not.

yassin-kammoun-sonarsource:
Fixes #2334


eslint-bridge/tests/rules/main.test.ts
```diff
@@ -0,0 +1,42 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+import * as fs from 'fs';
+import { rules } from 'rules/main';
+import { Linter, SourceCode } from 'eslint';
+import { parseTypeScriptSourceFile } from '../../src/parser';
+import { rule as secondaryLocation } from 'rules/secondary-location';
+
```
andrey-tyukin-sonarsource, line 25: I believe it would be advantageous to point out in the comment that this unit test does not analyse the runtime behaviour of any compiled rule, but instead applies a rule to the _source-code_ of other rules. It's very nice, but it's not a very typical thing to do for a unit test, if this fails, the one who has to fix it might initially find it a bit surprising.

andrey-tyukin-sonarsource (changes requested):
I like the overall approach.

And that's not because I couldn't come up with any other options. I actually considered several alternatives:

  * One might attempt to deprecate the `toEncodedMessage` method, and provide something that checks the `schema` at compile time. While not pretty, it's not impossible - the interested reader might want to experiment with the signature `function reportEncoded<E, A extends (ReadonlyArray<E> | [E]) & {[I in keyof A]: {[K in I]: {'enum': ['the-required-flag']}}}[number]>(rule: { meta: { schema: A, [k: string]: any }, [k: string]: any}, ctx: any, message: string, secondary: [number, number, string][])` and see how it reacts on various `meta`-settings. This, however, would not make the code necessarily easier to understand, and it also would require passing around the `rule` as the evidence of valid configuration, which is a rather counter-intuitive thing to do.
  * One might attempt to do the same with runtime-errors, by replacing the `toEncodedMessage` method by something that again accepts the `rule` and checks its metadata. This, again, would make the interface counter-intuitive, and it would also add some "impossible" branches in the implementation of 'toEncodedMessage'.

The current approach seems both non-intrusive and relatively simple to understand.
However, it seems to be limited. For example, commenting out the `enum: ['sonar-runtime']` in `SensitiveMiddlewarePropertyRule` raises no issues, despite the fact that it was one of the reasons why this rule was suggested in the first place.

There are several ways to fix it:

  1. One simple partial solution would be to extend the range of files over which one iterates (the problem with `utils-express` is that it's not actually a rule itself)
  2. A more robust solution (which, incidentally, is very much in the spirit of "analyze the tests!"-sprint), would be the following. One could define an ordinary test helper method, e.g. `ensureSonarRuntime(rule)`, and then use the rule to analyze the _test code_ instead of the source code. If an `invalid: [ { errors: /.*secondaryLocations.*/ } ]` is discovered, but the `ensureSonarRuntime`-test is absent, then one could raise an issue and point out that this important test is missing. Note how it would use the test code to reduce a complicated semantic problem (following data flow across multiple `utils-*`-files) to a simpler syntactic problem of detecting a missing `ensureSonarRuntime`-test.

yassin-kammoun-sonarsource:
Thank you for the great suggestions!

I really like this idea of analyzing test code rather than main code. As you correctly pointed out, it would solve the problem of rules that "instantiate" `SensitiveMiddlewarePropertyRule` template, which pass under the radar with the current solution. Still, unless I am missing something, I see two flaws with the solution you proposed:

1. Rule tests that do not assert secondary locations but whose implementation use them would not be detected if they happened not to enable `sonar-runtime`. Responsibility would lie between the developer who is expected to test secondary locations at least once and the reviewer who is supposed to report if such a test is missing.
2. There is unfortunately no standard way of testing secondary locations. The rule tests that currently do test secondary locations do so with three different flavours: [string literals](https://github.com/SonarSource/SonarJS/blob/10652906308aafbbeb2558db015986494fe64a90/eslint-bridge/tests/rules/destructuring-assignment-syntax.test.ts#L94), [object literals](https://github.com/SonarSource/SonarJS/blob/10652906308aafbbeb2558db015986494fe64a90/eslint-bridge/tests/rules/no-mime-sniff.test.ts#L168-L178), and [generated from helper functions](https://github.com/SonarSource/SonarJS/blob/10652906308aafbbeb2558db015986494fe64a90/eslint-bridge/tests/rules/cyclomatic-complexity.test.ts#L412-L427). A rule could detect those three patterns, but it seems to be cumbersome. Furthermore, it wouldn't be able to handle new ones.

I see pros and cons for both solutions, but none of them seems to be robust enough. If we put aside the "rule template" problem, the solution of analysing source code seems better. Rule templates were introduced for the very first time during last sprint. I don't expect it to happen often. It should be robust enough to detect that a rule implementation uses secondary locations, regardless of whether the corresponding test suite tests them. On the other hand, analysing test code would nicely work as long as we make sure to always test secondary locations and don't introduce too fancy ways of asserting them. In that regard, we could decide on a standard way of testing secondary locations (the `JSON.stringify` flavour seems to be the "right" way), but standards are not always followed.

No matter how I look at it, I can't make up my mind. I would say we should go with the `analyse the tests` solution, but first, what do you think about the flaws I pointed out ?

_(I didn't realised before understanding the purpose of your `ensureSonarRuntime(rule)` that we have access to rules' metadata from the rule itself. There is indeed no need to analyse that part of the source code.)_

andrey-tyukin-sonarsource (approved):
The outcome of this review was flipped from "Request changes" to "Approve", without any too significant changes having been made. This requires some explanation.

Back when I requested the changes, I assumed that it should be relatively straightforward to refactor a code in a way that would completely exclude the forgotten `sonar-runtime' setting.

However, further experiments revealed that:

  * Enforcing the configuration by a runtime check would require to propagate the `rule` (or the rule metadata) to every position where the problematic `toEncodedMessage` is used. This would require change in a large amount of helper methods, and also introduce difficult to understand circular dependencies between the rules and their helper methods.
  * Attempts to do the configuration check completely at compile time using type-level computations were not successful either, because if a rule (together with its metadata) is not passed directly as an object literal to the checking method, the exact types of the nested configuration settings are not inferred (e.g., instead of `['sonar-runtime']`, a broader `string[]` is inferred etc.). 

In the end, decoupling the configuration check from the code by implementing a separate rule seemed like the least intrusive option that still should prevent most such configuration issues.

Issue #1810, opened 2019-12-16
Deprecate low value rules


quentin-jaquier-sonarsource:
After a first round of deprecation (#1698), we realized that some additional rules are behaving poorly and don't provide a lot of value.

- [S3002](https://jira.sonarsource.com/browse/RSPEC-3002): _Unary operators "+" and "-" should not be used with objects_

This rule is hardly ever reporting any issue, and when it does, it's questionable since it can be correct to use unary operator on object, for example:
```
let x = {toString() {return "42";}};
+x;
```
does not result in NaN.

- [S2583](https://jira.sonarsource.com/browse/RSPEC-2583): _Conditionally executed blocks should be reachable_

In [S2589](https://jira.sonarsource.com/browse/RSPEC-2589), we already report when a condition is always true or false, if it results in dead code, it's superfluous to report it a second time.
We therefore decided to deprecate this rule, and keep only [S2583](https://jira.sonarsource.com/browse/RSPEC-2589).

- [S2873](https://jira.sonarsource.com/browse/RSPEC-2873): _Calls should not be made to non-callable values_

We realized that this rule was never behaving very well, most of the issues that we could observe were false positive. Even with the current migration improvement, we will not have the tools to improve the results to a state where it's worth the migration effort, we therefore decided to deprecate it.

-  [S2714](https://jira.sonarsource.com/browse/RSPEC-2714): _Element type selectors should not be used with class selectors_
This rule is related to JQuery, we missed it during the first path of deprecation. See #1698.

## Security Rules
- [S2611](https://jira.sonarsource.com/browse/RSPEC-2611), Untrusted content should not be included, see #1859
 
- [S3271](https://jira.sonarsource.com/browse/RSPEC-3271), see #1862

PR #2339, opened 2020-11-11
Drop deprecated rules S2550, S2583 and S2611

guillaume-dequenne-sonarsource:


andrea-guarino-sonarsource (commented):
[S2583](https://jira.sonarsource.com/browse/RSPEC-2583) doesn't seem to be deprecated. Could you please double check if we really want to drop it?

guillaume-dequenne-sonarsource:
S2583 doesn't look deprecated in Jira but in the deleted `S2583.html` there's this comment:
```
<!-- Make sure to not drop this deprecation section when doing bulk update (as well as updating S2583.json)  -->	
<!-- This rule is deprecated only for JS language, that's why we can't deprecate it in RSPEC -->
```
So I assumed it was correctly flagged to be deleted.

andrea-guarino-sonarsource:
> S2583 doesn't look deprecated in Jira but in the deleted `S2583.html` there's this comment:
> 
> ```
> 	
> 
> ```
> 
> So I assumed it was correctly flagged to be deleted.

Indeed, I found the ticket deprecating this rule: https://github.com/SonarSource/SonarJS/issues/1810
Thanks!

andrea-guarino-sonarsource (approved):

PR #2343, opened 2020-11-12
Migrate S2259 (NullDereferenceCheck) to eslint implementation

### LINKED ISSUE #2331: Migrate S2259 (NullDereferenceCheck) to eslint implementation

andrea-guarino-sonarsource:
Fixes #2331


eslint-bridge/tests/rules/null-dereference.test.ts
```diff
@@ -0,0 +1,414 @@
...
+              x.call();
+            }
+          
+            var y = foo();
+          
+            if (typeof y === 'undefined') {
+              y.call();  // Noncompliant
+            }
+          }`,
+      errors: 1,
```
guillaume-dequenne-sonarsource, line 360: Can you maybe add an assertion on the location for this issue?

eslint-bridge/tests/rules/null-dereference.test.ts
```diff
@@ -0,0 +1,414 @@
...
+      
+        var initializedArray = [];
+        for(let i of initializedArray) {      // OK
+        }
+      
+        var x;
+        for(x of obj) {                       // OK we should not care about x being undefined
+        }
+      }`,
+      errors: 2,
```
guillaume-dequenne-sonarsource, line 392: Same here, I feel it's better to be 100% sure the issue is at the right place.

guillaume-dequenne-sonarsource (approved):
LGTM.
Although I think that, while we lose quite a few FPs from the change of implementation, we also lose interesting TPs as well. Do you think it'd be worth documenting somehow the rules for which it happens for this last batch of migration? 
I feel it's  bound to happen for some of the remaining rules and while it shouldn't stop us from finishing the migration, it could be interesting to keep a list of rules that we know we can improve later on.

andrea-guarino-sonarsource:
As discussed, I will create tickets to document those introduced FN

PR #2357, opened 2020-11-17
Migrate S3798 to eslint eslint-based implementation

### LINKED ISSUE #1871: Add rule S3798 (DeclarationInGlobalScopeCheck) for JS/TS (new rule for TS)

guillaume-dequenne-sonarsource:
Fixes #1871

Some new ruling issues are due to environments that we supported but are not supported by eslint (e.g `rhino`) and some of the missing ones are FNs due to slight change in variable scopes (as we only consider module scopes variables for S3798)


its/ruling/src/test/expected/js/angular.js/javascript-S3798.json
```diff
@@ -187,6 +187,7 @@
 3,
 4,
 5,
+6,
```
andrea-guarino-sonarsource, line 190: just an idea: maybe we can avoid raising on assignments having `require('foo')` as rhs. wdyt?
It could be treated in a separate tickets in any way

andrea-guarino-sonarsource (commented):
on implementation side, LGTM. However I would like to hear the opinion from someone of JS bubble (cc @saberduck) regarding the introduction of those new properties to setup environment and globals.

its/ruling/src/test/expected/js/angular.js/javascript-S3798.json
guillaume-dequenne-sonarsource: I created https://github.com/SonarSource/SonarJS/issues/2366 to fix it (and maybe we can identify some other cases that we can improve on Peach in the meantime), I guess it's indeed not urgent as the previous implementation was already raising issues in such cases.

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServer.java
```diff
@@ -39,6 +39,8 @@
 
   void initLinter(Rule[] rules) throws IOException;
```
saberduck, line 40: it seems that there is only single usage of this method in tests. Instead of overload we can drop it and have only one signature with environments.

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServer.java
```diff
@@ -39,6 +39,8 @@
...
 
+  void initLinter(Rule[] rules, String[] environments, String[] globals) throws IOException;
```
saberduck, line 42: I would use `List` instead of array. Collections should be preferred before arrays when possible. (I know that rules used array before, but I think it's a mistake, so this is a chance to fix it).

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
```diff
@@ -343,4 +349,16 @@ private HttpUrl url(String endpoint) {
       this.tsconfig = tsconfig;
     }
   }
+
+  static class InitLinterRequest {
+    Rule[] rules;
```
saberduck, line 354: you can also use `List` here, GSON will handle it

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/TypeScriptSensor.java
```diff
@@ -107,7 +109,9 @@ void analyzeFiles() throws IOException, InterruptedException {
     boolean success = false;
     ProgressReport progressReport = new ProgressReport("Progress of TypeScript analysis", TimeUnit.SECONDS.toMillis(10));
     List<InputFile> inputFiles = getInputFiles();
-    eslintBridgeServer.initLinter(rules);
+    String[] environments = context.config().getStringArray(ENVIRONMENTS_PROPERTY_KEY);
```
saberduck, line 112: it seems that this initialization could be pulled up to the superclass and shared between sensors

saberduck (changes requested):
overall approach looks good, but I made some comments to improve implementation

saberduck (approved):

PR #2362, opened 2020-11-19
Migrate S3827 to eslint-based implementation

### LINKED ISSUE #1872: Add rule S3827 (ReferenceErrorCheck) for JS/TS (new rule for TS)

guillaume-dequenne-sonarsource:
Fixes #1872

New issues are mostly due to environments enabled by default on the Java frontend that are not supported by eslint (e.g `wsh` and `rhino`). We could add them by default as `globals` anyway, or simply let users do it by themselves if needed.

Should not be merged before https://github.com/SonarSource/SonarJS/pull/2357


eslint-bridge/src/rules/no-reference-error.ts
```diff
@@ -0,0 +1,84 @@
...
+        undeclaredIdentifiersByName.clear();
+        const globalScope = context.getScope();
+        globalScope.through.forEach(ref => {
+          const identifier = ref.identifier;
+          if (excludedNames.has(identifier.name)) {
+            return;
+          }
+          if (
+            ref.writeExpr ||
+            hasTypeOfOperator(identifier as TSESTree.Node) ||
```
andrea-guarino-sonarsource, line 50: I'm not sure it's always safe to cast to `TESTree.Node`. What if another parser is used instead of TS? Usually I check if `identifier.hasOwnProperty('parent')` before casting . Maybe we should put that in a utility function.

andrea-guarino-sonarsource (approved):
LGTM, just a comment about casting to `TSESTree`

eslint-bridge/src/rules/no-reference-error.ts
guillaume-dequenne-sonarsource: From a previous discussion, it seemed to me that `estree.Node` is a subset of `TESTree.Node` and that the cast is always safe, as the `parent` property of the latter is optional.
I see there are many usages (50+) in our rules where the cast is made without additional safety so I'd argue it's probably safe (and if it's not for some use case, we should know  about it very quickly as everything would break 😄 )

PR #2376, opened 2020-11-25
Migrate S2432 to eslint implementation

### LINKED ISSUE #1858: Migrate rule S2432 (ReturnInSetterCheck) for JS
Double check with https://eslint.org/docs/rules/no-setter-return if possible to use eslint core implementation.

guillaume-dequenne-sonarsource:
Fixes #1858 
Migrated only the use case covered by the existing Java-based rule. Created https://github.com/SonarSource/SonarJS/issues/2377 to use eslint rule once NodeJS upgrade is complete.


eslint-bridge/src/rules/no-return-in-setter.ts
```diff
@@ -17,17 +17,34 @@
...
-import org.sonar.javascript.checks.verifier.JavaScriptCheckVerifier;
+import { Rule } from 'eslint';
+import * as estree from 'estree';
+import { childrenOf } from '../utils/visitor';
 
-public class ReturnInSetterCheckTest {
+export const rule: Rule.RuleModule = {
+  create(context: Rule.RuleContext) {
+    return {
+      'Property[kind="set"]': (node: estree.Node) => {
```
andrea-guarino-sonarsource, line 29: I would suggest to simplify the code like:
```typescript
'Property[kind="set"] ReturnStatement': (node: estree.Node) => {
        if (!!(node as estree.ReturnStatement).argument) {
          context.report({
            node,
            message: 'Consider removing this return statement; it will be ignored.',
          });
        }
      },
```

andrea-guarino-sonarsource (approved):
LGTM, just a suggestion to simplify implementation

eslint-bridge/src/rules/no-return-in-setter.ts
guillaume-dequenne-sonarsource: Thanks for the suggestion! I always forget how powerful those selectors can be.

PR #2386, opened 2020-11-30
Simplify "reaching definitions" implementation and avoid non-terminating computation

andrea-guarino-sonarsource:


guillaume-dequenne-sonarsource (approved):
LGTM

PR #2392, opened 2020-11-30
Migrate S2189 to eslint-based implementation

### LINKED ISSUE #674: False Positive in LoopsShouldNotBeInfiniteCheck : support 'yield'
[RSPEC-2189](https://jira.sonarsource.com/browse/RSPEC-2189)

`yield` should be supported in the CFG.

```
function * generator() {
    while(true) { // False positive here
        foo();
        yield;
    }
}
```

As soon as this is implemented we should put this in an Exceptions section on the rule's description.

### LINKED ISSUE #1275: S2189 (Loops should not be infinite) raised on a variant condition
**I want to report a bug.**

**SonarJS version:** 5.2.1 (build 7778)<br>
**SonarQube version:** 7.7<br>
**SonarLint version:** 1.9.0<br>

**Rule key:** javascript:S2189

**Reproducer** 
```
const sleep = time => {
  let count = 0;
  const waitTill = new Date(new Date().getTime() + time);
  while (waitTill > new Date()) { // False-positive here
    count += 1;
  }
};
```

**Expected behavior**
I think the _javascript:S2189_ issue should not be raised for the given code, The issue description recommends to "_Correct this loop's end condition as to not be invariant_" but really the condition changes since "New Date()" is different at each evaluation. 

For example, if we make this change explicit as shown in the code below, the issue is not raised any more:

const sleep = time => {
  let count = 0;
  const waitTill = new Date(new Date().getTime() + time);
  let now = new Date();
  while (waitTill > now) {
    count += 1;
    now = new Date();
  }
};

### LINKED ISSUE #2278: Migrate rule S2189
Eslint https://eslint.org/docs/rules/no-unmodified-loop-condition 

After migration see also #1275 and #674 if they are still relevant (and eventually fix them)

guillaume-dequenne-sonarsource:
Fixes #2278
Fixes #1275
Fixes #674

Some new issues are arguably FPs (looking more like code smells than actual bugs) due to the fact the eslint rule has a slightly different scope than ours (reports on variables used as loop conditions despite not being directly modified in the loop - while ours is trying to report on "true" infinite loops) but it seems acceptable to report them.

Documented in the tests some new FNs that we used to report with the previous implementation.


javascript-checks/src/test/java/org/sonar/javascript/checks/LoopsShouldNotBeInfiniteCheckTest.java
```diff
@@ -25,9 +25,11 @@
 
 public class LoopsShouldNotBeInfiniteCheckTest {
 
+/*
```
saberduck, line 28: should this be deleted?

saberduck: also file `loopsShouldNotBeInfinite.js` should be deleted

saberduck (approved):
I think there is some forgotten code which should be deleted, otherwise lgtm

PR #2395, opened 2020-12-01
Issue 2334 minor refactoring

andrey-tyukin-sonarsource:


yassin-kammoun-sonarsource (approved):
LGTM!

Thanks for adding the small piece of code that throws an error if there is a mismatch.

andrey-tyukin-sonarsource:
> Thanks for adding the small piece of code that throws an error if there is a mismatch.

When the condition was satisfied, it would throw an error regardless, the manually thrown error merely provides a better comprehensible error message.

PR #2403, opened 2020-12-03
Remove Java based frontend

### LINKED ISSUE #2119: Migrate S1314 (no-octal) to eslint implementation (new rule for TS)

### LINKED ISSUE #2393: Migrate S2260 (parsing error check)
Technically this class exists only to register the rule, as parsing error mechanism for eslint rules is handled differently. To "migrate" this rule, we just need to change `extends` clause to implement directly `JavaScriptCheck`, which will be true marker interface (without any methods)

saberduck:
Fix #2119
Fix #2393


its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/NoSonarTest.java
```diff
@@ -54,6 +56,12 @@ public static void startServer() {
     orchestrator.executeBuild(build);
   }
 
+  static WsClient newWsClient(Orchestrator orchestrator) {
```
guillaume-dequenne-sonarsource, line 59: _Nitpick:_ Can we not use the one defined in `Tests` (`com.sonar.javascript.it.plugin.Tests.newWsClient`) instead ?
I see this one is used in `EslintCustomRulesTest` as well, providing a different instance of orchestrator, but perhaps we can still only keep one method?

sonar-javascript-plugin/pom.xml
```diff
@@ -73,6 +61,10 @@
       <groupId>org.sonarsource.sonarqube</groupId>
       <artifactId>sonar-testing-harness</artifactId>
     </dependency>
+    <dependency>
```
andrea-guarino-sonarsource, line 64: why do we need this dependency?

andrea-guarino-sonarsource (approved):
LGTM, as already discussed, there are still some js test files that could be removed.

sonar-javascript-plugin/pom.xml
guillaume-dequenne-sonarsource, line 64: I see it's already in the parent pom but I don't understand why it's needed here?

guillaume-dequenne-sonarsource (approved):
LGTM, left minor comments

sonar-javascript-plugin/pom.xml
saberduck: indeed, it seems it can be removed

saberduck:
This change was reverted in #2417 because custom rules were deprecated in current LTS cycle of SonarQube. Java based frontend will be removed in SQ 9.x cycle

PR #2418, opened 2020-12-08
Rule S5443: Using publicly writable directories is security-sensitive

### LINKED ISSUE #2397: Add rule S5443 for JS/TS
## Rule S5443: Using publicly writable directories is security-sensitive

[S5443](https://jira.sonarsource.com/browse/RSPEC-5443), part of [MMF-2154](https://jira.sonarsource.com/browse/MMF-2154).

### Code Samples

https://github.com/SonarSource/security-expected-issues/tree/master/javascript/rules/hotspots/RSPEC-5443

### Sensitive folder path in string literal

Highlight literal string starting with:
* `/tmp`
* `/var/tmp`
* `/usr/tmp`
* `/dev/shm`
* `/dev/mqueue`
* `/run/lock`
* `/var/run/lock`
* `/Library/Caches`
* `/Users/Shared`
* `/private/tmp`
* `/private/var/tmp`
* `\Windows\Temp`

Matching is case insensitive.
For Windows pathes, ignore the volume or drive letter before the first backslash.

### Sensitive environment variables

Highlight following environment variable accessed through {process.env`:

* `TMP` 
* `TMPDIR`
* `TEMP` 
* `TEMPDIR`

guillaume-dequenne-sonarsource:
Fixes #2397


eslint-bridge/src/rules/publicly-writable-directories.ts
```diff
@@ -0,0 +1,84 @@
...
+        const { object, property } = memberExpression;
+        if (
+          property.type !== 'Identifier' ||
+          !SENSITIVE_ENV_VARIABLES.includes(property.name) ||
+          object.type !== 'MemberExpression'
+        ) {
+          return;
+        }
+        if (
+          object.property.type === 'Identifier' &&
```
saberduck, line 74: can we use `isCallToFQN` from utils instead?

saberduck (approved):
lgtm. One small comment

eslint-bridge/src/rules/publicly-writable-directories.ts
guillaume-dequenne-sonarsource: `isCallToFQN` takes a `CallExpression` whereas here we have a `MemberExpression`. The function could be generalized a bit though, do you think it should be?

saberduck: sorry, I didn't see this ealier. We can keep as is, but perhaps we could improve the method in the future

PR #2420, opened 2020-12-09
Fix Sonar Way Recommended profile which should include all rules from Sonar Way profile

saberduck:
I also added automated test to assert this condition.


guillaume-dequenne-sonarsource (approved):
LGTM

PR #2428, opened 2020-12-14
Use "no-incorrect-string-concat" for S3402 in TypeScript

### LINKED ISSUE #2408: Use "no-incorrect-string-concat" for S3402 in TypeScript
S3402 is currently relying on [restrict-plus-operands](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/restrict-plus-operands.md) rule from typescript-eslint.

We can use instead our own (less noisy) implementation `no-incorrect-string-concat` that is already used for JavaScript.

guillaume-dequenne-sonarsource:
Fixes #2408


andrea-guarino-sonarsource (approved):
LGTM

PR #2430, opened 2020-12-14
Add script to generate boilerplate code for new rules

andrea-guarino-sonarsource:


eslint-bridge/tools/resources/unitTest.template_ts
```diff
@@ -0,0 +1,46 @@
...
+          message: 'TODO: add message',
+          line: 1,
+          column: 1,
+          endLine: 1,
+          endColumn: 1,
+        },
+      ],
+    },
+  ],
+});
```
saberduck, line 46: missing newline

eslint-bridge/tools/resources/rule.template_java
```diff
@@ -0,0 +1,37 @@
+/*
```
saberduck, line 1: looks like CRLF is used instead of LF as eol

eslint-bridge/tools/resources/rule.template_ts
```diff
@@ -0,0 +1,46 @@
+/*
```
saberduck, line 1: CRLF is used instead of LF as eol

eslint-bridge/tools/resources/unitTest.template_ts
```diff
@@ -0,0 +1,46 @@
+/*
```
saberduck, line 1: CRLF is used instead of LF as eol

saberduck (changes requested):
looks good, however it seems `CRLF` is used instead of `LF` as eol character.

eslint-bridge/tools/resources/unitTest.template_ts
andrea-guarino-sonarsource: good catch, thanks!

saberduck (approved):
perfect, thanks!

PR #2432, opened 2020-12-15
Fix FP on S2424 with TypeScript Enum constants

### LINKED ISSUE #2415: Fix FP on S2424 with TypeScript Enum constants
The rule should ignore Enum constants that override JavaScript standard built-in objects.

yassin-kammoun-sonarsource:
Fixes #2415

### DIFF
--- a/eslint-bridge/src/rules/no-built-in-override.ts
+++ b/eslint-bridge/src/rules/no-built-in-override.ts
@@ -22,6 +22,7 @@
 import { globalsByLibraries } from '../utils/globals';
 import { Rule, Scope } from 'eslint';
 import * as estree from 'estree';
+import { TSESTree } from '@typescript-eslint/experimental-utils';
 
 export const rule: Rule.RuleModule = {
   create(context: Rule.RuleContext) {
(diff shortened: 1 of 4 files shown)


eslint-bridge/src/rules/no-built-in-override.ts
```diff
@@ -45,17 +46,24 @@ export const rule: Rule.RuleModule = {
       scope.childScopes.forEach(checkScope);
     }
 
+    function isTSEnumMemberId(node: estree.Identifier) {
+      const id = node as TSESTree.Identifier;
+      return id.parent?.type === 'TSEnumMember' && id.parent.id === id;
```
saberduck, line 51: why do we need `id.parent.id === id` ? it seems not to be covered by tests

saberduck (changes requested):

eslint-bridge/src/rules/no-built-in-override.ts
yassin-kammoun-sonarsource: Indeed. I thought that both `Err` and `Error` in `enum E { Err = Error++ }` had the same `TSEnumMember ` parent, but it's obviously not the case.

saberduck (approved):

PR #2438, opened 2020-12-16
Fix FP: S4507 should not raise on alert() and debugger;

### LINKED ISSUE #2169: Fix FP on S1121 when the sub-expression is part of a while condition
Reported by community: https://community.sonarsource.com/t/rule-1121-assignments-in-sub-expressions-should-exclude-while-conditions/30413

S1121 should not raise if the assignment is not the controlling expression of a while condition, such as in the following case:
```
while ((line = reader.readLine()) !== null) {
  doSomething();
}
```
Once done, RSPEC should be updated to clearly mention this exception (as done in e.g PHP: https://jira.sonarsource.com/browse/RSPEC-2017).

guillaume-dequenne-sonarsource:
Fixes #2169


yassin-kammoun-sonarsource (approved):
LGTM!

PR #2449, opened 2020-12-30
Add support for timeout property

### LINKED ISSUE #1960: Timeout after analysing file for 60 seconds
I'm having an issue after upgrade to SQ 8.2 + SonarJS 6.2 + SonarTS 2.1 that looks identical to #1835.

ES-Lint analysis seems to get killed by a 60 sec timeout. It says 1190 files to be analyzed, which frankly is not that huge amount of files and adding 8GB of memory to node doesn't help at all.

I checked and launched the server command that demonized the es-lint and it seems to be starting properly.

Also tried removing `node_modules`  folder with no success.

Here's the full debug log:
[debug.log](https://github.com/SonarSource/SonarJS/files/4281959/debug.log)

Help would be greatly appreciated.

ricmatsui:
Fixes #1960

### DIFF
--- a/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
+++ b/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
@@ -54,6 +54,7 @@ public class EslintBridgeServerImpl implements EslintBridgeServer {
   private static final Profiler PROFILER = Profiler.createIfDebug(LOG);
 
   private static final int DEFAULT_TIMEOUT_SECONDS = 60;
+  private static final String TIMEOUT_SECONDS_PROPERTY = "sonar.javascript.timeoutSeconds";
   // internal property to set "--max-old-space-size" for Node process running this server
   private static final String MAX_OLD_SPACE_SIZE_PROPERTY = "sonar.javascript.node.maxspace";
   private static final String ALLOW_TS_PARSER_JS_FILES = "sonar.javascript.allowTsParserJsFiles";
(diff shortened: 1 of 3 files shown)


saberduck:
Thank you for your contribution @ricmatsui. However, I am not sure that making this option configurable is the way we want to "fix" this issue. I am considering that we will simply raise the default value so we allow longer processing by default. There were also nice performance improvements in recent TypeScript versions, which should alleviate this problem.

Can you share what kind of hardware do you use to analyze the code and at which timeout setting is your analysis successful? I assume that it's falling on default value, is that correct?

ricmatsui:
Thanks @saberduck I looked at making it configurable because the timeout value has been increased a couple times already:

- [Sep 28, 2018](https://github.com/SonarSource/SonarJS/commit/3be04bc743e96c03ed705b027a33aad7549ff7cb#diff-195dd59c3cd8e3335c38551504e78e04429f2b187ddc007fad00a8f0e3f6c0fb): 5s -> 10s
- [Sep 16, 2019](https://github.com/SonarSource/SonarJS/commit/50356d950e9df9d7715c56cd7c1f59cc71abfa74#diff-4c5e2408dc5152b41a8915f21e54e326c4a1fdef4f6796da7e7659db630cf595): 10s -> 30s 
- [Sep 16, 2019](https://github.com/SonarSource/SonarJS/commit/e4318917b2b0fad0387f557c2b687f501272027d#diff-4c5e2408dc5152b41a8915f21e54e326c4a1fdef4f6796da7e7659db630cf595): 30s -> 60s

And as far as I know the time needed depends on a few factors that could vary per project such as hardware speed, codebase size and structure, and rule count. So similar to how the memory limit can be raised once the analysis requires it, the idea would be that the timeout could also be raised.

> Can you share what kind of hardware do you use to analyze the code and at which timeout setting is your analysis successful? I assume that it's falling on default value, is that correct?

That's correct, the analysis is run with a MacBook Pro 2019 on 140k lines JS and TS (3.9.7). With 60s timeout it fails, and with 300s timeout it succeeds.

lukeapage:
We have 700k loc and I believe the value for us needs to be around 200s.

unless your going to make it configurable it needs to be much higher.

saberduck:
I created the PR to increase the limit to 300s https://github.com/SonarSource/SonarJS/pull/2505 , I don't think we need to make this option configurable. I will close this PR for now, if you have any issues please describe them on https://community.sonarsource.com/

ricmatsui:
Alright thanks @saberduck

PR #2455, opened 2021-01-17
Remove dependency on sslr-squid-bridge

### LINKED ISSUE #2454: Remove dependency on sslr-squid-bridge because it's deprecated
Should be replaced by sonar-analyzer-commons

margarita-nedzelska-sonarsource:
Fixes #2454


saberduck (approved):

margarita-nedzelska-sonarsource:
@saberduck It looks like this update leads to removing old custom rules api approach, which has already been deprecated since 4.2.1  Is it ok, if we remove it?

saberduck:
As we discussed it would be better to avoid this partial change before LTS and drop everything after LTS is out and we start to work on SQ 9.x

PR #2478, opened 2021-02-08
Deprecate SonarWay Recommended quality profile

### LINKED ISSUE #2450: Deprecate SonarWay Recommended quality profile
We want to remove SonarWay Recommended profile, as a first step we should change the name of the profile to be `SonarWay Recommended (deprecated)`. 

When this profile is used we should log a warning `SonarWay Recommended profile is deprecated, use SonarWay profile instead`. This warning should be in the logs and UI.

The change should be applied to both JS and TS.

This change should be implemented after SonarWay profile is updated.

yassin-kammoun-sonarsource:
Fixes #2450


andrey-tyukin-sonarsource (approved):
LGTM.

Tried various combinations (running analyses with different profiles with old version of plugin, then switching to new one, re-runninng etc.), it all seems to work.

PR #2484, opened 2021-02-11
Sanitize TypeScript ESLint rules requiring type information

### LINKED ISSUE #2483: Sanitize TypeScript ESLint rules requiring type information
When we write our own ESLint-based rules requiring type information, we systematically check that the type checker is available from the parser services, if not, such rules generally do nothing for most of them. This behaviour is not true for TypeScript ESLint rules, which don't perform that check but rather throw an error while retrieving the parser services. We then need to replicate this sanitization so that TypeScript ESLint rules relying on type checking but missing type information don't raise runtime errors.

yassin-kammoun-sonarsource:
Fixes #2483 

--
To sanitize TypeScript ESLint rules, we basically decorate the rules; each decorated rule now checks on creation whether it requires type information and if type information is available. If both conditions are met, then the original behaviour of the rule is preserved, otherwise the rule does nothing.

### DIFF
--- a/eslint-bridge/typings/typescript-eslint/index.d.ts
+++ b/eslint-bridge/typings/typescript-eslint/index.d.ts
@@ -1,3 +1,4 @@
 declare module '@typescript-eslint/eslint-plugin' {
-  export const rules: any;
+  import { Rule } from "eslint";
+  export const rules: { [name: string]: Rule.RuleModule };
 }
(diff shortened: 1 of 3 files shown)


eslint-bridge/typings/typescript-eslint/index.d.ts
```diff
@@ -1,3 +1,4 @@
 declare module '@typescript-eslint/eslint-plugin' {
-  export const rules: any;
+  import { Rule } from "eslint";
+  export const rules: { [name: string]: Rule.RuleModule };
```
saberduck, line 3: 👍

saberduck (approved):

PR #2485, opened 2021-02-11
Fix symbol highlighting for TS enums

saberduck:
`Variable#identifiers` doesn't containt ASTNode for TS enums. This will result in trying to find location for undefined value. 

As per https://eslint.org/docs/developer-guide/scope-manager-interface#identifiers is planned to be deprecated. Replacement is `Definition#name` which is generic ASTNode (although typings in `.d.ts` are incorrect) and contains `Literal` node for TS enum.

This PR has a fix on two levels 
- we use `Definition#name` instead of `Variable#identifiers`, which works correctly for TS enums
- we defensively check if there is at least one AST node for each variable instance before creating highlighting


yassin-kammoun-sonarsource (approved):
LGTM.

saberduck:
I will merge this with failing coverage, as it is not possible to cover the branch.

PR #2486, opened 2021-02-11
Upgrade S2432 to use eslint rule (no-setter-return)

### LINKED ISSUE #2377: Upgrade S2432 to use eslint rule (no-setter-return)
When upgrading to eslint 6.7, we can use the rule [no-setter-return](https://eslint.org/docs/rules/no-setter-return) instead of our custom `no-return-in-setter` rule as it supports more use cases than our own.

yassin-kammoun-sonarsource:
Fixes #2377


andrey-tyukin-sonarsource (approved):
LGTM

PR #2487, opened 2021-02-11
Use @babel/eslint-parser instead of babel-eslint

### LINKED ISSUE #2255: Use `@babel/eslint-parser` dependency instead of `babel-eslint`
Community: https://community.sonarsource.com/t/sonarjs-fails-parsing-a-file-unexpected-character/32621

saberduck:
Fixes #2255 

Babel in latest version was completely modularized and `core` module doesn't transpile any additional syntax. I added presets to support Flow and JSX, which we supported before. In the future we could consider adding additional experimental ES features.


eslint-bridge/tests/parser.test.ts
```diff
@@ -27,14 +27,15 @@ import {
...
   parseExceptionCodeOf,
-} from '../src/parser';
-import * as babel from 'babel-eslint';
+  babelConfig,
+} from 'parser';
+import * as babel from '@babel/eslint-parser';
 import { SourceCode } from 'eslint';
-import { ParsingError } from '../src/analyzer';
+import { ParsingError } from 'analyzer';
 import visit from '../src/utils/visitor';
```
andrey-tyukin-sonarsource, line 35: just wondering whether `utils/visitor` might work here as well, but that really doesn't seem important.

andrey-tyukin-sonarsource (approved):
LGTM

PR #2495, opened 2021-02-15
Break down rule utils into smaller pieces

yassin-kammoun-sonarsource:
The rule utilities also known as `eslint-bridge/rule/utils` are broken down into the following pieces:

- `ancestor-finder`: utility functions to find the ancestor(s) of a node.
- `ast-shape`: `is<Node>`-like functions to test a node syntax.
- `module-resolving`: functions resolving module names from identifiers and call expressions.
- `node-extractors`: extractor functions to get references and children from a node.
- `secondary-locations`: message encoder for secondary locations.
- `type-checking`: mainly `is<Type>`-like functions to test a node semantic with TypeScript compiler.
- (`isRequiredParserServices` renamed to `parser-services`)

I am open to any suggestions for improving the breakdown as well as the namings of the new pieces. The breakdown seems good to me, but I think the namings could be better.


saberduck (commented):
nice work! I think it would make sense to group all the files in the utils module, so we get more manageable files and it's a bit less cluttered when imported

saberduck (approved):
nice work!

PR #2502, opened 2021-02-24
Log warning when not running on LTS or latest version of Node.js

### LINKED ISSUE #2496: Log warning when running on unsupported Node.JS version
It seems that because of the issue in transitive dependency (https://github.com/feross/queue-microtask/issues/13) we are no longer able to run on Node.JS v11. Odd releases of Node.JS are not supported, so we can't expect the issue to be fixed. Also very few projects are using odd Node.JS releases and we are not actively testing them.

We should log a warning whenever we run on unsupported version of Node.JS (as of today supported versions are v10, v12, v14 and v15). We should always support LTS and LATEST versions.

Docs should also be updated accordingly.

saberduck:
Fixes #2496

### DIFF
--- a/sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/NodeDeprecationWarningTest.java
+++ b/sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/NodeDeprecationWarningTest.java
@@ -36,6 +36,7 @@ import static org.assertj.core.api.Assertions.assertThat;
 public class NodeDeprecationWarningTest {
 
   private static final String MSG = "You are using Node.js version 8, which reached end-of-life. Support for this version will be dropped in future release, please upgrade Node.js to more recent version.";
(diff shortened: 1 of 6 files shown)


sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/NodeDeprecationWarningTest.java
```diff
@@ -36,6 +36,7 @@
 public class NodeDeprecationWarningTest {
 
   private static final String MSG = "You are using Node.js version 8, which reached end-of-life. Support for this version will be dropped in future release, please upgrade Node.js to more recent version.";
+  private static final String UNSUPPORTED_MSG = "Node.js version 11 is not supported, you might experience issues. Please use supported version of Node.js [10, 12, 14, 15]";
```
andrey-tyukin-sonarsource, line 39: How about `"Please use a supported version of Node"` with `a`? Google-doc also insists on an `a`.

andrey-tyukin-sonarsource (approved):
LGTM

Proposed a tiny stylistic change in the formulation of a log message, though.

PR #2503, opened 2021-02-24
Don't throw exception when missing tsconfig

### LINKED ISSUE #2498: NodeCommandException should only be used when Node fails
We started to use `NodeCommandException` for some issues which are not fatal. This breaks contract with SonarLint and causes UX issues. 

> Reusing this exception for other erroneous cases, like tsconfig.json not being found, resulted in a significant degradation of the user experience: instead of an exception being thrown at most once, we get it at each analysis - so each keystroke in the worst case scenario.

More info https://discuss.sonarsource.com/t/degraded-ux-in-sonarlint-due-to-reuse-of-nodecommandexception/6998

saberduck:
Fixes #2498

### DIFF
--- a/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/TypeScriptSensor.java
+++ b/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/TypeScriptSensor.java
@@ -48,7 +48,6 @@ import org.sonar.plugins.javascript.TypeScriptLanguage;
 import org.sonar.plugins.javascript.eslint.EslintBridgeServer.AnalysisRequest;
 import org.sonar.plugins.javascript.eslint.EslintBridgeServer.AnalysisResponse;
 import org.sonarsource.analyzer.commons.ProgressReport;
-import org.sonarsource.nodejs.NodeCommandException;
 
 import static java.util.Collections.singletonList;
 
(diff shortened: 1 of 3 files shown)


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/TypeScriptSensor.java
```diff
@@ -108,7 +107,13 @@ void analyzeFiles() throws IOException, InterruptedException {
     ProgressReport progressReport = new ProgressReport("Progress of TypeScript analysis", TimeUnit.SECONDS.toMillis(10));
     List<InputFile> inputFiles = getInputFiles();
     eslintBridgeServer.initLinter(rules, environments, globals);
-    List<String> tsConfigs = tsConfigs();
+    List<String> tsConfigs = new TsConfigProvider(tempFolder).tsconfigs(context);
+    if (tsConfigs.isEmpty()) {
+      // This can happen in SonarLint context where we are not able to create temporary file for generated tsconfig.json
+      // note that JS files work even without generated tsconfig, perhaps we could do same for TS
```
yassin-kammoun-sonarsource, line 113: I think we can remove this comment here (i.e. the part with `note that...`) and create a ticket instead to materialise this investigating task.

yassin-kammoun-sonarsource (approved):

PR #2505, opened 2021-02-24
Increase default eslint-bridge timeout to 5min

### LINKED ISSUE #2504: Increase default timeout for eslint-bridge
There were various reports that increasing timeout for eslint-bridge helps to pass the analysis in certain cases.

In longer term we should try to understand why more time is needed for some projects.

saberduck:
Fixes #2504


andrey-tyukin-sonarsource (approved):
LGTM

PR #2529, opened 2021-03-29
Update S5527: add a message on secondary locations

### LINKED ISSUE #2295: Update S5527: add a message on secondary locations
[RSPEC-5527](https://jira.sonarsource.com/browse/RSPEC-5527) creates a secondary location on `rejectUnauthorized: false`. This secondary location can be a bit far from the primary location and developers will expect to navigate to them in SonarLint. Thus we need to add a message to these secondary location.

This should be the case for both JS and TS.

The secondary location message is already provided in the RSPEC.

yassin-kammoun-sonarsource:
Fixes #2295

### DIFF
--- a/eslint-bridge/src/rules/unverified-hostname.ts
+++ b/eslint-bridge/src/rules/unverified-hostname.ts
@@ -20,7 +20,6 @@
 // https://jira.sonarsource.com/browse/RSPEC-5667
 
 import { Rule } from 'eslint';
-import { toEncodedMessage } from 'eslint-plugin-sonarjs/lib/utils/locations';
 import * as estree from 'estree';
 import {
   isCallToFQN,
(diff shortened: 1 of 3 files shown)


eslint-bridge/src/rules/unverified-hostname.ts
```diff
@@ -50,13 +51,15 @@ export const rule: Rule.RuleModule = {
       }
       const sensitiveArgument = callExpression.arguments[sensitiveArgumentIndex];
       const secondaryLocations: estree.Node[] = [];
+      const secondaryMessages: (string | undefined)[] = [];
```
saberduck, line 54: I wonder if using `null` is more appropriate

saberduck (approved):
maybe consider using `null` instead of `undefined`

eslint-bridge/src/rules/unverified-hostname.ts
yassin-kammoun-sonarsource: It may be more appropriate, but I think I will keep that as it is because the API used for issue locations from both eslint-bridge and eslint-plugin-sonarjs defines issue location messages as something that can be undefined rather than nullable.

PR #2533, opened 2021-03-29
Update S3513: add a message on secondary locations

### LINKED ISSUE #2289: Update S3513: add a message on secondary locations
[RSPEC-3513](https://jira.sonarsource.com/browse/RSPEC-3513) creates secondary locations on every reference to `arguments`. These secondary locations can be quite far from the primary location and developers will expect to navigate to them in SonarLint. Thus we need to add a message to these secondary location.

This is the case for both JS and TS.

The secondary location message is already provided in the RSPEC.

yassin-kammoun-sonarsource:
Fixes #2289

### DIFF
--- a/eslint-bridge/src/rules/arguments-usage.ts
+++ b/eslint-bridge/src/rules/arguments-usage.ts
@@ -73,7 +73,7 @@ function checkArgumentsVariableWithoutDefinition(
       message: toEncodedMessage(
         MESSAGE,
         secondaryLocations,
-        secondaryLocations.map(_ => SECONDARY_MESSAGE),
+        Array(secondaryLocations.length).fill(SECONDARY_MESSAGE),
       ),
     });
   }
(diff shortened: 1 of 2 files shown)


eslint-bridge/src/rules/arguments-usage.ts
```diff
@@ -69,7 +70,11 @@ function checkArgumentsVariableWithoutDefinition(
     const secondaryLocations = references.slice(1).map(ref => ref.identifier) as TSESTree.Node[];
     context.report({
       node: firstReference.identifier,
-      message: toEncodedMessage(message, secondaryLocations),
+      message: toEncodedMessage(
+        MESSAGE,
+        secondaryLocations,
+        secondaryLocations.map(_ => SECONDARY_MESSAGE),
```
andrea-guarino-sonarsource, line 76: _minor_: I don't like too much using `map` here. I find it a bit unclear as you're ignoring the parameter.
I would prefer the approach you took in other PRs:
```javascript
Array(secondaryLocations.lenght).fill(SECONDARY_MESSAGE)
```

andrea-guarino-sonarsource (approved):
LGTM

PR #2534, opened 2021-03-29
Generate static file with rule metadata for SonarLint

### LINKED ISSUE #1982: Provide json file with mapping between SQ rule key and eslint rule key
SonarLint in VS will use eslint-bridge directly wihout Java to launch analysis. To be able to run correct rules they need mapping between SQ rule key (`Sxxxx`) and eslint rule key `no-alert`. This mapping is provided implicitly via annotations on rule classes. To make it easier for SonarLint to use it we will create simple json file during the build extracting this mapping and package it inside the sonarjs jar artifact. We should also provide mapping for legacy keys.

Precise location and structure of json file is TBD

### LINKED ISSUE #1983: Provide json file with rule parameters metadata
SonarLint in VS will not use Java api's to extract rule metadata from sonarjs plugin. Currently we already have json file with metadata for every rule in https://github.com/SonarSource/SonarJS/tree/master/javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript . This metadata doesn't include information about rule parameters, because these are provided with `@RuleParameter` annotation on the class.

Goal of this ticket is to create additional static json file with the metadata for rules. Besides description, name, .... we also need to include any default values which are hardcoded in `EslintBasedRule#configuration` method in every rule.

saberduck:
Fixes #1983 
Fixes #1982

### DIFF
--- a/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/utils/RulesMetadataForSonarLint.java
+++ b/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/utils/RulesMetadataForSonarLint.java
@@ -56,12 +56,12 @@ import static org.sonar.plugins.javascript.rules.JavaScriptRulesDefinition.METAD
  *  <li>rule parameter data from annotations on check classes</li>
  *  <li>eslint key</li>
  * </ul>
- *
+ * <p>
  * The file is packaged inside the jar artifact during the build
  */
 public class RulesMetadataForSonarLint {
 
-  private List<Rule> rules = new ArrayList<>();
+  private final List<Rule> rules = new ArrayList<>();
 
   RulesMetadataForSonarLint(String repositoryKey, List<Class<? extends JavaScriptCheck>> checks) {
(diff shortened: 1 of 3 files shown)


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/utils/RulesMetadataForSonarLint.java
```diff
@@ -0,0 +1,164 @@
...
+
+  void save(Path path) throws IOException {
+    Gson gson = new GsonBuilder().setPrettyPrinting().create();
+    String json = gson.toJson(rules);
+    Files.write(path, json.getBytes(StandardCharsets.UTF_8));
+  }
+
+  public static void main(String[] args) {
+    if (args.length == 0) {
+      throw new IllegalStateException("missig path to save metadata");
```
andrea-guarino-sonarsource, line 123: typo on `missig`

andrea-guarino-sonarsource (approved):
LGTM

PR #2537, opened 2021-03-30
Cherry-pick S3735 from Sonar Way Recommended into Sonar Way

### LINKED ISSUE #2526: Cherry-pick S3735 from Sonar Way Recommended into Sonar Way
- [ ] Add exception to [immediately invoked-function expressions](https://docs.google.com/spreadsheets/d/1yye3xZ1yfIjN9_LrqngXWdEkCjnRPwYMf-Nkx9kfa5w/edit#gid=0&range=P47)
- [ ] Document added exception in [RSPEC-3735](https://jira.sonarsource.com/browse/RSPEC-3735)

yassin-kammoun-sonarsource:
Fixes #2526


eslint-bridge/src/rules/void-use.ts
```diff
@@ -36,6 +36,17 @@ export const rule: Rule.RuleModule = {
       });
     }
 
+    function isVoid0(expr: estree.UnaryExpression) {
+      return expr.argument.type === 'Literal' && 0 === expr.argument.value;
+    }
+
+    function isIIFE(expr: estree.UnaryExpression) {
```
andrea-guarino-sonarsource, line 43: I don't fully understand why we want to exclude IIFE from this rule, is there an FP documented somewhere?

eslint-bridge/tests/rules/void-use.test.ts
```diff
@@ -41,23 +41,14 @@ const testCases = {
             void (0);
             `,
     },
-  ],
-  invalid: [
     {
       code: `
```
andrea-guarino-sonarsource, line 45: Should we also exclude arrow functions?

```
void (() => 42) () 
```

eslint-bridge/src/rules/void-use.ts
yassin-kammoun-sonarsource: Yes, it is documented in the [shared spreadsheet](https://docs.google.com/spreadsheets/d/1yye3xZ1yfIjN9_LrqngXWdEkCjnRPwYMf-Nkx9kfa5w/edit#gid=0&range=P47) prepared by Nicolas. All "cherry-picking" tickets were created based on this document.

eslint-bridge/tests/rules/void-use.test.ts
yassin-kammoun-sonarsource: That would makes sense.

eslint-bridge/src/rules/void-use.ts
andrea-guarino-sonarsource: Can we document this FP a bit better in #2526? In the future, I don't think it will be easy to find out this comes from a spreadsheet

andrea-guarino-sonarsource (approved):
LGTM

PR #2539, opened 2021-04-01
Cherry-pick S1121 from Sonar Way Recommended into Sonar Way

### LINKED ISSUE #2521: Cherry-pick S1121 from Sonar Way Recommended into Sonar Way
- [ ] Align with other analyzers implementations
  - lambdas
  - declarations
  - relational expressions
  - while conditions
  - ~parenthesised~ (ESLint AST does not include any parenthesised expression node)
- [ ] Add new exceptions
  - [do-while conditions](https://peach.sonarsource.com/issues?open=AXS_jxcGSsmidl6JktC6&projects=typescript_nodev12&resolved=false&rules=typescript%3AS1121)
  - [chained assignments](https://peach.sonarsource.com/project/issues?id=kibana_nodev12&issues=AXS_sR_QSsmidl6J1j-d&open=AXS_sR_QSsmidl6J1j-d)
  - [conditional assignments](https://peach.sonarsource.com/project/issues?id=vscode_nodev12&issues=AXS_qyBPSsmidl6Jxz86&open=AXS_qyBPSsmidl6Jxz86)
- [ ] Document new exceptions in [RSPEC-1121](https://jira.sonarsource.com/browse/RSPEC-2019)
- [ ] Document previous exceptions: while conditions, sequences, and lambdas

yassin-kammoun-sonarsource:
Fixes #2521 

**tl;dr**

RSPEC-1121 hasn't been updated yet because we need to agree first on the set of exceptions we actually want to consider for the rule. If we align with the other analyzers for the implementation of the rule, the list of exceptions becomes quite long. With this PR, we now exhaustively exclude the following patterns:

#### assignments at declaration-level

```javascript
let a = b = 0;
```

Newly introduced exception that already exists for some analyzers.

#### chained assignments

```javascript
a = b = c = 0;
```
Newly introduced exception because of some "FPs" observed on Peach. For me, it rather makes sense to allow chained assignments, but it's maybe worth discussing whether we really want such an exception.

#### relational assignments

```javascript
(a = 0) != b 
```

Newly introduced exception that already exists for some analyzers.

#### sequential assignments

```javascript
a = 0, b = 1, c = 2
```
Historical non-documented exception that was introduced before the migration to ESLint's frontend. We should probably keep it since it's quite common, in particular in the initialisation and update parts of for-loops.

#### assignments in "short" lambdas

```javascript
() => a = 0
```

Non-documented existing exception that is shared with some analyzers.

#### conditional assignments idiom

```javascript
a || (a = 0)
```

Newly introduced exception not only because of some "FPs" observed on Peach but also because it's a common idiom in JavaScript.

#### assignments in (do-)while conditions

```javascript
while (a = 0) {}
```

Non-documented existing exception that is shared with some analyzers. However, the exception for the do-while construct doesn't exist for the other analyzers. Still, applying the exception of the while construct to the do-while one is rather consistent here.


andrea-guarino-sonarsource (approved):
LGTM

PR #2540, opened 2021-04-01
Cherry-pick S1848 from Sonar Way Recommended into Sonar Way

### LINKED ISSUE #2524: Cherry-pick S1848 from Sonar Way Recommended into Sonar Way
- [ ] Add new exceptions
  - [try-catch construct](https://peach.sonarsource.com/project/issues?id=jest_nodev12&issues=AXS_kf9ESsmidl6JmDor&open=AXS_kf9ESsmidl6JmDor)
  - ~[test code](https://peach.sonarsource.com/project/issues?id=keystonejs&issues=AW4DW_ChpXOcTGKIbDyP&open=AW4DW_ChpXOcTGKIbDyP)~ Edit: extracted in #2544
- [ ] Document added exceptions in [RSPEC-1848](https://jira.sonarsource.com/browse/RSPEC-1522)

yassin-kammoun-sonarsource:
Fixes #2524


eslint-bridge/src/rules/constructor-for-side-effects.ts
```diff
@@ -45,6 +47,28 @@ export const rule: Rule.RuleModule = {
   },
 };
 
+function isTestCode(context: Rule.RuleContext) {
```
saberduck, line 50: I don't like to have this logic in the rule itself. There is `org.sonar.api.batch.fs.InputFile#type` which should determine whether the file is test file or not. We would need a new mechanism to have this information available in the context.

javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S1848.html
```diff
@@ -10,4 +10,13 @@ <h2>Compliant Solution</h2>
 <pre>
 var something = new MyConstructor();  // Compliant
 </pre>
+<h2>Exceptions</h2>
+<p>Immediately dropped new objects inside <code>try</code>-statements are ignored.</p>
```
saberduck, line 14: doesn't mention the test code part, is it somewhere else?

saberduck (changes requested):
condition about test code should come from Java API. If you don't want to implement that, I would prefer that we remove the condition for the time being.

eslint-bridge/src/rules/constructor-for-side-effects.ts
yassin-kammoun-sonarsource: As discussed, I will revert the workaround to detect test code. I created a ticket for making available file type information in the rule context. Also, I created another ticket to properly ignore issues raised by this rule on test code while using the aforementioned mechanism.

javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S1848.html
yassin-kammoun-sonarsource: Irrelevant for now.

saberduck (approved):

PR #2546, opened 2021-04-08
Fix highlighting on S3800 for FunctionLike declarations

### LINKED ISSUE #2480: Fix highlighting on S3800 for FunctionLike declarations
Currently we highlight whole node if  node is not `FunctionDeclaration`. This can be ugly for longer function expressions used as methods. 

2021-02-09 22_17_49-Refactor this function to always return the same type

yassin-kammoun-sonarsource:
Fixes #2480

### DIFF
--- a/eslint-bridge/src/rules/function-return-type.ts
+++ b/eslint-bridge/src/rules/function-return-type.ts
@@ -21,6 +21,8 @@
 
 import { TSESTree } from '@typescript-eslint/experimental-utils';
 import { Rule } from 'eslint';
+import { getMainFunctionTokenLocation } from 'eslint-plugin-sonarjs/lib/utils/locations';
+import { getParent } from 'eslint-plugin-sonarjs/lib/utils/nodes';
 import * as estree from 'estree';
 import * as ts from 'typescript';
 import { isRequiredParserServices, getTypeFromTreeNode, toEncodedMessage } from '../utils';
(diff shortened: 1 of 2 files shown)


eslint-bridge/src/rules/function-return-type.ts
```diff
@@ -87,7 +89,7 @@ export const rule: Rule.RuleModule = {
                   )}`,
               ),
             ),
-            node: node.type === 'FunctionDeclaration' && node.id ? node.id : node,
+            loc: getMainFunctionTokenLocation(node as estree.Function, getParent(context), context),
```
saberduck, line 92: I think you could use `TSESTree` types for `node` and access `.parent` field instead of this `getParent` call

yassin-kammoun-sonarsource: Type checking will fail because `getMainFunctionTokenLocation` from eslint-plugin-sonarjs expects ESTree node types, which are not compatible with TSESTree types. Unless we change the signature of the function and release a new version of the plugin, we won't be able to make the change.

eslint-bridge/src/rules/function-return-type.ts
```diff
@@ -21,6 +21,8 @@
 
 import { TSESTree } from '@typescript-eslint/experimental-utils';
 import { Rule } from 'eslint';
+import { getMainFunctionTokenLocation } from 'eslint-plugin-sonarjs/lib/utils/locations';
```
saberduck, line 24: just a remark, probably nothing to do, it would have been better to export the function in the package, so we don't need to import such a "deep" path. Not sure how to handle this in the long term

saberduck (approved):
small remark to use `TSESTree.Node.parent` instead of `getParent` , otherwise lgtm

PR #2549, opened 2021-04-08
Improve IndexOfCompareToPositiveNumberCheck: ignore strings

### LINKED ISSUE #590: Improve IndexOfCompareToPositiveNumberCheck: ignore strings
Most of the time using `indexOf` with string object is safe, as checked substring is for sure not a first part, e.g. `url.indexOf("?") > 0`
So we should limit this rule with just array usages.
Rule could be added to SW profile when ticket is fixed

RSPEC-2692

yassin-kammoun-sonarsource:
Fixes #590


saberduck (approved):

saberduck:
merged with missing coverage, because it's just about missing parsing services

PR #2551, opened 2021-04-08
Add S1226 no-parameter-reassignment to Sonar way profile

### LINKED ISSUE #2523: Cherry-pick S1226 from Sonar Way Recommended into Sonar Way
- [ ] [Add exception to parameters conditionally assigned](https://docs.google.com/spreadsheets/d/1yye3xZ1yfIjN9_LrqngXWdEkCjnRPwYMf-Nkx9kfa5w/edit#gid=0&range=P22)
- [ ] Document added exception in [RSPEC-1226](https://jira.sonarsource.com/browse/RSPEC-3818)

saberduck:
Fixes #2523 

I don't think this is causing too many FPs (I didn't find any clear FP on peach), and the proposed exception is too wide, exluding many valid issues.  I am more inclined to change the type of the issue to code smell


andrea-guarino-sonarsource (approved):
I agree about leaving the implementation as it is.
Regarding changing issue type, to avoid impacting other analyzers implementing this rule I think we can leave it as `bug` for the time being.
Let's maybe wait to get some feedback from users or from our PMs before changing it.

PR #2556, opened 2021-04-09
Fix FP on S3798 when require is used as rhs

### LINKED ISSUE #2366: Fix FP on S3798 when require is used as rhs
S3798 should not raise issues on cases such as:

`var grunt = require('grunt');`

yassin-kammoun-sonarsource:
Fixes #2366


eslint-bridge/src/rules/declarations-in-global-scope.ts
```diff
@@ -55,3 +56,12 @@ export const rule: Rule.RuleModule = {
 function findModuleScope(context: Rule.RuleContext) {
   return context.getSourceCode().scopeManager.scopes.find(s => s.type === 'module');
 }
+
+function isRequire(node: estree.Node | null | undefined) {
+  return (
+    node?.type === 'CallExpression' &&
+    node.arguments.length === 1 &&
+    node.callee.type === 'Identifier' &&
```
andrea-guarino-sonarsource, line 64: **Minor:** you could use `isIdentifier(node.callee, 'require')` from `utils`

andrea-guarino-sonarsource (approved):
LGTM

PR #2559, opened 2021-04-12
Fix code coverage reporting of conditions on non-executable lines

### LINKED ISSUE #2558: Fix code coverage reporting of conditions on non-executable lines
The community reported an issue about [uncovered lines not-shown for conditions in jsx](https://community.sonarsource.com/t/uncovered-lines-not-shown-for-conditions-in-jsx/32517). This problem actually generalises to conditions on non-executable lines regardless of jsx constructs.  For the particular case of SonarJS and LCOV coverage engine, SonarQube should highlight uncovered conditions of branches (`BRDA`) regardless of whether those branches belong to a particular executable line (`DA`) or not. Otherwise, users will have a hard time identifying at what line a condition is partially covered, especially if the source code happens to be big.

The culprit of the aforementioned issue is SonarJS's [LCOV parser](https://github.com/SonarSource/SonarJS/blob/8f7052b27970d037e86139335310ababbdec5cef/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/lcov/LCOVParser.java#L44) on code coverage reporting. For SonarQube to highlight partial coverage, it requires that [there is a hit and uncovered condition(s)](https://github.com/SonarSource/sonar-enterprise/blob/ded38c88c86242a3afdb562f48c8b04df831882a/server/sonar-web/src/main/js/components/SourceViewer/helpers/getCoverageStatus.ts#L21-L27) for a given line. Currently, a line is [marked as hit by the LCOV parser](https://github.com/SonarSource/SonarJS/blob/8f7052b27970d037e86139335310ababbdec5cef/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/lcov/LCOVParser.java#L206) only if it is an executable line. We should relax that requirement so that executable lines **and non-executable-lines that include conditions are highlighted on SonarQube when it comes to code coverage**. In other words, we should add synthetic line hits for `BDRA`s whose line number don't occur in any `DA`s.

SonarQube's code coverage reporting API says that [line hits should only apply to executable lines](https://github.com/SonarSource/sonarqube/blob/f22b3c80da086d7288669f3f267daf5f82d32020/sonar-plugin-api/src/main/java/org/sonar/api/batch/sensor/coverage/NewCoverage.java#L64). However, after a few experiments, it happens to be safe for non-executable ones as well. Also, it turns out that the JaCoCo coverage importer does exactly what we want to achieve here, i.e. [non-executable lines that include conditions are always marked as hit](https://github.com/SonarSource/sonar-jacoco/blob/ece00024105bbfa6c5e6cf8db9cba39c23b2c038/src/main/java/org/sonar/plugins/jacoco/ReportImporter.java#L44-L45). Discussions with SQ and SL folks led to the conclusion that such a change would not hurt. Beside having the UI red marker, the main impact would lie in the overall coverage since non-executable lines including conditions would then become executable ones.

yassin-kammoun-sonarsource:
Fixes #2558 

--
Using the code snippet provided by the community, here is what we get on SonarQube UI before/after the changes introduced by this pull request.

**Before**

before

There is an uncovered condition at line 5. The metrics do say that there are two conditions and that one of them is actually uncovered. However, there is no partial coverage highlighting on the line.

**After**

after

The metrics regarding condition coverage remain the same: there are two conditions and one of the two is uncovered. However, there is now one more executable line, which is line 5. Furthermore, the very same line is now highlighted with a red marker denoting a partially covered line. And finally, the overall code coverage changes since executable lines are considered in the computation.

### DIFF
--- a/its/plugin/projects/lcov-jsx/conditions-on-non-executable-lines.lcov
+++ b/its/plugin/projects/lcov-jsx/conditions-on-non-executable-lines.lcov
@@ -1,0 +1,15 @@
+TN:
+SF:./file.jsx
+FN:3,(anonymous_0)
+FNF:1
+FNH:1
+FNDA:1,(anonymous_0)
+DA:3,1
+DA:4,1
+LF:2
+LH:2
+BRDA:5,0,0,1
+BRDA:5,0,1,0
+BRF:2
+BRH:1
+end_of_record
--- a/its/plugin/projects/lcov-jsx/file.jsx
+++ b/its/plugin/projects/lcov-jsx/file.jsx
@@ -1,0 +1,11 @@
+import React from 'react'
+
+const Test = ({ condition }) => (
+  <div>
+    {condition && <span>Test1</span>}
+
+    <span>Test2</span>
+  </div>
+)
+
+export default Test
--- a/its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/CoverageTest.java
+++ b/its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/CoverageTest.java
@@ -193,10 +193,29 @@ public class CoverageTest {
       .contains("DEBUG: Problem during processing LCOV report: can't save DA data for line 12 of coverage report file")
       .contains("DEBUG: Problem during processing LCOV report: can't save BRDA data for line 18 of coverage report file");
 
-    assertThat(getMeasureAsInt(projectKey, "lines_to_cover")).isEqualTo(6);
+    assertThat(getMeasureAsInt(projectKey, "lines_to_cover")).isEqualTo(7);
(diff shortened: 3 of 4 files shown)


yassin-kammoun-sonarsource:
Using the code snippet provided by the community, here is what we get on SonarQube UI before and after the changes introduced by this pull request.

**Before**

before

There is an uncovered condition at line 5. The metrics do say that there are two conditions and that one of them is actually uncovered. However, there is no partial coverage highlighting on the line.

**After**

after

The metrics regarding condition coverage remain the same: there are two conditions and one of the two is uncovered. However, there is now one more executable line, which is line 5. Furthermore, the very same line is now highlighted with a red marker denoting a partially covered line. And finally, the overall code coverage changes since executable lines are considered in the computation.

saberduck (approved):
lgtm

PR #2565, opened 2021-04-13
Fix FP on S4335 when interfaces have computed properties

### LINKED ISSUE #2459: Fix FP on S4335 when interfaces have computed properties
Community: https://community.sonarsource.com/t/ts-interface-dynamic-property-names-cause-a-false-positive-report-of-rspec-4335/36832

--
After a quick investigation, the function helper `isEmptyInterface` considers interfaces with only computed properties as being empty. The implementation should use the declarations of the type symbol to check the emptiness of the interface, i.e. the absence of members.

saberduck:
Fixes #2459

### DIFF
--- a/eslint-bridge/src/rules/no-useless-intersection.ts
+++ b/eslint-bridge/src/rules/no-useless-intersection.ts
@@ -75,12 +75,20 @@ function isNullLike(tp: ts.Type): boolean {
 }
 
 function isEmptyInterface(tp: ts.Type): boolean {
-  return tp.getProperties().length === 0;
+  return (
+    tp.getProperties().length === 0 &&
+    !(tp as ts.InterfaceTypeWithDeclaredMembers).declaredStringIndexInfo &&
+    !(tp as ts.InterfaceTypeWithDeclaredMembers).declaredNumberIndexInfo
+  );
 }
 
-function isStandaloneInterface({ declarations }: ts.Symbol) {
+function isStandaloneInterface(typeSymbol: ts.Symbol) {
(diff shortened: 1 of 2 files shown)


eslint-bridge/src/rules/no-useless-intersection.ts
```diff
@@ -75,12 +75,20 @@ function isNullLike(tp: ts.Type): boolean {
...
+  );
 }
 
-function isStandaloneInterface({ declarations }: ts.Symbol) {
+function isStandaloneInterface(typeSymbol: ts.Symbol) {
   // there is no declarations for `{}`
   // otherwise check that none of declarations has a heritage clause (`extends X` or `implements X`)
+  if (!typeSymbol) {
+    return false;
+  }
```
yassin-kammoun-sonarsource, line 90: I am not sure I understand why there is this if-statement here. `typeSymbol` is supposed to be well-typed so it should never be undefined, right ? Is it an oversight or did you encounter some failures while running the ruling its that you are not able to reproduce in a unit test ?

saberduck: it's the latter, there were failures in ruling - and I was not able to extract them

yassin-kammoun-sonarsource: Okay, then can you maybe add a little comment about this before the return statement ?

yassin-kammoun-sonarsource (approved):

saberduck:
merged with red QG, because missing coverage is related to missing type information

PR #2572, opened 2021-04-16
Make available file type information in ESLint's rule context 

### LINKED ISSUE #2543: Make available file type information in ESLint's rule context
Existing rules or new rules might require avoiding raising issues on test code. For instance, [S1848](RSPEC-1848) raises a lot of false positives on test code. The idea here is to make available the file type provided by `org.sonar.api.batch.fs.InputFile#type` in the rule context that is injected on each rule during creation. This would require first to extend the analysis request sent to eslint-bridge with an additional field denoting the file type. Then, this field should somehow be passed to the rule context.

After a quick investigation, one easy way to do so could be extending the linter config on a file analysis with (fake) settings that include the file type. This should be done somewhere around these lines:

https://github.com/SonarSource/SonarJS/blob/6993729aa3e586231f93c1d0ea9a450a693c1e83/eslint-bridge/src/linter.ts#L149-L153

### LINKED ISSUE #2544: Improve S1848 by ignoring issues on test code
Depends on #2543

yassin-kammoun-sonarsource:
Fixes #2543
Fixes #2544

--
I changed my mind about updating the ruling ITs so that the analysis can distinguish main code from test code. The distinction is supposed to be based on the `sonar.sources` and `sonar.tests` properties of the scanner, whose values denoting directories must be disjoint.

Currently, `sonar.sources` is always set as the root of each project, which means that it includes the test directory, if any. Introducing the aforementioned creates to many problems, mainly due to the different code structure of each project.

To give one example, there are some projects, like [underscore.js](https://github.com/SonarSource/underscore/tree/3c6d93b254632c069bbd8921a760127d04aa560b), where the source files are at the root of the project while the test files are in a dedicate `test` directory. It would not be possible to analyse such source files because `sonar.sources=./` and `sonar.tests=./test` are not disjoint, and it doesn't seem possible to exclude the latter from the former.

Changing the ruling ITs is not worth the trouble for the sake of one rule that introduces a special behaviour for test code, at least for now. Therefore, I think it's enough to have one integration test that demonstrates that file type is indeed fed into the rule context.


saberduck (approved):
lgtm!

PR #2575, opened 2021-04-19
Drop Sonar Way Recommended creation

### LINKED ISSUE #2568: Drop Sonar Way Recommended creation
The "Sonar Way Recommended" quality profile can't be renamed in SQ or SC because it's a BUILT-IN one. If we change its name, a new QP will be created and the existing one will stay unchanged. We want to drop it so that new installations don't create it. This won't impact older installations since QPs will be kept after it's removed from the analyzer.

yassin-kammoun-sonarsource:
Fixes #2568


sonar-javascript-plugin/src/main/resources/static/documentation.md
```diff
@@ -30,9 +30,7 @@ Discover and update the JavaScript / TypeScript properties in: **<!-- sonarcloud
 
 ## Rule Profiles
 
-There are 2 built-in rule profiles for each JavaScript and TypeScript: `Sonar way` (default) and `Sonar way Recommended`.
-* `Sonar way` profile is activated by default. It defines a trimmed list of high-value/low-noise rules useful in almost any JS / TS development context.
-* `Sonar way Recommended` contains all rules from `Sonar way`, plus more rules that mandate high code readability and long-term project evolution.
+There is 1 built-in rule profile for each JavaScript and TypeScript: `Sonar way`. It defines a trimmed list of high-value/low-noise rules useful in almost any JS / TS development context.
```
yassin-kammoun-sonarsource, line 33: I am wondering whether we should just get rid of the whole section.

andrea-guarino-sonarsource: Indeed, I think we can remove it

andrea-guarino-sonarsource (approved):
LGTM. Manually tested: created a JS project with "Sonar Way Recommended" profile, then update the SonarJS plugin with this branch snapshot and no errors are raised.

PR #2581, opened 2021-04-20
Remove accidentally committed jar file

saberduck:


yassin-kammoun-sonarsource (approved):
Thank you for spotting that stupid mistake of mine.

PR #2589, opened 2021-04-21
Use babel parser to parse comments in no-commented-code rule

### LINKED ISSUE #2587: S125 no-commented-code rule should use babel to parse comments
Currently we use typescript parser to parse comments, which is unnecessarily slow. Using babel should be faster. We should also not attempt to parse comments as module files.

Also, we should not log any message when parsing fails.

saberduck:
Fixes #2587 

Note that babel parser will not parse TypeScript code which is not JS code, so there are some FN in TS ruling. I believe this is acceptable trade off.


yassin-kammoun-sonarsource (commented):
> Note that babel parser will not parse TypeScript code which is not JS code, so there are some FN in TS ruling. I believe this is acceptable trade off.

I didn't notice that in the previous PR, but I am not so sure about that because the [documentation](https://babeljs.io/docs/en/babel-parser) says the parser supports TypeScript. Did you try setting [babel-preset-typescript](https://babeljs.io/docs/en/babel-preset-typescript) in babel options ?

yassin-kammoun-sonarsource (approved):
As discussed, let's just forget about babel parsing TypeScript syntax, and merge the PR as it is.

PR #2590, opened 2021-04-22
Split Vue.js Single File Components Analysis for JS/TS

### LINKED ISSUE #2584: Associate .vue files to JavaScript language
When introducing the support of Vue.js Single File Components analysis for TypeScript, we decided to change the file association of `.vue` files to the TypeScript language. However, this change has actually created an issue with SonarLint because it has its own mapping. Users now experience parsing errors when linting JavaScript code in Vue.js Single File Components with SonarLint.

We want to revert the file association to how it originally was, i.e. `.vue` are associated to the JavaScript language.

### LINKED ISSUE #2585: Analyze JS code in Vue.js Single File Components as JavaScript
Code within Vue.js Single File Components (`.vue` files) can be written in JavaScript as well as in TypeScript. JS code must be analyzed as JavaScript by the plugin. The JavaScript sensor analyzes such files only if the contents of the `<script>` tag is JS code. The decision is based on the presence of `lang="ts"` in the tag. If it is present, then it is not JS code. Anything else means it is JS code. The detection mechanism is relies on regular expressions.

### LINKED ISSUE #2586: Analyze TS code in Vue.js Single File Components as TypeScript
Code within Vue.js Single File Components (`.vue` files) can be written in JavaScript as well as in TypeScript. TS code must be analyzed as TypeScript by the plugin. The TypeScript sensor analyzes such files only if the contents of the `<script>` tag is TS code. The decision is based on the presence of `lang="ts"` in the tag. If it is present, then it is TS code. The detection mechanism reuses the one described for #2585.

yassin-kammoun-sonarsource:
Fixes #2584
Fixes #2585
Fixes #2586


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/JavaScriptEslintBasedSensor.java
```diff
@@ -145,9 +146,7 @@ private void analyze(InputFile file, List<String> tsConfigs) throws IOException
   }
 
   static FilePredicate filePredicate(FileSystem fileSystem) {
-    return fileSystem.predicates().and(
-      fileSystem.predicates().hasType(Type.MAIN),
-      fileSystem.predicates().hasLanguage(JavaScriptLanguage.KEY));
+    return JavaScriptFilePredicate.getJavaScriptPredicate(fileSystem);
```
saberduck, line 149: I think you can remove this method and inline the invocation

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/TypeScriptSensor.java
```diff
@@ -96,9 +98,7 @@ public void describe(SensorDescriptor descriptor) {
   }
 
   static FilePredicate filePredicate(FileSystem fileSystem) {
-    return fileSystem.predicates().and(
-      fileSystem.predicates().hasType(Type.MAIN),
-      fileSystem.predicates().hasLanguage(TypeScriptLanguage.KEY));
+    return JavaScriptFilePredicate.getTypeScriptPredicate(fileSystem);
```
saberduck, line 101: same here, inline the invocation, and use static import

saberduck (approved):
nice work!

PR #2593, opened 2021-04-22
Enable S3523 for TypeScript

### LINKED ISSUE #2592: Enable S3523 for TypeScript

yassin-kammoun-sonarsource:
Fixes #2592


andrea-guarino-sonarsource (commented):
it looks like S3523 is deprecated: does it make sense to activate a deprecated rule for TS?

yassin-kammoun-sonarsource:
> it looks like S3523 is deprecated: does it make sense to activate a deprecated rule for TS?

Nice catch. Even though we found issues, I think it doesn't make sense to activate the rule for TS. Let's close the ticket and the PR.

PR #2595, opened 2021-04-23
S3801/no-inconstinent-return: avoid FP when functions returns never

### LINKED ISSUE #2579: FP S3801 should not raise when functions returns `never`
We should not raise issue when function returns `never` type

```
function doesRaiseIssueButShouldNot(a: any): boolean | never {
    if (a === 1) {
        return true;
    }
    throwError('False')
}
```

andrea-guarino-sonarsource:
Fixes #2579

### DIFF
--- a/eslint-bridge/tests/rules/no-inconsistent-returns.test.ts
+++ b/eslint-bridge/tests/rules/no-inconsistent-returns.test.ts
@@ -144,12 +144,24 @@ ruleTester.run(`Functions should use "return" consistently`, rule, {
     },
     {
       code: `
+      function withNeverType(a) {
+        if (a === 1) {
+            return true;
+        }
+        throw new Error('False')
+      }`,
+    },
+    {
+      code: `
+      function throwError(message: string): never {
+        throw new Error(message);
+      }
       function withNeverType(a): boolean | never {
         if (a === 1) {
             return true;
         }
         throwError('False')
-    }
+      }
     `,
     },
   ],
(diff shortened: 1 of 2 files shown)


eslint-bridge/tests/rules/no-inconsistent-returns.test.ts
```diff
@@ -142,6 +142,16 @@ ruleTester.run(`Functions should use "return" consistently`, rule, {
...
         var arrowWithExpressionBody = (p) => p ? true : false;`,
     },
+    {
+      code: `
+      function withNeverType(a): boolean | never {
+        if (a === 1) {
+            return true;
+        }
+        throwError('False')
+    }
```
yassin-kammoun-sonarsource, line 152: I think we should change a bit the code snippet and replace `throwError('False')` by `throw Error('False')`. Also, we could add another test case with the same (fixed) snippet without the type annotation for the return type.

yassin-kammoun-sonarsource (approved):
LGTM! I left a minor suggestion that is optional to address.

eslint-bridge/tests/rules/no-inconsistent-returns.test.ts
andrea-guarino-sonarsource: Let me clarify something. 
```typescript
function withNeverType(a) {
  if (a === 1) {
    return true;
  }
  throw new Error('False');
}
```

This snippet is not raising FP with the current implementation.

The FP comes when calling another function which in turn would throw an exception.


```typescript
function throwError(message: string): never {
  throw new Error(message);
}
function withNeverType(a): boolean | never {
 if (a === 1) {
     return true;
 }
 throwError('False')
}
```

The code snippet comes from this discuss post https://discuss.sonarsource.com/t/typescript-false-positive-on-rule-s3801-sq-8-8-sonarjs-7-3/7334

So I think I will add `throwError` declaration instead, to stay close to the reproducer.

> Also, we could add another test case with the same (fixed) snippet without the type annotation for the return type.

In this case my fix wouldn't work and I'm not sure if there is a simple way to prevent that. I can document it in the unit test anyway

yassin-kammoun-sonarsource: Ah, my bad. I wrongly thought that the rest of the implementation relied on type inference.

PR #2598, opened 2021-04-23
Add analysis option to not compute metrics for SonarLint

### LINKED ISSUE #2510: Add analysis option to only return issues (i.e. excluding metrics etc)
Metrics/highlighting/cpd data is calculated for each file that is analysed ([see here](https://github.com/SonarSource/SonarJS/blob/0dda9105bab520569708e230f4d2dffdca3cec74/eslint-bridge/src/analyzer.ts#L192)). However, this information is only used by the scanner (i.e. in batch mode). It's ignored by SonarLint.

From a performance point of view, it would better to avoid calculating and returning the metrics when called from SonarLint.

TODO: what about cognitive complexity? ESLintBridge treats this as a metric rather than an issue. SonarLint is capable of displaying complexity issues in the IDE.

saberduck:
Fixes #2510

### DIFF
--- a/eslint-bridge/bin/server
+++ b/eslint-bridge/bin/server
@@ -6,8 +6,10 @@
 *
 * port - port number on which server should listen
 * host - host address on which server should listen
-* additionalRules - ; or : delimited paths to additional rule bundles
 * workDir - working directory from SonarQube API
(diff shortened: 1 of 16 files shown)


yassin-kammoun-sonarsource (approved):
Very nice ! LGTM !

PR #2599, opened 2021-04-26
Improve S1874 by ignoring identifiers from object expression properties

yassin-kammoun-sonarsource:
Investigating the seemingly infinite loop problem during the analysis of the file [certifiedUserData.js](https://cirrus-ci.com/task/5176886268526592?logs=analyze#L83) led to identifying both a performance issue and a false positive issue with the rule [S1874](https://jira.sonarsource.com/browse/RSPEC-1874), which was recently enabled for JavaScript. The rule visits every identifier of an AST and checks whether the underlying symbol of each identifier is deprecated. It sometimes happens that the TypeScript compiler is involved in the process to retrieve type information.

The file certifiedUserData.js is a bit more than 4000-line long. It essentially consists of a rudimentary but gigantic object expression, in which there are as many identifiers as there are lines in the file. These identifiers come in the form of keys of object expression properties. The rule then takes an important amount of time for analysing this particular file because the logic relying on the TypeScript compiler is always executed, and this happens more than 4000 times.

It does not make sense to visit identifiers from keys of object expression properties. Those identifiers are actually declarations. The rule is supposed to raise issues on identifiers denoting occurrences of using deprecated symbols, not the declarations themselves. Otherwise, false positives could be reported if such properties were annotated with the `@deprecated` tag, which happens to be the case with the jshint project used in the ruling integration test.

This PR fixes the performance issue by actually fixing a false positive problem, that is, ignoring identifiers from keys of object expression properties.


saberduck (approved):
very nice finding!

PR #2615, opened 2021-05-03
Exclusion filter should exclude only JS/TS files and patterns should work only on relative path

### LINKED ISSUE #2612: Exclusion filter should exclude only JS/TS files and patterns should work only on relative path
In #2437 we added additional directories to the exclusion filter.

There are multiple problems with this filter
1. it doesn't work on relative paths, it will filter even if the pattern is matching directory above the project basedir
2. it excludes also files which don't have JS/TS as a language 
3. `build` pattern seems to be too non-specific to be excluded by default

saberduck:
Fixes #2612

### DIFF
--- a/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/JavaScriptExclusionsFileFilter.java
+++ b/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/JavaScriptExclusionsFileFilter.java
@@ -85,8 +85,7 @@ public class JavaScriptExclusionsFileFilter implements InputFileFilter {
       return false;
     }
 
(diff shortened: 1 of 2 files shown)


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/JavaScriptExclusionsFileFilter.java
```diff
@@ -38,7 +38,6 @@
     "**/node_modules/**",
     "**/bower_components/**",
     "**/dist/**",
-    "**/build/**",
```
vilchik-elena, line 41: update documentation

yassin-kammoun-sonarsource (approved):
lgtm

PR #2623, opened 2021-05-12
Add integration test for file filter behavior

saberduck:
This is to cover issue #2612 also with integration test

### DIFF
--- a/its/plugin/projects/file-filter/node_modules/project/main.js
+++ b/its/plugin/projects/file-filter/node_modules/project/main.js
@@ -1,0 +1,7 @@
+// issues in this file are detected even if parent path is node_modules
+// project base dir is "project" directory
+if (cond) {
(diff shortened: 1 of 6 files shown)


its/plugin/projects/file-filter/node_modules/project/main.js
```diff
@@ -0,0 +1,7 @@
+// issues in this file are detected even if parent path is node_modules
```
vilchik-elena, line 1: I would mention here something about considering relative path in the filter

vilchik-elena (commented):
I'm not sure what exactly we want to test with this test, seems like the fact that we rely on relative path. But I would suggest to modify default `exclusions` property to make sure that we rely on its value for exclusion.

vilchik-elena (approved):

PR #2644, opened 2021-06-01
Fix rule key in rule description for S4784

### LINKED ISSUE #2639: Fix rule link in rule description for S4784
https://github.com/SonarSource/SonarJS/blob/69397dc1443136b91608b358f6f6a250d5962833/javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S4784.html#L50

Should be `jssecurity`

vilchik-elena:
Fixes #2639

### DIFF
--- a/javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S4784.html
+++ b/javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S4784.html
@@ -47,5 +47,6 @@ str.split("(a+)+b"); // Sensitive
   <li> OWASP Regular expression Denial of Service - ReDoS </li>
 </ul>
 <h2>Deprecated</h2>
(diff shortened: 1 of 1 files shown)


javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S4784.html
```diff
@@ -47,5 +47,5 @@ <h2>See</h2>
   <li> OWASP Regular expression Denial of Service - ReDoS </li>
 </ul>
 <h2>Deprecated</h2>
-<p>This rule is deprecated; use {rule:javascript:S2631} instead.</p>
-
+<p>This rule is deprecated; use {rule:jssecurity:S2631} instead.</p>
```
saberduck, line 50: can you pls add a comment, that this change should be kept when generating automatically?

saberduck (approved):
see the comment

PR #2647, opened 2021-06-02
Update eslint and @typescript-eslint

### LINKED ISSUE #2634: Update eslint and @typescript-eslint
ESLint https://github.com/eslint/eslint/releases/tag/v7.27.0
Typescript-Eslint https://github.com/typescript-eslint/typescript-eslint/releases/tag/v4.25.0

Notable updates
- removal of lodash (should kill the noise around vulnerable deps)

saberduck:
Fixes #2634


its/ruling/src/test/expected/js/javascript-test-sources/javascript-S3798.json
```diff
@@ -1914,6 +1914,9 @@
 'javascript-test-sources:src/ecmascript6/snoode/src/models/stylesheet.es6.js':[
 4,
 ],
+'javascript-test-sources:src/ecmascript6/snoode/test/models/account.js':[
+12,
```
vilchik-elena, line 1918: did you understand why we have this new issue? Why we don't have other issues in this file?

its/ruling/src/test/expected/ts/rxjs/typescript-S1117.json
```diff
@@ -1,7 +1,4 @@
 {
-'rxjs:spec/Observable-spec.ts':[
-662,
```
vilchik-elena, line 3: and why this one is removed? Or why was it reported in the first place, seems like a FP

vilchik-elena (approved):

its/ruling/src/test/expected/js/javascript-test-sources/javascript-S3798.json
saberduck: we have it because we don't detect this is call to `require` which is normally excluded https://github.com/SonarSource/SonarJS/blob/a3f23224ab3258d689dd2f34f09e91d586cfde8c/eslint-bridge/src/rules/declarations-in-global-scope.ts#L63 

I am not sure why we didn't have it before, maybe scoping of the variable improved?

its/ruling/src/test/expected/ts/rxjs/typescript-S1117.json
saberduck: I think before it was reporting on type parameter `T` as being shadowed, and this was probably modified in the rule implementation (rule is not ours). Anyway looks like a change for the better to me

PR #2651, opened 2021-06-03
Log deprecation when Node.JS 10 or 15 is used

### LINKED ISSUE #2557: Log deprecation when Node.JS v10 is used
Node.JS 10 will reach EOL on 2021-04-30 (see https://nodejs.org/en/about/releases/ ) while we don't expect things to break, we should start logging deprecation message because our dependencies (ESLint, TS) might stop supporting it in a future release and we don't want to be blocked with the upgrade.

Also, there are improvements in memory management and performance in more recent versions of Node.JS

Node.JS v15 is also no longer supported, as it reached EOL on 2021-06-01

saberduck:
Fixes #2557


vilchik-elena (approved):
Could you mention in the ticket why v 15 is not supported anymore?

PR #2652, opened 2021-06-03
[CI] Use Node.JS v16.3.0 in CI

saberduck:


vilchik-elena (approved):

vilchik-elena:
@saberduck could you describe why moving to 16?

PR #2656, opened 2021-06-04
Ignore React functional components for cognitive complexity rule

### LINKED ISSUE #2238: Add exception to Cognitive Complexity for React functional components
Consider React functional component such as

```js
function ChatBoardComponent(props: ChatBoardComponentProps) {
    
    function handleShowAddRoomModal() {
        if(something){
            do something
    }

    function handleCloseAddRoomModal() {
        if(something){
            do something
    }

    function handleSomethingElse() {
        if(something){
            do something
    }
     
    if(something){} // we need some complexity-statement here to make top-level function considered by the rule
    return <p>Hello world</p>;
}
````

Currently, cognitive complexity of such component would be 7, while the equivalent component using class syntax will have only one. We should avoid adding complexity of nested functions to the outer function in such case.

See thread on community forum https://community.sonarsource.com/t/sonarqube-counts-incorrect-cognitive-complexity-for-react-functional-components-bug-no-functionality/30852

vilchik-elena:
Fixes #2238 

Actual fix is provided by https://github.com/SonarSource/eslint-plugin-sonarjs/pull/219 as the rule is provided by eslint-plugin-sonarjs

### DIFF
--- a/eslint-bridge/package.json
+++ b/eslint-bridge/package.json
@@ -52,7 +52,7 @@
     "bytes": "3.1.0",
     "eslint": "7.27.0",
     "eslint-plugin-chai-friendly": "0.6.0",
-    "eslint-plugin-sonarjs": "0.7.1-102",
+    "eslint-plugin-sonarjs": "0.8.0-125",
     "express": "4.17.1",
     "run-node": "2.0.0",
     "typescript": "4.3.2",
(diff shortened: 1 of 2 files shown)


saberduck (approved):

nathanhannig:
When can we expect these updated to be ported to SonarQube?

saberduck:
@nathanhannig this was shipped with SQ 9.0

user4302:
> @nathanhannig this was shipped with SQ 9.0

hi, im on v9.3.3,
and im still getting this issue, 
do i need to toggle something or change a setting somewhere for this to be enabled?

image

image

PR #2657, opened 2021-06-04
Update rules metadata

vilchik-elena:


javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S2092.html
```diff
@@ -1,10 +1,10 @@
 <p>When a cookie is protected with the <code>secure</code> attribute set to <em>true</em> it will not be send by the browser over an unencrypted HTTP
-request and thus cannot be observed by an unauthorized person during a man-in-the-middle attack. </p>
+request and thus cannot be observed by an unauthorized person during a man-in-the-middle attack.</p>
 <h2>Ask Yourself Whether</h2>
 <ul>
   <li> the cookie is for instance a <em>session-cookie</em> not designed to be sent over non-HTTPS communication. </li>
-  <li> it's not sure that the website contains <a href="https://developer.mozilla.org/fr/docs/Web/Security/Mixed_content">mixed content</a> or not (ie
-  HTTPS everywhere or not) </li>
+  <li> it’s not sure that the website contains <a href="https://developer.mozilla.org/en-US/docs/Web/Security/Mixed_content">mixed content</a> or not
```
vilchik-elena, line 6: In fact I think `’` fits better here
From wiki:

> In English, it is used for four purposes... marking of the omission of one or more letters, e.g. the contraction of "do not" to "don't".

saberduck: so, it should be apostrophe not backtick? I don't follow

saberduck (approved):

javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S2092.html
vilchik-elena: Apostrophe: ’
Backtick: `
Quote: '
We used quote, it's correctly changed to apostrophe

PR #2659, opened 2021-06-07
Remove unnecessary dependencies

saberduck:


nodejs-utils/src/main/java/org/sonarsource/nodejs/NodeCommand.java
```diff
@@ -179,7 +179,8 @@ public void destroyForcibly(Process process) {
 
     @Override
     public boolean isMac() {
-      return SystemUtils.IS_OS_MAC;
+      String osName = System.getProperty("os.name").toLowerCase(Locale.ROOT);
+      return osName.contains("mac");
```
christophe-zurn-sonarsource, line 183: Even if it should work on all recent os, `System.getProperty("os.name")` does not guarantee to return a non-null value, we should check for this case to be thorough.
Also, not sure if it can have a real impact, but the `SystemUtils.IS_OS_MAC` implementation relied on `startsWith` instead of `contains`, which is a bit more permissive ("Weird Linux Emac Os" -> would match) (https://commons.apache.org/proper/commons-lang/apidocs/src-html/org/apache/commons/lang3/SystemUtils.html#line.1899)

christophe-zurn-sonarsource (changes requested):

christophe-zurn-sonarsource (approved):
LGTM !

PR #2664, opened 2021-06-10
Rule S6299: Disabling Vue XSS protection is security sensitive

### LINKED ISSUE #2665: S6299: Disabling Vue.js built-in escaping is security-sensitive
New rule to implement

S6299 should raise on:
* [v-html directive](https://github.com/SonarSource/security-expected-issues/blob/master/javascript/nuxt.js/pages/RSPEC-5696/noncompliant1.vue#L9) when using Vue.js templates.
* [domProps.innerHTML property](https://github.com/SonarSource/security-expected-issues/blob/master/javascript/express.js/RSPEC-5696/vue/noncompliant.js#L54) when rendering function.
* [domPropsInnerHTML](https://github.com/SonarSource/security-expected-issues/blob/master/javascript/nuxt.js/pages/RSPEC-5696/noncompliant2.vue#L15) attribute when using JSX.

saberduck:
Fixes #2665


javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S6299.html
```diff
@@ -0,0 +1,83 @@
...
+  <li> is difficult to understand how it was constructed. </li>
+</ul>
+<p>There is a risk if you answered yes to any of those questions.</p>
+<h2>Recommended Secure Coding Practices</h2>
+<ul>
+  <li> Avoid injecting HTML content with <code>v-html</code> directive unless the content can be considered 100% safe, instead try to rely as much as
+  possible on built-in auto-escaping Vue.js features. </li>
+  <li> Take care when using the <code>v-bind:href</code> directive to set URLs which can contain malicious Javascript
+  (<code>javascript:onClick(...)</code>). </li>
+  <li> Event directives like <code>:onmouseover</code> are also prone to Javascript injection and should not be used with unsafe values. </li>
```
yassin-kammoun-sonarsource, line 19: As discussed in the slack channel, we agreed that the current implementation won't raise issues on event directives. Still, does it makes sense to keep that line in the RSPEC?

I would say yes since it's valuable information for the users, but I am just asking in case the RSPEC needs to be consistent with the implementation.

eslint-bridge/tests/rules/no-vue-bypass-sanitization.test.ts
```diff
@@ -0,0 +1,197 @@
...
+  parserOptions,
+});
+const ruleTester = new RuleTesterTs();
+
+const message = `Make sure bypassing Vue built-in sanitization is safe here.`;
+const testName = 'Disabling Vue.js built-in escaping is security-sensitive';
+
+ruleTesterForVue.run(testName, rule, {
+  valid: [
+    {
```
yassin-kammoun-sonarsource, line 41: I would add a test case that validates the fact that we don't make noise for `href` attributes that aren't directives:

```
{
  code: `
    <template>
      <a href="tainted">Click me</a>
    </template>
  `
}
```

yassin-kammoun-sonarsource (approved):
Very nice! The implementation looks good to me, and I particularly liked the part with the declarative selectors, which are really convenient here. I left a few comments that you are kindly invited to address. Also, since ruling doesn't raise any issues for this rule, we can maybe add a test file in [SonarJS/its/sources/file-for-rules](https://github.com/SonarSource/SonarJS/tree/master/its/sources/file-for-rules). FYI, Andrea introduced this folder for that purpose.

javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S6299.html
saberduck: I think we can keep as is, so there is the information for the user even if issue is not raised.

saberduck:
I didn't add the file for ruling, since current ruling is not raising issues for hotspots, perhaps something to improve in our ruling impl

Issue #2636, opened 2021-05-27
FP no-invalid-await on JS file in SonarLint


peterpoliwoda:
Reflecting on a forum issue spoken about here:
https://community.sonarsource.com/t/refactor-this-redundant-await-on-a-non-promise-is-not-correct/28800

My VSCode extension wrongly tells me that I don't need to await for a non-promise. Problem is, my function is an async function and by all means does return a promise as per ES6 standards.

```JavaScript
async function getAuthenticationToken() {
    try {
        const tokenUri = `${keycloakAuthServerUrl}realms/${keycloakRealm}/protocol/openid-connect/token`;

        const params = new URLSearchParams();
        params.append('client_id', DB_KEYCLOAK_CLIENT_ID);
        params.append('client_secret', DB_KEYCLOAK_CLIENT_SECRET);
        params.append('grant_type', MY_GRANT_TYPE);

        const { data } = await axios.post(
            tokenUri,
            params,
            {
                headers: {
                    'Content-Type': 'application/x-www-form-urlencoded'
                }
            }
        );
        return data.access_token;
    } catch (err) {
        logger.error('[getAuthenticationToken] Error getting authentication token', err);
        throw err;
    }
}

async function insertToMd4DatabaseApi(dbTable, data) {
    logger.info('[insertToMd4DatabaseApi] Insert data to Database API', { dbTable });
    logger.debug('[insertToMd4DatabaseApi] Data being saved', { dbTable, data });
    try {
        const authToken = await getAuthenticationToken(); // This part gets underlined by Sonar lint
        return axios.post(`/${dbTable}`, data, {
            baseURL: apiBaseUri,
            headers: {
                'some-auth-stuff': `Bearer ${authToken}`
            }
        });
    } catch (err) {
        logger.error('[insertToMd4DatabaseApi] Error saving data to Database API', err);
        throw err;
    }
}
```

This rule is causing some stir in my code and has caused some bugs. Please add support for async functions in JavaScript as this is a bug in linting at the moment.

vilchik-elena:
Hi @peterpoliwoda, 

It seems I rushed to add the ticket to the milestone as I can't reproduce it. Could you share self contained code snippet reproducing FP?

peterpoliwoda:
Hi @vilchik-elena I've updated the code above to reflect the two actual functions in case the code wasn't showing up for you. I updated to show that my underlined code is also wrapped within a try/catch block which might be the reason it's not handled. Please try now.

[...]

PR #2667, opened 2021-06-11
Fixes #2658 - allow await on transient properties

### LINKED ISSUE #2658: FP S4123 (no-invalid-await): TypeScript `Pick` type
Creating new issue as requested by @vilchik-elena in #2636 

The following TypeScript code snippet creates a “Refactor this redundant ‘await’ on a non-promise.” warning, which is incorrect.

```typescript
interface MyQuery<T> extends Pick<Promise<T>, keyof Promise<T>> {
  toQuery(): string;
}

export async function demo(query: MyQuery<string>) {
  const result = await query;
  console.log(result);
}
```

image

I see the error both in the SonarLint VS Code plugin (v2.0.0) and the current version of SonarCloud.

The [`Pick`](https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys) type is a Typescript "utility" type.  It creates a dynamic interface.  For example `Pick<Todo, "title" | "completed">` would be an interface with only the "title" and "completed" fields in the "Todo" interface.

The [`keyof`](https://www.typescriptlang.org/docs/handbook/2/keyof-types.html) operator returns all of the keys/members of a type.

As far as I can tell, these should be equivalent:

```typescript
interface MyQuery<T> extends Pick<Promise<T>, keyof Promise<T>> {
  toQuery(): string;
}

interface MyQuery<T> extends Promise<T> {
  toQuery(): string;
}
```

To be candid, I don't use either of these features myself.  This is just my understanding.  

My project uses [Knex](https://github.com/knex/knex), which is a popular SQL builder.  Knex uses this pattern extensively, and then creates many false positives. 

Here is the root declaration in Knex:

https://github.com/knex/knex/blob/dfbc76b915fff9f40e806ac2e0aafe8e4e5b6965/types/index.d.ts#L1742

```typescript
interface ChainableInterface<T = any> extends Pick<Promise<T>, keyof Promise<T> & ExposedPromiseKeys>, StringTagSupport {
  generateDdlCommands(): Promise<{ pre: string[], sql: string[], post: string[] }>;
  toQuery(): string;
```

_Originally posted by @codyebberson in https://github.com/SonarSource/SonarJS/issues/2636#issuecomment-854097749_

codyebberson:
Fixes #2658 - see issue for more details

### DIFF
--- a/eslint-bridge/src/rules/no-invalid-await.ts
+++ b/eslint-bridge/src/rules/no-invalid-await.ts
@@ -49,7 +49,10 @@ export const rule: Rule.RuleModule = {
 
 function hasThenMethod(type: ts.Type) {
   const thenProperty = type.getProperty('then');
-  return Boolean(thenProperty && thenProperty.flags & ts.SymbolFlags.Method);
+  return Boolean(
+    thenProperty &&
+      (thenProperty.flags & ts.SymbolFlags.Method || thenProperty.flags & ts.SymbolFlags.Transient),
+  );
 }
 
 function isAny(type: ts.Type) {
--- a/eslint-bridge/tests/rules/no-invalid-await.test.ts
+++ b/eslint-bridge/tests/rules/no-invalid-await.test.ts
@@ -110,6 +110,17 @@ ruleTester.run('await should only be used with promises.', rule, {
       }
       `,
     },
+    {
+      code: `
+      interface MyQuery<T> extends Pick<Promise<T>, keyof Promise<T>> {
+        toQuery(): string;
+      }
+      async function foo(query: MyQuery<string>) {
+        const result = await query;
+        console.log(result);
+      }
+      `,
+    },
   ],
   invalid: [
     {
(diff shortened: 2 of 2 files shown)


codyebberson:
From the build failure:

```
yarn config set registry "${ARTIFACTORY_URL}/api/npm/npm"
/tmp/scripts52fdfc072182654f163f5f0f9a621d72.sh: line 6: ARTIFACTORY_URL: unbound variable
```

This seems unrelated to the files changed.

spartanatreyu:
Is there something holding this back?

Not being able to use promises/async/await in JS/TS files is a big deal when all modern libraries are built upon them.

I can't trust the linter until this is fixed.

codyebberson:
@spartanatreyu - see https://github.com/SonarSource/SonarJS/issues/2636#issuecomment-878330083

My understanding is that they found some deeper issues with the tsconfig, so we're waiting for that.

spartanatreyu:
Ah, I can see it on the [8.3 milestone](https://github.com/SonarSource/SonarJS/milestone/42) now, awesome.

vilchik-elena:
Closing PR, it will be merged as part of https://github.com/SonarSource/SonarJS/pull/2755 (with a different approach applied)

PR #2669, opened 2021-06-14
Support syntax highlighting for Vue.js Single File Components templates

### LINKED ISSUE #2668: Support syntax highlighting for Vue.js Single File Components templates

yassin-kammoun-sonarsource:
Fixes #2668 

Syntax highlighting for the template section of Vue Single File Components takes inspiration from [sonar-html](https://github.com/SonarSource/sonar-html/blob/dbfc23fa354db91c784786c6a45baf054e612521/sonar-html-plugin/src/main/java/org/sonar/plugins/html/core/HtmlTokensVisitor.java#L73-L92) and how it implements such highlighting. 

Below is a table that summarises how Vue.js ESLint represents HTML in terms of AST nodes and how such nodes will be highlighted:

| HTML  |  Vue ESLint Node    | Highlighting |
|---|------|----|
| `<tag`  |   `HTMLTagOpen `   | `KEYWORD`  |
|  `>` |   `HTMLTagClose `   | `KEYWORD`  |
| `</tag`  | `HTMLEndTagOpen ` | `KEYWORD`  |
|  `/>` |   `HTMLSelfClosingTagClose `  | `KEYWORD`  |
|  `<!-- foo -->` |   `HTMLComment `   | `COMMENT`  |
|  `<!DOCTYPE ->` |   `HTMLBogusComment `  | `STRUCTURED_COMMENT `  |
|  attr=`"value"` |   `HTMLLiteral `   | `STRING`  |

> The value of a Vue.js directive-like attribute, e.g. `:href="myLink"`, is a JavaScript expression. Therefore, it is not highlighted as a string but as plain JavaScript code.

This PR also introduces symbol highlighting for the template part of Vue Single File Components. It only considers tags and highlights tags that are not self closing, i.e., tags that come in pair like `<foo></foo>`.

### DIFF
--- a/eslint-bridge/src/runner/symbol-highlighter.ts
+++ b/eslint-bridge/src/runner/symbol-highlighter.ts
@@ -108,11 +108,14 @@ export const rule: Rule.RuleModule = {
           } else if (token.type === 'HTMLSelfClosingTagClose') {
             openHtmlTagsStack.pop();
           } else if (token.type === 'HTMLEndTagOpen') {
-            const highlightedSymbol: HighlightedSymbol = {
-              declaration: location(openHtmlTagsStack.pop()!.loc),
-              references: [location(token.loc)],
-            };
-            result.push(highlightedSymbol);
+            const openHtmlTag = openHtmlTagsStack.pop();
+            if (openHtmlTag) {
+              const highlightedSymbol: HighlightedSymbol = {
+                declaration: location(openHtmlTag.loc),
+                references: [location(token.loc)],
+              };
+              result.push(highlightedSymbol);
+            }
           }
         });
 
--- a/eslint-bridge/src/runner/symbol-highlighter.ts
+++ b/eslint-bridge/src/runner/symbol-highlighter.ts
@@ -92,30 +92,34 @@ export const rule: Rule.RuleModule = {
         const openCurlyBracesStack: AST.Token[] = [];
         const openHtmlTagsStack: AST.Token[] = [];
         extractTokensAndComments(context.getSourceCode()).tokens.forEach(token => {
-          if (token.type === 'Punctuator') {
-            if (token.value === '{') {
-              openCurlyBracesStack.push(token);
-            }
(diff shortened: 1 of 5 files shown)


yassin-kammoun-sonarsource:
I forgot to mention that SonarQube Web API doesn't expose any endpoints from which one could fetch data about highlighting. As a consequence, it turned out to be impossible to write an integration test for that feature. Still, I ran locally an analysis on the [directus](https://github.com/directus/directus) project and the highlighting of `.vue` files was nice.

eslint-bridge/src/runner/symbol-highlighter.ts
```diff
@@ -100,6 +103,16 @@ export const rule: Rule.RuleModule = {
               };
               result.push(highlightedSymbol);
             }
+          } else if (token.type === 'HTMLTagOpen') {
+            openHtmlTagsStack.push(token);
+          } else if (token.type === 'HTMLSelfClosingTagClose') {
+            openHtmlTagsStack.pop();
+          } else if (token.type === 'HTMLEndTagOpen') {
+            const highlightedSymbol: HighlightedSymbol = {
+              declaration: location(openHtmlTagsStack.pop()!.loc),
```
vilchik-elena, line 112: it will fail if input is not consistent (which we can't guarantee):
```
<template>
</p>
</template>

```

eslint-bridge/tests/runner/symbol-highlighter.test.ts
```diff
@@ -134,6 +134,35 @@ it('should highlight TS enums', () => {
   });
 });
 
+it('should highlight Vue templates', () => {
+  const filePath = '/some/path/file.vue';
+  const fileContent = `
+  <template>
```
vilchik-elena, line 140: I would say just 2 levels of nested pairs would be enough

vilchik-elena (changes requested):

vilchik-elena (approved):

PR #2675, opened 2021-06-22
Update dependencies

saberduck:
- changes in rule S3683 is because of this change in ESLint https://github.com/eslint/eslint/pull/14238 they are FPs strictly speaking


vilchik-elena:
done in #2757

PR #2676, opened 2021-06-23
Update rules metadata

vilchik-elena:
Many formatting changes due to move to RSPECv2
Also removing CERT standard from everywhere as it's not applicable to JS/TS


javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S1067.html
```diff
@@ -1,4 +1,4 @@
 <p>The complexity of an expression is defined by the number of <code>&amp;&amp;</code>, <code>||</code> and <code>condition ? ifTrue : ifFalse</code>
 operators it contains.</p>
-<p>A single expression's complexity should not become too high to keep the code readable.</p>
+<p>A single expression’s complexity should not become too high to keep the code readable.</p>
```
saberduck, line 3: @arseniy-sonar is this change of apostrophe to backtick desired?

arseniy-sonar: Yes. It is the [default behavior](https://docs.asciidoctor.org/asciidoc/latest/text/quotation-marks-and-apostrophes/#apostrophe-syntax) of Asciidoctor.

saberduck (approved):

PR #2704, opened 2021-07-19
Add rule S6324: Regular expressions should not contain control characters

saberduck:


yassin-kammoun-sonarsource (approved):
Looks good to me. There is just a conflict to resolve, though.

PR #2719, opened 2021-07-29
Do not show logs when running jest tests during mvn build

vilchik-elena:
I suggest to hide logging when executing jest tests in MVN build. They are many and decrease readability of logs greatly


yassin-kammoun-sonarsource (approved):
Makes sense indeed.

saberduck (approved):

PR #2720, opened 2021-07-29
Support precise location for RegExp literals

### LINKED ISSUE #2712: Support precise issue location for regexp literals
We want to introduce the support of raising issues whose locations are inside regular expressions. Currently, we raise issues on the whole regular expression node instead of fragments of it. The locations of nodes parsed by regexpp are relative to the regular expression, be it a string literal or a regular expression literal.

See also #2721

saberduck:
Fixes #2712


eslint-bridge/tests/rules/no-empty-alternatives.test.ts
```diff
@@ -47,14 +47,29 @@ ruleTester.run('Alternation with empty alternatives', rule, {
           message: 'Remove this empty alternative.',
           line: 1,
           endLine: 1,
-          column: 1,
+          column: 2,
```
vilchik-elena, line 50: I would put issue on the pipe for this rule. And IMO it's even ok to have 2 issues on the same pipe for this case

eslint-bridge/tests/rules/no-empty-group.test.ts
```diff
@@ -47,7 +47,7 @@ ruleTester.run('Empty groups', rule, {
           message: 'Remove this empty group.',
           line: 1,
           endLine: 1,
-          column: 1,
+          column: 2,
           endColumn: 5,
```
vilchik-elena, line 51: why endColumn is still 5? should not it be 4?

eslint-bridge/tests/rules/single-character-alternation.test.ts
```diff
@@ -56,14 +56,22 @@ ruleTester.run('Single-character alternation', rule, {
           message: 'Replace this alternation with a character class.',
           line: 2,
           endLine: 2,
-          column: 22,
+          column: 23,
           endColumn: 29,
```
vilchik-elena, line 60: and here as well, why end column is at closing `/`?

eslint-bridge/src/rules/regex-rule-template.ts
```diff
@@ -18,24 +18,30 @@
...
  * Rule context for regex rules that also includes the original ESLint node
  * denoting the regular expression (string) literal.
  */
-export type RegexRuleContext = Rule.RuleContext & { node: estree.Node };
+export type RegexRuleContext = Rule.RuleContext & {
+  node: estree.Node;
+  reportRegExpNode: (descriptor: RegexReportDescriptor) => void;
+};
+
+type RegexReportDescriptor = { message: string; regexpNode?: regexpp.AST.Node; node: estree.Node };
```
vilchik-elena, line 44: why making `regexpNode` optional? IMO if `reportRegExpNode` is called, `regexpNode` should be available

eslint-bridge/src/rules/regex-rule-template.ts
```diff
@@ -58,7 +64,24 @@ export function createRegExpRule(
...
+      function reportRegExpNode(descriptor: RegexReportDescriptor) {
+        let loc: AST.SourceLocation;
+        const { node, regexpNode, message } = descriptor;
+        if (regexpNode) {
+          const source = context.getSourceCode();
+          const [start] = node.range!;
+          const [reStart, reEnd] = getRegexpRange(node, regexpNode);
+          loc = {
+            start: source.getLocFromIndex(start + reStart),
+            end: source.getLocFromIndex(start + reEnd + 1),
```
vilchik-elena, line 79: I think we don't need this `+1`

eslint-bridge/src/rules/regex-rule-template.ts
```diff
@@ -58,7 +64,24 @@ export function createRegExpRule(
...
+          };
+        } else {
+          loc = node.loc!;
```
vilchik-elena, line 82: I don't get why not simply executing this line when `node` is not regex literal

vilchik-elena (changes requested):

eslint-bridge/src/rules/regex-rule-template.ts
saberduck: indeed, this was only needed when reporting empty alternations, it doesn't make sense otherwise

eslint-bridge/src/rules/regex-rule-template.ts
```diff
@@ -58,11 +69,28 @@ export function createRegExpRule(
...
           return;
         }
-        regexpp.visitRegExpAST(regExpAST, handlers({ ...context, node }));
+        regexpp.visitRegExpAST(regExpAST, handlers({ ...context, node, reportRegExpNode }));
+      }
+
+      function reportRegExpNode(descriptor: RegexReportDescriptor) {
+        let loc: AST.SourceLocation;
+        const { node, regexpNode, message, offset = [0, 0] } = descriptor;
+        if (regexpNode && isRegexLiteral(node)) {
```
vilchik-elena, line 78: we don't need to check for `regexpNode`, right?

vilchik-elena (approved):

eslint-bridge/src/rules/regex-rule-template.ts
saberduck: indeed, I will fix this in the next PR, so I can merge this one already

PR #2722, opened 2021-07-29
Add rule S5860: Names of regular expressions named groups should be used

### LINKED ISSUE #2693: S5860: Names of regular expressions named groups should be used
https://github.com/SonarSource/rspec/pull/206

yassin-kammoun-sonarsource:
Fixes #2693


eslint-bridge/src/rules/unused-named-groups.ts
```diff
@@ -0,0 +1,356 @@
...
+    ],
+  },
+  create(context: Rule.RuleContext) {
+    const services = context.parserServices;
+    if (!isRequiredParserServices(services)) {
+      return {};
+    }
+    const intellisense = new RegexIntelliSense(services, context);
+    return {
+      'Literal[regex]:exit': (literal: estree.Literal) => {
```
vilchik-elena, line 55: we should update `regex-rule-template` with this `[regex]` flag

eslint-bridge/src/rules/unused-named-groups.ts
```diff
@@ -0,0 +1,356 @@
...
+        /* /regex/ */
+        intellisense.collectKnowledge(literal);
+      },
+      'NewExpression:exit': (newExpr: estree.NewExpression) => {
+        /* new RegExp(regex) */
+        intellisense.collectKnowledge(newExpr);
+      },
+      'CallExpression:exit': (callExpr: estree.CallExpression) => {
+        /* RegExp(regex) */
+        intellisense.collectKnowledge(callExpr);
```
vilchik-elena, line 65: we are not considering when string is used in `str.match`, `str.search`. I think you should somehow duplicate here the logic from `checkCallExpression` in `regex-rule-template`. That way you will support something like `str.match("(?<name>John)")`

If doing this I would also drop condition on line 316

eslint-bridge/src/rules/unused-named-groups.ts
```diff
@@ -0,0 +1,356 @@
...
+    }
+  }
+
+  findRegex(node: estree.Node): RegexKnowledge | undefined {
+    if (isRegexLiteral(node) || isRegExpConstructor(node)) {
+      return this.knowledge.find(regex => regex.node === node);
+    }
+    const variable = this.findVariable(node);
+    if (variable) {
+      for (const def of variable.defs) {
```
vilchik-elena, line 320: I was doing similar thing, and using `getUniqueWriteUsage` from utils for it

eslint-bridge/src/rules/unused-named-groups.ts
```diff
@@ -0,0 +1,356 @@
...
+    const regex = intellisense.findRegex(pattern);
+    if (regex) {
+      const references = extractReferences(substr);
+      const indexes = new Set<number>();
+      const names = new Set<string>();
+      references.forEach(ref =>
+        isNaN(Number(ref.value)) ? names.add(ref.value) : indexes.add(Number(ref.value)),
+      );
+      regex.groups.forEach(group => {
+        group.used ||= names.has(group.name) || indexes.has(group.index);
```
vilchik-elena, line 102: more readable if split into 2 `||=` statements

eslint-bridge/src/rules/unused-named-groups.ts
```diff
@@ -0,0 +1,356 @@
...
+    }
+  }
+}
+
+function checkIndexBasedGroupReference(
+  memberExpr: estree.MemberExpression,
+  intellisense: RegexIntelliSense,
+) {
+  const { object: matcher, property } = memberExpr;
+  const pattern = intellisense.resolve(matcher);
```
vilchik-elena, line 113: call it `regex` to be consistent with the rest of code

eslint-bridge/src/rules/unused-named-groups.ts
```diff
@@ -0,0 +1,356 @@
...
+  }
+}
+
+function checkNonExistingGroupReference(
+  memberExpr: estree.MemberExpression,
+  intellisense: RegexIntelliSense,
+) {
+  const { object: matcher, property, computed } = memberExpr;
+  const pattern = intellisense.resolve(matcher);
+  if (pattern && property.type === 'Identifier' && !computed) {
```
vilchik-elena, line 143: you already checked `computed` in visitor method

eslint-bridge/src/rules/unused-named-groups.ts
```diff
@@ -0,0 +1,356 @@
...
+  const ancestors = intellisense.context.getAncestors();
+  let parent = ancestors.pop();
+  if (
+    parent?.type === 'MemberExpression' &&
+    parent.property.type === 'Identifier' &&
+    parent.property.name === 'groups' &&
+    !parent.computed
+  ) {
+    parent = ancestors.pop();
+    if (
```
vilchik-elena, line 198: these lines are duplicated with `extractGroupNameFromMatchMatcher`

eslint-bridge/src/rules/unused-named-groups.ts
```diff
@@ -0,0 +1,356 @@
...
+
+function checkNonExistingGroupReference(
+  memberExpr: estree.MemberExpression,
+  intellisense: RegexIntelliSense,
+) {
+  const { object: matcher, property, computed } = memberExpr;
+  const pattern = intellisense.resolve(matcher);
+  if (pattern && property.type === 'Identifier' && !computed) {
+    let name: string | null = null;
+    switch (property.name) {
```
vilchik-elena, line 145: I would move this switch into one function `extractGroupName` and avoid having those 2 other `extract...` functions

eslint-bridge/src/rules/unused-named-groups.ts
```diff
@@ -0,0 +1,356 @@
...
+      const group = pattern.groups.find(grp => grp.name === name);
+      if (group) {
+        group.used = true;
+      } else {
+        // Temporary workaround
+        //
+        // Add secondary location on group nodes instead of the regex node (#2712, #2721)
+        intellisense.context.report({
+          message: toEncodedMessage(
+            `There is no group named '${name}' in the regular expression.`,
```
vilchik-elena, line 165: message should be action-oriented. E.g. `Make sure group named '${name}' exists in the regular expression.`

vilchik-elena (changes requested):

eslint-bridge/tests/rules/unused-named-groups.test.ts
```diff
@@ -0,0 +1,350 @@
...
+          const matched = 'str'.matchAll(unknownPattern);
+          if (matched) {
+            matched.groups.foo;
+          }
+        }
+      `,
+    },
+    {
+      code: `
+        const pattern = /(?<foo>\\w)/;
```
vilchik-elena, line 97: How about reporting here? There is name but index is used

vilchik-elena (changes requested):

eslint-bridge/tests/rules/unused-named-groups.test.ts
```diff
@@ -0,0 +1,350 @@
...
+    },
+    {
+      code: `
+        const result = 'str'.replace(unknownPattern, '$<foo>');
+      `,
+    },
+  ],
+  invalid: [
+    {
+      code: `/(?<foo>\\w)/`,
```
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: It's not weird since the regular expression includes a named capturing group that is never used. Regular expressions that are never matched are still considered by the rule. For example, we would also raise an issue for:

```typescript
const pattern = /(?<foo>\w)/; // Noncompliant: unused 'foo'
if ('str'.test(pattern)) {
  // ...
}
```

The named group is useless regardless of the pattern being tested. In a sense, the pattern-matcher logic with `RegExp.prototype.match()` just helps to kill the noise if a group is used. The same applies with `String.prototype.replace()`.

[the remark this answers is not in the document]
vilchik-elena: what is pattern is exported and used in another file?

eslint-bridge/src/rules/unused-named-groups.ts
yassin-kammoun-sonarsource: Are you sure you what to change the message to that ? The current one comes from the Java implementation, and the one you are suggesting sounds like a hotspot message, i.e. `Make sure that something is XXX`.

eslint-bridge/tests/rules/unused-named-groups.test.ts
yassin-kammoun-sonarsource: Nice catch! I did felt I forgot to implement something, and you found it for me.

eslint-bridge/src/rules/unused-named-groups.ts
vilchik-elena: ok then, let's be consistent with Java :)

eslint-bridge/tests/rules/unused-named-groups.test.ts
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: Indeed, this leaves the door open for many false-positives, which is a problem. To kill the noise, and as agreed, we will change the strategy: unused group issues will be only reported for patterns that are actually matched.

A matched pattern here refers to a pattern that is used with the following APIs:

- `RegExp.prototype.match()`
- `RegExp.prototype.matchAll()`
- `RegExp.prototype.exec()`
- `String.prototype.replace()`
- `String.prototype.replaceAll()`

eslint-bridge/src/rules/unused-named-groups.ts
yassin-kammoun-sonarsource: @vilchik-elena  Do you know why we need the if-statement at line 105 ?

https://github.com/SonarSource/SonarJS/blob/8c13aaf2d0d4769a888e18f44713bdf464ff6d18/eslint-bridge/src/rules/regex-rule-template.ts#L101-L111

I am asking because `isStringRegexMethodCall` checks whether the first argument is a string at line 130

https://github.com/SonarSource/SonarJS/blob/8c13aaf2d0d4769a888e18f44713bdf464ff6d18/eslint-bridge/src/rules/regex-rule-template.ts#L122-L132

If `isStringRegexMethodCall` returns true, then we know for a fact that the argument can't possibly be a regular expression literal or a regular expression object, right ?

And it seems, that we never take the then-branch at line 106 according to https://next.sonarqube.com/sonarqube/code?id=org.sonarsource.javascript%3Ajavascript&selected=org.sonarsource.javascript%3Ajavascript%3Aeslint-bridge%2Fsrc%2Frules%2Fregex-rule-template.ts&line=106

eslint-bridge/src/rules/unused-named-groups.ts
```diff
@@ -0,0 +1,367 @@
...
+      return this.bindings.get(variable) || null;
+    } else {
+      return null;
+    }
+  }
+
+  findRegex(node: estree.Node): RegexKnowledge | undefined {
+    const variable = this.findVariable(node);
+    if (variable) {
+      const value = getUniqueWriteUsage(this.context, variable.name);
```
vilchik-elena, line 342: there will be SOF for 
```
        let pattern;
        pattern = pattern;
        const matched = 'str'.matchAll(pattern);
```

I am not sure how to fix it though :(

vilchik-elena (changes requested):

eslint-bridge/src/rules/unused-named-groups.ts
```diff
@@ -108,12 +109,14 @@ function checkStringReplaceGroupReferences(
...
       const indexedGroups = regex.groups.filter(group => indexes.has(group.index));
       if (indexedGroups.length > 0) {
-        // Temporary workaround
-        //
-        // Remove names from message and add secondary location for each indexed group node
-        const indexedGroupNames = indexedGroups.map(group => group.name).join(', ');
         intellisense.context.report({
-          message: `Directly use '${indexedGroupNames}' instead of their group number.`,
+          message: toEncodedMessage(
+            `Directly use the names instead of their group number.`,
```
vilchik-elena, line 114: `Directly use the group names instead of their numbers.`

vilchik-elena (changes requested):

eslint-bridge/src/rules/unused-named-groups.ts
yassin-kammoun-sonarsource: Nice catch !

vilchik-elena (approved):

Issue #2616, opened 2021-05-04
EslintBridgeServer should handle unresponsive node process in SonarLint context


saberduck:
Sometimes `eslint-bridge` node process might get stuck. SonarLint might still continue to attempt to run the analysis and we are not able to recover from such failure gracefully. Current implementation will again deploy the bundle and mistakenly append the command line arguments when launching the process again.

When `startServerLazily` is invoked and `isAlive()` call returns `false` we should handle two possibilties
1. we are starting the server for the first time, so we should deploy and start (this is what's done currently)
2. there is already unresponsive node process, in which case we should do cleanup or fail

(as reported by @jblievremont)
Steps to reproduce:

Open IntelliJ (or WebStorm) with SonarLint installed
Open a TS file: the ESLint bridge server starts with command:
```
/home/user/.nvm/versions/node/v14.3.0/bin/node /home/user/.cache/JetBrains/IntelliJIdea2021.1/tmp/sonarlint/.sonartmp_13219161808921726661/eslint-bridge-bundle/package/bin/server 34317 127.0.0.1 /opt/Dev/SonarSource/sonarlint-vscode true true 
```
Wait for end of analysis
Kill the node process
Close/Reopen the TS file: the ESLint bridge tries to start and times out with command:
```
/home/user/.nvm/versions/node/v14.3.0/bin/node /home/user/.cache/JetBrains/IntelliJIdea2021.1/tmp/sonarlint/.sonartmp_13219161808921726661/eslint-bridge-bundle/package/bin/server 34317 127.0.0.1 /opt/Dev/SonarSource/sonarlint-vscode true true  44885 127.0.0.1 /opt/Dev/SonarSource/sonarlint-vscode true true
```

Note that all arguments after the path to the server script are duplicated.

PR #2723, opened 2021-07-30
EslintBridgeServer restarts always fail

bripkens:
# Why

the ESLint bridge server might need restarts as part of a single
scan. The scanner is trying to achieve this and currently always fails.
This is the case because the script arguments passed to the Node.js
process from the Java process are duplicated on subsequent starts. The
following excerpt from a scan log nicely shows this:

```
14:08:04.672 DEBUG: Starting Node.js process to start eslint-bridge server at port 39139
14:08:04.673 DEBUG: Launching command node --max-old-space-size=8192 --max-old-space-size=8192 /usr/src/.scannerwork/.sonartmp/eslint-bridge-bundle/package/bin/server 37023 127.0.0.1 /usr/src/.scannerwork true  39139 127.0.0.1 /usr/src/.scannerwork true
14:08:11.554 DEBUG: starting eslint-bridge server at port 37023
14:08:11.673 DEBUG: eslint-bridge server is running at port 37023
14:13:04.771 ERROR: Failed to start server (300s timeout)
org.sonarsource.nodejs.NodeCommandException: Failed to start server (300s timeout)
```

Note how the `Launching command node` log line lists all Node.js and script
arguments twice. This is the result of a missing reset of the `NodeCommandBuilder`.

# What

In order to keep my changes for this bugfix as small as possible, I opted
for an API to reset the `NodeCommandBuilder` state. This API is now used
before the `EslintBridgeServerImpl` constructs the Node.js command. This
then fixes `EslintBridgeServer` restarts.

# Future Work

`NodeCommandBuilder` is currently part of the dependency injection context.
This in itself is problematic and should be reconsidered. Maybe instead of
injecting a `NodeCommandBuilder`, instead inject `NodeCommand.ProcessWrapper`
and manually create a `NodeCommandBuilder` instance whenever necessary?

# References

 - [Bug Report](https://community.sonarsource.com/t/eslint-bridge-server-restarts-fail-due-to-missing-script-argument-resets/47292)

Fixes #47292

### DIFF
--- a/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
+++ b/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
@@ -56,7 +56,7 @@ public class EslintBridgeServerImpl implements EslintBridgeServer {
 
   private static final int DEFAULT_TIMEOUT_SECONDS = 5 * 60;
   // internal property to set "--max-old-space-size" for Node process running this server
-  private static final String MAX_OLD_SPACE_SIZE_PROPERTY = "sonar.javascript.node.maxspace";
+  static final String MAX_OLD_SPACE_SIZE_PROPERTY = "sonar.javascript.node.maxspace";
   private static final String ALLOW_TS_PARSER_JS_FILES = "sonar.javascript.allowTsParserJsFiles";
   private static final Gson GSON = new Gson();
   private static final int MIN_NODE_VERSION = 8;
@@ -152,6 +152,7 @@ public class EslintBridgeServerImpl implements EslintBridgeServer {
       LOG.info("Running in SonarLint context, metrics will not be computed.");
     }
     nodeCommandBuilder
+      .reset()
       .outputConsumer(message -> {
         if (message.startsWith("DEBUG")) {
           LOG.debug(message.substring(5).trim());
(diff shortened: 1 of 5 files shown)


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
```diff
@@ -56,7 +56,7 @@
 
   private static final int DEFAULT_TIMEOUT_SECONDS = 5 * 60;
   // internal property to set "--max-old-space-size" for Node process running this server
-  private static final String MAX_OLD_SPACE_SIZE_PROPERTY = "sonar.javascript.node.maxspace";
+  static final String MAX_OLD_SPACE_SIZE_PROPERTY = "sonar.javascript.node.maxspace";
```
bripkens, line 59: Visibility increased for testing purposes

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImplTest.java
```diff
@@ -256,6 +258,25 @@ public void test_lazy_start() throws Exception {
...
+  @Test
+  public void test_restarts() throws Exception {
+    context.settings().setProperty(MAX_OLD_SPACE_SIZE_PROPERTY, "4096");
+    eslintBridgeServer = createEslintBridgeServer("startServer.js");
+    eslintBridgeServer.startServerLazily(context);
+    assertThat(getNumberOfOccurrencesInDebugLogs("max-old-space-size")).isEqualTo(1);
+    eslintBridgeServer.stop();
+    logTester.clear();
+    eslintBridgeServer.startServerLazily(context);
+    assertThat(getNumberOfOccurrencesInDebugLogs("max-old-space-size")).isEqualTo(1);
```
bripkens, line 270: This assert is technically not even necessary. The server restart fails the unit test without the `.reset()` call

saberduck:
This is described in issue https://github.com/SonarSource/SonarJS/issues/2616

saberduck:
@bripkens thank you for this contribution, we will have a look at this in the next weeks when we plan to do hardening. We will likely implement it ourselves, but I will take some ideas from your PR

vilchik-elena:
Closing PR as the problem was fixed with https://github.com/SonarSource/SonarJS/pull/2748

PR #2724, opened 2021-07-30
Use precise issue location for string literals used as regexp

### LINKED ISSUE #2721: Support precise issue location in regexp strings
We want to support precise issue location inside strings which are used as regular expressions - ie `RegExp('pattern1|pattern2')`  we want to be able to raise issue inside the string passed to the constructor.

This is followup of #2712

saberduck:
Fixes #2721

### DIFF
--- a/eslint-bridge/src/rules/regex-rule-template.ts
+++ b/eslint-bridge/src/rules/regex-rule-template.ts
@@ -29,6 +29,7 @@ import {
   isRegExpConstructor,
   isRequiredParserServices,
   isString,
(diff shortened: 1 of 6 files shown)


yassin-kammoun-sonarsource (approved):
LGTM! I left a suggestion about having precise location unit tests through issue reporting. Feel free to ignore it if you feel that it's enough to have unit tests for `tokenizeString`.

saberduck:
See also https://github.com/SonarSource/rspec/pull/239

PR #2729, opened 2021-07-30
S6323: Add exception when empty alternation is used to make group optional

### LINKED ISSUE #2728: Add exception to S6323 about optional group

saberduck:
Fixes #2728


saberduck:
See also https://github.com/SonarSource/rspec/pull/239

eslint-bridge/src/rules/no-empty-alternatives.ts
```diff
@@ -40,6 +41,15 @@ export const rule: Rule.RuleModule = createRegExpRule(context => {
         });
       }
     }
+
+    function isLastEmptyInGroup(alt: regexpp.AST.Alternative) {
```
vilchik-elena, line 45: I am not fan of nested functions (it's level 3)

vilchik-elena (approved):

PR #2734, opened 2021-08-02
Fix error when computing precise issue location in regexp string literal

saberduck:


eslint-bridge/src/utils/utils-regex.ts
```diff
@@ -137,9 +138,21 @@ function getRegexpRange(node: estree.Node, regexpNode: regexpp.AST.Node): AST.Ra
...
+    if (regexpNode.start === regexpNode.end) {
+      if (regexpNode.start - 1 < tokens.length) {
+        return [
+          tokens[regexpNode.start - 1].range[0] + 1,
+          tokens[regexpNode.start - 1].range[0] + 1,
+        ];
+      } else {
+        return [last(tokens).range[1] + 1, last(tokens).range[1] + 1];
+      }
+    }
```
yassin-kammoun-sonarsource, line 150: We should document this corner case logic using the example of empty alternatives.

eslint-bridge/src/utils/utils-regex.ts
```diff
@@ -137,9 +138,21 @@ function getRegexpRange(node: estree.Node, regexpNode: regexpp.AST.Node): AST.Ra
...
+    const startToken = regexpNode.start - 1;
+    const start = tokens[startToken].range[0];
+    const endToken = Math.min(regexpNode.end - 2, tokens.length - 1);
+    const end = tokens[endToken].range[1];
+    return [start + 1, end + 1];
```
yassin-kammoun-sonarsource, line 155: We should document the general case because it's not easy to follow what's going on with these '+1, '-1', and '-2' without sketching on a piece of paper what's happening.

yassin-kammoun-sonarsource (approved):
LGTM! I think we should document a bit how `getRegexpRange` works because I am afraid it will become difficult to maintain to whoever needs to dive into it later on (this could be done in a different PR if merging is urgent).

PR #2741, opened 2021-08-05
FP S1128: ignore unused imported symbols denoting JSX factories and JSX fragment factories

### LINKED ISSUE #2701: FP S1128: ignore unused imported symbols denoting JSX factories and JSX fragment factories
JSX requires `h` function to be imported, however we report the import to be unused. 

We should detect if `jsxFactory` setting is present in `tsconfig` and ignore `h` function import.

See more details in the user report

https://community.sonarsource.com/t/false-positive-remove-this-unused-import-of-h-in-all-stenciljs-components/42367

yassin-kammoun-sonarsource:
Fixes #2701

### DIFF
--- a/eslint-bridge/src/rules/unused-import.ts
+++ b/eslint-bridge/src/rules/unused-import.ts
@@ -49,13 +49,19 @@ export const rule: Rule.RuleModule = {
       return variable.name === 'jsx' && isJsxPragmaSet;
     }
 
-    function getJsxFactory() {
+    function getJsxFactories() {
+      const factories = new Set<string>();
       const parserServices = context.parserServices;
       if (isRequiredParserServices(parserServices)) {
         const compilerOptions = parserServices.program.getCompilerOptions();
-        return compilerOptions.jsxFactory;
+        if (compilerOptions.jsxFactory) {
+          factories.add(compilerOptions.jsxFactory);
+        }
+        if (compilerOptions.jsxFragmentFactory) {
+          factories.add(compilerOptions.jsxFragmentFactory);
+        }
       }
-      return undefined;
+      return factories;
     }
 
     return {
@@ -83,7 +89,7 @@ export const rule: Rule.RuleModule = {
         );
       },
       'Program:exit': () => {
-        const jsxFactory = getJsxFactory();
+        const jsxFactories = getJsxFactories();
         const jsxIdentifiers = context
           .getSourceCode()
           .ast.tokens.filter(token => token.type === 'JSXIdentifier')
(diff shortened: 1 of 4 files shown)


eslint-bridge/tests/rules/unused-import.test.ts
```diff
@@ -258,3 +259,57 @@ ruleTesterTS.run('Unnecessary imports should be removed', rule, {
...
+});
+
+const filename = path.resolve(`${__dirname}/../fixtures/unused-import/sample.lint.tsx`);
+
+ruleTesterJsxFactory.run('Unused imports denoting jsx factory should be ignored', rule, {
+  valid: [
+    {
+      filename,
+      code: `
+        import { h } from 'jsxlib';
```
vilchik-elena, line 280: it does not matter from which lib it's imported, right? (in the docs it's `preact`)

vilchik-elena, line 280: let's add test to show that we report on `h` when jsxFactory is not defined (or we can use simply js project)

eslint-bridge/tests/rules/unused-import.test.ts
```diff
@@ -258,3 +259,57 @@ ruleTesterTS.run('Unnecessary imports should be removed', rule, {
...
+        export class Component {
+          render() {
+            return <div>Hello, world!</div>
```
vilchik-elena, line 283: I am not sure about JSX in the code here, on one side we imitate real code but then it's not clear from the test that we don't care about jsx presence. How about having both?)

eslint-bridge/src/rules/unused-import.ts
```diff
@@ -48,6 +49,15 @@ export const rule: Rule.RuleModule = {
       return variable.name === 'jsx' && isJsxPragmaSet;
     }
 
+    function getJsxFactory() {
+      const parserServices = context.parserServices;
+      if (isRequiredParserServices(parserServices)) {
+        const compilerOptions = parserServices.program.getCompilerOptions();
+        return compilerOptions.jsxFactory;
```
vilchik-elena, line 56: let's also support this property https://www.typescriptlang.org/tsconfig#jsxFragmentFactory (ticket and PR title to be updated then)

vilchik-elena (changes requested):

eslint-bridge/tests/rules/unused-import.test.ts
yassin-kammoun-sonarsource: Indeed, it doesn't matter. I will rename the lib to avoid the confusion.

vilchik-elena (approved):

PR #2746, opened 2021-08-09
Rely on full type information only

### LINKED ISSUE #2636: FP no-invalid-await on JS file in SonarLint

vilchik-elena:
Fixes #2636


eslint-bridge/package.json
```diff
@@ -52,7 +52,7 @@
     "bytes": "3.1.0",
     "eslint": "7.27.0",
     "eslint-plugin-chai-friendly": "0.6.0",
-    "eslint-plugin-sonarjs": "0.9.1",
+    "eslint-plugin-sonarjs": "0.10.0-290",
```
yassin-kammoun-sonarsource, line 55: Why don't we just release a new version of the plugin ?

vilchik-elena: I hope we will fix smth else in that version and release by the end of the week

yassin-kammoun-sonarsource (approved):

PR #2748, opened 2021-08-10
Do not try to restart Node process if it's not answering

### LINKED ISSUE #2616: EslintBridgeServer should handle unresponsive node process in SonarLint context

vilchik-elena:
Fixes #2616


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
```diff
@@ -38,6 +38,7 @@
 import okhttp3.Response;
 import org.sonar.api.SonarProduct;
 import org.sonar.api.batch.sensor.SensorContext;
+import org.sonar.api.internal.google.common.annotations.VisibleForTesting;
```
saberduck, line 41: I don't like importing shaded artifacts from sonar-api package. Imo just replace this annotation with the comment.

saberduck (approved):

PR #2754, opened 2021-08-11
FP S3723: should check presence of comma only on multiline contructions

### LINKED ISSUE #2689: FP S3723: should check presence of comma only on multiline contructions
When configuring eslint [`comma-dangle`](https://eslint.org/docs/rules/comma-dangle) rules (which is actual implementation of S3723) `always-multiline` should be set for all syntax types, otherwise default `never` is used for anything which is not array or object literal:

```
foo(a, b, c,); // FP is reported here
```

Reported in[ internal thread](https://discuss.sonarsource.com/t/function-with-rest-parameters-and-rules-detecting-presence-or-lack-of-trailing-coma/8148)

yassin-kammoun-sonarsource:
Fixes #2689


vilchik-elena:
@yassin-kammoun-sonarsource I realized that the rule description is only about objects and arrays. Should we mention all constructs reported by eslint (function args, imports, etc)? Or should we report only on objects and arrays may be? That seems possible with eslint options but I tend to think it's better to extend the description

vilchik-elena (changes requested):
See my comment on PR

PR #2755, opened 2021-08-11
Improve S4123 (no-invalid-await): allow Pick<Promise> type

### LINKED ISSUE #2658: FP S4123 (no-invalid-await): TypeScript `Pick` type

vilchik-elena:
Fixes #2658

### DIFF
--- a/eslint-bridge/src/rules/no-invalid-await.ts
+++ b/eslint-bridge/src/rules/no-invalid-await.ts
@@ -49,7 +49,10 @@ export const rule: Rule.RuleModule = {
 
 function hasThenMethod(type: ts.Type) {
   const thenProperty = type.getProperty('then');
-  return Boolean(thenProperty && thenProperty.flags & ts.SymbolFlags.Method);
+  return Boolean(
+    thenProperty &&
+      (thenProperty.flags & ts.SymbolFlags.Method || thenProperty.flags & ts.SymbolFlags.Transient),
+  );
 }
 
 function isAny(type: ts.Type) {
--- a/eslint-bridge/src/rules/no-invalid-await.ts
+++ b/eslint-bridge/src/rules/no-invalid-await.ts
@@ -49,14 +49,9 @@ export const rule: Rule.RuleModule = {
 
 function hasThenMethod(type: ts.Type) {
   const thenProperty = type.getProperty('then');
-  if (thenProperty && thenProperty.declarations && thenProperty.declarations?.length > 0) {
-    const declaration = thenProperty.declarations[0];
-    return (
-      declaration.kind === ts.SyntaxKind.MethodSignature ||
-      declaration.kind === ts.SyntaxKind.MethodDeclaration
-    );
-  }
-  return false;
+  return thenProperty?.declarations?.some(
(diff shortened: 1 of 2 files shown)


eslint-bridge/src/rules/no-invalid-await.ts
```diff
@@ -49,7 +49,14 @@ export const rule: Rule.RuleModule = {
...
   const thenProperty = type.getProperty('then');
-  return Boolean(thenProperty && thenProperty.flags & ts.SymbolFlags.Method);
+  if (thenProperty && thenProperty.declarations && thenProperty.declarations?.length > 0) {
+    const declaration = thenProperty.declarations[0];
+    return (
+      declaration.kind === ts.SyntaxKind.MethodSignature ||
+      declaration.kind === ts.SyntaxKind.MethodDeclaration
+    );
+  }
+  return false;
```
yassin-kammoun-sonarsource, line 59:
```suggestion
  return thenProperty?.declarations?.some(
    d => d.kind === ts.SyntaxKind.MethodSignature || d.kind === ts.SyntaxKind.MethodDeclaration,
  );
```

yassin-kammoun-sonarsource (approved):
LGTM. I left a minor suggestion. Also, I would rename the PR (and the ticket maybe) because it's more about fixing the detection of `Promise.prototype.then()` than allowing `Pick<Promise> type`.

PR #2768, opened 2021-08-16
Migrate to JUnit 5 API and SQ 8.9 API

saberduck:
Most of the changes are boring such as changing

- annotation `org.junit.Test` to `org.junit.jupiter.api.Test`
- using `@TempDir` for temporary folders
- using `LogTesterJUnit5` for logtester (this required update of SQ API)

Plugin QA tests required a bit of rework because JUnit 5 doesn't support `@TestSuite` annotation which was required to start the Orchestrator only once for all QA tests. Now this is implemented with `@Extension` logic from JUnit 5 as described in https://stackoverflow.com/questions/43282798/in-junit-5-how-to-run-code-before-all-tests . Advantage of the new way is that we don't need to explicitly list test in the test suite anymore.

Ruling tests are using `@ParameterizedTest` and `@MethodSource` as described in https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests-sources-MethodSource

### DIFF
--- a/its/plugin/tests/pom.xml
+++ b/its/plugin/tests/pom.xml
@@ -74,7 +74,7 @@
         <artifactId>maven-surefire-plugin</artifactId>
         <configuration>
           <includes>
-            <include>**/Test.java</include>
+            <include>**/*Test.java</include>
           </includes>
         </configuration>
       </plugin>
--- a/its/plugin/tests/pom.xml
+++ b/its/plugin/tests/pom.xml
@@ -36,12 +36,18 @@
       <version>${sonar.version}</version>
     </dependency>
     <dependency>
+      <groupId>org.sonarsource.sonarqube</groupId>
+      <artifactId>sonar-plugin-api-impl</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
       <groupId>org.sonarsource.sonarlint.core</groupId>
       <artifactId>sonarlint-core</artifactId>
     </dependency>
     <dependency>
-      <groupId>junit</groupId>
-      <artifactId>junit</artifactId>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter</artifactId>
+      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>org.assertj</groupId>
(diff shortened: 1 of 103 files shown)


its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/SonarLintTest.java
```diff
@@ -171,8 +166,7 @@ private static StandaloneGlobalConfiguration getSonarLintConfig(Path nodePath, V
     return StandaloneGlobalConfiguration.builder()
       .addEnabledLanguage(Language.JS)
       .addEnabledLanguage(Language.TS)
-      .addPlugin(Tests.JAVASCRIPT_PLUGIN_LOCATION.getFile().toURI().toURL())
-      .setSonarLintUserHome(temp.newFolder().toPath())
```
yassin-kammoun-sonarsource, line 175: Was it removed by accident or did it turn out to be useless ?

its/plugin/tests/pom.xml
```diff
@@ -68,9 +74,7 @@
         <artifactId>maven-surefire-plugin</artifactId>
         <configuration>
           <includes>
-            <include>**/Tests.java</include>
-            <include>**/TypeScriptRuleTest.java</include>
-            <include>**/EslintCustomRulesTest.java</include>
```
yassin-kammoun-sonarsource, line 73: Do you know why those were there in the first place ?

yassin-kammoun-sonarsource (commented):
Almost LGTM. I just left two questions for you to answer.

its/plugin/tests/pom.xml
saberduck: This was need because those two test were not part of the TestSuite, because they used another configuration of orchestrator. This is no longer necessary to be separate and they can be included using same wildcard

its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/SonarLintTest.java
saberduck: it seems to be useless in fact

saberduck: I changed it back, because when I removed it, this was using system user home as home directory, which could intefere on local machine. I also fixed the issue when test would fail on Windows, because this directory was not cleaned up properly

yassin-kammoun-sonarsource (approved):

PR #2770, opened 2021-08-18
Require Node.js v10 and update documentation about supported Node.js versions

### LINKED ISSUE #2771: Require Node.js v10 as minimum
We still enforce only v8 even though the version of eslint we are using requires v10 as a minimum.

See also #2501 and #2502 

Note that v10 is already EOL and v12 is recommended minimum #2651

saberduck:
Fix #2771 

I moved all constants to NodeDeprecationWarning, so they are in one place.

### DIFF
--- a/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
+++ b/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
@@ -65,7 +65,7 @@ public class EslintBridgeServerImpl implements EslintBridgeServer {
   private static final String MAX_OLD_SPACE_SIZE_PROPERTY = "sonar.javascript.node.maxspace";
   private static final String ALLOW_TS_PARSER_JS_FILES = "sonar.javascript.allowTsParserJsFiles";
   private static final Gson GSON = new Gson();
-  private static final int MIN_NODE_VERSION = 8;
+
   private static final String DEPLOY_LOCATION = "eslint-bridge-bundle";
 
   private final OkHttpClient client;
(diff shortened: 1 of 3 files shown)


sonar-javascript-plugin/src/main/resources/static/documentation.md
```diff
@@ -11,11 +11,12 @@ key: javascript
 ## Prerequisites
 
 In order to analyze JavaScript or TypeScript code, you need to have supported version of Node.js installed on the
-machine running the scan. Supported versions are current LTS versions (v10, v12, v14) and latest version v15. Odd
+machine running the scan. Supported versions are current LTS versions (v12, v14) and the latest version - v16. Odd
```
vilchik-elena, line 14: don't we still support 8? https://github.com/SonarSource/SonarJS/blob/master/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java#L68

sonar-javascript-plugin/src/main/resources/static/documentation.md
```diff
@@ -11,11 +11,12 @@ key: javascript
...
 (non LTS) versions might work, but are not actively tested. We recommend using the latest available LTS version 
-(v14 as of today) for optimal stability and performance.
+(v14 as of today) for optimal stability and performance. v10 is still supported, but it already reached end-of-life and 
+is deprecated.
```
vilchik-elena, line 17: do we log a deprecation note?

vilchik-elena: found it

sonar-javascript-plugin/src/main/resources/static/documentation.md
saberduck: No, it should be 10, because eslint we use requires 10. MIN_VERSION was not updated. I created ticket to fix it.

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
```diff
@@ -171,7 +171,7 @@ private void initNodeCommand(SensorContext context, File scriptFile, File workDi
         }
       })
       .pathResolver(bundle)
-      .minNodeVersion(MIN_NODE_VERSION)
+      .minNodeVersion(NodeDeprecationWarning.MIN_NODE_VERSION)
```
vilchik-elena, line 174: a naive question, can we unit test it?

vilchik-elena (approved):

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
saberduck: not sure how, will think about it

PR #2780, opened 2021-08-24
Performance monitoring

saberduck:


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/Monitoring.java
```diff
@@ -0,0 +1,207 @@
...
+    sensorMetric.clock = new Clock();
+  }
+
+  void stopSensor() {
+    if (!enabled) {
+      return;
+    }
+    sensorMetric.duration = sensorMetric.clock.stop();
+    sensorMetric.component = sensor.getClass().getCanonicalName();
+    sensorMetric.projectKey = sensorContext.project().key();
```
vilchik-elena, line 79: If we move these 2 lines to `startSensor` then no need to store `sensor` and `sensorContext` in the `Monitoring`. You can inject `Configuration` directly into `Monitoring`. IMO a bit cleaner that way and tests are more concise

eslint-bridge/src/analyzer.ts
```diff
@@ -134,9 +141,13 @@ function analyze(input: AnalysisInput, language: 'ts' | 'js'): AnalysisResponse
...
   }
-  const result = buildSourceCode(input, language);
+  const { result, duration: parseTime } = measureDuration(() => buildSourceCode(input, language));
   if (result instanceof SourceCode) {
-    return analyzeFile(result, input);
+    const { result: response, duration: analysisTime } = measureDuration(() =>
+      analyzeFile(result, input),
+    );
+    response.perf = { parseTime, analysisTime };
+    return response;
```
vilchik-elena, line 150: not a big improvement, but still:
```js
    const { result, duration: analysisTime } = measureDuration(() =>
      analyzeFile(result, input),
    );
    return {... result, parseTime, analysisTime };
```

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/MonitoringTest.java
```diff
@@ -0,0 +1,155 @@
...
+    EslintBridgeServer.Perf perf = new EslintBridgeServer.Perf();
+    perf.analysisTime = 2;
+    perf.parseTime = 3;
+    monitoring.stopFile(inputFile, 4, perf);
+    monitoring.stopSensor();
+    monitoring.stop();
+    Path metricsPath = monitoringPath.resolve("metrics.json");
+    assertThat(metricsPath).exists();
+    try (BufferedReader br = Files.newBufferedReader(metricsPath)) {
+      Monitoring.FileMetric fileMetric = gson.fromJson(br.readLine(), Monitoring.FileMetric.class);
```
vilchik-elena, line 108: imho asserting raw json is more readable

vilchik-elena (approved):

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/MonitoringTest.java
saberduck: yeah, however values measuring time are not stable, we can't assert them reliably

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/AbstractEslintSensor.java
```diff
@@ -83,7 +84,7 @@
 
   AbstractEslintSensor(AbstractChecks checks, NoSonarFilter noSonarFilter,
                        FileLinesContextFactory fileLinesContextFactory, EslintBridgeServer eslintBridgeServer,
-                       @Nullable AnalysisWarnings analysisWarnings) {
+                       AnalysisWarnings analysisWarnings, Monitoring monitoring) {
```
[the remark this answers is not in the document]
saberduck: indeed, it would be more consistent that way

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/Monitoring.java
saberduck: afaik it's not possible to inject `Configuration` in the constructor, because properties are not set yet, they will become visible only in configuration from `SensorContext`

saberduck: forget it, I was wrong and had issue in another piece of code. It works with `Configuration` injected

PR #2782, opened 2021-08-24
[CI] Windows QA testing the antivirus setting

saberduck:
Fixes #


saberduck:
it seems antivirus is not a problem

PR #2784, opened 2021-08-27
Add rule S6351: Regular expressions with the global flag should be used with caution

### LINKED ISSUE #2637: Add rule S6351: Regular expressions with the global flag should be used with caution
https://sonarsource.github.io/rspec/#/rspec/S6351/javascript

yassin-kammoun-sonarsource:
Fixes #2637


eslint-bridge/src/rules/stateful-regex.ts
```diff
@@ -0,0 +1,212 @@
...
+    schema: [
+      {
+        // internal parameter for rules having secondary locations
+        enum: ['sonar-runtime'],
+      },
+    ],
+  },
+  create(context: Rule.RuleContext) {
+    const invocations = new Map<Scope.Variable, estree.CallExpression[]>();
+    const regexes: { node: estree.Node; flags: string }[] = [];
```
saberduck, line 49: I think it would be handy to have explicit type for this `{ node: estree.Node; flags: string }` , seems like it's used in multiple places

saberduck (approved):
small comment, otherwise lgtm!

PR #2786, opened 2021-08-31
Add rule S6353: Regular expression quantifiers and character classes should be used concisely

### LINKED ISSUE #2778: Add rule S6353: Regular expression quantifiers and character classes should be used concisely
https://sonarsource.github.io/rspec/#/rspec/S6353/javascript

yassin-kammoun-sonarsource:
Fixes #2778


saberduck (changes requested):
it seems that `[\s\S]` is an idiom used to match also newline character, which is by default not matched by `.`  I think we can still keep the rule, but I would detail this in the rule description and also adjust the message that if newline should be matched, this should be set by using `s` flag

yassin-kammoun-sonarsource:
> it seems that `[\s\S]` is an idiom used to match also newline character, which is by default not matched by `.` I think we can still keep the rule, but I would detail this in the rule description and also adjust the message that if newline should be matched, this should be set by using `s` flag

I was not aware of that, and I think it's better to update the rule implementation so that `[\s\S]` is only considered when the `s` flag is enabled.

saberduck (approved):

saberduck:
@yassin-kammoun-sonarsource can you also update description of the rule about `\s\S` and `s` flag?

yassin-kammoun-sonarsource:
> @yassin-kammoun-sonarsource can you also update description of the rule about `\s\S` and `s` flag?

@saberduck Sure thing. Could you review first [this change](https://github.com/SonarSource/rspec/pull/297) ?

PR #2791, opened 2021-09-03
Add rule S5869: Character classes in regular expressions should not contain the same character twice

### LINKED ISSUE #2773: Add rule S5869: Character classes in regular expressions should not contain the same character twice
https://sonarsource.github.io/rspec/#/rspec/S5869/javascript

yassin-kammoun-sonarsource:
Fixes #2773


eslint-bridge/tests/rules/duplicates-in-character-class.test.ts
```diff
@@ -0,0 +1,226 @@
...
+      code: `/[A-_d-{]/i`, // FN because we ignore case insensitivity unless both ends of the ranges are letters
+    },
+    {
+      code: `/[A-z_]/i`, // FN because A-z gets misinterpreted as A-Za-z due to the way we handle case insensitivity
+    },
+    {
+      code: `/[\\p{IsLatin}x]/`, // FN because we don't support \p at the moment
+    },
+  ],
+  invalid: [
```
vilchik-elena, line 96: no assertion of location and messages

eslint-bridge/src/utils/utils-regex.ts
```diff
@@ -188,3 +199,232 @@ export function isStringRegexMethodCall(call: estree.CallExpression, services: P
...
+    }
+
+    isAscii(c: number) {
+      return c < 128;
+    }
+
+    codePoint(c: string) {
+      const cp = c.codePointAt(0);
+      if (cp === undefined) {
+        throw new Error(`failed to compute code point for: ${c}`);
```
vilchik-elena, line 425: are you sure we want to throw in this case? I think the whole analysis will be cancelled

eslint-bridge/src/utils/utils-regex.ts
```diff
@@ -188,3 +199,232 @@ export function isStringRegexMethodCall(call: estree.CallExpression, services: P
     isString(call.arguments[0], services)
   );
 }
+
+const MAX_CODE_POINT = 0x10ffff;
+
+export class SimplifiedRegexCharacterClass {
```
vilchik-elena, line 205: I think it should be in a separate file

eslint-bridge/src/utils/utils-regex.ts
```diff
@@ -188,3 +199,232 @@ export function isStringRegexMethodCall(call: estree.CallExpression, services: P
...
+  /**
+   * This map defines the contents of the character class in the following way:<br>
```
vilchik-elena, line 207: I think methods mentioned in this comment are java-specific

eslint-bridge/src/utils/utils-regex.ts
```diff
@@ -188,3 +199,232 @@ export function isStringRegexMethodCall(call: estree.CallExpression, services: P
...
+    tree: createTree.Tree<number, regexpp.AST.Node | undefined>,
+  ) {
+    const entry = tree.find(key);
+    if (entry.valid) {
+      return entry.update(value);
+    }
+    return tree.insert(key, value);
+  }
+
+  private static readonly Builder = class {
```
vilchik-elena, line 292: I don't get why we need this class

eslint-bridge/src/rules/duplicates-in-character-class.ts
```diff
@@ -0,0 +1,77 @@
...
+        const characterClass = new SimplifiedRegexCharacterClass(flags);
+        node.elements.forEach(element => {
+          const intersections = new SimplifiedRegexCharacterClass(flags, element).findIntersections(
+            characterClass,
+          );
+          if (intersections.length > 0) {
+            intersections.forEach(intersection => {
+              if (!duplicates.has(intersection)) {
+                // The element the current element is intersecting with should be included as well.
+                duplicates.add(intersection);
```
vilchik-elena, line 45: `duplicates` is a set, can't we add here unconditionally?

vilchik-elena (changes requested):

eslint-bridge/src/utils/utils-regex.ts
yassin-kammoun-sonarsource: Well, the easy answer is because sonar-java defined this class. I think it's okay to keep this class and our implementation as similar as sonar-java. It will make things easier later on if we need to fix something.

eslint-bridge/src/utils/utils-regex.ts
yassin-kammoun-sonarsource: I could also have used the non-null assertion operator, but I preferred to throw an error instead. In any case, it should never happen.

vilchik-elena (approved):

PR #2793, opened 2021-09-06
Replace rule S5868 with our implementation providing precise location

### LINKED ISSUE #2725: Implement S5868 to have proper issue location and message

saberduck:
Fixes #2725


eslint-bridge/src/rules/sonar-no-misleading-character-class.ts
```diff
@@ -0,0 +1,137 @@
...
+    }
+  }
+
+  return {
+    onCharacterClassEnter(ccNode) {
+      for (const chars of iterateCharacterSequence(ccNode.elements)) {
+        testChars(
+          surrogatePairWithoutUFlag,
+          chars,
+          "Unexpected surrogate pair in character class. Use 'u' flag.",
```
vilchik-elena, line 124: I think we should make messages more sonar-like (actionable). 
Also those terms (like surrogate pair, combined character etc) are not explained or mentioned in the rule description. We should either extend description or use other terms in the messages

eslint-bridge/src/rules/sonar-no-misleading-character-class.ts
```diff
@@ -0,0 +1,137 @@
...
+    test: (chars: Character[]) => Character | undefined,
+    chars: Character[],
+    message: string,
+  ) {
+    const regexpNode = test(chars);
+    if (regexpNode !== undefined) {
+      context.reportRegExpNode({
+        regexpNode,
+        message,
+        node: context.node,
```
vilchik-elena, line 113: we will report on the node if regex is not literal (like here https://peach.sonarsource.com/project/issues?id=ruby%3Adiscourse&issues=AXu7El7BMrVfiJHibBed&open=AXu7El7BMrVfiJHibBed). So I think message should contain the reported character.

eslint-bridge/tests/rules/sonar-no-misleading-character-class.test.ts
```diff
@@ -0,0 +1,281 @@
...
+    {
+      code: String.raw`var r = new RegExp("[\\uD83D\\uDC68\\u200D\\uD83D\\uDC69\\u200D\\uD83D\\uDC66]", "u")`,
+      errors: [{ message: zwj }],
+    },
+    {
+      code: String.raw`var r = new RegExp("[\\u{1F468}\\u{200D}\\u{1F469}\\u{200D}\\u{1F466}]", "u")`,
+      errors: [{ message: zwj }],
+    },
+    {
+      code: String.raw`var r = new globalThis.RegExp("[❇️]", "")`,
```
vilchik-elena, line 261: I have never heard of `globalThis` and it's never used in our ITs' sources. Are you sure we want to support it?

eslint-bridge/src/rules/sonar-no-misleading-character-class.ts
```diff
@@ -0,0 +1,137 @@
...
+  function* iterateCharacterSequence(nodes: CharacterClassElement[]) {
+    let seq: Character[] = [];
+
+    for (const node of nodes) {
+      switch (node.type) {
+        case 'Character':
+          seq.push(node);
+          break;
+
+        case 'CharacterClassRange':
```
vilchik-elena, line 42: Why we need this case? Nothing fails if I comment it out

eslint-bridge/src/rules/sonar-no-misleading-character-class.ts
```diff
@@ -0,0 +1,137 @@
...
+          seq.push(node.min);
+          yield seq;
+          seq = [node.max];
+          break;
+
+        case 'CharacterSet':
```
vilchik-elena, line 48: same here

vilchik-elena (changes requested):

eslint-bridge/tests/rules/sonar-no-misleading-character-class.test.ts
saberduck: it's part of the standard, so I don't see why we wouldn't want to support it https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis

eslint-bridge/src/rules/sonar-no-misleading-character-class.ts
saberduck: this is needed in cases such as `/[\u0041\u0301-\u0301]/` , where the range would be another node in the regexp AST

eslint-bridge/src/rules/sonar-no-misleading-character-class.ts
saberduck: this is because we don't want to report in case like this `/[\\uD83D\\d\\uDC4D]/`

vilchik-elena (approved):
@saberduck could you update the ticket to be explicit about the change in the rule scope

PR #2794, opened 2021-09-07
Add rule S6019: Reluctant quantifiers in regular expressions should be followed by an expression that can't match the empty string

### LINKED ISSUE #2774: Add rule S6019: Reluctant quantifiers in regular expressions should be followed by an expression that can't match the empty string
https://sonarsource.github.io/rspec/#/rspec/S6019/javascript

vilchik-elena:
Fixes #2774


javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S6019.json
```diff
@@ -0,0 +1,20 @@
+{
+  "title": "Reluctant quantifiers in regular expressions should be followed by an expression that can\u0027t match the empty string",
```
yassin-kammoun-sonarsource, line 2: Any idea why the unicode character instead of the single quote ?

yassin-kammoun-sonarsource (changes requested):

javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S6019.json
vilchik-elena: That's how rule-api is generating, for better compatibility probably

PR #2798, opened 2021-09-08
Fix handling of forward slash when parsing regexp

### LINKED ISSUE #2788: Improve regex range computation by accounting for implicitly escaped slashes
The parser for regular expressions `regexpp` implicitly escapes forward slashes when parsing string literals. To illustrate this point, `'//x'` in `RegExp('//x')` is actually rewritten as `'\/\/x'` at parsing time.

This distorts the computation of regular expression ranges because the tokeniser that is used under the hood relies on the raw value of the string literal, not the raw value of the `regexpp` AST node.

In the end, every range of nodes after a slash-node would always be off by one column per anterior slash.

saberduck:
Fixes #2788 

The improvement makes two changes
 - in string tokenizer, when we process `/` we insert an additional token to correct for the implicit escape
 - when computing offsets for secondary locations, this should be done in the coordinates of the regexp, which will account for / thanks to the previous point

Computation of offsets will not work if the regexp spans multiple lines. I believe we can ignore such corner case.

### DIFF
--- a/eslint-bridge/src/rules/regex-complexity.ts
+++ b/eslint-bridge/src/rules/regex-complexity.ts
@@ -35,6 +35,7 @@ import {
 import {
   getParsedRegex,
   getRegexpLocation,
+  getRegexpRange,
   getUniqueWriteUsage,
   isBinaryPlus,
   isIdentifier,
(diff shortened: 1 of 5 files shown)


yassin-kammoun-sonarsource (approved):
lgtm!

PR #2860, opened 2021-11-01
Remove '.only' from 'analyzer' test

vilchik-elena:
Also fixing one test (one more token to highlight is `// nosonar` comment)

### DIFF
--- a/eslint-bridge/tests/analyzer.test.ts
+++ b/eslint-bridge/tests/analyzer.test.ts
@@ -94,7 +94,7 @@ describe('#analyzeJavaScript', () => {
     expect(result.highlights).toHaveLength(11);
   });
 
-  it.only('should analyze both main and test files', () => {
+  it('should analyze both main and test files', () => {
     initLinter([
       { key: 'no-one-iteration-loop', configurations: [], fileTypeTarget: ['TEST', 'MAIN'] },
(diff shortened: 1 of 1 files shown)


eslint-bridge/tests/analyzer.test.ts
```diff
@@ -94,7 +94,7 @@ describe('#analyzeJavaScript', () => {
     expect(result.highlights).toHaveLength(11);
   });
 
-  it.only('should analyze both main and test files', () => {
```
yassin-kammoun-sonarsource, line 97: I am wondering whether we should have a rule reporting on `only`-like unit tests. It's likely a mistake. WDYT ?

yassin-kammoun-sonarsource (approved):
LGTM. We should probably merge #2858 first.

eslint-bridge/tests/analyzer.test.ts
vilchik-elena: we definitely should :) Is there such thing in mocha btw?

yassin-kammoun-sonarsource: Yes, there is: see https://mochajs.org/#exclusive-tests

saberduck: I think this is great rule idea

vilchik-elena: https://github.com/SonarSource/SonarJS/issues/2861

PR #2862, opened 2021-11-01
Use npm instead of yarn

vilchik-elena:


eslint-bridge/pom.xml
```diff
@@ -27,10 +27,10 @@
...
         <executions>
           <execution>
-            <id>yarn install</id>
+            <id>npm install</id>
             <phase>generate-resources</phase>
             <configuration>
-              <executable>yarn</executable>
+              <executable>npm</executable>
               <arguments>
                 <argument>install</argument>
```
saberduck, line 35: we should use `ci` instead of `install` here to make sure we are not overwriting lock file. See https://docs.npmjs.com/cli/v7/commands/npm-ci

saberduck: in fact, `install` is part of the build - maybe we don't need this phase at all

eslint-bridge/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "1.0.0",
   "description": "bridge between SonarJS and ESLint",
   "scripts": {
-    "build": "yarn install && yarn check-format && yarn clear && yarn compile",
+    "build": "npm install && npm run check-format && npm run clear && npm run compile",
```
saberduck, line 6: we should use `ci` instead of `install`

saberduck (approved):

PR #2868, opened 2021-11-11
Use Java 11

saberduck:


sonar-javascript-plugin/pom.xml
```diff
@@ -122,6 +122,7 @@
           <sonarLintSupported>true</sonarLintSupported>
           <sonarQubeMinVersion>${sonarQubeMinVersion}</sonarQubeMinVersion>
           <nodeJsMinVersion>10.12.0</nodeJsMinVersion>
+          <jreMinVersion>${jdk.min.version}</jreMinVersion>
```
vilchik-elena, line 125: why are we using `jdk` for `jre`?

vilchik-elena (approved):

sonar-javascript-plugin/pom.xml
saberduck: it's the same thing, just sonarlint decided to name it differently than we do in parent pom

PR #2874, opened 2021-11-12
Various improvements after CSS merge

vilchik-elena:
* Update README
* Update documentation.md (merge JS and CSS, TS 4.4)
* create separate interfaces for AnalysisRequest for JS and CSS
* drop mock-server.test.ts and move its tests into `analyzer.test.ts`
* Rename `configFile` into `stylelintConfig` in AnalysisRequest
* Move CssRuleSensor to `eslint` package
* Fix imports formatting 
* Other quality issues


eslint-bridge/tests/analyzer.test.ts
```diff
@@ -69,6 +70,7 @@ describe('#analyzeJavaScript', () => {
       filePath,
       fileContent: codeToTest,
       fileType: 'MAIN',
+      tsConfigs: [],
```
vilchik-elena, line 73: to reproduce production analysis we should provide tsconfig here.

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/css/metrics/CssMetricSensor.java
```diff
@@ -37,13 +37,13 @@
 import org.sonar.api.utils.log.Loggers;
 import org.sonar.plugins.javascript.css.CssLanguage;
 
-public class MetricSensor implements Sensor {
+public class CssMetricSensor implements Sensor {
```
yassin-kammoun-sonarsource, line 40: I have a vague memory of hardcoded sensor names somewhere for speed of analysis. Will this change impact it ?

yassin-kammoun-sonarsource (approved):
LGTM. I suggested minor changes that you are free to ignore if irrelevant. Also, we need to create tickets for the pieces of code (message trimming, error parsing, metric) that we expect to move from CSS sensor to the bridge.

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/css/metrics/CssMetricSensor.java
vilchik-elena: Nope, we are relying there on sensor's name, set in `describe` method

sonar-javascript-plugin/src/main/resources/static/documentation.md
```diff
@@ -1,39 +1,42 @@
 ---
-title: JS / TS / CSS
+title: JavaScript / TypeScript / CSS
```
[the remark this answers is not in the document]
vilchik-elena: I will remove spaces but prefer to keep CSS last to keep JavaScript more visible

PR #2877, opened 2021-11-12
Improve S4790: report only on insecure algorithms

### LINKED ISSUE #1997: Rule S4790: its content should be replaced by S2070
For hashing algorithms, several rules exist, in particular these two:
- A security-hotspot: https://jira.sonarsource.com/browse/RSPEC-4790
- That supersedes this vulnerability: https://jira.sonarsource.com/browse/RSPEC-2070 (not available in SonarJS)

It's not possible to maintain two rules on exactly the same subject, for us and the end users, so:
- S2070 will be deprecated
- the content/implementation of S2070 is more relevant than S4790 because S4790 raises everywhere a hash function is used (even when secure hash function, like SHA-256 is used) and S2070 raises only when a weak hash function is used (like MD5)
- so the content/implementation of S2070 should "be moved" to S4790 (the key of the rule should be updated in SonarJs), because the type of issue (hotspot) is more relevant

yassin-kammoun-sonarsource:
Fixes #1997


eslint-bridge/src/rules/hashing.ts
```diff
@@ -17,18 +17,64 @@
...
+    return {
+      'CallExpression[arguments.length > 0]': (node: estree.Node) => {
+        const { callee, arguments: args } = node as estree.CallExpression;
+        const hashAlgorithm = getUniqueWriteUsageOrNode(context, args[0]);
+        if (
+          isStringLiteral(hashAlgorithm) &&
+          UNSECURE_HASH_ALGORITHMS.has(hashAlgorithm.value.toLocaleLowerCase())
+        ) {
+          const { module, method } = getModuleAndCalledMethod(callee, context);
+          if (module?.value === 'crypto' && isIdentifier(method, 'createHash')) {
```
vilchik-elena, line 70: that seems weird to me to first check the argument and only then the API, I would revert

eslint-bridge/tests/rules/hashing.test.ts
```diff
@@ -18,108 +18,84 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
```
vilchik-elena, line 19: How about moving the test to be comment-based?

eslint-bridge/tests/rules/hashing.test.ts
```diff
@@ -18,108 +18,84 @@
...
-      code: `import * as crypto from 'foo'; crypto.digest();`,
+      code: `
+        const crypto = require('crypto');
+        crypto.createHash('sha512');`,
     },
     {
-      code: `import * as crypto from 'crypto'; crypto.encrypt();`,
+      code: `
+        const crypto = require('crypto');
+        crypto.createHsah('sha1');`,
```
vilchik-elena, line 49: it took me ages to understand why we are not reporting here :D Can we make this test more explicit?

eslint-bridge/tests/rules/hashing.test.ts
```diff
@@ -18,108 +18,84 @@
...
+        import * as crypto from 'crypto';
+        crypto.createHash('sha1');
+      `,
       errors: 1,
     },
     {
-      code: `import { scryptSync } from 'crypto'; scryptSync();`,
+      code: `
+        import {createHash} from 'crypto';
+        createHash('sha1');
```
vilchik-elena, line 97: can we test other weak algorithms and also assert message and location?

eslint-bridge/src/rules/hashing.ts
vilchik-elena, line 70: Why did you decide to report only on this API?

eslint-bridge/src/rules/hashing.ts
```diff
@@ -17,18 +17,64 @@
...
+import {
+  getModuleAndCalledMethod,
+  getUniqueWriteUsageOrNode,
+  isIdentifier,
+  isStringLiteral,
+} from '../utils';
 
-const message = `Make sure that hashing data is safe here.`;
+const MESSAGE = 'Make sure this weak hash algorithm is not used in a sensitive context here.';
+const UNSECURE_HASH_ALGORITHMS = new Set([
```
vilchik-elena, line 32: where did you get this list from? I don't see `md2` here which is mentioned in the rule description

vilchik-elena (changes requested):

eslint-bridge/src/rules/hashing.ts
yassin-kammoun-sonarsource: I only saw this API in the [rspec](https://sonarsource.github.io/rspec/#/rspec/S4790/javascript). Did I missing another one ?

eslint-bridge/src/rules/hashing.ts
yassin-kammoun-sonarsource: I had to run `openssl list-message-digest-algorithms` as mentioned in Node.js documentation for [crypto#createHash](https://nodejs.org/api/crypto.html#cryptocreatehashalgorithm-options), but it's a bit annoying because the result depends on the platform. That's why I didn't include `'md2'` in the set. Should I include it as a precaution ?

eslint-bridge/src/rules/hashing.ts
yassin-kammoun-sonarsource: Never mind. I missed the other API in the RSPEC comment.

vilchik-elena (changes requested):
Please please what we discussed

vilchik-elena (approved):

PR #2880, opened 2021-11-15
Fix minification computation and CSS its

saberduck:


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/minify/AverageLineLengthCalculator.java
```diff
@@ -76,7 +76,7 @@ public boolean isLineInHeaderComment(String line) {
   }
 
   private boolean isFirstLineInHeaderComment(String line) {
-    if (line.startsWith("/*")) {
+    if (line.startsWith("/*") && (!line.contains("*/") || line.endsWith("*/"))) {
```
vilchik-elena, line 79: tbh I am not sure about the impact of this fix on those computations for JS. Are we sure there will be no change in behaviour?

vilchik-elena (approved):
LGTM 
But please update PR title and commit to contain the fix for minification.
Also I am not 100% that the fix you did for minification is harmless.

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/minify/AverageLineLengthCalculator.java
saberduck: tests are OK, do we trust our tests? :)

vilchik-elena: Who is guarding the guards?)

PR #2882, opened 2021-11-15
Update metadata

saberduck:


javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S5843.html
```diff
@@ -3,6 +3,43 @@
 <p>The complexity of a regular expression is determined as follows:</p>
 <p>Each of the following operators increases the complexity by an amount equal to the current nesting level and also increases the current nesting
 level by one for its arguments:</p>
+<ul>
```
vilchik-elena, line 6: @saberduck smth is wrong here, it's java

vilchik-elena (changes requested):

vilchik-elena (approved):

PR #2887, opened 2021-11-17
Improve S3498: fix FP for Aura Lightning Components and change location highlighting

### LINKED ISSUE #2884: Improve S3498: highlight only the key part of an object expression property
The rule currently highlights the whole property node when it raises an issue (see [Peach](https://peach.sonarsource.com/issues?open=AWB2EDErIVt3g_SeiMBF&resolved=false&rules=javascript%3AS3498)), which makes other issues unreadable.

### LINKED ISSUE #2885: Fix FP S3498: should not raise on Aura Lightning Components
The rule currently raises issues on Salesforce Aura Lightning Components (see [Peach](https://peach.sonarsource.com/project/issues?id=sf-LightningFlowComponents_nodev10&open=AX0khViHTPFHcz6tHEez&resolved=false&rules=javascript%3AS3498)) like the following:

```javascript
({
  action: function(component, event, helper) {
    /* .... */
  }
});
```

Although such issues are true positives, all snippets from [Salesforce documentation](https://developer.salesforce.com/docs/atlas.en-us.lightning.meta/lightning/expr_resource_value_provider.htm) use this old syntax as well as code generators, and make the rule noisy for Aura Lightning Components. Therefore, we should add an exception to the rule for these components.

yassin-kammoun-sonarsource:
Fixes #2885
Fixes #2884

### DIFF
--- a/eslint-bridge/src/linter.ts
+++ b/eslint-bridge/src/linter.ts
@@ -35,6 +35,7 @@ import { getContext } from './context';
 import { decoratePreferTemplate } from './rules/prefer-template-decorator';
 import { decorateAccessorPairs } from './rules/accessor-pairs-decorator';
 import { decorateNoRedeclare } from './rules/no-redeclare-decorator';
+import { decorateObjectShorthand } from './rules/object-shorthand-decorator';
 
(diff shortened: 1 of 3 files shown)


vilchik-elena (approved):
LGTM, but please improve coverage before merge

PR #2907, opened 2021-12-07
Control `ts.Program` creation for TS files

### LINKED ISSUE #2073: Files not referenced in tsconfig are not analyzed
We don't analyze files that are part of the project, but they are not directly referenced in `tsconfig.json`. If we have `tsconfig.json` such as 

```
{
  "files": ["root.ts"]
}
```

where `root.ts` imports another file `module.ts`, which is not referenced in `tsconfig.json` however it is part of the project build

```
import * as m from 'module'
```

`module.ts` will not be recognized as belonging to project by SonarJS. To be able to recognize it as belonging to the project we would need to create the project using TypeScript APIs (`project.getSourceFiles()`), and not only parse `tsconfig.json` file.

As a workaround, project should list all TypeScript files in `tsconfig.json` file

vilchik-elena:
Currently `ts.Program` is created by `typescript-eslint`. We want to control its creation in order to:
* improve UX so that Program creation is a separate step and not part of first file analysis
* analyze files not being root files of the Program (fixes #2073)


eslint-bridge/src/parser.ts
```diff
@@ -23,18 +23,21 @@ import { Linter, SourceCode } from 'eslint';
...
 import { getContext } from './context';
-import { JsAnalysisInput } from './analyzer';
+import { JsAnalysisInput, ProgramBasedAnalysisInput } from './analyzer';
+import { getProgramById } from './programManager';
 
 const babelParser = { parse: babel.parseForESLint, parser: '@babel/eslint-parser' };
 const vueParser = { parse: VueJS.parseForESLint, parser: 'vue-eslint-parser' };
 const tsParser = { parse: tsEslintParser.parseForESLint, parser: '@typescript-eslint/parser' };
 
+type AnalysisInputNotCss = JsAnalysisInput | ProgramBasedAnalysisInput;
```
vilchik-elena, line 33: I don't like the name

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/AnalysisWithProgram.java
```diff
@@ -0,0 +1,114 @@
...
+    Set<InputFile> skippedFiles = new HashSet<>(inputFiles);
+    skippedFiles.removeAll(analyzedFiles);
+    LOG.debug("Skipped {} files because they were not part of any tsconfig", skippedFiles.size());
+    skippedFiles.forEach(f -> LOG.debug("File not part of any tsconfig: {}", f));
+  }
+
+  private void analyzeProgram(TsProgram program, Set<InputFile> analyzedFiles) throws IOException {
+    var fs = context.fileSystem();
+    for (var file : program.files) {
+      var inputFile = fs.inputFile(fs.predicates().hasAbsolutePath(file));
```
vilchik-elena, line 83: I think we should be stricter here, to predicate for the language as well. Otherwise we could happen to analyze some JS file twice (if it's part of TS project)

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/AnalysisWithProgram.java
```diff
@@ -0,0 +1,114 @@
...
+      if (inputFile == null) {
+        LOG.debug("File not part of the project: '{}'", file);
```
vilchik-elena, line 85: I wonder if we should do something with this log as it will be probably extremely noisy

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/AnalysisWithProgram.java
```diff
@@ -0,0 +1,114 @@
...
+        continue;
+      }
+      if (analyzedFiles.add(inputFile)) {
+        analyze(inputFile, program);
+      } else {
+        LOG.debug("File already analyzed: '{}'", file);
```
vilchik-elena, line 91: this implies some misconfiguration in the project, right? If yes, how about logging `info`?

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/AnalysisWithProgram.java
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  private void analyze(InputFile file, TsProgram tsProgram) throws IOException {
+    if (context.isCancelled()) {
+      throw new CancellationException("Analysis interrupted because the SensorContext is in cancelled state");
+    }
+    try {
+      LOG.debug("Analyzing {}", file);
+      monitoring.startFile(file);
+      String fileContent = contextUtils.shouldSendFileContent(file) ? file.contents() : null;
```
vilchik-elena, line 103: didn't we say that we don't want to send content here?

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/AnalysisWithProgram.java
```diff
@@ -0,0 +1,114 @@
...
+      EslintBridgeServer.JsAnalysisRequest request = new EslintBridgeServer.JsAnalysisRequest(file.absolutePath(), file.type().toString(), fileContent,
+        contextUtils.ignoreHeaderComments(), emptyList(), tsProgram.id);
```
vilchik-elena, line 105: I rely on the absence of `tsConfigs` in the request (https://github.com/SonarSource/SonarJS/pull/2907/files#diff-09bb0c3f74a024d48398cff79e8e92b568085b5170ed3d1f8862858e89d9f3e5R124), so it should be really absent, not `null` and not empty list. 

In fact probably `typescript-eslint` will work with the empty list and `programs` provided, but I would like to keep it cleaner.

WDYT we should do? I would update the request to not include `tsConfigs` property

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServer.java
```diff
@@ -200,5 +186,26 @@ TextRange toTextRange(InputFile inputFile) {
...
+  class TsProgramRequest {
+    final String tsConfig;
+    final List<String> files;
+
+    public TsProgramRequest(String tsConfig) {
+      this.tsConfig = tsConfig;
+      this.files = null;
+    }
+
+    public TsProgramRequest(List<String> files) {
```
vilchik-elena, line 205: this one is not supported by `eslint-bridge`

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/ProcessAnalysis.java
```diff
@@ -0,0 +1,262 @@
...
+import org.sonar.api.measures.Metric;
+import org.sonar.api.rule.RuleKey;
+import org.sonar.api.utils.log.Logger;
+import org.sonar.api.utils.log.Loggers;
+import org.sonar.plugins.javascript.eslint.EslintBridgeServer.AnalysisResponse;
+
+import static org.sonar.plugins.javascript.eslint.EslintBridgeServer.Issue;
+import static org.sonar.plugins.javascript.eslint.EslintBridgeServer.IssueLocation;
+
+public class ProcessAnalysis {
```
vilchik-elena, line 51: I would think of another name, e.g. `AnalysisResultsProvider`. Anyway it should be noun-based

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
```diff
@@ -319,6 +324,18 @@ public TsConfigFile loadTsConfig(String filename) {
...
 
+  @Override
+  public TsProgram createProgram(TsProgramRequest tsProgramRequest) throws IOException {
+    var response = request(GSON.toJson(tsProgramRequest), "create-program");
+    return GSON.fromJson(response, TsProgram.class);
+  }
+
+  @Override
+  public boolean deleteProgram(TsProgram tsProgram) throws IOException {
+    var response = request(GSON.toJson(tsProgram), "delete-program");
```
vilchik-elena, line 335: did you send entire `TsProgram` not avoid new class creation?

its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/TypeScriptAnalysisTest.java
```diff
@@ -164,7 +191,7 @@ public void should_support_solution_tsconfig() {
       tuple(4, "typescript:S3923", "solution-tsconfig:src/file.ts")
     );
 
-    assertThat(result.getLogsLines(l -> l.contains("Skipped files: src/unlisted.ts"))).hasSize(1);
+    assertThat(result.getLogsLines(l -> l.contains("Skipped 0 files because they were not part of any tsconfig"))).hasSize(1);
```
vilchik-elena, line 194: in fact having this test we could drop the new one you added? `implicit_import_test`

eslint-bridge/src/parser.ts
saberduck: isn't it you who created it?

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/AnalysisWithProgram.java
saberduck: it was updated

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/AnalysisWithProgram.java
saberduck: I prefer to keep it because it helps a lot to troubleshoot. Should be ok on `DEBUG` level

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/ProcessAnalysis.java
saberduck: changed to `AnalysisProcessor`

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
saberduck: yes, and I optimized now a bit, to not send the files again in the request

its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/TypeScriptAnalysisTest.java
saberduck: yeah, we could indeed

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/TypeScriptSensorTest.java
```diff
@@ -332,44 +331,17 @@ public void should_log_when_failing_typescript() throws Exception {
...
     createInputFile(context, "dir/file1.ts");
     createInputFile(context, "dir/file2.ts");
+    createVueInputFile();
     createSensor().execute(context);
     assertThat(logTester.logs(LoggerLevel.ERROR)).contains("Failed to analyze file [dir/file1.ts] from TypeScript: Debug Failure. False expression.");
     assertThat(logTester.logs(LoggerLevel.ERROR)).contains("Failed to analyze file [dir/file2.ts] from TypeScript: Debug Failure. False expression.");
   }
 
-  @Test
-  public void should_create_ui_warning_missing_typescript() throws Exception {
```
vilchik-elena, line 338: @saberduck should we drop some logic responsible for these tests?

saberduck: done in #2909

saberduck:
replaced by #2910

PR #2911, opened 2021-12-09
Add progress report for AnalysisWithProgram

saberduck:


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/AnalysisWithProgram.java
```diff
@@ -64,25 +69,37 @@ void analyzeFiles(SensorContext context, AbstractChecks checks, List<InputFile>
...
     }
-    Deque<String> workList = new ArrayDeque<>(tsConfigs);
-    Set<String> analyzedProjects = new HashSet<>();
-    Set<InputFile> analyzedFiles = new HashSet<>();
-    while (!workList.isEmpty()) {
-      var tsConfig = workList.pop();
-      if (!analyzedProjects.add(tsConfig)) {
-        continue;
+    progressReport = new ProgressReport(PROGRESS_REPORT_TITLE, PROGRESS_REPORT_PERIOD);
+    progressReport.start(inputFiles.size(), inputFiles.iterator().next().absolutePath());
```
vilchik-elena, line 73: it will be confusing to see that first input file is analyzed while the program for the first tsconfig is created. The same will happen in between tsconfigs. How about adding tsconfigs to the list of progress report and have it as "next" when program is created?

vilchik-elena, line 73: also with such start we will increment twice for one file

vilchik-elena (changes requested):

vilchik-elena:
Let's merge with those minor drawbacks, we can improve on that later

vilchik-elena (approved):

PR #2915, opened 2021-12-13
Declare CSS sensors as sensors processing files independently

### LINKED ISSUE #2908: Declare CSS sensors as sensors processing files independently
This is to analyze only changed files in PR.

In order to have the Scanner aware that CSS is compatible with processing files independently, we need to call the API: `processesFilesIndependently`

See https://github.com/SonarSource/sonar-xml/pull/212/files as an example

vilchik-elena:
Fixes #2908

### DIFF
--- a/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/CssRuleSensor.java
+++ b/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/CssRuleSensor.java
@@ -23,6 +23,7 @@ import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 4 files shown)


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/CssRuleSensor.java
```diff
@@ -221,4 +228,17 @@ private static String normalizeMessage(String message) {
       return message;
     }
   }
+
+  private void processesFilesIndependently(SensorDescriptor descriptor) {
```
mpaladin, line 232: @vilchik-elena in case you want to avoid duplicating this method you can add `sonarRuntime` as an argument and make it `static`, I let you decide and merge.

mpaladin (approved):

Issue #2861, opened 2021-11-01
Rule idea: '.only' should not be used in tests


vilchik-elena:
Most probably  `.only` is a leftover after debugging and should be removed before committing

vilchik-elena:
Such syntax exists in both Mocha and Jest

gabssnake:
Hello there :)

In case you're interested, there's one implementation of this case in [eslint-plugin-mocha](https://github.com/lo1tuma/eslint-plugin-mocha/blob/master/docs/rules/no-exclusive-tests.md).

Some of the other rules there are also quite useful!

Happy new year!

PR #2918, opened 2021-12-27
Fix test - remove `only` and increase timeout

saberduck:
we should really implement https://github.com/SonarSource/SonarJS/issues/2861 :)


vilchik-elena:
Is it the test which makes build on windows fail?

vilchik-elena (approved):

PR #2923, opened 2022-01-10
Add performance monitoring for program and rules

saberduck:


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/Monitoring.java
```diff
@@ -40,11 +43,14 @@
...
-import static org.sonar.plugins.javascript.eslint.Monitoring.Metric.MetricType.SENSOR;
+import static org.sonar.plugins.javascript.eslint.Monitoring.MetricType.FILE;
+import static org.sonar.plugins.javascript.eslint.Monitoring.MetricType.PROGRAM;
+import static org.sonar.plugins.javascript.eslint.Monitoring.MetricType.RULE;
+import static org.sonar.plugins.javascript.eslint.Monitoring.MetricType.SENSOR;
+import static org.sonarsource.api.sonarlint.SonarLintSide.MULTIPLE_ANALYSES;
 
 @ScannerSide
-@SonarLintSide
+@SonarLintSide(lifespan = MULTIPLE_ANALYSES)
```
vilchik-elena, line 53: why do we need this to be available in Sonarlint?

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
```diff
@@ -407,4 +407,52 @@ public InitLinterRequest(List<EslintRule> rules, List<String> environments, List
       this.globals = globals;
     }
   }
+
+  static class MonitoringOutputConsumer implements Consumer<String> {
+
+    private static final String HEADER = "Rule                                 | Time (ms) | Relative";
```
vilchik-elena, line 413: we should be careful, as this format can change after ESLint update. Can we throw if monitoring is enabled and we don't detect header?

vilchik-elena (approved):

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/Monitoring.java
```diff
@@ -40,11 +43,14 @@
 import org.sonar.api.utils.log.Loggers;
 import org.sonarsource.api.sonarlint.SonarLintSide;
 
-import static org.sonar.plugins.javascript.eslint.Monitoring.Metric.MetricType.FILE;
-import static org.sonar.plugins.javascript.eslint.Monitoring.Metric.MetricType.SENSOR;
+import static org.sonar.plugins.javascript.eslint.Monitoring.MetricType.FILE;
```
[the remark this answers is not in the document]
saberduck: I think it's more concise to use short name in the code, imports are handled automatically and I never see them, so I don't really care about these lines - it's purely technical.

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/Monitoring.java
saberduck: this is need to be able to import it in `EslintBridgeServer` , otherwise I would need to have speciall constructor in EslintBridgeServer when it's not available

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/AnalysisWithProgram.java
```diff
@@ -90,6 +91,7 @@ void analyzeFiles(SensorContext context, AbstractChecks checks, List<InputFile>
           continue;
```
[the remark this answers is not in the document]
saberduck: yep, I don't think we should collect failed programs now.

nodejs-utils/src/main/java/org/sonarsource/nodejs/NodeCommand.java
```diff
@@ -65,16 +66,21 @@
    * Start external NodeJS process
    *
    * @throws NodeCommandException when start of the external process fails
+   * @param env environment for started process
    */
-  public void start() {
+  public void start(Map<String, String> env) {
```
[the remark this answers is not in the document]
saberduck: indeed, seems to be more consistent

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
saberduck: it's rather complicated to do, I think we are going to notice manually

PR #2928, opened 2022-01-18
Ignore promises in S3735 void-use

### LINKED ISSUE #2629: Ignore promises in S3735 `void-use`
eslint@typescript-eslint/no-floating-promises is reporting when promise is used not being used and suggests to add `void` to be explicit that value is ignore.
```
  const exampleFunction = () => { return new Promise(() => {});};
  void exampleFunction();
```

For such code rule S3735 will report that `void` should not be used. We should improve S3735 to not report when void argument has promise type.
 
Reported in https://community.sonarsource.com/t/typescript-void-should-not-be-used-triggers-wrongly/41893

yassin-kammoun-sonarsource:
Fixes #2629

### DIFF
--- a/eslint-bridge/src/rules/no-try-promise.ts
+++ b/eslint-bridge/src/rules/no-try-promise.ts
@@ -22,8 +22,7 @@
 import { Rule, SourceCode } from 'eslint';
 import * as estree from 'estree';
 import { TSESTree } from '@typescript-eslint/experimental-utils';
-import * as ts from 'typescript';
-import { isRequiredParserServices, RequiredParserServices, toEncodedMessage } from '../utils';
+import { isRequiredParserServices, isThenable, toEncodedMessage } from '../utils';
 
 type CallLikeExpression =
   | TSESTree.CallExpression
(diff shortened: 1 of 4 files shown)


vilchik-elena (approved):
LGTM. 

There are plenty issues like this on Peach, could you verify tomorrow that they are gone?

yassin-kammoun-sonarsource:
Sure thing.

navtejsohal7:
Hey Yassin, I am wondering, do we need to add anything into our code or sonarQube to make it work? OR is that something already implemented in sonarQube because I am still seeing this error a lot in my project

yassin-kammoun-sonarsource:
Hello @navtejsohal7,

This change was introduced in SonarQube 8.8. Maybe you are using an older version. If this is not the case, I suggest you report the false positive you are encountering in https://community.sonarsource.com.

mmospanenko:
Add void - sonarlint(typescript:S3735)
Remove void - @typescript-eslint/no-floating-promises

how to be?

PR #2951, opened 2022-02-09
Enable quick fixes for 22 rules with customized messages

saberduck:


eslint-bridge/src/quickfix.ts
```diff
@@ -21,7 +21,52 @@
 import { Linter, Rule as ESLintRule, SourceCode } from 'eslint';
 import { IssueLocation } from './analyzer';
 
-const rulesWithQuickFix = new Set(['no-extra-semi', 'no-unsafe-negation']);
+const rulesWithQuickFix = new Set([
```
yassin-kammoun-sonarsource, line 24: Since we have a constant `quickFixMessages` below, I would rename `rulesWithQuickFix` to `quickFixRules`.

eslint-bridge/src/quickfix.ts
```diff
@@ -21,7 +21,52 @@
...
+  'comma-dangle',
+  'eol-last',
+  'no-extra-semi',
+  'no-trailing-spaces',
+  'no-var',
+  'no-unsafe-negation',
```
yassin-kammoun-sonarsource, line 30:
```suggestion
  'no-unsafe-negation',
  'no-var',
```

eslint-bridge/src/quickfix.ts
```diff
@@ -21,7 +21,52 @@
...
+  'object-shorthand',
+  'prefer-const',
+  'prefer-template',
+  'quotes',
+  'radix',
+  'semi',
+  'sonarjs/prefer-immediate-return',
+  'sonarjs/prefer-while',
+  '@typescript-eslint/no-empty-interface',
+  '@typescript-eslint/no-explicit-any',
```
yassin-kammoun-sonarsource, line 40: These two rules are flagged as "not fixed nicely" in the spreadsheet. Was there some discussion about enabling them in the end ?

eslint-bridge/src/quickfix.ts
```diff
@@ -76,3 +121,11 @@ function fixToEdit(source: SourceCode, fix: ESLintRule.Fix): QuickFixEdit {
     text: fix.text,
   };
 }
+
+function getMessageForFix(ruleKey: string): string {
+  if (!quickFixMessages.has(ruleKey)) {
+    console.log(`DEBUG Missing message for ${ruleKey}`);
```
yassin-kammoun-sonarsource, line 127:
```suggestion
    console.log(`DEBUG Missing message for quick fix ${ruleKey}`);
```

eslint-bridge/src/quickfix.ts
```diff
@@ -21,7 +21,52 @@
...
+  'no-trailing-spaces',
+  'no-var',
+  'no-unsafe-negation',
+  'object-shorthand',
+  'prefer-const',
+  'prefer-template',
+  'quotes',
+  'radix',
+  'semi',
+  'sonarjs/prefer-immediate-return',
```
yassin-kammoun-sonarsource, line 37: Are we going to enable `sonarjs/no-inverted-boolean-check` as well given [this fix](https://github.com/SonarSource/eslint-plugin-sonarjs/pull/322) ? If yes, we'll need to release eslint-plugin-sonarjs and upgrade the dependency in SonarJS.

yassin-kammoun-sonarsource (approved):
I left minor suggestions and a couple of questions for you, but lgtm otherwise.

eslint-bridge/src/quickfix.ts
```diff
@@ -21,7 +21,52 @@
...
+  ['no-extra-semi', 'Remove extra semicolon'],
+  ['no-trailing-spaces', 'Remove trailing space'],
+  ['no-var', "Replace 'var' with 'let'"],
+  ['object-shorthand', 'Use shorthand property'],
+  ['prefer-const', "Replace with 'const'"],
+  ['quotes', 'Fix quotes'],
+  ['radix', 'Add 10 as radix'],
+  ['semi', 'Add semicolon'],
+  ['sonarjs/prefer-immediate-return', 'Return value immediately'],
+  ['sonarjs/prefer-while', 'Use while'],
```
vilchik-elena, line 61: `Replace with 'while' loop`

eslint-bridge/src/quickfix.ts
```diff
@@ -21,7 +21,52 @@
...
+  ['@typescript-eslint/no-empty-interface', 'Use type alias instead'],
+  ['@typescript-eslint/no-inferrable-types', 'Remove type declaration'],
+  ['@typescript-eslint/no-unnecessary-type-arguments', 'Remove type arguments'],
```
vilchik-elena, line 64: `Remove type argument` (I think it should be in signular)

eslint-bridge/src/quickfix.ts
```diff
@@ -21,7 +21,52 @@
...
+  ['@typescript-eslint/no-unnecessary-type-assertion', 'Remove type assertion'],
+  ['@typescript-eslint/prefer-namespace-keyword', "Use 'namespace' keyword"],
```
vilchik-elena, line 66: `Replace with 'namespace' keyword`

eslint-bridge/src/quickfix.ts
```diff
@@ -21,7 +21,52 @@
...
+  ['@typescript-eslint/prefer-readonly', "Use 'readonly'"],
```
vilchik-elena, line 67: `Add 'readonly'`

eslint-bridge/src/quickfix.ts
```diff
@@ -21,7 +21,52 @@
...
+  ['@typescript-eslint/no-non-null-assertion', "Use '.?' instead"],
```
vilchik-elena, line 68: `Replace with optional chaining '?.'`

vilchik-elena (changes requested):

eslint-bridge/src/quickfix.ts
```diff
@@ -21,7 +21,52 @@
...
+  ['no-trailing-spaces', 'Remove trailing space'],
+  ['no-var', "Replace 'var' with 'let'"],
+  ['object-shorthand', 'Use shorthand property'],
+  ['prefer-const', "Replace with 'const'"],
+  ['quotes', 'Fix quotes'],
+  ['radix', 'Add 10 as radix'],
+  ['semi', 'Add semicolon'],
+  ['sonarjs/prefer-immediate-return', 'Return value immediately'],
+  ['sonarjs/prefer-while', 'Use while'],
+  ['@typescript-eslint/no-empty-interface', 'Use type alias instead'],
```
vilchik-elena, line 62: `Replace with type alias`

eslint-bridge/src/quickfix.ts
```diff
@@ -21,7 +21,52 @@
...
+const rulesWithQuickFix = new Set([
+  'comma-dangle',
+  'eol-last',
+  'no-extra-semi',
+  'no-trailing-spaces',
+  'no-var',
+  'no-unsafe-negation',
+  'object-shorthand',
+  'prefer-const',
+  'prefer-template',
```
vilchik-elena, line 33: there is no message for this rule

eslint-bridge/src/quickfix.ts
vilchik-elena: yeah, we said that we prefer to keep only 'unknown' suggestion as suggestion to use 'never' has limited scope. But as it would be too much trouble we decided to provide fixes as they are.

vilchik-elena (approved):

eslint-bridge/src/quickfix.ts
```diff
@@ -76,3 +122,11 @@ function fixToEdit(source: SourceCode, fix: ESLintRule.Fix): QuickFixEdit {
     text: fix.text,
   };
 }
+
+function getMessageForFix(ruleKey: string): string {
+  if (!quickFixMessages.has(ruleKey)) {
+    console.log(`DEBUG Missing message for quick fix '${ruleKey}'`);
```
vilchik-elena, line 128: IMHO we should log error or even throw as I don't expect us to have rules without proper message.

PR #2953, opened 2022-02-11
Fix activation of quick fixes for rules from plugins and throw when no customized message

vilchik-elena:


yassin-kammoun-sonarsource (approved):
LGTM!

Maybe we could have to separate commits instead of squashing everything:

1. one commit for the activation fix
2. one commit for throwing error when missing message

PR #2954, opened 2022-02-14
Exclude import statements from duplication computation

### LINKED ISSUE #2920: Exclude import statements from duplication computation
Reported in https://community.sonarsource.com/t/import-considered-as-duplicated-lines-in-sonarqube/54141

yassin-kammoun-sonarsource:
Fixes #2920

### DIFF
--- a/eslint-bridge/src/runner/cpd.ts
+++ b/eslint-bridge/src/runner/cpd.ts
@@ -31,9 +31,13 @@ export default function getCpdTokens(sourceCode: SourceCode): { cpdTokens: CpdTo
   tokens.forEach(token => {
     let text = token.value;
 
-    if (text.trim().length === 0 || importTokens.includes(token)) {
+    if (text.trim().length === 0) {
       // for EndOfFileToken and JsxText tokens containing only whitespaces
-      // as well as import statements
+      return;
+    }
+
+    if (importTokens.includes(token)) {
+      // for tokens from import statements
       return;
     }
 
@@ -73,7 +77,6 @@ function extractTokens(sourceCode: SourceCode): {
         }
         break;
       case 'ImportDeclaration':
-      case 'ImportExpression':
         importTokens.push(...sourceCode.getTokens(tsNode as estree.Node));
         break;
     }
--- a/eslint-bridge/src/runner/cpd.ts
+++ b/eslint-bridge/src/runner/cpd.ts
@@ -26,13 +26,14 @@ import { Location } from './location';
 export default function getCpdTokens(sourceCode: SourceCode): { cpdTokens: CpdToken[] } {
   const cpdTokens: CpdToken[] = [];
   const tokens = sourceCode.ast.tokens;
-  const jsxTokens = extractJSXTokens(sourceCode);
+  const { jsxTokens, importTokens } = extractTokens(sourceCode);
 
   tokens.forEach(token => {
     let text = token.value;
 
-    if (text.trim().length === 0) {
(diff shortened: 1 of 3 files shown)


eslint-bridge/src/runner/cpd.ts
```diff
@@ -26,13 +26,14 @@ import { Location } from './location';
...
   const cpdTokens: CpdToken[] = [];
   const tokens = sourceCode.ast.tokens;
-  const jsxTokens = extractJSXTokens(sourceCode);
+  const { jsxTokens, importTokens } = extractTokens(sourceCode);
 
   tokens.forEach(token => {
     let text = token.value;
 
-    if (text.trim().length === 0) {
+    if (text.trim().length === 0 || importTokens.includes(token)) {
```
vilchik-elena, line 34: that's subtle but I would prefer to have a separate condition and return for import tokens as it's different from an empty token.

eslint-bridge/src/runner/cpd.ts
```diff
@@ -57,15 +58,27 @@ export default function getCpdTokens(sourceCode: SourceCode): { cpdTokens: CpdTo
...
-    if (tsNode.type === 'JSXAttribute' && tsNode.value?.type === 'Literal') {
-      tokens.push(...sourceCode.getTokens(tsNode.value as estree.Node));
+    switch (tsNode.type) {
+      case 'JSXAttribute':
+        if (tsNode.value?.type === 'Literal') {
+          jsxTokens.push(...sourceCode.getTokens(tsNode.value as estree.Node));
+        }
+        break;
+      case 'ImportDeclaration':
+      case 'ImportExpression':
```
vilchik-elena, line 76: Are you sure about import expression? I checked quickly, to me it seems something which should not be excluded

vilchik-elena (changes requested):
I also commented on the ticket that import statements
Should not increment the "statements" metric
Should not be considered as "executable" lines

TBH now I don't remember why I wrote that, probably for consistency. Could you check with another analyzer (e.g. sonar-java) what they do with import statements?

yassin-kammoun-sonarsource:
> I also commented on the ticket that import statements Should not increment the "statements" metric Should not be considered as "executable" lines

Yes, I saw those comments, and we are already ignoring import declarations for these metrics.

> TBH now I don't remember why I wrote that, probably for consistency. Could you check with another analyzer (e.g. sonar-java) what they do with import statements?

Just checked sonar-java; they don't consider import statements as statements and they don't count them as executable lines.

eslint-bridge/src/runner/cpd.ts
yassin-kammoun-sonarsource: Makes sense.

eslint-bridge/tests/runner/cpd.test.ts
```diff
@@ -129,6 +129,18 @@ it('should preserve string literals inside JSX constructs', () => {
   expect(result).toContainEqual(token(7, 16, 7, 21, 'LITERAL'));
 });
 
+it('should exclude import statements', () => {
+  const result = actual(`
+  import a from "x";
+  import * as b from "y";
+  import { c } from "z";
+  import "lib";
+  'hello';
```
vilchik-elena, line 138: should we add import expression to be explicit that it's kept?

yassin-kammoun-sonarsource: That's a good point. I will refine the test case's snippet.

vilchik-elena (approved):

PR #2958, opened 2022-02-14
Introduce back sonar.typescript.lcov.reportPaths as alias for sonar.javascript.lcov.reportPaths

### LINKED ISSUE #2938: Introduce sonar.typescript.lcov.reportPaths as alias for sonar.javascript.lcov.reportPaths
Many users still use `sonar.typescript.lcov.reportPaths` , we should support it as alias for `sonar.javascript.lcov.reportPaths`

yassin-kammoun-sonarsource:
Fixes #2938

### DIFF
--- a/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/lcov/CoverageSensor.java
+++ b/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/lcov/CoverageSensor.java
@@ -48,7 +48,7 @@ public class CoverageSensor implements Sensor {
   public void describe(SensorDescriptor descriptor) {
     descriptor
       .onlyOnLanguages(JavaScriptLanguage.KEY, TypeScriptLanguage.KEY)
-      .onlyWhenConfiguration(conf -> conf.hasKey(JavaScriptPlugin.LCOV_REPORT_PATHS))
+      .onlyWhenConfiguration(conf -> conf.hasKey(JavaScriptPlugin.LCOV_REPORT_PATHS) || conf.hasKey(JavaScriptPlugin.LCOV_REPORT_PATHS_ALIAS))
       .name("JavaScript/TypeScript Coverage")
       .onlyOnFileType(Type.MAIN);
   }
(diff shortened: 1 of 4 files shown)


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/lcov/CoverageSensor.java
```diff
@@ -48,14 +48,18 @@
...
       .name("JavaScript/TypeScript Coverage")
       .onlyOnFileType(Type.MAIN);
   }
 
   @Override
   public void execute(SensorContext context) {
     Set<String> reports = new HashSet<>(Arrays.asList(context.config().getStringArray(JavaScriptPlugin.LCOV_REPORT_PATHS)));
+    reports.addAll(Arrays.asList(context.config().getStringArray(JavaScriptPlugin.LCOV_REPORT_PATHS_ALIAS)));
+    if (context.config().hasKey(JavaScriptPlugin.LCOV_REPORT_PATHS) && context.config().hasKey(JavaScriptPlugin.LCOV_REPORT_PATHS_ALIAS)) {
+      LOG.warn(String.format("Merging coverage reports from %s and %s.", JavaScriptPlugin.LCOV_REPORT_PATHS, JavaScriptPlugin.LCOV_REPORT_PATHS_ALIAS));
```
vilchik-elena, line 61: why warning? I would put `info`

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/lcov/CoverageSensor.java
```diff
@@ -48,14 +48,18 @@
...
+      .onlyWhenConfiguration(conf -> conf.hasKey(JavaScriptPlugin.LCOV_REPORT_PATHS) || conf.hasKey(JavaScriptPlugin.LCOV_REPORT_PATHS_ALIAS))
       .name("JavaScript/TypeScript Coverage")
       .onlyOnFileType(Type.MAIN);
   }
 
   @Override
   public void execute(SensorContext context) {
     Set<String> reports = new HashSet<>(Arrays.asList(context.config().getStringArray(JavaScriptPlugin.LCOV_REPORT_PATHS)));
+    reports.addAll(Arrays.asList(context.config().getStringArray(JavaScriptPlugin.LCOV_REPORT_PATHS_ALIAS)));
+    if (context.config().hasKey(JavaScriptPlugin.LCOV_REPORT_PATHS) && context.config().hasKey(JavaScriptPlugin.LCOV_REPORT_PATHS_ALIAS)) {
```
vilchik-elena, line 60: I feel now it will not be redundant to log the used property even if only one is used. It might ease the debugging in future

vilchik-elena (changes requested):

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/lcov/CoverageSensor.java
yassin-kammoun-sonarsource: I am not convinced about logging which property is used. I mean, if we want to debug, all we need is the project configuration to know which property was set by the user. If we start logging that one particular property is used, why shouldn't we do the same for others ?

vilchik-elena: there are many sources of property values (UI, sonar-project.properties, CI config), so it might be not so easy to get actual config. As for other properties, we don't have aliases there :)
@saberduch WDYT?

saberduck: I think it's ok to log in `DEBUG` when property is used, it's really easier when someone posts the logs on community

vilchik-elena (approved):

PR #2960, opened 2022-02-14
FP S1135: surrounding accented characters should be considered

### LINKED ISSUE #2957: FP S1135: surrounding accented characters should be considered
The rule currently ignores todo occurrences iff it is surrounded by a character accepted by the regular expression [A-Za-z]. This is incorrect as the rule could reports false-positives when there are surrounding accented characters, like for example:

```javascript
class C {
  // Método que calcula a média dos números
  m() { /* ... */}
}
```

Here `Método` is flagged because the _todo_ occurrence is preceded by _é_, that is, an accented character.

Reported in https://community.sonarsource.com/t/l25-typescript-catches-todo-in-part-of-other-words/56672

yassin-kammoun-sonarsource:
Fixes #2957

### DIFF
--- a/eslint-bridge/src/rules/todo-tag.ts
+++ b/eslint-bridge/src/rules/todo-tag.ts
@@ -24,7 +24,7 @@ import { TSESTree } from '@typescript-eslint/experimental-utils';
 
 const todoMessage = 'Complete the task associated to this "TODO" comment.';
 const todoPattern = 'todo';
-const letterPattern = /[A-Za-zÀ-ÖØ-öø-ÿ]/;
+const letterPattern = /[\p{Letter}]/u;
 
 export const rule: Rule.RuleModule = {
   create(context: Rule.RuleContext) {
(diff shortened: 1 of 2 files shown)


eslint-bridge/src/rules/todo-tag.ts
```diff
@@ -24,6 +24,7 @@ import { TSESTree } from '@typescript-eslint/experimental-utils';
 
 const todoMessage = 'Complete the task associated to this "TODO" comment.';
 const todoPattern = 'todo';
+const letterPattern = /[A-Za-zÀ-ÖØ-öø-ÿ]/;
```
vilchik-elena, line 27: I found this one `/[\p{Letter}]/u`. Seems more reliable and matches `ā` not matched by your regex

vilchik-elena (changes requested):

vilchik-elena:
@yassin-kammoun-sonarsource could you add a code example to the fixed issue?

eslint-bridge/src/rules/todo-tag.ts
yassin-kammoun-sonarsource: Thanks.

vilchik-elena (approved):

PR #2962, opened 2022-02-15
Fix FP ('no-try-promise'): consider yield as an alternative to await

### LINKED ISSUE #2683: Consider `yield` in S4822 (no-try-promise)
```js
function * foo() {
  try {
    yield somePromise(); // FP
  } catch (e) {
  }
}
```

`yield` is an alternative to the `await`

Reported in https://community.sonarsource.com/t/typescript-s4822-yield-keyword-should-be-accepted/26778

yassin-kammoun-sonarsource:
Fixes #2683

### DIFF
--- a/eslint-bridge/src/rules/no-try-promise.ts
+++ b/eslint-bridge/src/rules/no-try-promise.ts
@@ -71,13 +71,7 @@ function visitTryStatement(
         return;
       }
 
-      if (
-        (callLikeExpr.parent &&
-          (callLikeExpr.parent.type === 'AwaitExpression' ||
-            callLikeExpr.parent.type === 'YieldExpression')) ||
-        isThened(callLikeExpr) ||
-        isCatch(callLikeExpr)
-      ) {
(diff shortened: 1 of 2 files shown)


eslint-bridge/src/rules/no-try-promise.ts
```diff
@@ -72,7 +72,9 @@ function visitTryStatement(
       }
 
       if (
-        (callLikeExpr.parent && callLikeExpr.parent.type === 'AwaitExpression') ||
+        (callLikeExpr.parent &&
```
vilchik-elena, line 75: this condition was already complex:) how about extracting part of it?

eslint-bridge/tests/rules/no-try-promise.test.ts
```diff
@@ -85,6 +85,18 @@ ruleTester.run(`Promise rejections should not be caught by 'try' block`, rule, {
       }
       `,
     },
+    {
+      code: `
+      function returningPromise() { return Promise.reject(); }
+      async function okWithYield() {
```
vilchik-elena, line 91: let's add `*` to function to be explicit that it's generator

PR #2964, opened 2022-02-16
Match S1172 (`no-unused-function-argument`) behavior to TypeScript compiler for JS and enable for TS

### LINKED ISSUE #1832: Change S1172 (UnusedFunctionArgument) to match TS compiler behavior (new rule for TS)
When migrating rule [S1172](https://jira.sonarsource.com/browse/RSPEC-1172) to the new front-end (#1822), we observed that the current rule scope is fine for JS, but not good enough to enable it for TS.

We also realized that TS compiler provides an option (noUnusedParameters), to report unused parameter. Ideally, our rule should match this behavior. Since we want to avoid having two implementations, and the fact that this also works well for JS, we should make it as the default one.

**When this behaviour is changed, we should add the rule to the recommended profile.**

_For reference, the behavior of TS compiler is the following:_
- Report any unused parameter, including ones that are before an other used parameter.
- Parameter prefixed with underscore are ignored.

yassin-kammoun-sonarsource:
Fixes #1832

Be careful not to squash commits.


yassin-kammoun-sonarsource:
Rule description's update: https://github.com/SonarSource/rspec/pull/827.

eslint-bridge/src/rules/no-unused-function-argument.ts
```diff
@@ -66,13 +66,13 @@ function reportUnusedArgument(
...
-  while (i >= 0 && isUnusedVariable(parametersVariable[i])) {
-    context.report({
-      message: `Remove the unused function parameter "${parametersVariable[i].name}".`,
-      node: parametersVariable[i].identifiers[0],
-    });
-    i--;
+  for (const param of parametersVariable) {
+    if (isUnusedVariable(param) && !isIgnoredParameter(param)) {
+      context.report({
+        message: `Remove the unused function parameter "${param.name}".`,
```
vilchik-elena, line 72: I think we should change the message to `Remove the unused function parameter "${param.name}" or rename it to "_${param.name}" to make intention explicit.`

eslint-bridge/tests/rules/no-unused-function-argument.test.ts
```diff
@@ -116,7 +115,7 @@ ruleTester.run('Unused function parameters should be removed', rule, {
       code: `function fun(a, b, c) {
              b = 1;
            }`,
-      errors: 1,
+      errors: 2,
```
vilchik-elena, line 128: how about asserting messages to be sure we report the right variables?

vilchik-elena (changes requested):
I think we should migrate the rule to messageId or to comment based test :)

eslint-bridge/src/rules/no-unused-function-argument.ts
yassin-kammoun-sonarsource: It's quite long, but I am fine with that change :)

vilchik-elena (approved):

PR #2969, opened 2022-02-18
Call setQuickFixAvailable for SQ

saberduck:
It seems this can't be easily tested, because SonarLint doesn't implement the API


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/AnalysisProcessor.java
```diff
@@ -232,7 +232,8 @@ void saveIssue(EslintBridgeServer.Issue issue) {
       newIssue.gap(issue.cost);
     }
 
-    if (isQuickFixCompatible() && issue.quickFixes != null) {
+    if (isQuickFixCompatible() && issue.quickFixes != null && !issue.quickFixes.isEmpty()) {
+      newIssue.setQuickFixAvailable(true);
```
vilchik-elena, line 236: this API is available with SQ only, it's not the same condition as in `isQuickFixCompatible`

vilchik-elena (changes requested):
Build is failing

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/AnalysisProcessor.java
saberduck: that's just stupid 🤦  fixing it

vilchik-elena (approved):

vilchik-elena:
@saberduck I think we should add integration test for SQ with quick fix and run ITs on SQ[9.0] to make SC support more robust. WDYT?

saberduck:
> @saberduck I think we should add integration test for SQ with quick fix and run ITs on SQ[9.0] to make SC support more robust. WDYT?

OK, good idea. But maybe we should test LTS and not 9.0?

vilchik-elena (changes requested):
As discussed we need ITs for QF on SQ LTS

vilchik-elena (approved):

PR #2973, opened 2022-02-21
Prepare next development iteration

yassin-kammoun-sonarsource:


vilchik-elena:
We do a major release as we decided to keep major version in sync with SQ releases.

vilchik-elena (approved):

PR #2975, opened 2022-02-22
Migrate S1135, S1128, S1154, S3758, S117, S3735, S4423, S2817, S5689, S4817 to messageId

yassin-kammoun-sonarsource:


eslint-bridge/src/rules/variable-name.ts
```diff
@@ -30,6 +30,11 @@ interface FunctionLike {
 }
 
 export const rule: Rule.RuleModule = {
+  meta: {
+    messages: {
+      renameSymbol: `Rename this {{symbolType}} "{{symbol}}" to match the regular expression {{format}}.`,
```
vilchik-elena, line 35: I would not use template string here

vilchik-elena (approved):

eslint-bridge/src/rules/variable-name.ts
vilchik-elena: it is so minor, I am merging PR

PR #2979, opened 2022-02-23
Migrate S4043, S1121, S3358, S881, S1226, S1533, S4165, S3533, S4324, S105 to messageId

yassin-kammoun-sonarsource:


eslint-bridge/src/rules/no-misleading-array-reverse.ts
```diff
@@ -65,14 +73,14 @@ export const rule: Rule.RuleModule = {
   },
 };
 
-function formatMessage(mutatingMethod: string) {
+function formatMethod(mutatingMethod: string) {
   let mutatingMethodText;
   if (mutatingMethod.startsWith('"') || mutatingMethod.startsWith("'")) {
     mutatingMethodText = mutatingMethod.substr(1, mutatingMethod.length - 2);
```
vilchik-elena, line 79: I would return directly without local var

vilchik-elena (approved):

vilchik-elena (approved):

PR #2982, opened 2022-02-23
Migrate S1527, S3531, S5691, S2692, S1439, S4622, S1994, S2871, S6268, S2870 to messageId

yassin-kammoun-sonarsource:


eslint-bridge/src/rules/misplaced-loop-counter.ts
```diff
@@ -33,6 +33,11 @@ class ForInfo {
 }
 
 export const rule: Rule.RuleModule = {
+  meta: {
+    messages: {
+      misplacedCounter: `This loop's stop condition tests "{{test}}" but the incrementer updates "{{update}}".`,
```
vilchik-elena, line 38: I suggest to use regular string template here

vilchik-elena (approved):

eslint-bridge/src/rules/misplaced-loop-counter.ts
vilchik-elena: it's so minor, I merge PR

PR #2984, opened 2022-02-23
Migrate S4798, S1472, S101, S3525, S1523, S3616, S1848, S5693, S2255, S3798 to messageId

yassin-kammoun-sonarsource:


vilchik-elena (approved):
Let's revert tests

vilchik-elena (approved):

PR #2985, opened 2022-02-23
Migrate S2189, S2681, S3782, S3796, S1528, S1529, S3524 to messageId

yassin-kammoun-sonarsource:


eslint-bridge/src/rules/bitwise-operators.ts
```diff
@@ -42,6 +42,12 @@ const BITWISE_OPERATORS = [
 ];
 
 export const rule: Rule.RuleModule = {
+  meta: {
+    messages: {
+      reviewBitwise:
+        'Review this use of bitwise "{operator}" operator; conditional "{{intended}}" might have been intended.',
```
vilchik-elena, line 48: why only one set of curly braces?

eslint-bridge/src/rules/no-unenclosed-multiline-block.ts
```diff
@@ -140,9 +155,11 @@ function raiseInlineAndIndentedIssue(
   context: Rule.RuleContext,
 ) {
   context.report({
-    message:
-      `This line will not be executed ${chainedStatements.includedStatementQualifier()}; only the first statement will be. ` +
-      `The rest will execute ${chainedStatements.excludedStatementsQualifier()}.`,
+    messageId: 'onlyFirstStatement',
```
vilchik-elena, line 158: should be `onlyFirstLine`?

vilchik-elena (changes requested):

eslint-bridge/src/rules/no-unenclosed-multiline-block.ts
yassin-kammoun-sonarsource: I don't think so. It raises issues on the following pattern:

```javascript

if (condition) action1();
  action2();
```

vilchik-elena:
Closing PR as we realized that we don't have to migrate to message IDs

PR #3038, opened 2022-03-14
Add quick fix for S1172 ('no-unused-function-argument')

### LINKED ISSUE #3033: Add quick fix for S1172: Unused function parameters should be removed
SonarJS rule.

yassin-kammoun-sonarsource:
Fixes #3033

### DIFF
--- a/eslint-bridge/src/rules/no-unused-function-argument.ts
+++ b/eslint-bridge/src/rules/no-unused-function-argument.ts
@@ -29,7 +29,8 @@ export const rule: Rule.RuleModule = {
     messages: {
       removeOrRenameParameter:
         'Remove the unused function parameter "{{param}}" or rename it to "_{{param}}" to make intention explicit.',
-      removeOrRenameParameterFix: 'Rename {{param}} to _{{param}}',
+      suggestRemoveParameter: 'Remove {{param}} (beware of call sites)',
+      suggestRenameParameter: 'Rename {{param}} to _{{param}}',
     },
   },
   create(context: Rule.RuleContext) {
@@ -74,7 +75,7 @@ function reportUnusedArgument(
     parametersVariable = parametersVariable.filter(v => v.name !== functionId.name);
   }
 
-  for (const param of parametersVariable) {
+  for (const [idx, param] of parametersVariable.entries()) {
     if (isUnusedVariable(param) && !isIgnoredParameter(param) && !isParameterProperty(param)) {
       context.report({
         messageId: 'removeOrRenameParameter',
@@ -82,18 +83,46 @@ function reportUnusedArgument(
         data: {
           param: param.name,
         },
-        suggest: [
-          {
-            messageId: 'removeOrRenameParameterFix',
-            data: {
-              param: param.name,
-            },
-            fix: fixer => fixer.replaceText(param.identifiers[0], `_${param.name}`),
-          },
-        ],
+        suggest: getSuggestions(param, idx),
       });
     }
   }
+}
+
+function getSuggestions(paramVariable: Scope.Variable, paramIndex: number) {
+  const paramIdentifier = paramVariable.identifiers[0];
+  const suggestions: Rule.SuggestionReportDescriptor[] = [
+    {
+      messageId: 'suggestRenameParameter',
+      data: {
+        param: paramVariable.name,
+      },
+      fix: fixer => fixer.replaceText(paramIdentifier, `_${paramVariable.name}`),
+    },
+  ];
+  const funct = paramVariable.defs[0].node as TSESTree.FunctionLike;
+  if ((paramIdentifier as TSESTree.Node).parent === funct) {
+    suggestions.push({
+      messageId: 'suggestRemoveParameter',
+      data: {
+        param: paramVariable.name,
+      },
+      fix: fixer => {
+        if (funct.params.length === 1) {
+          return fixer.remove(funct.params[paramIndex] as estree.Node);
+        } else if (funct.params.length - 1 === paramIndex) {
+          const [, start] = funct.params[paramIndex - 1].range;
+          const [, end] = funct.params[paramIndex].range;
+          return fixer.removeRange([start, end]);
+        } else {
+          const [start] = funct.params[paramIndex].range;
+          const [end] = funct.params[paramIndex + 1].range;
+          return fixer.removeRange([start, end]);
+        }
+      },
+    });
+  }
+  return suggestions;
 }
 
 function isUnusedVariable(variable: Scope.Variable) {
--- a/eslint-bridge/src/rules/no-unused-function-argument.ts
+++ b/eslint-bridge/src/rules/no-unused-function-argument.ts
@@ -23,6 +23,11 @@ import { Rule, Scope } from 'eslint';
 import * as estree from 'estree';
 import { TSESTree } from '@typescript-eslint/experimental-utils';
 
+type FunctionLike =
+  | TSESTree.ArrowFunctionExpression
+  | TSESTree.FunctionDeclaration
+  | TSESTree.FunctionExpression;
+
 export const rule: Rule.RuleModule = {
   meta: {
     hasSuggestions: true,
@@ -83,13 +88,13 @@ function reportUnusedArgument(
         data: {
           param: param.name,
         },
-        suggest: getSuggestions(param),
+        suggest: getSuggestions(param, context),
       });
     }
   }
 }
 
-function getSuggestions(paramVariable: Scope.Variable) {
+function getSuggestions(paramVariable: Scope.Variable, context: Rule.RuleContext) {
   const paramIdentifier = paramVariable.identifiers[0];
   const suggestions: Rule.SuggestionReportDescriptor[] = [
     {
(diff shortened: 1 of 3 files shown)


eslint-bridge/src/rules/no-unused-function-argument.ts
```diff
@@ -80,11 +83,49 @@ function reportUnusedArgument(
...
+  const suggestions: Rule.SuggestionReportDescriptor[] = [
+    {
+      messageId: 'suggestRenameParameter',
+      data: {
+        param: paramVariable.name,
+      },
+      fix: fixer => fixer.replaceText(paramIdentifier, `_${paramVariable.name}`),
+    },
+  ];
+  const funct = paramVariable.defs[0].node as TSESTree.FunctionLike;
```
vilchik-elena, line 103: why such name for variable?

eslint-bridge/src/rules/no-unused-function-argument.ts
```diff
@@ -80,11 +83,49 @@ function reportUnusedArgument(
...
 
+function getSuggestions(paramVariable: Scope.Variable) {
+  const paramIdentifier = paramVariable.identifiers[0];
+  const suggestions: Rule.SuggestionReportDescriptor[] = [
+    {
+      messageId: 'suggestRenameParameter',
+      data: {
+        param: paramVariable.name,
+      },
+      fix: fixer => fixer.replaceText(paramIdentifier, `_${paramVariable.name}`),
```
vilchik-elena, line 100: may be using `insertTextBefore` will be more efficient?

eslint-bridge/src/rules/no-unused-function-argument.ts
vilchik-elena, line 103: why such name for variable?

eslint-bridge/tests/rules/no-unused-function-argument.test.ts
```diff
@@ -196,5 +196,54 @@ ruleTester.run('Unused function parameters should be removed', rule, {
             }`,
       errors: 1,
     },
+    {
```
vilchik-elena, line 199: can we add unit test with parameter type to verify that type is also removed?

eslint-bridge/src/rules/no-unused-function-argument.ts
```diff
@@ -80,11 +83,49 @@ function reportUnusedArgument(
...
+  if ((paramIdentifier as TSESTree.Node).parent === funct) {
+    suggestions.push({
+      messageId: 'suggestRemoveParameter',
+      data: {
+        param: paramVariable.name,
+      },
+      fix: fixer => {
+        const paramIndex = funct.params.indexOf(paramIdentifier as TSESTree.Parameter);
+        if (funct.params.length === 1) {
+          return fixer.remove(funct.params[paramIndex] as estree.Node);
```
vilchik-elena, line 113: what will happen with trailing comma?

eslint-bridge/src/rules/no-unused-function-argument.ts
```diff
@@ -80,11 +83,49 @@ function reportUnusedArgument(
...
+        } else if (funct.params.length - 1 === paramIndex) {
+          const [, start] = funct.params[paramIndex - 1].range;
+          const [, end] = funct.params[paramIndex].range;
+          return fixer.removeRange([start, end]);
```
vilchik-elena, line 117: this should be fine for trailing comma but can we add a test for it?

vilchik-elena (changes requested):

eslint-bridge/src/rules/no-unused-function-argument.ts
yassin-kammoun-sonarsource: `funct` for `function`, but the latter is a keyword.

vilchik-elena: ah, indeed:)) it's just usually we use `fun` or `func`. I have never seen `funct`

eslint-bridge/src/rules/no-unused-function-argument.ts
```diff
@@ -80,11 +88,90 @@ function reportUnusedArgument(
...
+          case 'ArrowFunctionExpression': {
+            const param = func.params[paramIndex] as estree.Node;
+            const token = context.getSourceCode().getTokenAfter(param);
+            if (token?.value === ')') {
+              return fixer.remove(param);
+            } else {
+              return fixer.replaceText(param, '()');
+            }
+          }
+          case 'FunctionDeclaration': {
```
vilchik-elena, line 139: this case and next one are very similar, can't we generalize the logic?

eslint-bridge/src/rules/no-unused-function-argument.ts
```diff
@@ -23,11 +23,19 @@ import { Rule, Scope } from 'eslint';
...
+  | TSESTree.FunctionDeclaration
+  | TSESTree.FunctionExpression;
+
 export const rule: Rule.RuleModule = {
   meta: {
+    hasSuggestions: true,
     messages: {
       removeOrRenameParameter:
         'Remove the unused function parameter "{{param}}" or rename it to "_{{param}}" to make intention explicit.',
+      suggestRemoveParameter: 'Remove {{param}} (beware of call sites)',
```
vilchik-elena, line 37: I think we should wrap parameter names into quotes.

eslint-bridge/src/rules/no-unused-function-argument.ts
```diff
@@ -80,11 +88,90 @@ function reportUnusedArgument(
...
+    data: {
+      param: paramVariable.name,
+    },
+    fix: fixer => {
+      const paramIndex = func.params.indexOf(paramIdentifier as TSESTree.Parameter);
+      if (func.params.length === 1) {
+        switch (func.type) {
+          case 'ArrowFunctionExpression': {
+            const param = func.params[paramIndex] as estree.Node;
+            const token = context.getSourceCode().getTokenAfter(param);
```
vilchik-elena, line 132: theoretically there could be comma (`(a,) => foo()`)

eslint-bridge/tests/rules/no-unused-function-argument.test.ts
```diff
@@ -196,5 +196,180 @@ ruleTester.run('Unused function parameters should be removed', rule, {
...
+      code: `function fun(a, b, c, d, {e}, f, g) { b = d = f; }`,
+      errors: [
+        {
+          suggestions: [
```
vilchik-elena, line 203: I think you don't have to assert all suggestions and their messages. Assert only what matters for the case, otherwise it's hard to follow

vilchik-elena (changes requested):
And same thing as with another PR, I think we should keep the trailing comma style

eslint-bridge/src/rules/no-unused-function-argument.ts
```diff
@@ -80,11 +88,83 @@ function reportUnusedArgument(
...
+      param: paramVariable.name,
+    },
+    fix: fixer => {
+      const paramIndex = func.params.indexOf(paramIdentifier as TSESTree.Parameter);
+      const param = func.params[paramIndex] as estree.Node;
+      if (func.params.length === 1) {
+        const openingParenthesis = context
+          .getSourceCode()
+          .getTokenBefore(param, token => token.value === '(');
+        const closingParenthesis = context
```
vilchik-elena, line 133: this will not work for `a => foo()` :(

eslint-bridge/src/rules/no-unused-function-argument.ts
```diff
@@ -80,11 +88,83 @@ function reportUnusedArgument(
...
+          .getSourceCode()
+          .getTokenAfter(param, token => token.value === ')');
+        let [start, end] = param.range!;
+        if (openingParenthesis) {
+          start = openingParenthesis.range[0];
+        }
+        if (closingParenthesis) {
+          end = closingParenthesis.range[1];
+        }
```
vilchik-elena, line 142:
```suggestion
        const openingParenthesis = context
          .getSourceCode()
          .getTokenBefore(param);
        const closingParenthesis = context
          .getSourceCode()
          .getTokenAfter(param, token => token.value === ')');
        let [start, end] = param.range!;
        if (openingParenthesis && openingParenthesis.value === '(') {
          start = openingParenthesis.range[0];
          end = closingParenthesis!.range[1];
        }
```

vilchik-elena (changes requested):

eslint-bridge/tests/rules/no-unused-function-argument.test.ts
```diff
@@ -196,5 +196,163 @@ ruleTester.run('Unused function parameters should be removed', rule, {
...
+    },
+    {
+      // arrow with trailing
+      code: `( a, ) => {}`,
+      errors: [
+        {
+          suggestions: [
+            {
+              desc: 'Rename "a" to "_a"',
+              output: '( _a, ) => {}',
```
vilchik-elena, line 347: I see we can't remove this suggestion from assertion, but let's at least remove the messages to reduce some readablility noise

vilchik-elena (approved):

PR #3040, opened 2022-03-14
Add quick fix for S4623: "undefined" should not be passed as the value of optional parameters

### LINKED ISSUE #3028: Add quick fix for S4623: "undefined" should not be passed as the value of optional parameters
SonarJS rule.

yassin-kammoun-sonarsource:
Fixes #3028


eslint-bridge/src/rules/no-undefined-argument.ts
```diff
@@ -47,6 +49,21 @@ export const rule: Rule.RuleModule = {
             context.report({
               messageId: 'removeUndefined',
               node: lastArgument,
+              suggest: [
+                {
+                  messageId: 'suggestRemoveUndefined',
+                  fix: fixer => {
+                    const singleArgument = call.arguments.length === 1;
```
vilchik-elena, line 56: I would inline this

eslint-bridge/src/rules/no-undefined-argument.ts
```diff
@@ -27,8 +27,10 @@ import * as ts from 'typescript';
 
 export const rule: Rule.RuleModule = {
   meta: {
+    hasSuggestions: true,
     messages: {
       removeUndefined: 'Remove this redundant "undefined".',
+      suggestRemoveUndefined: 'Remove "undefined"',
```
vilchik-elena, line 33: I would say `Remove this redundant argument`

eslint-bridge/tests/rules/no-undefined-argument.test.ts
```diff
@@ -85,15 +85,56 @@ ruleTester.run(`"undefined" should not be passed as the value of optional parame
       foo(1, undefined, undefined);
       foo(1, undefined);
       `,
-      errors: 2,
+      errors: [
+        {
+          messageId: 'removeUndefined',
```
vilchik-elena, line 90: I would drop this extra assertion of issue message

eslint-bridge/tests/rules/no-undefined-argument.test.ts
```diff
@@ -85,15 +85,56 @@ ruleTester.run(`"undefined" should not be passed as the value of optional parame
...
+              desc: 'Remove "undefined"',
+              output: `
+      let funcExprWithOneParameter = function(p = 42) {}
+      funcExprWithOneParameter();
+      funcExprWithOneParameter(1);
+      `,
+            },
+          ],
+        },
+      ],
```
vilchik-elena, line 137: Can we have test with trailing comma?

vilchik-elena (changes requested):

eslint-bridge/tests/rules/no-undefined-argument.test.ts
```diff
@@ -136,5 +133,24 @@ ruleTester.run(`"undefined" should not be passed as the value of optional parame
...
+      fun(undefined,);
+      `,
+      errors: [
+        {
+          suggestions: [
+            {
+              desc: 'Remove this redundant argument',
+              output: `
+      function fun(p = 42) {}
+      fun(,);
```
vilchik-elena, line 148: this is syntax error, we can't have such fix

vilchik-elena: add test with trailing comma when there are several arguments

vilchik-elena (changes requested):

eslint-bridge/tests/rules/no-undefined-argument.test.ts
```diff
@@ -85,15 +85,105 @@ ruleTester.run(`"undefined" should not be passed as the value of optional parame
...
+      fun(1, 2, undefined, );
+      `,
+      errors: [
+        {
+          suggestions: [
+            {
+              desc: 'Remove this redundant argument',
+              output: `
+      function fun(p, q, r = 42) {}
+      fun(1, 2);
```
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: I am fine with the behaviour you are suggesting.

vilchik-elena (changes requested):
@yassin-kammoun-sonarsource I think you misunderstood me. I wanted to keep the styling: so if trailing comma is there we should keep it, if there is no trailing comma, there should be none in the result as well.

vilchik-elena (approved):

PR #3044, opened 2022-03-15
Add quick fix for S2990 ('no-global-this')

### LINKED ISSUE #3013: Add quick fix for S2990: The global "this" object should not be used
SonarJS rule.

yassin-kammoun-sonarsource:
Fixes #3013

### DIFF
--- a/eslint-bridge/tests/rules/no-global-this.test.ts
+++ b/eslint-bridge/tests/rules/no-global-this.test.ts
@@ -104,6 +104,10 @@ const testCases = {
               desc: 'Remove "this"',
               output: 'foo = bar;',
             },
+            {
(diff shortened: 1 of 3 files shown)


eslint-bridge/tests/rules/no-global-this.test.ts
```diff
@@ -95,6 +95,40 @@ const testCases = {
       }`,
       errors: 2,
     },
+    {
+      code: `this.foo = bar;`,
```
vilchik-elena, line 99: I think we should also suggest to replace with `window`

vilchik-elena (changes requested):

vilchik-elena (approved):

PR #3045, opened 2022-03-16
Add quick fix for S3403 ('different-types-comparison')

### LINKED ISSUE #3014: Add quick fix for S3403: Strict equality operators should not be used with dissimilar types
SonarJS rule.

yassin-kammoun-sonarsource:
Fixes #3014

### DIFF
--- a/eslint-bridge/src/rules/different-types-comparison.ts
+++ b/eslint-bridge/src/rules/different-types-comparison.ts
@@ -26,6 +26,7 @@ import { isRequiredParserServices, getTypeFromTreeNode, toEncodedMessage } from 
 
 export const rule: Rule.RuleModule = {
   meta: {
+    hasSuggestions: true,
     schema: [
       {
         // internal parameter for rules having secondary locations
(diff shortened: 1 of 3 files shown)


eslint-bridge/src/rules/different-types-comparison.ts
```diff
@@ -94,6 +95,17 @@ export const rule: Rule.RuleModule = {
               .getSourceCode()
               .getTokensBetween(left, right)
               .find(token => token.type === 'Punctuator' && token.value === operator)!.loc,
+            suggest: [
+              {
+                desc: `Replace "${actual}" with "${expected}"`,
+                fix: fixer => {
+                  const operatorToken = context
```
vilchik-elena, line 102: we compute it twice (see `loc`)

vilchik-elena (changes requested):

eslint-bridge/src/rules/different-types-comparison.ts
yassin-kammoun-sonarsource: Nice catch

vilchik-elena (approved):

PR #3046, opened 2022-03-16
Add quick fix for S3415 ('inverted-assertion-arguments')

### LINKED ISSUE #3015: Add quick fix for S3415: Assertion arguments should be passed in the correct order
SonarJS rule.

yassin-kammoun-sonarsource:
Fixes #3015

### DIFF
--- a/eslint-bridge/tests/rules/inverted-assertion-arguments.test.ts
+++ b/eslint-bridge/tests/rules/inverted-assertion-arguments.test.ts
@@ -20,8 +20,10 @@
 import { rule } from 'rules/inverted-assertion-arguments';
 import { RuleTester } from 'eslint';
 
-const ruleTester = new RuleTester({ parserOptions: { ecmaVersion: 2018, sourceType: 'module' } });
+// the behaviour of the rule is tested in another file; here we are just testing its quick fixes
+// until https://github.com/SonarSource/SonarJS/issues/3047 is done
 
+const ruleTester = new RuleTester({ parserOptions: { ecmaVersion: 2018, sourceType: 'module' } });
 ruleTester.run('Assertion arguments should be passed in the correct order', rule, {
   valid: [
     {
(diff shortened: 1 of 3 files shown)


yassin-kammoun-sonarsource:
The newly added testing file should be dropped once #3047 is done, and the quick fix assertions should be moved to the comment-based test file.

eslint-bridge/tests/rules/inverted-assertion-arguments.test.ts
```diff
@@ -0,0 +1,112 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+import { rule } from 'rules/inverted-assertion-arguments';
+import { RuleTester } from 'eslint';
+
+const ruleTester = new RuleTester({ parserOptions: { ecmaVersion: 2018, sourceType: 'module' } });
```
vilchik-elena, line 23: let's add comment that this rule is tested in another file and that here we just testing fixes.

eslint-bridge/tests/rules/inverted-assertion-arguments.test.ts
```diff
@@ -0,0 +1,112 @@
...
+          ],
+        },
+      ],
+    },
+    {
+      code: `
+const assert = require('chai').should();
+describe('suite', () => {
+  it('test', () => {
+    should.fail(42, aNumber);
```
vilchik-elena, line 92: tbh instead of testing different APIs, I would focus on more complex expressions (multiline, spaces)

vilchik-elena (changes requested):

vilchik-elena (approved):

PR #3050, opened 2022-03-16
Add quick fix for S3696 ('no-throw-literal')

### LINKED ISSUE #3017: Add quick fix for S3696: Literals should not be thrown
ESLint rule. Only for string literals and concatenation.

yassin-kammoun-sonarsource:
Fixes #3017

### DIFF
--- a/eslint-bridge/src/rules/decorators/no-throw-literal-decorator.ts
+++ b/eslint-bridge/src/rules/decorators/no-throw-literal-decorator.ts
@@ -1,0 +1,54 @@
+/*
+ * SonarQube JavaScript Plugin
+ * Copyright (C) 2011-2022 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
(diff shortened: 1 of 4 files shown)


eslint-bridge/src/rules/decorators/no-throw-literal-decorator.ts
```diff
@@ -0,0 +1,54 @@
...
+  rule.meta!.hasSuggestions = true;
+  return interceptReport(rule, (context, reportDescriptor) => {
+    const suggest: Rule.SuggestionReportDescriptor[] = [];
+    if ('node' in reportDescriptor) {
+      const { argument: thrown } = reportDescriptor.node as estree.ThrowStatement;
+      if (isStringLike(thrown)) {
+        const thrownText = context.getSourceCode().getText(thrown);
+        suggest.push({
+          desc: 'Throw an error object',
+          fix: fixer => fixer.replaceText(thrown, `new Error(${thrownText})`),
```
vilchik-elena, line 37: I wonder if it's more reliable to use `insertTextAfter` and `insertTextBefore` instead of `context.getSourceCode().getText()`. WDYT?

yassin-kammoun-sonarsource: Hard to say... I propose we keep it like this for now. If we manage to break the fix during validation, depending on the problem, we can then try those methods.

PR #3051, opened 2022-03-16
Add quick fix for S1763 ('no-unreachable')

### LINKED ISSUE #3008: Add quick fix for S1763: All code should be reachable
ESLint rule.

yassin-kammoun-sonarsource:
Fixes #3008

### DIFF
--- a/eslint-bridge/tests/rules/decorators/no-unreachable-decorator.test.ts
+++ b/eslint-bridge/tests/rules/decorators/no-unreachable-decorator.test.ts
@@ -1,0 +1,88 @@
+/*
+ * SonarQube JavaScript Plugin
+ * Copyright (C) 2011-2022 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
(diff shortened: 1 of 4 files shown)


eslint-bridge/tests/rules/decorators/no-unreachable-decorator.test.ts
```diff
@@ -0,0 +1,88 @@
...
+`,
+      errors: [
+        {
+          suggestions: [
+            {
+              output: `
+while (a()) {
+  b();
+  break;
+
```
vilchik-elena, line 78: not great to have this empty line but that's minor and I don't think it's worth it.

vilchik-elena (approved):

eslint-bridge/tests/rules/decorators/no-unreachable-decorator.test.ts
yassin-kammoun-sonarsource: Not great indeed -_-. I actually tried to remove it, but there were weird side-effects

PR #3052, opened 2022-03-17
Add quick fix for S4322 ('prefer-type-guard')

### LINKED ISSUE #3024: Add quick fix for S4322: Type guards should be used
SonarJS rule.

yassin-kammoun-sonarsource:
Fixes #3024

### DIFF
--- a/eslint-bridge/src/rules/prefer-type-guard.ts
+++ b/eslint-bridge/src/rules/prefer-type-guard.ts
@@ -141,10 +141,9 @@ function getTypePredicateSuggestion(
   if (node.returnType) {
     fix = fixer => fixer.replaceText(node.returnType as unknown as estree.Node, predicate);
   } else {
-    const functionBody = node.body as estree.Node;
     const closingParenthesis = context
       .getSourceCode()
-      .getTokenBefore(functionBody, token => token.value === ')')!;
+      .getTokenBefore(node.body as estree.Node, token => token.value === ')')!;
     fix = fixer => fixer.insertTextAfter(closingParenthesis, predicate);
   }
   suggestions.push({ messageId: 'suggestTypePredicate', fix });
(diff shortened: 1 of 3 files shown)


eslint-bridge/src/rules/prefer-type-guard.ts
```diff
@@ -121,11 +125,37 @@ function checkCastedType(
...
+  const suggestions: Rule.SuggestionReportDescriptor[] = [];
+  let fix: (fixer: Rule.RuleFixer) => Rule.Fix;
+  if (node.returnType) {
+    fix = fixer => fixer.replaceText(node.returnType as unknown as estree.Node, `: ${predicate}`);
+  } else {
+    const functionBody = node.body as estree.Node;
+    const closingParenthesis = context
+      .getSourceCode()
+      .getTokenBefore(functionBody, token => token.value === ')')!;
+    const openingBracket = context
```
vilchik-elena, line 148: it's `brace`

eslint-bridge/src/rules/prefer-type-guard.ts
```diff
@@ -121,11 +125,37 @@ function checkCastedType(
...
+      .getSourceCode()
+      .getFirstToken(functionBody, token => token.value === '{')!;
+    const [, begin] = closingParenthesis.range;
+    const [end, _] = openingBracket.range;
+    fix = fixer => fixer.replaceTextRange([begin, end], `: ${predicate} `);
```
vilchik-elena, line 153: I think it's more reliable to use `insertTextAfter` to rely only on closing parenthesis

vilchik-elena (changes requested):

eslint-bridge/src/rules/prefer-type-guard.ts
yassin-kammoun-sonarsource: Good point.

vilchik-elena (approved):

PR #3053, opened 2022-03-17
Add quick fix for S4619 ('no-in-misuse')

### LINKED ISSUE #3026: Add quick fix for S4619: "in" should not be used on arrays
SonarJS rule.

yassin-kammoun-sonarsource:
Fixes #3026

### DIFF
--- a/eslint-bridge/tests/rules/no-in-misuse.test.ts
+++ b/eslint-bridge/tests/rules/no-in-misuse.test.ts
@@ -93,11 +93,11 @@ ruleTester.run('"in" should not be used on arrays"', rule, {
         {
           suggestions: [
             {
-              desc: `Use "indexOf"`,
+              desc: `Replace with "indexOf" method`,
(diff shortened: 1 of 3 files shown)


eslint-bridge/tests/rules/no-in-misuse.test.ts
```diff
@@ -87,5 +87,22 @@ ruleTester.run('"in" should not be used on arrays"', rule, {
                 }`,
       errors: 1,
     },
+    {
+      code: `if ("bar" in ["foo", "bar", "baz"]) {}`,
+      errors: [
+        {
+          suggestions: [
+            {
+              desc: `Use "indexOf"`,
```
vilchik-elena, line 96: Message is kind of too short :) how about `Replace with "indexOf" method` (check our messages in `quickfix.ts`, I think it would be more consistent)

vilchik-elena (approved):
I suggest to change the message, otherwise ok

PR #3056, opened 2022-03-17
Add quick fix for S1186 ('no-empty-function')

### LINKED ISSUE #3005: Add quick fix for S1186: Functions should not be empty
Eslint rule. Fix available in SonarJava.

yassin-kammoun-sonarsource:
Fixes #3005

### DIFF
--- a/eslint-bridge/src/rules/decorators/no-empty-function-decorator.ts
+++ b/eslint-bridge/src/rules/decorators/no-empty-function-decorator.ts
@@ -1,0 +1,63 @@
+/*
+ * SonarQube JavaScript Plugin
+ * Copyright (C) 2011-2022 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+// https://sonarsource.github.io/rspec/#/rspec/S1186/javascript
+
+import { Rule } from 'eslint';
+import * as estree from 'estree';
+import { interceptReport } from '../../utils';
+
+type FunctionLike =
+  | estree.ArrowFunctionExpression
+  | estree.FunctionDeclaration
+  | estree.FunctionExpression;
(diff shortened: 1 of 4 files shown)


yassin-kammoun-sonarsource:
Inspired from [EmptyMethodsCheck](https://github.com/SonarSource/sonar-java/blob/9cb487451cbb3df1ce5715128b5c65d68f4ac90a/java-checks/src/main/java/org/sonar/java/checks/EmptyMethodsCheck.java#L101-L120) (SonarJava).

eslint-bridge/src/rules/decorators/no-empty-function-decorator.ts
```diff
@@ -0,0 +1,63 @@
...
+      const closingBrace = context
+        .getSourceCode()
+        .getLastToken(func.body, token => token.value === '}')!;
+      let commentPlaceholder: string;
+      if (openingBrace.loc.start.line === closingBrace.loc.start.line) {
+        commentPlaceholder = ` /* TODO document why this ${name} is empty */ `;
+      } else {
+        const columnOffset = closingBrace.loc.start.column;
+        const padding = ' '.repeat(columnOffset);
+        commentPlaceholder = `\n${padding}  // TODO document why this ${name} is empty\n${padding}`;
```
vilchik-elena, line 50: I see that java impl has that, but I don't get why we need second new line. Do you?

eslint-bridge/tests/rules/decorators/no-empty-function-decorator.test.ts
```diff
@@ -0,0 +1,115 @@
...
+      ],
+    },
+    {
+      code: `
+const obj = {
+  foo: function() {
+    }
+};
+`,
+      errors: [
```
vilchik-elena, line 97: can we add the arrow function to see what will be the name of it?

vilchik-elena (changes requested):

eslint-bridge/src/rules/decorators/no-empty-function-decorator.ts
```diff
@@ -0,0 +1,63 @@
...
+type FunctionLike =
+  | estree.ArrowFunctionExpression
+  | estree.FunctionDeclaration
+  | estree.FunctionExpression;
+
+// core implementation of this rule does not provide quick fixes
+export function decorateNoEmptyFunction(rule: Rule.RuleModule): Rule.RuleModule {
+  rule.meta!.hasSuggestions = true;
+  return interceptReport(rule, (context, reportDescriptor) => {
+    if ('node' in reportDescriptor) {
```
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: Yes, the issue won't be reported. I am starting to think that this condition is useless because once we intercept a report, we know for sure that there is a node, at least for this rule. Should I just remove the condition ?

eslint-bridge/src/rules/decorators/no-empty-function-decorator.ts
yassin-kammoun-sonarsource: No idea. [SONARJAVA-3906](https://jira.sonarsource.com/browse/SONARJAVA-3918) mentions that there are situations when a newline is needed, but I didn't see an example in the unit tests.

vilchik-elena (approved):

PR #3058, opened 2022-03-21
Add quick fix for S2688 ('use-isnan')

### LINKED ISSUE #3009: Add quick fix for S2688: "NaN" should not be used in comparisons
ESLint rule.

yassin-kammoun-sonarsource:
Fixes #3009

### DIFF
--- a/eslint-bridge/src/rules/decorators/use-isnan-decorator.ts
+++ b/eslint-bridge/src/rules/decorators/use-isnan-decorator.ts
@@ -45,10 +45,17 @@ export function decorateUseIsNan(rule: Rule.RuleModule): Rule.RuleModule {
       if (negate !== null) {
         const arg = isNaNIdentifier(left) ? right : left;
         const argText = context.getSourceCode().getText(arg);
-        suggest.push({
-          desc: 'Use "Number.isNaN()"',
-          fix: fixer => fixer.replaceText(node, `${negate ? '!' : ''}isNaN(${argText})`),
-        });
+        const prefix = negate ? '!' : '';
+        suggest.push(
+          {
+            desc: 'Use "isNaN()"',
+            fix: fixer => fixer.replaceText(node, `${prefix}isNaN(${argText})`),
+          },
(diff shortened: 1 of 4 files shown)


eslint-bridge/tests/rules/decorators/use-isnan-decorator.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    },
+  ],
+  invalid: [
+    {
+      code: `if (n == NaN) {}`,
+      errors: [
+        {
+          suggestions: [
+            {
+              desc: `Use "Number.isNaN()"`,
```
vilchik-elena, line 39: I think `Number.isNaN` and `isNaN` is not the same function, so we can't use this message for such fix

eslint-bridge/src/rules/decorators/use-isnan-decorator.ts
```diff
@@ -0,0 +1,60 @@
...
+        case '!=':
+        case '!==':
+          negate = true;
+          break;
+        case '==':
+        case '===':
+          negate = false;
+          break;
+      }
+      if (negate !== null) {
```
vilchik-elena, line 45: I don't get which test case covers this condition's else branch

vilchik-elena (changes requested):

eslint-bridge/tests/rules/decorators/use-isnan-decorator.test.ts
yassin-kammoun-sonarsource: My bad, I thought they were the same. I will suggest both then.

eslint-bridge/src/rules/decorators/use-isnan-decorator.ts
yassin-kammoun-sonarsource: There is none, but it's surprising to see 100% coverage.

vilchik-elena (approved):

PR #3061, opened 2022-03-22
Add quick fix for S4621 ('no-duplicate-in-composite')

### LINKED ISSUE #3027: Add quick fix for S4621: Union and intersection types should not be defined with duplicated elements
SonarJS rule.

yassin-kammoun-sonarsource:
Fixes #3027

### DIFF
--- a/eslint-bridge/src/rules/no-duplicate-in-composite.ts
+++ b/eslint-bridge/src/rules/no-duplicate-in-composite.ts
@@ -17,7 +17,7 @@
  * along with this program; if not, write to the Free Software Foundation,
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
-// https://jira.sonarsource.com/browse/RSPEC-4621
+// https://sonarsource.github.io/rspec/#/rspec/S4621/javascript
 
 import { Rule } from 'eslint';
 import * as estree from 'estree';
@@ -26,6 +26,7 @@ import { toEncodedMessage } from '../utils';
 
 export const rule: Rule.RuleModule = {
   meta: {
+    hasSuggestions: true,
     schema: [
       {
         // internal parameter for rules having secondary locations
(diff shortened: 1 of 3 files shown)


eslint-bridge/src/rules/no-duplicate-in-composite.ts
```diff
@@ -54,6 +55,7 @@ export const rule: Rule.RuleModule = {
 
         groupedTypes.forEach(duplicates => {
           if (duplicates.length > 1) {
+            const suggest = getSuggestions(compositeType, duplicates.slice(1), context);
```
vilchik-elena, line 58: minor but I would move slicing logic inside of the function

eslint-bridge/src/rules/no-duplicate-in-composite.ts
```diff
@@ -66,10 +68,38 @@ export const rule: Rule.RuleModule = {
...
+): Rule.SuggestionReportDescriptor[] {
+  const uniqueTypes = composite.types.filter(tpe => !duplicates.includes(tpe));
+  const uniqueTexts = uniqueTypes.map(tpe =>
+    context.getSourceCode().getText(tpe as unknown as estree.Node),
+  );
+  const compositeNode = composite as unknown as estree.Node;
+  const firstToken = context.getSourceCode().getFirstToken(compositeNode);
+  const lastToken = context.getSourceCode().getLastToken(compositeNode);
+  let [prefix, suffix] = ['', ''];
+  if (firstToken?.value === '(' && lastToken?.value === ')') {
```
vilchik-elena, line 93: I don't see test for this

vilchik-elena (changes requested):
TBH I am not big fan of such solution based on full text replacement (and not on removal of duplicates), but indeed it's pretty straight-forward.

eslint-bridge/src/rules/no-duplicate-in-composite.ts
yassin-kammoun-sonarsource: https://github.com/SonarSource/SonarJS/blob/777bc114e0fca7a251be1fcafe01d1dc20a5e5d3/eslint-bridge/tests/rules/no-duplicate-in-composite.test.ts#L164-L198

vilchik-elena: hm... here first and last parentheses are not a pair, that's why I didn't get that it covers the condition
Then for `(number | string) & (number | string) & Foo` we gonna get `number | string & Foo` which is wrong.

vilchik-elena (changes requested):

vilchik-elena (approved):

PR #3075, opened 2022-03-28
Update rules metadata

vilchik-elena:


javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S106.html
```diff
@@ -11,4 +11,4 @@ <h2>See</h2>
   <li> <a href="https://www.owasp.org/index.php/Top_10-2017_A3-Sensitive_Data_Exposure">OWASP Top 10 2017 Category A3</a> - Sensitive Data Exposure
   </li>
 </ul>
-
```
vilchik-elena, line 14: @SonarSource/languages-team-jsts  do you know what is changed here?

vilchik-elena, line 14: It seems that there were windows line endings before

yassin-kammoun-sonarsource (approved):

PR #3081, opened 2022-03-29
Add CSS to plugin description

vilchik-elena:


yassin-kammoun-sonarsource (approved):

agigleux:
Can you also update this line: https://github.com/SonarSource/SonarJS/blob/master/sonar-javascript-plugin/pom.xml#L16 ?

agigleux (changes requested):
Can you also update this line https://github.com/SonarSource/SonarJS/blob/master/sonar-javascript-plugin/pom.xml#L16 ?

agigleux (approved):

PR #3087, opened 2022-04-04
Improve S4328 (no-implicit-dependencies): consider 'baseUrl'

### LINKED ISSUE #1929: S4328: covering projects using "baseUrl" compiler option
Check this thread for details https://community.sonarsource.com/t/typescript-baseurl-and-relative-paths-configuration-issue-dependencies-should-be-explicit-typescript-s4328/19135

Basically it's possible to say TS compiler the base URL for relative paths when importing and that way avoiding `./` in the import paths

```
import * as foo from "foo"; // here "foo" might be referencing project file "./foo.ts"
```

vilchik-elena:
Fixes #1929

### DIFF
--- a/eslint-bridge/src/rules/no-implicit-dependencies.ts
+++ b/eslint-bridge/src/rules/no-implicit-dependencies.ts
@@ -17,7 +17,7 @@
  * along with this program; if not, write to the Free Software Foundation,
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
-// https://jira.sonarsource.com/browse/RSPEC-4328
+// https://sonarsource.github.io/rspec/#/rspec/S4328/javascript
 
 import { Rule } from 'eslint';
 import * as estree from 'estree';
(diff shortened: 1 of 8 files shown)


eslint-bridge/src/rules/no-implicit-dependencies.ts
```diff
@@ -107,6 +111,19 @@ function raiseOnImplicitImport(
...
   }
 
+  if (baseUrl) {
+    const underBaseUrlPath = path.join(baseUrl, moduleName);
+    const alternatives = [
+      underBaseUrlPath,
+      underBaseUrlPath + '.ts',
+      underBaseUrlPath + '.d.ts',
+      underBaseUrlPath + '.tsx',
+    ];
```
yassin-kammoun-sonarsource, line 121: I am wondering we should also consider `.js`, `.jsx`, and `.vue` files as well. It's possible to include them in tsconfig files. WDYT?

yassin-kammoun-sonarsource, line 121: Should we consider `.vue` files as well ?

yassin-kammoun-sonarsource (approved):
LGTM.

PR #3095, opened 2022-04-08
Update S4649 ('font-family-no-missing-generic-family-keyword'): Support 'ignoreFontFamilies' option

yassin-kammoun-sonarsource:
Fixes https://github.com/SonarSource/sonar-css/issues/356
Fixes https://github.com/SonarSource/sonar-css/pull/357

### DIFF
--- a/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/css/rules/FontFamilyNoMissingGenericFamilyKeyword.java
+++ b/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/css/rules/FontFamilyNoMissingGenericFamilyKeyword.java
@@ -19,13 +19,41 @@
  */
 package org.sonar.plugins.javascript.css.rules;
 
+import java.util.Arrays;
+import java.util.List;
+
 import org.sonar.check.Rule;
+import org.sonar.check.RuleProperty;
+
+import static org.sonar.plugins.javascript.css.rules.RuleUtils.splitAndTrim;
 
 @Rule(key = "S4649")
 public class FontFamilyNoMissingGenericFamilyKeyword implements CssRule {
 
+  private static final String DEFAULT_IGNORE_FONT_FAMILIES = "custom-font";
+
+  @RuleProperty(
+    key = "ignoreFontFamilies",
+    description = "Comma-separated list of regular expressions for font families exempt from this rule.",
+    defaultValue = "" + DEFAULT_IGNORE_FONT_FAMILIES)
+  String ignoreFontFamilies = DEFAULT_IGNORE_FONT_FAMILIES;
+
   @Override
   public String stylelintKey() {
     return "font-family-no-missing-generic-family-keyword";
+  }
+
+  @Override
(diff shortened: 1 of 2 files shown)


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/css/rules/FontFamilyNoMissingGenericFamilyKeyword.java
```diff
@@ -19,13 +19,41 @@
...
+
 import org.sonar.check.Rule;
+import org.sonar.check.RuleProperty;
+
+import static org.sonar.plugins.javascript.css.rules.RuleUtils.splitAndTrim;
 
 @Rule(key = "S4649")
 public class FontFamilyNoMissingGenericFamilyKeyword implements CssRule {
 
+  private static final String DEFAULT_IGNORE_FONT_FAMILIES = "custom-font";
```
vilchik-elena, line 33: we should not put example into default value

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/css/rules/FontFamilyNoMissingGenericFamilyKeyword.java
```diff
@@ -19,13 +19,41 @@
...
+
+  @RuleProperty(
+    key = "ignoreFontFamilies",
+    description = "Comma-separated list of regular expressions for font families exempt from this rule.",
```
vilchik-elena, line 37: As not only regex are support but also plain strings I would rephrase: `Comma-separated list of font families exempt from this rule (regular expressions supported).`

vilchik-elena (approved):

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/css/rules/CssRuleTest.java
```diff
@@ -142,7 +142,7 @@ void declaration_block_no_duplicate_properties_custom() {
   @Test
   void font_family_no_missing_generic_family_keyword_default() {
     String optionsAsJson = new Gson().toJson(new FontFamilyNoMissingGenericFamilyKeyword().stylelintOptions());
-    assertThat(optionsAsJson).isEqualTo("[true,{\"ignoreFontFamilies\":[\"custom-font\"]}]");
+    assertThat(optionsAsJson).isEqualTo("[true,{\"ignoreFontFamilies\":[\"\"]}]");
```
vilchik-elena, line 145: hm.. usually we have non-empty default, so we didn't have such problem before, but I believe we should have an empty array here, and not array with empty string literal.

vilchik-elena (changes requested):

farashaikh:
@vilchik-elena I am using SonarQube 9.4 and I can still see this issue
image

Please can you let me know if this update is available in version 9.4?

vilchik-elena:
@farashaikh this change only add a parameter for the rule. You should set it to the values you need.

PR #3097, opened 2022-04-08
Improve S4653 ('unit-no-unknown'): Support ignoring any kind of unit

yassin-kammoun-sonarsource:
Fixes https://github.com/SonarSource/sonar-css/pull/367

### DIFF
--- a/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/css/rules/UnitNoUnknown.java
+++ b/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/css/rules/UnitNoUnknown.java
@@ -22,9 +22,20 @@ package org.sonar.plugins.javascript.css.rules;
 import java.util.Arrays;
 import java.util.List;
 import org.sonar.check.Rule;
+import org.sonar.check.RuleProperty;
+
+import static org.sonar.plugins.javascript.css.rules.RuleUtils.splitAndTrim;
 
 @Rule(key = "S4653")
 public class UnitNoUnknown implements CssRule {
+
+  private static final String DEFAULT_IGNORED_UNITS = "x";
+
+  @RuleProperty(
+    key = "ignoredUnits",
+    description = "Comma-separated list of \"units\" to consider as valid.",
+    defaultValue = "" + DEFAULT_IGNORED_UNITS)
+  String ignoredUnits = DEFAULT_IGNORED_UNITS;
 
   @Override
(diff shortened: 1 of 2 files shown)


vilchik-elena (changes requested):
I didn't find any place describing why we need this parameter. Also why default should be `x`

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/css/rules/UnitNoUnknown.java
```diff
@@ -22,22 +22,37 @@
...
+import static org.sonar.plugins.javascript.css.rules.RuleUtils.splitAndTrim;
 
 @Rule(key = "S4653")
 public class UnitNoUnknown implements CssRule {
 
+  private static final String DEFAULT_IGNORED_UNITS = "x";
+
+  @RuleProperty(
+    key = "ignoredUnits",
+    description = "Comma-separated list of \"units\" to consider as valid.",
```
vilchik-elena, line 36: should we mention that regex are supported?

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/css/rules/UnitNoUnknown.java
```diff
@@ -22,22 +22,37 @@
...
   }
 
   @Override
   public List<Object> stylelintOptions() {
-    return Arrays.asList(true, new StylelintIgnoreOption());
+    return Arrays.asList(true, new StylelintIgnoreOption(splitAndTrim(ignoredUnits)));
   }
 
   private static class StylelintIgnoreOption {
     // Used by GSON serialization
```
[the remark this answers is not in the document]
vilchik-elena: Ok, I see now that we use GSON when sending the request

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/css/rules/UnitNoUnknown.java
yassin-kammoun-sonarsource: We should, similarly to what you suggested in the other PR.

vilchik-elena:
> Which parameter are you referring to?

The parameter you have added in this PR :)

yassin-kammoun-sonarsource:
Closing this PR as discussed.

PR #3102, opened 2022-04-08
Improve S4328 (no-implicit-dependencies): check package.json files in all levels up

### LINKED ISSUE #2682: S4328 (no-implicit-dependencies): check package.json files in all levels up
Currently we stop as soon as one package.json is found, yet NPM will search the dependency up until it finds it or the root is reached. This problem is common for monorepo projects.

Reported in https://community.sonarsource.com/t/dependencies-only-searched-one-package-json-up/42816

vilchik-elena:
Fixes #2682


eslint-bridge/src/rules/no-implicit-dependencies.ts
```diff
@@ -146,33 +151,54 @@ function getPackageName(name: string) {
...
+      dep.forEach(d => result.add(d));
+      dirCache.set(dirname, dep);
+    }
+
+    const upperDir = path.dirname(dirname);
+    if (upperDir === dirname) {
+      break;
+    } else {
+      dirname = upperDir;
+    }
```
yassin-kammoun-sonarsource, line 181: _Food for thought..._

Just saying, and not sure if it would make the code much more readable, but we could get rid of the if-break-else thing by turning the infinite while loop into a do-while loop.

yassin-kammoun-sonarsource (approved):
lgtm

eslint-bridge/src/rules/no-implicit-dependencies.ts
vilchik-elena: I see the point but not sure it's clearer with do-while

gpessa:
Do you know when this will be released?

yassin-kammoun-sonarsource:
Hello @gpessa, this improvement is already available on SonarCloud, and it will be part of SonarQube 9.5, whose release is planned for early June.

PR #3111, opened 2022-04-13
Require Node.js 12.22.0

### LINKED ISSUE #3100: SonarJS should check also minor version of Node.JS
It seems that it's important to also check the minor version of Node.JS to be sure that we are compatible with it. See 
https://community.sonarsource.com/t/getting-an-error-in-sonarcloud-analyzer/61095/8

yassin-kammoun-sonarsource:
Fixes #3100

### DIFF
--- a/nodejs-utils/src/main/java/org/sonarsource/nodejs/NodeCommandBuilderImpl.java
+++ b/nodejs-utils/src/main/java/org/sonarsource/nodejs/NodeCommandBuilderImpl.java
@@ -33,6 +33,7 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import javax.annotation.Nullable;
 import org.sonar.api.config.Configuration;
+import org.sonar.api.utils.Version;
 import org.sonar.api.utils.log.Logger;
 import org.sonar.api.utils.log.Loggers;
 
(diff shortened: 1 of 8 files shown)


nodejs-utils/src/main/java/org/sonarsource/nodejs/NodeCommandBuilderImpl.java
```diff
@@ -166,27 +167,39 @@ private void checkNodeCompatibility(String nodeExecutable) throws NodeCommandExc
...
     }
   }
 
   private String getVersion(String nodeExecutable) throws NodeCommandException {
     StringBuilder output = new StringBuilder();
-    NodeCommand nodeCommand = new NodeCommand(processWrapper, nodeExecutable, actualNodeVersion, singletonList("-v"), null, emptyList(), output::append, LOG::error, Map.of());
+    NodeCommand nodeCommand = new NodeCommand(
+      processWrapper,
+      nodeExecutable,
+      Version.create(0, 0),
```
yassin-kammoun-sonarsource, line 196: Was previously relying on an uninitialized state of `actualNodeVersion`, that is 0.

sonar-javascript-plugin/src/main/resources/static/documentation.md
```diff
@@ -11,7 +11,7 @@ key: javascript
 
 In order to analyze JavaScript, TypeScript or CSS code, you need to have supported version of Node.js installed on the
 machine running the scan. Recommended versions are previous LTS version v14 and the latest version - v16. We recommend using 
-the latest available LTS version (v16 as of today) for optimal stability and performance. v12 is still supported, but it already reached end-of-life and
+the latest available LTS version (v16 as of today) for optimal stability and performance. +v12.22.0 is still supported, but it already reached end-of-life and
```
yassin-kammoun-sonarsource, line 14: The choice of this particular version comes from the minimal Node.js version expected by ESLint 8.10.0 (see [ESLint supported engines](https://github.com/eslint/eslint/blob/f66890ba0f778e46b6dd6ab6e88d49ff0bf1f78b/package.json#L148)).

PR #3133, opened 2022-05-09
Fix S3863 ('no-duplicate-imports'): Suggest quick fixes iff there is a top-level import

yassin-kammoun-sonarsource:


eslint-bridge/src/rules/decorators/no-duplicate-imports-decorator.ts
```diff
@@ -124,10 +127,10 @@ function getModule(decl: estree.ImportDeclaration): string {
...
   context: Rule.RuleContext,
-): estree.ImportDeclaration {
+): estree.ImportDeclaration | undefined {
   return context
     .getSourceCode()
-    .ast.body.find(
-      node => node.type === 'ImportDeclaration' && module === getModule(node),
-    ) as estree.ImportDeclaration;
+    .ast.body.find(node => node.type === 'ImportDeclaration' && module === getModule(node)) as
+    | estree.ImportDeclaration
```
vilchik-elena, line 134: wow, I didn't know that such syntax with leading pipe exists

vilchik-elena (approved):

PR #3141, opened 2022-05-23
Create S6426: Exclusive tests should not be commited to version control

### LINKED ISSUE #2861: Rule idea: '.only' should not be used in tests

ilia-kebets-sonarsource:
Fixes #2861


eslint-bridge/tests/rules/rule-test-launcher.test.ts
```diff
@@ -57,6 +57,8 @@ function runRuleTests(rules: Record<string, Rule.RuleModule>, ruleTester: RuleTe
     }
     describe(`Running tests for rule ${rule}`, () => {
       files.forEach(filename => {
+        //console.log('readin test', filename);
+        if (!filename.includes('no-exclusive-tests.js')) return;
```
ilia-kebets-sonarsource, line 61: I will remove this

eslint-bridge/src/rules/no-exclusive-tests.ts
```diff
@@ -0,0 +1,63 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+// https://sonarsource.github.io/rspec/#/rspec/S6092/javascript
```
yassin-kammoun-sonarsource, line 20:
```suggestion
// https://sonarsource.github.io/rspec/#/rspec/S6426/javascript
```

eslint-bridge/src/rules/no-exclusive-tests.ts
```diff
@@ -0,0 +1,63 @@
...
+
+import { Rule } from 'eslint';
+import * as estree from 'estree';
+
+export const rule: Rule.RuleModule = {
+  meta: {
+    hasSuggestions: true,
+    messages: {
+      issue: 'Remove .only() from your test case.',
```
yassin-kammoun-sonarsource, line 29:
```suggestion
      issue: 'Remove ".only()" from your test case.',
```

eslint-bridge/src/rules/no-exclusive-tests.ts
```diff
@@ -0,0 +1,63 @@
...
+      quickfix: 'Remove .only().',
```
yassin-kammoun-sonarsource, line 30:
```suggestion
      quickfix: 'Remove ."only()".',
```

eslint-bridge/src/rules/no-exclusive-tests.ts
yassin-kammoun-sonarsource: You may need to update the unit tests.

eslint-bridge/src/rules/no-exclusive-tests.ts
yassin-kammoun-sonarsource: You may need to update the unit tests.

eslint-bridge/src/rules/no-exclusive-tests.ts
```diff
@@ -0,0 +1,63 @@
...
+    return {
+      CallExpression: (node: estree.CallExpression) => {
+        if (
+          node.callee.type === 'MemberExpression' &&
+          node.callee.property.type === 'Identifier' &&
+          node.callee.property.name === 'only' &&
+          node.callee.object.type === 'Identifier' &&
+          ['describe', 'it', 'test'].includes(node.callee.object.name)
+        ) {
+          const onlyProp: estree.MemberExpression = node.callee;
```
yassin-kammoun-sonarsource, line 43: You can make shorter this whole logic:
1. You can test that `node` denotes a method invocation with `isMethodCall` from `utils-ast` aka `utils`
2. You can use object destructuring to extract both `object` and `property` from `node.callee` after the previous test
3. You can check that `object` is either `'describe', 'it', 'test'` with `isIdentifier`  from `utils-ast` aka `utils`, the same applies for `property` but with `only`

yassin-kammoun-sonarsource: Pointers:

https://github.com/SonarSource/SonarJS/blob/7db147fd0ae395a125fb465531ee71f7eb2d82d2/eslint-bridge/src/utils/utils-ast.ts#L137

https://github.com/SonarSource/SonarJS/blob/7db147fd0ae395a125fb465531ee71f7eb2d82d2/eslint-bridge/src/utils/utils-ast.ts#L50-L53

eslint-bridge/src/rules/no-exclusive-tests.ts
yassin-kammoun-sonarsource: For your information, we generally leave this kind of URL as a comment for each rule we implement. It helps us quickly find the RSPEC of a given rule.

eslint-bridge/src/rules/no-exclusive-tests.ts
```diff
@@ -0,0 +1,63 @@
...
+          context.report({
+            messageId: 'issue',
+            node: node.callee.property,
+            suggest: [
+              {
+                fix: (fixer: Rule.RuleFixer) => {
+                  const fixes = [fixer.remove(onlyProp.property)];
+                  const dotBeforeOnly = context.getSourceCode().getTokenBefore(onlyProp.property);
+                  if (dotBeforeOnly != null) fixes.push(fixer.remove(dotBeforeOnly));
```
yassin-kammoun-sonarsource, line 52: We try to avoid one-line if-statements like this one. In that regard, you happen to have a code smell on this line because you are missing curly braces.

eslint-bridge/tests/rules/fixtures/no-exclusive-tests.js
```diff
@@ -0,0 +1,28 @@
+describe.only("only on root describe"); // Noncompliant {{Remove .only() from your test case.}}
```
yassin-kammoun-sonarsource, line 1:
```suggestion
describe.only("only on root describe"); // Noncompliant {{Remove ".only()" from your test case.}}
```

eslint-bridge/tests/rules/no-exclusive-tests.test.ts
```diff
@@ -0,0 +1,71 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+import { RuleTester } from 'eslint';
+import { rule } from 'rules/no-exclusive-tests';
+
+// Main test cases are in the file no-exclusive-tests.js.js (comment-based)
```
yassin-kammoun-sonarsource, line 23:
```suggestion
// Main test cases are in the file no-exclusive-tests.js (comment-based)
```

eslint-bridge/tests/rules/no-exclusive-tests.test.ts
```diff
@@ -0,0 +1,71 @@
...
+  invalid: [
+    {
+      code: 'describe.only("a describe with a .only()", function () {})',
+      errors: [
+        {
+          messageId: 'issue',
+          line: 1,
+          column: 10,
+          endLine: 1,
+          endColumn: 14,
```
yassin-kammoun-sonarsource, line 38: We can get rid of those location details since the comment-based test file already asserts issue locations. Here the whole purpose of this test file is to focus on quickfixes.

eslint-bridge/tests/rules/no-exclusive-tests.test.ts
```diff
@@ -0,0 +1,71 @@
...
+      code: `
+it// skipping a line here
+.only("should be flagged with a comment slipped in the middle.");`,
+      errors: [
+        {
+          messageId: 'issue',
+          line: 3,
+          column: 2,
+          endLine: 3,
+          endColumn: 6,
```
yassin-kammoun-sonarsource, line 58: Same remark as above.

eslint-bridge/tests/rules/no-exclusive-tests.test.ts
```diff
@@ -0,0 +1,71 @@
...
+      errors: [
+        {
+          messageId: 'issue',
+          line: 1,
+          column: 10,
+          endLine: 1,
+          endColumn: 14,
+          suggestions: [
+            {
+              messageId: 'quickfix',
```
yassin-kammoun-sonarsource, line 41: We should assert the quickfix description. Therefore, for this particular test case, let's assert the value of the `desc` property rather than the `messageId` one.

javascript-checks/src/main/java/org/sonar/javascript/checks/NoExclusiveTestsCheck.java
```diff
@@ -0,0 +1,36 @@
...
+
+@JavaScriptRule
+@TypeScriptRule
+@Rule(key="S6426")
+public class NoExclusiveTestsCheck extends TestFileCheck {
+
+    @Override
+    public String eslintKey() {
+      return "no-exclusive-tests";
+    }
```
yassin-kammoun-sonarsource, line 35: Not sure what happened there, but the quality gate reports two code smells here that you should probably address.

yassin-kammoun-sonarsource (changes requested):
You will need to manually declare the quickfix here:

https://github.com/SonarSource/SonarJS/blob/7db147fd0ae395a125fb465531ee71f7eb2d82d2/eslint-bridge/src/quickfix.ts#L71-L87

We require this step because we want to control for which rules we want to enable quickfixes.

yassin-kammoun-sonarsource (approved):

PR #3151, opened 2022-05-25
Fix FP S3403 ('different-types-comparison'): Symbols should be comparable

### LINKED ISSUE #3132: Fix FP S3403 ('different-types-comparison'): Symbol types should be assignable
https://discuss.sonarsource.com/t/s3403-false-positive-when-dependencies-are-provided/10492

yassin-kammoun-sonarsource:
Fixes #3132

### DIFF
--- a/eslint-bridge/src/rules/different-types-comparison.ts
+++ b/eslint-bridge/src/rules/different-types-comparison.ts
@@ -21,7 +21,6 @@
 
 import { Rule } from 'eslint';
 import * as estree from 'estree';
-import * as ts from 'typescript';
 import { isRequiredParserServices, getTypeFromTreeNode, toEncodedMessage } from '../utils';
 
 export const rule: Rule.RuleModule = {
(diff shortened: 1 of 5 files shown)


eslint-bridge/tests/rules/different-types-comparison.test.ts
```diff
@@ -135,6 +135,23 @@ ruleTesterTs.run(`Strict equality operators should not be used with dissimilar t
...
+    {
+      code: `
+        let str = 'str', obj = {};
+        str === obj;`,
+    },
+    {
+      code: `
+        let str = 'str', obj = {};
+        str !== obj;`,
+    },
```
yassin-kammoun-sonarsource, line 147: TypeScript seems to accept comparisons between strings and empty object literals.

eslint-bridge/src/rules/different-types-comparison.ts
```diff
@@ -40,50 +39,20 @@ export const rule: Rule.RuleModule = {
...
-        !isSameSymbol(lhsType, rhsType) &&
-        !isSubType(lhsType, rhsType) &&
-        !isSubType(rhsType, lhsType) &&
-        !isAny(lhsType) &&
-        !isAny(rhsType) &&
-        !isUndefinedOrNull(lhsType) &&
-        !isUndefinedOrNull(rhsType) &&
-        !isThis(lhs) &&
-        !isThis(rhs)
+        checker.isTypeAssignableTo(lhsType, rhsType) || checker.isTypeAssignableTo(rhsType, lhsType)
```
yassin-kammoun-sonarsource, line 48: I was not aware that we could use `TypeChecker#isTypeAssignableTo` although it's not exposed by TypeScript API. It's better to entirely rely on TypeScript type assignability rules rather than trying to implement them ourselves.

saberduck (approved):

PR #3159, opened 2022-06-02
BUILD-1625 Fix the build on Windows CirrusCI task

drautureau-sonarsource:
Fixes #


julien-carsique-sonarsource:
Not the wanted solution. This is a revert.

PR #3164, opened 2022-06-07
Support filtering for comment-based unit tests

### LINKED ISSUE #3162: Support filtering for comment-based unit tests
The [rule-test-launcher](https://github.com/SonarSource/SonarJS/blob/master/eslint-bridge/tests/rules/rule-test-launcher.test.ts) does not support Jest's test filtering. It should, as when we develop a rule, it is convenient to run the tests for your rule only, for readability and performance reasons.

ilia-kebets-sonarsource:
Fixes #3162

I have added a new script `ntest` (as "new test") as adding the filtering to `npm test` such as `npm test -t=your-rule` takes around 15s to run since it is going through all files defined by `<rootDir>/tests/**/*.test.t`.


eslint-bridge/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+## Test
+
+The ESLint bridge has unit tests that use [ESLint's RuleTester API](https://eslint.org/docs/developer-guide/nodejs-api#ruletester).
+
+### Comment-based tests
+
+We implemented a new format to write unit tests for ESLint using a more readable format.
+
+- You can launch them using: `npm run ntest`
```
vilchik-elena, line 12: I don't think we should document this one. Why someone would want to run all comment-based tests?

eslint-bridge/package.json
```diff
@@ -7,6 +7,7 @@
     "clear": "tsc -b src tests --clean",
     "check-format": "prettier --list-different \"{src,tests}/**/!(*.lint).ts\"",
     "test": "jest",
+    "ntest": "npm test tests/rules/rule-test-launcher.test.ts",
```
vilchik-elena, line 10: I don't get reasoning behind this name `ntest`. It stands for `new test` but what is new about it? How about `rule-test`?

vilchik-elena (changes requested):
It does not seem to work for me, all rules are executed.

eslint-bridge/package.json
ilia-kebets-sonarsource: Yeah, I'm not so confident with this alias `ntest`. 
It filters out the comments-based tests from all unit tests. I'll ask Victor if he feels creative.

ilia-kebets-sonarsource: I propose `ctest` for "comment-based tests"

vilchik-elena (changes requested):

vilchik-elena (approved):

PR #3165, opened 2022-06-07
fix tests on Windows

victor-diez-sonarsource:
basePath contained tilde path (C:\Users\VICTOR~1.DIE\AppData\...) instead of full path (C:/Users/victor.diez/AppData/...)

Using `baseDir.toRealPath()` fixes the issue, but to seems it actually checks FS

### DIFF
--- a/sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/TypeScriptSensorTest.java
+++ b/sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/TypeScriptSensorTest.java
@@ -230,10 +230,12 @@ class TypeScriptSensorTest {
     verify(eslintBridgeServerMock, never()).analyzeTypeScript(any());
     verify(eslintBridgeServerMock, never()).analyzeWithProgram(any());
 
-    assertThat(logTester.logs(LoggerLevel.ERROR)).contains("Provided tsconfig.json path doesn't exist. Path: '" + baseDir.toRealPath().resolve("wrong.json") + "'");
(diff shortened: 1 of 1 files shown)


sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/TypeScriptSensorTest.java
```diff
@@ -230,11 +230,11 @@ void should_log_and_stop_with_wrong_tsconfig() throws Exception {
...
     verify(eslintBridgeServerMock, never()).analyzeWithProgram(any());
 
-    assertThat(logTester.logs(LoggerLevel.ERROR)).contains("Provided tsconfig.json path doesn't exist. Path: '" + baseDir.resolve("wrong.json") + "'");
+    assertThat(logTester.logs(LoggerLevel.ERROR)).contains("Provided tsconfig.json path doesn't exist. Path: '" + baseDir.toRealPath().resolve("wrong.json") + "'");
   }
 
-  private SensorContextTester createSensorContext(Path baseDir) {
-    SensorContextTester ctx = SensorContextTester.create(baseDir);
+  private SensorContextTester createSensorContext(Path baseDir) throws IOException {
+    SensorContextTester ctx = SensorContextTester.create(baseDir.toRealPath());
```
vilchik-elena, line 237: Could you comment this line explaining why we need `toRealPath`?

vilchik-elena (changes requested):

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/TypeScriptSensorTest.java
victor-diez-sonarsource: done!

vilchik-elena (approved):

PR #3166, opened 2022-06-07
Add rule S4125: "typeof" expressions should only be compared to valid values

### LINKED ISSUE #625: Add rule S4125: "typeof" expressions should only be compared to valid values
Implements [RSPEC-4125)](https://jira.sonarsource.com/browse/RSPEC-4125)

ilia-kebets-sonarsource:
Fixes #625


ilia-kebets-sonarsource:
Based on the [implementation](https://github.com/eslint/eslint/blob/f31216a90a6204ed1fd56547772376a10f5d3ebb/lib/rules/valid-typeof.js#L105), the fix covers the case when:
```javascript
typeof foo === undefined
```
is changed to
```javascript
typeof foo === 'undefined'
```

We can update the S4125 rspec quickfix parameter to [`partial`](https://github.com/SonarSource/rspec/blob/master/docs/metadata.adoc#quickfix-field). Although, I have not been able to trigger this quickfix on a debug repo with eslint. The rule was flagged, but the quickfix not proposed.

javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S4125.html
```diff
@@ -0,0 +1,24 @@
...
+  <li> "number" </li>
+  <li> "string" </li>
+  <li> "symbol" (since ECMAScript 2015) </li>
+  <li> "function" </li>
+  <li> "object" (for <code>null</code> and any other object) </li>
+</ul>
+<p>Compare a <code>typeof</code> expression to anything else, and the result is predefined: <code>false</code>.</p>
+<h2>Noncompliant Code Example</h2>
+<pre>
+function someFunc(x: any): boolean {
```
vilchik-elena, line 14: I am not sure where this is coming from, but I don't think we should use type annotations as rule is enabled for JS

javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S4125.json
```diff
@@ -0,0 +1,17 @@
...
+  "remediation": {
+    "func": "Constant\/Issue",
+    "constantCost": "5min"
+  },
+  "tags": [],
+  "defaultSeverity": "Critical",
+  "ruleSpecification": "RSPEC-4125",
+  "sqKey": "S4125",
+  "scope": "All",
+  "quickfix": "unknown",
```
vilchik-elena, line 14: ESLint rule provides quick fixes for some cases (https://github.com/eslint/eslint/blob/21d647904dc30f9484b22acdd9243a6d0ecfba38/lib/rules/valid-typeof.js#L101), we should enable them. Let me know if you need help with that.

don't forget to update this metadata accordingly (here and in RSPEC repo)

vilchik-elena (changes requested):

javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S4125.html
```diff
@@ -0,0 +1,24 @@
+<p>The <code>typeof</code> operator returns a string indicating the type of its argument, and the set of returned values is limited:</p>
+<ul>
+  <li> "undefined" </li>
```
vilchik-elena, line 3: There is a new [`BigInt`](https://developer.mozilla.org/en-US/docs/Glossary/BigInt) value

javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S4125.json
vilchik-elena: Ah, I see now that you checked that already.  On my side I didn't even manage to have that issue triggered at all (on `if (typeof bar !== undefined) {}`), which is weird as it's in unit test of this rule. I suggest we don't waste time here as anyway quick fix application is very limited.

ilia-kebets-sonarsource: ACK, leaving this to `unknown`.

ilia-kebets-sonarsource:
@vilchik-elena I have applied the changes in RSPEC: https://github.com/SonarSource/rspec/pull/1044

javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S4125.html
vilchik-elena: why did you resolve it? there is still old description here. You should merge the change on RSPEC and regenerate here.

vilchik-elena (changes requested):
We need up to date description in this PR

PR #3167, opened 2022-06-08
Add rule S5148: Authorizing an opened window to access back to the originating window is security-sensitive

### LINKED ISSUE #2621: Add rule S5148: Authorizing an opened window to access back to the originating window is security-sensitive
Implements [RSPEC-5148](https://jira.sonarsource.com/browse/RSPEC-5148) 

For both JS/TS

Report on `window.open` when:
* the first argument is a string literal starting by "http" (external urls)
* and the third argument is missing or doesn't contain `noopener`

```
window.open("https://example.com/dangerous"); // Noncompliant (s5148)
window.open("http://example.com/dangerous"); // Noncompliant (s5148)
window.open("test.html"); // Compliant (s5148)
window.open("https://example.com/dangerous", "windowname", "resizable"); // Noncompliant (s5148)
window.open("https://example.com/dangerous", "windowname", "noopener"); // Compliant (s5148)
window.open("https://example.com/dangerous", "windowname", "resizable,scrollbars,status,noopener"); // Compliant (s5148)
```

ilia-kebets-sonarsource:
Fixes #2621


eslint-bridge/src/rules/link-with-target-blank.ts
```diff
@@ -0,0 +1,83 @@
...
+import { isIdentifier, isMethodCall, isStringLiteral } from '../utils';
+
+const REQUIRED_OPTION = 'noopener';
+const REQUIRED_OPTION_INDEX = 2;
+const URL_INDEX = 0;
+
+export const rule: Rule.RuleModule = {
+  meta: {
+    messages: {
+      issue: 'Make sure not using "noopener" is safe here.',
```
vilchik-elena, line 34: usually we try to describe the message with `messageId`, `issue` is too generic. How about `missingNoopener`?

eslint-bridge/src/rules/link-with-target-blank.ts
```diff
@@ -0,0 +1,83 @@
...
+    },
+  },
+  create(context: Rule.RuleContext) {
+    return {
+      CallExpression: (node: estree.CallExpression) => {
+        if (!isMethodCall(node)) {
+          return;
+        }
+        const { property, object } = node.callee;
```
vilchik-elena, line 43: I would swap `property` and `object`

eslint-bridge/src/rules/link-with-target-blank.ts
```diff
@@ -0,0 +1,83 @@
...
+
+        if (!(isIdentifier(property, 'open') && isIdentifier(object, 'window'))) {
```
vilchik-elena, line 45: `!(a && b)` ->  `!a || !b`

vilchik-elena, line 45: I see that you like early return, but IMO here it makes more sense to not have it.

eslint-bridge/src/rules/link-with-target-blank.ts
```diff
@@ -0,0 +1,83 @@
...
+          return;
+        }
+        const args = node.arguments;
+        if (args.length >= URL_INDEX && !isUrl(args[URL_INDEX])) {
```
vilchik-elena, line 49: IMHO you are abusing early return :) it does not improve the readability here.

eslint-bridge/src/rules/link-with-target-blank.ts
```diff
@@ -0,0 +1,83 @@
...
+    return false;
+  }
+  return argument.value.includes(REQUIRED_OPTION);
+}
+
+function isUrl(argument: estree.Node): boolean {
+  if (!isStringLiteral(argument)) {
+    return false;
+  }
+  try {
```
vilchik-elena, line 77: I think it's safe to simply check if string starts with `http`

eslint-bridge/src/rules/link-with-target-blank.ts
```diff
@@ -0,0 +1,83 @@
...
+};
+
+function hasRequiredOption(argument: estree.Node) {
+  if (!isStringLiteral(argument)) {
+    return false;
+  }
+  return argument.value.includes(REQUIRED_OPTION);
+}
+
+function isUrl(argument: estree.Node): boolean {
```
vilchik-elena, line 73: you can use `getValueOfExpression` from utils to cover more cases. E.g. 
```
let x = "http://hey.com"
window.open(x);
```

eslint-bridge/src/rules/link-with-target-blank.ts
```diff
@@ -0,0 +1,83 @@
...
+  if (!isStringLiteral(argument)) {
+    return false;
+  }
+  try {
+    new URL(argument.value);
+    return true;
+  } catch (_) {
```
vilchik-elena, line 80: you can simply write `} catch {`

eslint-bridge/src/rules/link-with-target-blank.ts
```diff
@@ -0,0 +1,83 @@
...
+            node: property,
+          });
+        }
+      },
+    };
+  },
+};
+
+function hasRequiredOption(argument: estree.Node) {
+  if (!isStringLiteral(argument)) {
```
vilchik-elena, line 67: `getValueOfExpression` can be used here as well

eslint-bridge/src/rules/link-with-target-blank.ts
vilchik-elena, line 45: this is good for checking `window.open`. I would suggest to support `this.window.open` and then it's good.

You don't need `isWindowObject` here as it's checking if variable has [`Window`](https://developer.mozilla.org/en-US/docs/Web/API/Window) type

vilchik-elena (changes requested):

eslint-bridge/src/rules/link-with-target-blank.ts
ilia-kebets-sonarsource: OK, so the current implementation is wrong as it flags `file://...`
However, there might be a file starting with `http`, like `httpprocol.png`.
Therefore I propose to check if it starts with `http://` or `https://`.
@vilchik-elena WDYT?

vilchik-elena: sounds good

eslint-bridge/src/rules/link-with-target-blank.ts
```diff
@@ -0,0 +1,92 @@
...
+            messageId: 'missingNoopener',
+            node: property,
+          });
+        }
+      },
+    };
+  },
+};
+
+function isThisWindow(node: estree.Node) {
```
vilchik-elena, line 69: there is `isMemberExpression` in utils

eslint-bridge/src/rules/link-with-target-blank.ts
```diff
@@ -0,0 +1,92 @@
...
+  create(context: Rule.RuleContext) {
+    return {
+      CallExpression: (node: estree.CallExpression) => {
+        if (!isMethodCall(node)) {
+          return;
+        }
+        const { object, property } = node.callee;
+
+        if (
+          !isIdentifier(property, 'open') ||
```
vilchik-elena, line 45: if you use early return then make the condition as simple as possible. Here you can extract it into another function or variable (to have opportunity to name the condition)

eslint-bridge/src/rules/link-with-target-blank.ts
```diff
@@ -0,0 +1,92 @@
...
+          (!isIdentifier(object, 'window') && !isThisWindow(object))
+        ) {
+          return;
+        }
+
+        const args = node.arguments;
+        if (args.length >= URL_INDEX && !isHttpUrl(context, args[URL_INDEX])) {
```
vilchik-elena, line 52: same here:
```suggestion
        const hasHttpUrl = URL_INDEX < args.length && isHttpUrl(context, args[URL_INDEX]);
        if (!hasHttpUrl) {
```

Note that I didn't invert `&&`. That way we will also return early if you have 0 arguments.

eslint-bridge/src/rules/link-with-target-blank.ts
```diff
@@ -0,0 +1,92 @@
...
+  return (
+    node.type === 'MemberExpression' &&
+    node.object.type === 'ThisExpression' &&
+    isIdentifier(node.property, 'window')
+  );
+}
+
+function hasRequiredOption(context: Rule.RuleContext, argument: estree.Node) {
+  const literalNode = getValueOfExpression(context, argument, 'Literal');
+  if (literalNode === undefined || !isStringLiteral(literalNode)) {
```
vilchik-elena, line 79: `literalNode === undefined` -> `!literalNode`

eslint-bridge/src/rules/link-with-target-blank.ts
```diff
@@ -0,0 +1,92 @@
...
+    return false;
+  }
+  return literalNode.value.includes(REQUIRED_OPTION);
+}
+
+function isHttpUrl(context: Rule.RuleContext, argument: estree.Node): boolean {
+  const literalNode = getValueOfExpression(context, argument, 'Literal');
```
vilchik-elena, line 86: how about removing duplication here and above?

eslint-bridge/src/rules/post-message.ts
```diff
@@ -57,12 +57,6 @@ export const rule: Rule.RuleModule = {
   },
 };
 
-function isWindowObject(node: estree.Node, context: Rule.RuleContext) {
```
vilchik-elena, line 60: bring all this back

eslint-bridge/tests/rules/comment-based/link-with-target-blank.ts
```diff
@@ -0,0 +1,36 @@
+window.open('https://example.com/dangerous1'); // Noncompliant {{Make sure not using "noopener" is safe here.}}
+    // ^^^^
+window.open('http://example.com/dangerous2'); // Noncompliant
+    // ^^^^
```
vilchik-elena, line 4: IMO no need to assert location for every issue

eslint-bridge/tests/rules/comment-based/link-with-target-blank.ts
```diff
@@ -0,0 +1,36 @@
...
+window.open('https://example.com/dangerous3', 'windowname', 'resizable'); // Noncompliant
+    // ^^^^
+window.open('https://example.com/dangerous4', 'windowname', 123); // Noncompliant
+    // ^^^^
+
+window.open('test.html');
```
vilchik-elena, line 10: add test with 0 arguments

eslint-bridge/tests/rules/comment-based/link-with-target-blank.ts
```diff
@@ -0,0 +1,36 @@
...
+const requiredOption = 'noopener';
+const missingRequiredOption = 'resizable';
+window.open(httpUrl, 'windowname', missingRequiredOption); // Noncompliant
+    // ^^^^
+window.open(otherUrl, 'windowname', missingRequiredOption);
+
+window.open(httpUrl, 'windowname', requiredOption);
+
+this.window.open('https://example.com/dangerous10', 'windowname', 'resizable'); // Noncompliant
+         // ^^^^
```
vilchik-elena, line 36: new line ;)
Just review your changes in github before you assign for review

javascript-checks/src/main/java/org/sonar/javascript/checks/LinkWithTargetBlankCheck.java
```diff
@@ -0,0 +1,38 @@
...
+import org.sonar.plugins.javascript.api.EslintBasedCheck;
+import org.sonar.plugins.javascript.api.JavaScriptRule;
+import org.sonar.plugins.javascript.api.TypeScriptRule;
+
+@JavaScriptRule
+@TypeScriptRule
+@Rule(key = "S5148")
+public class LinkWithTargetBlankCheck implements EslintBasedCheck{
+
+  public static final boolean CONSIDER_JSDOC = false;
```
vilchik-elena, line 32: huh?

vilchik-elena (changes requested):

eslint-bridge/src/rules/link-with-target-blank.ts
ilia-kebets-sonarsource: `extract it into another function or variable (to have the opportunity to name the condition)` awesome!

eslint-bridge/tests/rules/comment-based/link-with-target-blank.ts
ilia-kebets-sonarsource: I need to setup my IDE to flag this.

eslint-bridge/src/rules/link-with-target-blank.ts
ilia-kebets-sonarsource: I'm not sure how you propose to refactor it.

eslint-bridge/tests/rules/comment-based/link-with-target-blank.ts
```diff
@@ -0,0 +1,36 @@
...
+window.open('https://example.com/dangerous6', 'windowname', 'resizable,scrollbars,status,noopener');
+
+function open() {
+    console.log('hello')
+}
+open('https://example.com/dangerous7', 'windowname', 'resizable');
+
+window.notOpen('https://example.com/dangerous8', 'windowname', 'resizable');
+
+const httpUrl = 'https://example.com/dangerous9';
```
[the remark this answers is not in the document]
ilia-kebets-sonarsource: I added https://github.com/SonarSource/SonarJS/pull/3167/files#diff-d110b60ab02d53976ba21ce2ff17b333ed2e97beae59f6b886cc11b69244634aR36

Not sure it's what you suggested.

javascript-checks/src/main/java/org/sonar/javascript/checks/LinkWithTargetBlankCheck.java
ilia-kebets-sonarsource: mistake when copying

eslint-bridge/tests/rules/comment-based/link-with-target-blank.ts
ilia-kebets-sonarsource: changed it in VScode

eslint-bridge/src/rules/link-with-target-blank.ts
vilchik-elena: indeed you are right

vilchik-elena (approved):
Well done!

PR #3171, opened 2022-06-13
Fix comments-based testing: support inline conflict with `//` - triggered by an URL in this case

### LINKED ISSUE #3168: Fix comments-based testing: support inline conflict with `//` - triggered by an URL in this case
The comments-based framework does not behave correctly when you have an URL such as `https://example.com`, because it builds the error based on comments looking for `//`  which are found in the URL.

So, the logic is implemented in 2 phases:
- look for // to extract comments
- look for a regex with Noncompliant  in these comments
We could merge those 2 behaviours, although, we would have to change some other code, such as the secondary which probably uses the same code for comments extraction.

A quickfix (that shifts the problem) is to use `line.lastIndexOf()` instead of `line.indexOf()` here: https://github.com/SonarSource/SonarJS/blob/6798d21cd9fec8da929334460b364d548b0a608c/eslint-bridge/tests/testing-framework/comments.ts#L48

ilia-kebets-sonarsource:
Fixes #3168


eslint-bridge/src/utils/utils-ast.ts
```diff
@@ -128,12 +128,24 @@ export function isFunctionInvocation(
   );
 }
 
+/**
```
vilchik-elena, line 131: drop these docs

eslint-bridge/tests/rules/comment-based/single-char-in-character-classes.ts
```diff
@@ -0,0 +1,73 @@
+let valid = [
```
vilchik-elena, line 1: Let's exclude `comment-based` dir from Prettier

eslint-bridge/tests/rules/rule-test-launcher.test.ts
```diff
@@ -88,5 +88,5 @@ export function parseForESLint(
   );
 }
 
-const ruleTester = new RuleTester({ parser: __filename });
+const ruleTester = new RuleTester({ parser: __filename }); // loading the above parseForESLint() function
```
vilchik-elena, line 91: don't put comments on the code line

eslint-bridge/tests/testing-framework/issues.ts
```diff
@@ -30,7 +33,25 @@ export class FileIssues {
    * @param fileContent
    */
   constructor(fileContent: string) {
-    const comments = extractComments(fileContent);
+    const parsed = buildSourceCode(
```
vilchik-elena, line 36: Put all this logic inside `extractComments`

vilchik-elena (changes requested):

vilchik-elena (approved):

PR #3173, opened 2022-06-15
Fix windows-fix side effects on MacOS: plugin test failing

ilia-kebets-sonarsource:


ilia-kebets-sonarsource:
closed as per https://github.com/SonarSource/SonarJS/issues/3175

PR #3174, opened 2022-06-16
Document that Ruling ITs should be run separately

vilchik-elena:


ilia-kebets-sonarsource (approved):
Tested and validated locally!

PR #3179, opened 2022-06-16
Improve S2259 to cover cases from S1697

victor-diez-sonarsource:
Fixes #3123


eslint-bridge/src/utils/utils-ast.ts
```diff
@@ -47,6 +47,10 @@ export const functionLike = new Set([
   'MethodDefinition',
 ]);
 
+export const equal = new Set(['==', '===']);
```
vilchik-elena, line 50: don't try to put everything into utils, you don't know if this will be useful for someone else. Put it into your rule's module.

eslint-bridge/src/utils/utils-ast.ts
```diff
@@ -190,10 +194,18 @@ export function isLiteral(n: estree.Node | null): n is estree.Literal {
   return n != null && n.type === 'Literal';
 }
 
+export function isNullish(n: estree.Node): boolean {
```
vilchik-elena, line 197: put this into your rule module as well.

eslint-bridge/src/utils/utils-ast.ts
```diff
@@ -190,10 +194,18 @@ export function isLiteral(n: estree.Node | null): n is estree.Literal {
...
+  return isNullLiteral(n) || isUndefined(n);
+}
+
 export function isNullLiteral(n: estree.Node): boolean {
   return isLiteral(n) && n.value === null;
 }
 
+export function isUndefined(node: estree.Node): boolean {
```
vilchik-elena, line 205: if you declare this util function, replace other usages of this logic in the whole project by `isUndefined`.

eslint-bridge/tests/rules/null-dereference.test.ts
```diff
@@ -304,6 +304,22 @@ ruleTesterJsWithTypes.run('', rule, {
       x?.y?.z?.foo;
       `,
     },
+    {
+      code: `
+      if (x != null && x.prop == 0) {
+        console.log("String is empty");
```
vilchik-elena, line 310: remove these `log`s. It does not help here. You can have empty block instead `{}`

eslint-bridge/tests/rules/null-dereference.test.ts
```diff
@@ -304,6 +304,22 @@ ruleTesterJsWithTypes.run('', rule, {
...
+      if (x == null || x.prop == 0) {
+        console.log("String is empty");
+      }
+      if (x != undefined && x.prop == 0) {
+        console.log("String is empty");
+      }
+      if (x == undefined || x.prop == 0) {
+        console.log("String is empty");
+      }
+      `,
```
vilchik-elena, line 321: I would expect a test when it's not the same variable `x == undefined && y.prop == 42`

eslint-bridge/tests/rules/null-dereference.test.ts
```diff
@@ -425,6 +441,48 @@ ruleTesterJsWithTypes.run('', rule, {
...
+        if (undefined == x && x.prop == 0) {
+          console.log("String is empty");
+        }
+        if (x.prop != undefined || x.prop.prop2 == 0) {
+          console.log("String is empty");
+        }
+      `,
+      errors: [
+        {
+          line: 2,
```
vilchik-elena, line 461: assert message

vilchik-elena, line 461: You don't have to assert everything for every error. I would assert full location for first error, and then just start `line` and `column`

eslint-bridge/tests/rules/null-dereference.test.ts
```diff
@@ -425,6 +441,48 @@ ruleTesterJsWithTypes.run('', rule, {
         x.foo`,
       errors: 1,
     },
+    {
+      code: `
+        if (x == null && x.prop == 0) {
```
vilchik-elena, line 446: it's better to keep invalid cases separately. If it fails, it's clearer what's wrong.

eslint-bridge/src/rules/null-dereference.ts
```diff
@@ -27,71 +27,140 @@ import {
...
+      node,
+    });
+  }
+}
 
 export const rule: Rule.RuleModule = {
   meta: {
     messages: {
       nullDereference: 'TypeError can be thrown as "{{symbol}}" might be null or undefined here.',
+      shortCircuitError: 'MemberExpression on null or undefined reference',
```
vilchik-elena, line 135: Take the message from https://sonarsource.github.io/rspec/#/rspec/S1697/javascript

eslint-bridge/src/rules/null-dereference.ts
```diff
@@ -27,71 +27,140 @@ import {
...
+import { areEquivalent } from 'eslint-plugin-sonarjs/lib/utils/equivalence';
+
+export enum Nullish {
+  confirmed,
+  discarded,
+  unknown,
+}
+
+//Check nullish value implicitly by Binary Expression
+function binaryExpressionImplicitNullish(
```
vilchik-elena, line 44: we usually put all the internal functions after rule export statement

eslint-bridge/src/rules/null-dereference.ts
```diff
@@ -27,71 +27,140 @@ import {
...
 } from '../utils';
+import { areEquivalent } from 'eslint-plugin-sonarjs/lib/utils/equivalence';
+
+export enum Nullish {
+  confirmed,
+  discarded,
+  unknown,
+}
+
+//Check nullish value implicitly by Binary Expression
```
vilchik-elena, line 43: use jsdoc `/**` format you want to describe what your function is doing. But better find the name which is explicit enough.

eslint-bridge/src/rules/null-dereference.ts
```diff
@@ -27,71 +27,140 @@ import {
...
   isUndefinedOrNull,
   findFirstMatchingAncestor,
+  RuleContext,
+  not_equal,
+  equal,
+  isNullish,
 } from '../utils';
+import { areEquivalent } from 'eslint-plugin-sonarjs/lib/utils/equivalence';
+
+export enum Nullish {
```
vilchik-elena, line 37: I saw this word used only in the context of `??` operator (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator). I would find something else here (`UndefinedState`).

vilchik-elena, line 37: why exporting it?

eslint-bridge/src/utils/utils-ast.ts
```diff
@@ -47,6 +47,10 @@ export const functionLike = new Set([
   'MethodDefinition',
 ]);
 
+export const equal = new Set(['==', '===']);
+
+export const not_equal = new Set(['!=', '!==']);
```
vilchik-elena, line 52: bad casing, `notEqual`

eslint-bridge/src/utils/utils-ast.ts
vilchik-elena, line 50: I would use some more explicit name, e.g. `equalOperators`, especially keeping in mind you wanted it in utils.

eslint-bridge/src/rules/null-dereference.ts
```diff
@@ -27,71 +27,140 @@ import {
...
+      const enclosingFn = findFirstMatchingAncestor(ref.identifier as TSESTree.Node, node =>
+        functionLike.has(node.type),
+      );
+      return enclosingFn && enclosingFn !== fn;
+    }
+    return false;
+  });
+}
+
+//Check NullDereference in scope declaration
```
vilchik-elena, line 98: I am not sure what this phase means

eslint-bridge/src/rules/null-dereference.ts
```diff
@@ -27,71 +27,140 @@ import {
...
       MemberExpression(node: estree.Node) {
         const { object, optional } = node as estree.MemberExpression;
         if (!optional) {
-          checkNullDereference(object);
+          checkNullDereference(object, context, alreadyRaisedSymbols);
+        }
+      },
+      'LogicalExpression MemberExpression'(node: estree.Node) {
+        const { object, optional } = node as estree.MemberExpression;
+        if (!optional) {
```
vilchik-elena, line 153: I don't see test for it

eslint-bridge/src/rules/null-dereference.ts
```diff
@@ -27,71 +27,140 @@ import {
...
+    (isNullish(left) &&
+      areEquivalent(right as TSESTree.Node, id as TSESTree.Node, context.getSourceCode()))
+  ) {
+    if (not_equal.has(expr.operator)) return Nullish.discarded;
+    if (equal.has(expr.operator)) return Nullish.confirmed;
+  }
+  return Nullish.unknown;
+}
+
+//Check Null Dereference by implicit logical expression
```
vilchik-elena, line 62: I don't get this one as well. Why logical expression is implicit?

eslint-bridge/src/rules/null-dereference.ts
```diff
@@ -27,71 +27,140 @@ import {
...
+function checkLogicalNullDereference(
+  expr: estree.LogicalExpression,
+  id: estree.Node,
```
vilchik-elena, line 65: rename, it's not identifier

eslint-bridge/src/rules/null-dereference.ts
vilchik-elena, line 44: add verb to this function name. I would name it `getUndefinedState`

eslint-bridge/src/rules/prefer-type-guard.ts
```diff
@@ -67,9 +67,9 @@ function checkFunctionLikeDeclaration(
 
   if (isInequalityBinaryExpression(returnedExpression)) {
     const { left, right } = returnedExpression;
-    if (isUndefined(right)) {
+    if (isUndefined(right as estree.Node)) {
```
vilchik-elena, line 70: I wonder if we can have both `estree.Node` and `TSESTree.Node` as parameter type to avoid such casts. Could you check?

eslint-bridge/tests/rules/null-dereference.test.ts
```diff
@@ -304,6 +304,24 @@ ruleTesterJsWithTypes.run('', rule, {
...
+    {
+      code: `
+      if (x != null && x.prop == 0) {}
+      if (x == null || x.prop == 0) {}
+      if (x != undefined && x.prop == 0) {}
+      if (x == undefined || x.prop == 0) {}
+      `,
+    },
+    {
+      code: `
```
vilchik-elena, line 316: make this literal (and next ones when it's a one line of code only) on one line 
```
code: `if (x == null && x?.prop == 0) {}`
```

eslint-bridge/tests/rules/null-dereference.test.ts
```diff
@@ -425,6 +443,62 @@ ruleTesterJsWithTypes.run('', rule, {
...
+          endColumn: 27,
+          message: 'TypeError can be thrown as expression might be null or undefined here.',
+        },
+      ],
+    },
+    {
+      code: `
+        if (null != x || x.prop == 0) {}
+      `,
+      errors: [
```
vilchik-elena, line 464: btw you can use `messageId` for the assertion. I would assert only it as it's not so interesting to assert the whole message and location again and again. It decreases the readability of the test. (same for the cases below)

eslint-bridge/tests/rules/null-dereference.test.ts
```diff
@@ -425,6 +443,62 @@ ruleTesterJsWithTypes.run('', rule, {
...
+          endLine: 2,
+          column: 31,
+          endColumn: 32,
+          message: 'TypeError can be thrown as expression might be null or undefined here.',
+        },
+      ],
+    },
+    {
+      code: `
+        if (x.prop != undefined || x.prop.prop2 == 0) {}
```
vilchik-elena, line 490: here it might be interesting to assert start and end columns

eslint-bridge/src/rules/null-dereference.ts
```diff
@@ -27,75 +27,143 @@ import {
...
+  return Null.unknown;
+}
+
+function checkLogicalNullDereference(
+  expr: estree.LogicalExpression,
+  node: estree.Node,
+  context: Rule.RuleContext,
+) {
+  if (expr.left.type === 'BinaryExpression') {
+    const nullish = getNullState(expr.left, node, context as unknown as RuleContext);
```
vilchik-elena, line 114: rename variable

vilchik-elena (changes requested):

PR #3181, opened 2022-06-21
Fix 3170 Ignore Redux reducers for S1788 (default-param-last) 

### LINKED ISSUE #3170: Ignore Redux reducers for S1788 (`default-param-last`)
```js
export default function appReducer(state = initialState, action) {
// ..
}
```

https://redux.js.org/tutorials/fundamentals/part-3-state-actions-reducers#creating-the-root-reducer

We can make this exclusion work based on the parameters number and names (if the first parameter has name `state` and second `action`, default for the first one should be ignored).

Raised in internal thread https://discuss.sonarsource.com/t/s1788-function-parameters-with-default-values-should-be-last-not-applied-to-redux-methods/10797

victor-diez-sonarsource:
Fixes #3170

### DIFF
--- a/eslint-bridge/src/rules/decorators/default-param-last-decorator.ts
+++ b/eslint-bridge/src/rules/decorators/default-param-last-decorator.ts
@@ -1,0 +1,57 @@
+/*
+ * SonarQube JavaScript Plugin
+ * Copyright (C) 2011-2022 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+import { Rule } from 'eslint';
+import { interceptReport, isIdentifier } from '../../utils';
+import { BaseFunction } from 'estree';
+import { TSESTree } from '@typescript-eslint/experimental-utils';
+
+export function decorateDefaultParamLast(rule: Rule.RuleModule): Rule.RuleModule {
+  return interceptReport(rule, reportExempting(isReduxReducer));
+}
+
+function reportExempting(
+  exemptionCondition: (enclosingFunction: BaseFunction) => boolean,
+): (context: Rule.RuleContext, reportDescriptor: Rule.ReportDescriptor) => void {
+  return (context, reportDescriptor) => {
+    const scope = context.getScope();
+    if ('node' in reportDescriptor) {
+      const node = reportDescriptor['node'] as TSESTree.AssignmentPattern;
+      const variable = scope.variables.find(
+        value => node.left.type === 'Identifier' && node.left.name === value.name,
+      );
+      if (!exemptionCondition(variable?.defs?.[0]?.node as BaseFunction)) {
+        context.report(reportDescriptor);
+      }
+    }
+  };
+}
+
+function isReduxReducer(enclosingFunction: BaseFunction) {
+  if (enclosingFunction?.params?.length === 2) {
+    const [firstParam, secondParam] = enclosingFunction.params;
+    if (
+      firstParam.type === 'AssignmentPattern' &&
+      isIdentifier(firstParam.left, 'state') &&
+      isIdentifier(secondParam, 'action')
+    )
+      return true;
+  }
+  return false;
+}
--- a/eslint-bridge/tests/rules/decorators/default-param-last-decorator.test.ts
+++ b/eslint-bridge/tests/rules/decorators/default-param-last-decorator.test.ts
@@ -41,5 +41,13 @@ ruleTester.run(`Function parameters with default values should be last`, rule, {
       code: `function multiply(a = 1, b) {}`,
       errors: 1,
     },
+    {
+      code: `function appReducer(state = initialState, action, param) {}`,
+      errors: 1,
+    },
+    {
+      code: `function appReducer(status = initialState, action) {}`,
+      errors: 1,
+    },
   ],
 });
--- a/eslint-bridge/tests/rules/decorators/default-param-last-decorator.test.ts
+++ b/eslint-bridge/tests/rules/decorators/default-param-last-decorator.test.ts
@@ -1,0 +1,45 @@
+/*
+ * SonarQube JavaScript Plugin
+ * Copyright (C) 2011-2022 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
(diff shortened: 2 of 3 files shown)


eslint-bridge/src/rules/decorators/default-param-last-decorator.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export function decorateDefaultParamLast(rule: Rule.RuleModule): Rule.RuleModule {
+  return interceptReport(rule, reportExempting(isReduxReducer));
+}
+
+function reportExempting(
+  exemptionCondition: (enclosingFunction: BaseFunction) => boolean,
+): (context: Rule.RuleContext, reportDescriptor: Rule.ReportDescriptor) => void {
+  return (context, reportDescriptor) => {
+    const scope = context.getScope();
```
yassin-kammoun-sonarsource, line 35: We can move this declaration before its use.

eslint-bridge/src/rules/decorators/default-param-last-decorator.ts
```diff
@@ -0,0 +1,61 @@
...
+    if ('node' in reportDescriptor) {
+      const node = reportDescriptor['node'] as TSESTree.AssignmentPattern;
+      const variable = scope.variables.find(
+        value => node.left.type === 'Identifier' && node.left.name === value.name,
+      );
```
yassin-kammoun-sonarsource, line 40:
```suggestion
      const variable = scope.variables.find(value => isIdentifier(node.left, value.name));
```

eslint-bridge/src/rules/decorators/default-param-last-decorator.ts
```diff
@@ -0,0 +1,61 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+import { Rule } from 'eslint';
+import { interceptReport, isIdentifier } from '../../utils';
+import { BaseFunction } from 'estree';
+import { TSESTree } from '@typescript-eslint/experimental-utils';
```
yassin-kammoun-sonarsource, line 23: We are mixing two different APIs to manipulate the abstract syntax tree, which is okay. However, looking at your implementation, it seems that you don't really need to use `'@typescript-eslint/experimental-utils'` at all. I suggest you drop this import and rely entirely instead on `'estree'`.

yassin-kammoun-sonarsource: You might need to change the `'estree'` import to `import * as estree ...` and adapt a few type annotations here and there.

eslint-bridge/tests/rules/decorators/default-param-last-decorator.test.ts
```diff
@@ -0,0 +1,53 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+import { Linter } from 'eslint';
+import { decorateDefaultParamLast } from 'rules/decorators/default-param-last-decorator';
+import { RuleTesterJsWithTypes } from '../../RuleTesterJsWithTypes';
```
yassin-kammoun-sonarsource, line 22: Since we don't test something specific about type information, we can just use the `RuleTester` from `eslint` instead like we do here for example:

https://github.com/SonarSource/SonarJS/blob/a25b7ac77906b72bcb92120ad1f866588f43f690/eslint-bridge/tests/rules/decorators/no-empty-decorator.test.ts#L20-L24

eslint-bridge/src/rules/decorators/default-param-last-decorator.ts
```diff
@@ -0,0 +1,61 @@
...
+function isReduxReducer(enclosingFunction: BaseFunction) {
+  if (enclosingFunction.params.length === NUM_ARGS_REDUX_REDUCER) {
+    const [firstParam, secondParam] = enclosingFunction.params;
+    if (
+      firstParam.type === 'AssignmentPattern' &&
+      isIdentifier(firstParam.left, 'state') &&
+      isIdentifier(secondParam, 'action')
+    ) {
+      return true;
+    }
```
yassin-kammoun-sonarsource, line 58:
```suggestion
    return (
      firstParam.type === 'AssignmentPattern' &&
      isIdentifier(firstParam.left, 'state') &&
      isIdentifier(secondParam, 'action')
    );
```

yassin-kammoun-sonarsource: I am nitpicking here. Feel free to ignore that one if you prefer to explicitly return `true`.

eslint-bridge/src/rules/decorators/default-param-last-decorator.ts
```diff
@@ -0,0 +1,61 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+import { Rule } from 'eslint';
```
yassin-kammoun-sonarsource, line 20:
```suggestion
// https://sonarsource.github.io/rspec/#/rspec/S1788/javascript

import { Rule } from 'eslint';
```

yassin-kammoun-sonarsource: For your information, for each rule implemented or decorated, we tend to include in the source code a link to the rule description. This helps quickly find the corresponding documentation.

yassin-kammoun-sonarsource (changes requested):

eslint-bridge/src/rules/decorators/default-param-last-decorator.ts
victor-diez-sonarsource: no, you are totally right. I refactored that function so many times I did't realize I was just returning the result of the condition

eslint-bridge/src/rules/decorators/default-param-last-decorator.ts
```diff
@@ -0,0 +1,58 @@
...
+}
+
+function reportExempting(
+  exemptionCondition: (enclosingFunction: BaseFunction) => boolean,
+): (context: Rule.RuleContext, reportDescriptor: Rule.ReportDescriptor) => void {
+  return (context, reportDescriptor) => {
+    if ('node' in reportDescriptor) {
+      const node = reportDescriptor['node'] as AssignmentPattern;
+      const scope = context.getScope();
+      const variable = scope.variables.find(value => isIdentifier(node.left as Node, value.name));
```
yassin-kammoun-sonarsource, line 39:
```suggestion
      const variable = scope.variables.find(value => isIdentifier(node.left, value.name));
```

yassin-kammoun-sonarsource: I tried locally, and it seems unnecessary to cast to `Node` unless I missed something.

eslint-bridge/src/rules/decorators/default-param-last-decorator.ts
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+// https://sonarsource.github.io/rspec/#/rspec/S1788/javascript
+
+import { Rule } from 'eslint';
+import { interceptReport, isIdentifier } from '../../utils';
+import { AssignmentPattern, BaseFunction, Node } from 'estree';
```
yassin-kammoun-sonarsource, line 24:
```suggestion
import { AssignmentPattern, BaseFunction } from 'estree';
```

yassin-kammoun-sonarsource (changes requested):

eslint-bridge/src/rules/decorators/default-param-last-decorator.ts
victor-diez-sonarsource: indeed, it's unnecessary with estree.AssignmentPattern. When I changed the line I was still using the tstree type instead, which raised this: 
image

yassin-kammoun-sonarsource: I see :)

yassin-kammoun-sonarsource:
Let's update the RSPEC before merging.

PR #3184, opened 2022-06-22
Fix FP S3616 ('comma-or-logical-or-case'): Ignore `||`-operator in `switch(true)`

### LINKED ISSUE #2645: FP on S3616 when using `||`-operator in `switch(true)`
We should not report on all `||` expressions in switch case:
Here is an example of such FP, when `switch (true)` construct is used, still it's not clear if we should ignore only that or any boolean-typed argument. I would go only for `switch(true)` as it's a clear pattern and anyway no much noise is generated (0 issues on Peach)
```
switch (true) {
  case cond1() || cond2(): 
    // works as expected, `cond2()` will be evaluated if `cond1()` is false
    break;
}
```

[source](https://discuss.sonarsource.com/t/potential-fp-on-javscript-s3616/7842/2)

victor-diez-sonarsource:
Fixes #2645


yassin-kammoun-sonarsource (approved):
Looks good to me. I left a minor suggestion to address. We can then document the newly introduced exception in the RSPEC, fetch the changes in this same PR, and eventually merge it.

PR #3197, opened 2022-06-29
Fix JavaScript execution environments description

yassin-kammoun-sonarsource:
We are including the list of default JavaScript execution environments in the JavaScript analyzer's description on SonarQube (see [Peach(internal link)](https://peach.sonarsource.com/admin/settings?category=JavaScript+%2F+TypeScript)). It seems that there is a CSS issue on the SonarQube UI that we need to dogfood. Still, it does not make sense to include this list of defaults in the description itself as SQ will take care displaying it if user changed to a non-default.

Screenshot 2022-06-28 at 11 20 16


yassin-kammoun-sonarsource:
Screenshot 2022-06-30 at 10 46 04

vilchik-elena (approved):

PR #3220, opened 2022-07-11
Add S6439: React components should not render non-boolean condition values

### LINKED ISSUE #3205: Add rule S6439: React components should not render non-boolean condition values
[RSPEC PR](https://github.com/SonarSource/rspec/pull/1091)

We should try first `react/jsx-no-leaked-render` as implementation but there is a big chance it will be noisy as it's not relying on type information.

victor-diez-sonarsource:
Fixes #3205


saberduck (changes requested):

eslint-bridge/src/rules/sonar-jsx-no-leaked-render.ts
```diff
@@ -0,0 +1,105 @@
...
+import { Rule, SourceCode } from 'eslint';
+import * as estree from 'estree';
+import { getTypeFromTreeNode, isBigIntType, isNumberType, isStringType } from '../utils';
+
+export const rule: Rule.RuleModule = {
+  meta: {
+    hasSuggestions: true,
+    messages: {
+      noPotentialLeakedRender:
+        'Potential leaked value that might cause unintentionally rendered values or rendering crashes',
```
saberduck, line 31: Issue message should be actionable, from this message it is not clear to the user how to fix the problem. I would rephrase it to something like "Convert value to a boolean type to avoid rendering issues"

eslint-bridge/src/rules/sonar-jsx-no-leaked-render.ts
```diff
@@ -0,0 +1,105 @@
...
+      suggestCoercion: 'Coerce the conditional to a boolean',
```
saberduck, line 32: I would replace "coerce" to "convert", it's a more common and easier-to-understand word.

saberduck (changes requested):

eslint-bridge/src/rules/sonar-jsx-no-leaked-render.ts
victor-diez-sonarsource: just kept the message from original eslint plugin implementation. I'll modify that

eslint-bridge/src/rules/sonar-jsx-no-leaked-render.ts
victor-diez-sonarsource: same as the message, I kept the suggestion from the original. I'll change that now

saberduck (approved):

PR #3224, opened 2022-07-12
Add rule S6442: React's setState hook should only be used in the render function or body of a component

### LINKED ISSUE #3219: Add rule S6442: React's setState hook should only be used in the render function or body of a component
React's setState hook setter function should not be called directly in the body of a component, otherwise it would produce an infinite render loop. This can happen by mistake. Most commonly, the setter function is called from an event handler.

Non compliant code:
```js
import { useState } from "react";

function ShowLanguage() {

    // Makes an infinite loop
    setLanguage(navigator.language);

    return (
      <section>
        <h1>Your language is {language}!</h1>
        <button onClick={() => setLanguage("fr-FR")}>Je préfère le Français</button>
      </section>
    );
};
```

Compliant code:
```js
import { useState } from "react";

function ShowLanguage() {
    const [language, setLanguage] = useState(navigator.language);

    return (
      <section>
        <h1>Your language is {language}!</h1>
        <button onClick={() => setLanguage("fr-FR")}>Je préfère le Français</button>
      </section>
    );
};
```

saberduck:
Fixes #3219

### DIFF
--- a/eslint-bridge/src/rules/no-hook-setter-in-body.ts
+++ b/eslint-bridge/src/rules/no-hook-setter-in-body.ts
@@ -50,7 +50,8 @@ const HOOK_FUNCTION = 'useState';
 export const rule: Rule.RuleModule = {
   meta: {
     messages: {
-      noHookSetterInBody: 'Move the state setter call to a callback',
+      noHookSetterInBody:
+        'Remove this state setter call, perhaps move it to an event handler or JSX attribute',
     },
   },
   create(context: Rule.RuleContext) {
--- a/eslint-bridge/tests/rules/comment-based/no-hook-setter-in-body.js
+++ b/eslint-bridge/tests/rules/comment-based/no-hook-setter-in-body.js
@@ -12,7 +12,7 @@ function notAReactComponent() {
 function ShowLanguageInvalid() {
   const [language, setLanguage] = useState("fr-FR");
 
-  setLanguage(navigator.language); // Noncompliant {{Move the state setter call to a callback}}
+  setLanguage(navigator.language); // Noncompliant {{Remove this state setter call, perhaps move it to an event handler or JSX attribute}}
 //^^^^^^^^^^^
 
   return (
--- a/javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S6442.html
+++ b/javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S6442.html
@@ -1,4 +1,4 @@
-<p>React’s setState hook setter function should not be called directly in the body of a component, otherwise it would produce an infinite render loop.
+<p>React’s useState hook setter function should not be called directly in the body of a component, otherwise it would produce an infinite render loop.
 This can happen by mistake. Most commonly, the setter function is called from an event handler.</p>
 <h2>Noncompliant Code Example</h2>
 <pre>


eslint-bridge/src/rules/no-hook-setter-in-body.ts
```diff
@@ -0,0 +1,225 @@
...
+
+type SetterCall = estree.CallExpression & {
+  callee: estree.Identifier;
+};
+
+export const rule: Rule.RuleModule = {
+  meta: {
+    messages: {
+      noHookSetterInBody:
+        "React's setState hook should only be used in the render function or body of a component",
```
saberduck, line 52: we should change the message to suggest the action which should be taken (to be actionable). Something like "Move the setState hook to the render function or body of a component"

eslint-bridge/src/rules/no-hook-setter-in-body.ts
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    /**
+     * Returns the current scope if it corresponds to a React component function or undefined otherwise.
+     */
+    function getReactComponentScope(): Scope | undefined {
+      const scope = context.getScope();
+      const isReact =
+        scope.type === 'function' &&
+        isFunctionNode(scope.block) &&
```
saberduck, line 110: this condition seems to be redundant to `scope.type === function`, do we need it?

eslint-bridge/src/rules/no-hook-setter-in-body.ts
```diff
@@ -0,0 +1,225 @@
...
+        }
+      },
+    };
+  },
+};
+
+/**
+ * Check if the node is an eslint Rule.Node that's to say it has a parent property.
+ */
+function isRuleNode(node: estree.Node): node is Rule.Node {
```
saberduck, line 188: I don't think we need a function for this, we can just inline the check for `node.parent`

eslint-bridge/src/rules/no-hook-setter-in-body.ts
francoismora: Indeed it is redundant. I will remove it. Thanks.

eslint-bridge/src/rules/no-hook-setter-in-body.ts
francoismora: Changed

eslint-bridge/src/rules/no-hook-setter-in-body.ts
francoismora: The 'isRuleNode' function avoids casting 'node' twice. Without it, the inlined code looks like the below. Do you prefer me to remove it still?
image

eslint-bridge/src/rules/no-hook-setter-in-body.ts
```diff
@@ -0,0 +1,209 @@
...
+    function getReactComponentScope(): Scope | undefined {
+      const scope = context.getScope();
+      const isReact = isFunctionNode(scope.block) && matches(scope.block, REACT_PATTERN, 1);
+      return isReact ? scope : undefined;
+    }
+
+    /**
+     * Returns the closest enclosing scope of type 'function' or undefined if there isn't any.
+     */
+    function isInsideFunction(scope: Scope | undefined): boolean {
```
vilchik-elena, line 104: I think we should stick to only `null` or  only `undefined` (looking at signatures of these functions)

vilchik-elena, line 104: I would rename parameter `scope` to `functionScope`

javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S6442.json
```diff
@@ -0,0 +1,19 @@
+{
+  "title": "React\u0027s setState hook should only be used in the render function or body of a component",
+  "type": "BUG",
+  "status": "ready",
+  "remediation": {
+    "func": "Constant\/Issue",
+    "constantCost": "5min"
+  },
+  "tags": [],
```
vilchik-elena, line 9: the `react` tag is missing

eslint-bridge/tests/rules/comment-based/no-hook-setter-in-body.js
```diff
@@ -0,0 +1,220 @@
...
+  const [, forceUpdate] = React.useReducer(x => x + 1, 0);
+  return forceUpdate;
+}
+
+// Function Declarations..............
+
+function ShowLanguageInvalid() {
+  const [language, setLanguage] = useState("fr-FR");
+
+  // Makes an infinite loop
```
vilchik-elena, line 15: remove those comments

eslint-bridge/tests/rules/comment-based/no-hook-setter-in-body.js
```diff
@@ -0,0 +1,220 @@
...
+  setLanguage(navigator.language); // Noncompliant {{Move the setState hook to the render function or body of a component}}
+//^^^^^^^^^^^
+
+  return (
```
vilchik-elena, line 19: I would simplify everything which is possible in unit test cases, we need such a big jsx

eslint-bridge/tests/rules/comment-based/no-hook-setter-in-body.js
```diff
@@ -0,0 +1,220 @@
...
+
+    return (
+        <section>
+            <h1>Your language is {language}!</h1>
+            <button onClick={() => setLanguage("fr-FR")}>Je préfère le Français</button>
+        </section>
+    );
+}
+
+function ShowLanguageValid() {
```
vilchik-elena, line 42: in fact we tested it already in previous cases

eslint-bridge/tests/rules/comment-based/no-hook-setter-in-body.js
```diff
@@ -0,0 +1,220 @@
...
+  );
+}
+
+// Arrow Functions..............
+
+const ShowLanguageInvalidArrow = () => {
+  const [language, setLanguage] = useState("fr-FR");
+
+  // Makes an infinite loop
+  setLanguage(navigator.language); // Noncompliant {{Move the setState hook to the render function or body of a component}}
```
vilchik-elena, line 59: fyi there is no need to assert message and location everytime

eslint-bridge/tests/rules/comment-based/no-hook-setter-in-body.js
```diff
@@ -0,0 +1,220 @@
...
+    <section>
+      <h1>Your language is {language}!</h1>
+      <button onClick={() => setLanguage("fr-FR")}>Je préfère le Français</button>
+    </section>
+  );
+};
+
+// MemberInvalid..............
+
+exports.MemberInvalid = () => {
```
vilchik-elena, line 111: isn't it just arrow function?

eslint-bridge/src/rules/no-hook-setter-in-body.ts
```diff
@@ -0,0 +1,209 @@
...
+        }
+      },
+    };
+  },
+};
+
+/**
+ * Check if the node is an eslint Rule.Node that's to say it has a parent property.
+ */
+function isRuleNode(node: estree.Node): node is Rule.Node {
```
vilchik-elena, line 176: just rename `hasParent` and no need for any comment

eslint-bridge/src/rules/no-hook-setter-in-body.ts
```diff
@@ -0,0 +1,209 @@
...
+ * <ul>
+ *   <li>Identifier: it uses the name
+ *   <li>FunctionDeclaration: it uses the id.
+ *   <li>VariableDeclarator: it uses the id.
+ *   <li>AssignmentExpression: it uses the left part.
+ *   <li>MemberExpression: it uses the property part.
+ * </ul>
+ * If the node is not supported calls itself recursively on the parent node up to max level.
+ */
+function matches(node: estree.Node | null, pattern: RegExp, max = 0): node is estree.Identifier {
```
vilchik-elena, line 191: you only call this one with react regex, I would remove this parameter to make the semantic of the function explicit (`isReactComponent`)

vilchik-elena, line 191: that's a weird return type... it's not true only when node is Identifier

vilchik-elena (dismissed):
Try to keep to minimum the amount of comments, try to find good variable names, so that just reading the code one can understand what's going on

eslint-bridge/src/rules/no-hook-setter-in-body.ts
```diff
@@ -0,0 +1,209 @@
...
+  getModuleNameOfImportedIdentifier,
+  getVariableFromName,
+  isFunctionNode,
+  isIdentifier,
+  isMemberExpression,
+} from '../utils';
+import Variable = EsLintScope.Variable;
+import Scope = EsLintScope.Scope;
+
+// Types used in the hook declaration callback signature which reflects the expectations coming from the selector.
```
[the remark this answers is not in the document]
francoismora: Ok, I was not too sure the intent for this types would be clear. So I guess I will just remove it then.

eslint-bridge/src/rules/no-hook-setter-in-body.ts
```diff
@@ -0,0 +1,209 @@
...
+     * module name. In the example below <code>isReactName(node)</code> will return true.
+     * @example
+     * import { useState } from "react";
+     *
+     * function ShowLanguageInvalid() {
+     *   const [language, setLanguage] = useState('fr-FR'); // 'useState' id is a React name.
+     *   // [...]
+     * }
+     */
+    function isReactName(identifier: estree.Identifier): boolean {
```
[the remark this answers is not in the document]
francoismora: Ok, I will remove it then.

eslint-bridge/src/rules/no-hook-setter-in-body.ts
```diff
@@ -0,0 +1,209 @@
...
+     * Returns the current scope if it corresponds to a React component function or undefined otherwise.
+     */
+    function getReactComponentScope(): Scope | undefined {
+      const scope = context.getScope();
+      const isReact = isFunctionNode(scope.block) && matches(scope.block, REACT_PATTERN, 1);
+      return isReact ? scope : undefined;
+    }
+
+    /**
+     * Returns the closest enclosing scope of type 'function' or undefined if there isn't any.
```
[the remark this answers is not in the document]
francoismora: As suggested above, I will simply remove the comments.

eslint-bridge/src/rules/no-hook-setter-in-body.ts
```diff
@@ -0,0 +1,209 @@
...
+};
+
+type SetterCall = estree.CallExpression & {
+  callee: estree.Identifier;
+};
+
+export const rule: Rule.RuleModule = {
+  meta: {
+    messages: {
+      noHookSetterInBody: 'Move the setState hook to the render function or body of a component',
```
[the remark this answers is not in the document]
francoismora: Updating to this until there's a better version: "Move the state setter call to a callback"

eslint-bridge/src/rules/no-hook-setter-in-body.ts
```diff
@@ -0,0 +1,209 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+// https://sonarsource.github.io/rspec/#/rspec/S6442/javascript
+
+import { Rule, Scope as EsLintScope } from 'eslint';
```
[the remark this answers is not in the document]
francoismora: Renaming to ESLintScope...

javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S6442.json
francoismora: Added. What's the goal of this tag?

eslint-bridge/tests/rules/comment-based/no-hook-setter-in-body.js
francoismora: Removed

eslint-bridge/tests/rules/comment-based/no-hook-setter-in-body.js
francoismora: Ok, removing

eslint-bridge/tests/rules/comment-based/no-hook-setter-in-body.js
francoismora: Keeping only the first one.

eslint-bridge/src/rules/no-hook-setter-in-body.ts
francoismora: Renamed. I would have imagined that a type predicate should be named isXxx() in general.

eslint-bridge/src/rules/no-hook-setter-in-body.ts
francoismora: Inlined.

francoismora: Indeed, that's an error. Thanks.

eslint-bridge/tests/rules/comment-based/no-hook-setter-in-body.js
francoismora: The implementation is different if this is an 'AssignmentExpression' or a 'VariableDeclarator' as we need to check the name the arrow function is assigned to. So the test aim at covering both cases.

gabriel-vivas-sonarsource (approved):
The rule description LGTM :+1:

eslint-bridge/src/rules/no-hook-setter-in-body.ts
```diff
@@ -0,0 +1,169 @@
...
+
+const REACT_MODULE = 'react';
+const REACT_ROOT = 'React';
+const REACT_PATTERN = /^[^a-z]/;
+const HOOK_FUNCTION = 'useState';
+
+export const rule: Rule.RuleModule = {
+  meta: {
+    messages: {
+      noHookSetterInBody: 'Move the state setter call to a callback',
```
gabriel-vivas-sonarsource, line 53: Maybe:
```diff
- noHookSetterInBody: 'Move the state setter call to a callback',
+ noHookSetterInBody: 'Remove this state setter call, perhaps move it to an event handler or JSX attribute',
```

gabriel-vivas-sonarsource (approved):
Reviewed the description and messages, some small tweaks, otherwise LGTM 👍

Issue #3194, opened 2022-06-28
tsconfig files should be analyzed in the order from deeper directories


vilchik-elena:
When we added [`moose`](https://github.com/ritz078/moose) project to the Ruling, we faced following problem: when running on Windows (@victor-diez-sonarsource machine) and Cirrus QA (Windows OS), rule S4328 (`no-implicit-dependencies`) starts to raise 57 FPs. 

The root problem is that many files are analyzed with the root `tsconfig.json` and not `renderer/tsconfig.json` which contains `paths` configuration. This happens as tsconfig discovery is [system-dependent](https://github.com/SonarSource/SonarJS/blob/55ff7e7a9585133641bc519ad7c73325b85f5f4b/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/TsConfigProvider.java#L122) and it can happen that the more specific or a less specific `tsconfig` is first in the list.

To avoid this problem for `moose` project the root `tsconfig.json` [was renamed](https://github.com/SonarSource/typescript-test-sources/commit/0058034ff8f81ccbbcec0cc3818444cb106fe3f0) to `tsconfig.base.json`. So to prove that the problem is fixed we can rename it back.

vilchik-elena:
I close this issue as after some investigation it appears that moose project's configuration is buggy (top level inherited tsconfig defines the files to include `**/*.ts`, so every file will be part of every sub-project). So it seems like a corner case for a badly configured project.

vilchik-elena:
The same happens for `jest`. The root `tsconfig.json` is used only as a parent for other tsconfigs. In a perfect world it should be called smth like `tsconfig.base.json`.

PR #3225, opened 2022-07-12
Fix project structure for 'moose' in TS Ruling

vilchik-elena:
4 missing issues on the file 'CastDeviceType.ts' which is not seems to
be used

relates to #3194


vilchik-elena:
@victor-diez-sonarsource could you confirm that this change makes ruling green on your machine?

victor-diez-sonarsource:
> @victor-diez-sonarsource could you confirm that this change makes ruling green on your machine?
all good!
image

victor-diez-sonarsource (approved):

PR #3227, opened 2022-07-13
Add rule S6443: React state setter function should not be called with its matching state variable

### LINKED ISSUE #3218: Add rule S6443: React state setter function should not be called with its matching state variable
React's setState hook gives you a value and a setter function. When the setter function is called with the value, nothing will happen.

This is can happen by mistake when attempting to reset a default value, invert a boolean, among others.

Non compliant code:
```js
import { useState } from "react";

function ShowLanguage() {
    return (
      <section>
        <h1>Your language is {language}!</h1>
        <button onClick={() => setLanguage(navigator.language)}>Detect language</button>
    	{/* this button does nothing */}
        <button onClick={() => setLanguage(language)}>Je préfère le Français</button>
      </section>
    );
};
```

Compliant code:
```js
import { useState } from "react";

function ShowLanguage() {
    return (
      <section>
        <h1>Your language is {language}!</h1>
        <button onClick={() => setLanguage(navigator.language)}>Detect language</button>
        <button onClick={() => setLanguage("fr-FR")}>Je préfère le Français</button>
      </section>
    );
};
```

victor-diez-sonarsource:
Fixes #3218


vilchik-elena (changes requested):
When requesting review check following:
* go through files in github PR to make sure nothing wrong is committed (like ruling projects change)
* make sure coverage is close to 100%
* make sure there are 0 issues on Next

eslint-bridge/src/rules/no-useless-react-setstate.ts
```diff
@@ -0,0 +1,113 @@
...
+};
+
+function isReactCall(context: Rule.RuleContext, callExpr: estree.CallExpression): boolean {
+  let usesReactState = false;
+
+  if (callExpr.callee.type === 'Identifier') {
+    const module = getModuleNameOfImportedIdentifier(context, callExpr.callee);
+    usesReactState = module?.value === 'react';
+  } else if (callExpr.callee.type === 'MemberExpression') {
+    const module = getModuleNameOfIdentifier(context, callExpr.callee.object as estree.Identifier);
```
vilchik-elena, line 109: it's not guaranteed that object has identifier type

eslint-bridge/src/rules/no-useless-react-setstate.ts
```diff
@@ -0,0 +1,113 @@
...
+    };
+  },
+};
+
+function isReactCall(context: Rule.RuleContext, callExpr: estree.CallExpression): boolean {
+  let usesReactState = false;
+
+  if (callExpr.callee.type === 'Identifier') {
+    const module = getModuleNameOfImportedIdentifier(context, callExpr.callee);
+    usesReactState = module?.value === 'react';
```
vilchik-elena, line 107: I would directly return here

eslint-bridge/src/rules/no-useless-react-setstate.ts
```diff
@@ -0,0 +1,113 @@
...
+    const callSelector = [
+      'CallExpression[callee.type="Identifier"]',
+      '[arguments.length=1]',
+      '[arguments.0.type="Identifier"]',
+    ].join('');
+
+    return {
+      [declaratorSelector](node: estree.VariableDeclarator) {
+        if (isReactCall(context, node.init as estree.CallExpression)) {
+          const ids = node.id as estree.ArrayPattern;
```
vilchik-elena, line 66: you only use `elements` from it

eslint-bridge/src/rules/no-useless-react-setstate.ts
```diff
@@ -0,0 +1,113 @@
...
+          const setter = (ids.elements[1] as estree.Identifier).name;
+          referencesBySetterName[setter] = {
+            setter: null,
+            value: null,
+          };
+          const scope = context.getScope();
+          scope.references.forEach(ref => {
```
vilchik-elena, line 73: why didn't you use `getVariableFromName` here as well?

eslint-bridge/src/rules/no-useless-react-setstate.ts
```diff
@@ -0,0 +1,113 @@
...
+        }
+      },
+      [callSelector](node: estree.CallExpression) {
+        const setter = getVariableFromName(context, (node.callee as estree.Identifier).name);
+        const value = getVariableFromName(context, (node.arguments[0] as estree.Identifier).name);
+        const key = setter?.name as string;
+        if (
+          referencesBySetterName.hasOwnProperty(key) &&
+          referencesBySetterName[key].setter === setter &&
+          referencesBySetterName[key].value === value
```
vilchik-elena, line 90: what if value variable is not resolved and both of them will be `null`? Strictly speaking one will be `null` and another `undefined` so they will never be equal but still a dirty state to improve.

eslint-bridge/src/rules/no-useless-react-setstate.ts
```diff
@@ -0,0 +1,113 @@
...
+export const rule: Rule.RuleModule = {
+  meta: {
+    messages: {
+      uselessSetState: 'Change the argument of this setter to not use its matching state variable',
+    },
+  },
+  create(context: Rule.RuleContext) {
+    const referencesBySetterName: { [key: string]: Reference } = {};
+
+    const declaratorSelector = [
```
vilchik-elena, line 44: I would move it to the very top, no point in computing it for every file

vilchik-elena, line 44: -> `declarationSelector`

vilchik-elena (changes requested):

eslint-bridge/src/rules/no-useless-react-setstate.ts
victor-diez-sonarsource: :man_facepalming:

vilchik-elena (approved):

PR #3229, opened 2022-07-14
Improve S6439 (sonar-jsx-no-leaked-render): report on a string only for react-native

### LINKED ISSUE #3228: Improve S6439 (`sonar-jsx-no-leaked-render`):  report on a string only for react-native
We should update description (also making sure examples of code are covered by the implementation).
We should also report on the condition only

victor-diez-sonarsource:
Fixes #3228


eslint-bridge/src/rules/sonar-jsx-no-leaked-render.ts
```diff
@@ -41,69 +50,72 @@ export const rule: Rule.RuleModule = {
...
+          fixes.push(fixer.replaceText(node, `!!(${sourceCode.getText(node)})`));
+          return fixes;
+        },
+      },
+    ],
+  });
 }
 
 function isStringOrNumber(node: estree.Node, context: Rule.RuleContext) {
   const type = getTypeFromTreeNode(node, context.parserServices);
```
vilchik-elena, line 101: we are computing this type twice, here and on line 106

eslint-bridge/src/rules/sonar-jsx-no-leaked-render.ts
```diff
@@ -41,69 +50,72 @@ export const rule: Rule.RuleModule = {
...
-    previousToken.value === '(' &&
-    previousToken.range[1] <= node.range[0] &&
-    nextToken.value === ')' &&
-    nextToken.range[0] >= node.range[1]
-  );
+          const fixes = [];
+          if (
+            !!previousToken &&
+            !!nextToken &&
+            typeof node.range !== 'undefined' &&
```
vilchik-elena, line 83: can't we just check the value? why `typeof`?

its/sources/file-for-rules/S6439.js
```diff
@@ -0,0 +1,27 @@
+const Component1 = collection => {
```
vilchik-elena, line 1: add comments to show where you expect issues here

vilchik-elena (changes requested):
What about rule description?

eslint-bridge/src/rules/sonar-jsx-no-leaked-render.ts
victor-diez-sonarsource: what do you mean? I would say typeof is necessary here to check before comparing the values

saberduck: why not checking just `node.range !== 'undefined' , what's the reason for using `typeof`? you can be sure that 'range' property exists

victor-diez-sonarsource: oh, you mean:

node.range !== undefined

instead of:

typeof node.range !== 'undefined'

I can change that, in javascript I would say typeof is the preferred solution, but the second is more ts-like

vilchik-elena (approved):

Issue #3191, opened 2022-06-27
Analyze JS inside YAML for AWS Lambdas


ilia-kebets-sonarsource:
- [ ] Plugin
  - [ ] new YAML sensor
    - [x] filter YAML input files
    - [ ] generate TSConfig (to include type-dependent rules)
    - [x] analyze YAML
      - [x] analyze file
      - [x] save issues
      - [x] handle errors
- [x] Bridge
  - [x] new YAML endpoint
  - [x] YAML parser
    - [x] test Node.js 12 version (requires refactor because different API in yaml v1.x, possibly not doable)
    - [x] test Node.js 14 version
    - [x] Fork `yaml` repository and transpile it into Node12 compatible
  - [x] parse YAML file
    - [x] YAML syntax parsing
      - [x] single document
      - [x] multiple document
    - [x] parsing error
  - [x] extract AWS lambdas
    - [x] AWS::Lambda::Function
    - [x] AWS::Serverless:Function
  - [x] parse AWS lambdas
    - [x] JS snippet parsing
    - [x] parsing error
  - [x] normalize AWS lambda ast
    - [x] tree
      - [x] node
      - [x] token
      - [x] comment
    - [x] normalization
      - [x] loc
      - [x] offset
      - [x] range
      - [x] helpers
      - [x] parsing error (patch location)
      - [x] explain our patching of location in the code
      - [x] refactor YAML parsing code in separate file
      - [x] don't analyze YAML file content outside of JS snippets (leads to irrelevant issues and duplicates in case of multiple snippets per file)
      - [x] write test for header issue with multiple snippets
      - [x] write test for trailing comma issue with multiple snippets
    - [x] format (https://yaml-multiline.info/) - we handle only single item in items array
      - [x] block folded ">"
      - [x] block literal "|"
      - [x] flow with plain
      - [ ] flow with single-quotes
      - [ ] flow with double-quotes
      - [x] fix 1-off offset
    - [x] checks
      - [x] primary location
      - [x] secondary location
      - [x] quickfixes
      - [x] regex rules
      - [x] parent node location reached by visit()
  - [x] analyze AWS lambdas
    - [x] ~disable quickfixes (if possible)~
    - [x] build source codes
    - [x] lint source codes
    - [x] aggregate issues
    - [x] aggregate parsing errors
- [x] ITs
  - [x] plugin
    - [x] IaC coexistence
      - [x] issues
      - [x] metrics
[...]

PR #3237, opened 2022-07-18
implement support for single- and double-quote flow string format

ilia-kebets-sonarsource:
Extra features for #3191


ilia-kebets-sonarsource:
closing because there are conflicts with the branch `mmf-2828`

PR #3261, opened 2022-07-21
Add new eslint-bridge structure

yassin-kammoun-sonarsource:


eslint-bridge/src/linting/eslint/index.ts
```diff
@@ -0,0 +1,83 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import { debug } from '../../helpers';
+import { LinterWrapper } from './linter';
+import { RuleConfig } from './linter/config';
+import { CustomRule } from './linter/custom-rules/custom-rule';
```
saberduck, line 24: looks like imports could be simplified

```suggestion
import { LinterWrapper, RuleConfig } from './linter';
import { CustomRule } from './linter/custom-rules';
```

eslint-bridge/src/linting/eslint/index.ts
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Loads rule bundles
+ *
+ * A rule bundle is a set of external custom rules (like the taint analysis rule)
+ * including rule keys and rule definitions that cannot be provided to the linter
+ * wrapper using the same feeding channel as rules from the active quality profile.
+ *
+ * @param bundles the path of rule bundles to load
+ */
+export function loadBundles(bundles: string[]) {
```
saberduck, line 76: I don't like that this function needs to be exported, it feels strange. I think we could refactor this logic by changing `bundles` to not be provided as parameter to `start` in `bin/server` but to put them in `context` instead. That way we can pull them out of the context just when we invoke `initializeLinter` , wdyt?

eslint-bridge/src/linting/eslint/linter/custom-rules/symbol-highlighting.ts
```diff
@@ -0,0 +1,36 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import { CustomRule } from './custom-rule';
+import { rule as symbolHighlightingRule } from '../visitors/symbol-highlighting';
+
+/**
+ * The ESLint rule key of the internal _symbol-highlighting_ custom rule
+ */
+export const ruleId = 'internal-symbol-highlighting';
```
saberduck, line 27: same as above, instead of exporting `ruleId` as separate constant, we can use the `ruleId` property of `rule` object.

eslint-bridge/src/linting/eslint/linter/issues/normalize.ts
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Normalizes an issue location
+ *
+ * SonarQube uses 0-based column indexing when it comes to issue locations
+ * while ESLint uses 1-based column indexing for message locations.
+ *
+ * @param issue the issue to normalize
+ * @returns the normalized issue
+ */
+export function normalizeLocation(issue: Issue): Issue {
```
saberduck, line 32: this is only used in `transform` , I think we can just move it there as local function and it doesn't need to exported

eslint-bridge/src/linting/eslint/linter/issues/index.ts
```diff
@@ -0,0 +1,27 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+export * from './decode';
```
saberduck, line 21: I don't think `decode` should be exported outside of this module

eslint-bridge/src/parsing/jsts/builders/build-ts.ts
```diff
@@ -0,0 +1,45 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import { JsTsAnalysisInput } from '../../../services/analysis/analyzers/js';
+import { buildParserOptions, parsers, parseForESLint } from '../parsers/';
```
saberduck, line 22:
```suggestion
import { buildParserOptions, parsers, parseForESLint } from '../parsers';
```

eslint-bridge/src/parsing/jsts/builders/build-vue.ts
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import { SourceCode } from 'eslint';
+import { debug } from '../../../helpers';
+import { JsTsAnalysisInput } from '../../../services/analysis/analyzers/js';
+import { buildParserOptions, parsers, parseForESLint } from '../parsers/';
```
saberduck, line 24:
```suggestion
import { buildParserOptions, parsers, parseForESLint } from '../parsers';
```

eslint-bridge/src/parsing/jsts/build.ts
```diff
@@ -0,0 +1,57 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import { getContext } from '../../helpers';
+import { JsTsAnalysisInput } from '../../services/analysis/analyzers/js';
```
saberduck, line 22: it seems strange that `parsing` layer needs to understand `services` layer so deeply , IMO imports between layers should be limited to top level exported entities, otherwise we don't have the right level of abstraction

eslint-bridge/src/parsing/jsts/build.ts
```diff
@@ -0,0 +1,57 @@
...
+ * Builds an ESLint SourceCode for JavaScript / TypeScript
+ *
+ * This functions routes the parsing of the input based on the input language,
+ * the file extension, and some contextual information.
+ *
+ * @param input the JavaScript / TypeScript analysis input
+ * @param language the language of the input
+ * @returns the parsed source code
+ */
+export function buildSourceCode(input: JsTsAnalysisInput, language: Language) {
```
saberduck, line 38: imo this function should be in the `builders` dir , so you don't need to import all kinds of builders and expose internal structure of the `builders` "package"

saberduck (commented):
I left some general comments, I will give it another go tomorrow

eslint-bridge/src/linting/eslint/index.ts
```diff
@@ -0,0 +1,83 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import { debug } from '../../helpers';
```
saberduck, line 21: let's use tsc `paths` tsconfig option so we don't need to rely on relative paths when importing modules

ilia-kebets-sonarsource:
Can you replace `routing/index.ts` with, thus removing `routing/handlers/index.ts` and moving all remaining `routing/handlers/*` files to `routing/`? I think the extra layer `handlers/` is heavy and doesn't bring anything. Also, "handler" is a bad name, like "manager".:

```javascript
import express from 'express';

import onAnalyzeCss from './on-analyze-css';
import onAnalyzeJs from './on-analyze-js';
import onAnalyzeTs from './on-analyze-ts';
import onAnalyzeWithProgram from './on-analyze-with-program';
import onAnalyzeYaml from './on-analyze-yaml';
import onCreateProgram from './on-create-program';
import onDeleteProgram from './on-delete-program';
import onInitLinter from './on-init-linter';
import onNewTSConfig from './on-new-tsconfig';
import onStatus from './on-status';
import onTSConfigFiles from './on-tsconfig-files';

const router = express.Router();

router.post('/analyze-css', onAnalyzeCss);
router.post('/analyze-js', onAnalyzeJs);
router.post('/analyze-ts', onAnalyzeTs);
router.post('/analyze-with-program', onAnalyzeWithProgram);
router.post('/analyze-yaml', onAnalyzeYaml);
router.post('/create-program', onCreateProgram);
router.post('/delete-program', onDeleteProgram);
router.post('/init-linter', onInitLinter);
router.post('/new-tsconfig', onNewTSConfig);
router.get('/status', onStatus);
router.post('/tsconfig-files', onTSConfigFiles);

export default router;
```

ilia-kebets-sonarsource:
This might be part of another refactor, but since we're using express, we could rely on its - IMO pretty good - error handling: https://expressjs.com/en/guide/error-handling.html

ilia-kebets-sonarsource:
what does DTO in `services/analysis/analyzers/**/dto.ts` stand for?

ilia-kebets-sonarsource:
Am I the only one who can't load any file beyond linting to add inline comments?

Screenshot 2022-08-03 at 11 34 18

ilia-kebets-sonarsource:
Renaming comment from other PR: https://github.com/SonarSource/SonarJS/pull/3298/files/63a54bd1af901278c5da6fe2829363c54624000e#r936596785

eslint-bridge/src/linting/eslint/linter/custom-rules/cognitive-complexity.ts
```diff
@@ -0,0 +1,38 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import { rules as sonarjsESLintRules } from 'eslint-plugin-sonarjs';
+import { CustomRule } from './custom-rule';
+
+/**
+ * The ESLint rule key of the internal _cognitive complexity_ custom rule
+ */
+export const ruleId = 'internal-cognitive-complexity';
```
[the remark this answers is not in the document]
ilia-kebets-sonarsource: so you can provide `rule` as a `export default` here?

eslint-bridge/src/linting/eslint/linter/issues/issue.ts
```diff
@@ -0,0 +1,50 @@
...
+ * @param endLine the issue ending line
+ * @param endColumn the issue ending column
+ * @param message the issue message
+ * @param cost the cost to fix the issue
+ * @param secondaryLocations the issue secondary locations
+ * @param quickFixes the issue quick fixes
+ */
+export interface Issue {
+  ruleId: string;
+  line: number;
```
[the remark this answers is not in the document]
ilia-kebets-sonarsource: why not

eslint-bridge/src/linting/eslint/linter/issues/index.ts
```diff
@@ -0,0 +1,27 @@
...
+export * from './extract';
+export * from './issue';
+export * from './location';
```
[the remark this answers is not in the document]
ilia-kebets-sonarsource: there's another Location object in `linting/eslint/linter/visitors/metrics/helpers/`

eslint-bridge/src/linting/eslint/linter/issues/index.ts
ilia-kebets-sonarsource: I agree that in general, stuff that is only used inside a module should not be exported

ilia-kebets-sonarsource:
Can we rename `src/helpers/context.ts` and its associated methods to `config.ts` as it is indeed the eslint-bridge's configuration?
Also, we could use something like [nconf](https://www.npmjs.com/package/nconf) for it.

francoismora:
There's a minor inconsistency in the names `analyzers` and `analysis`.

ilia-kebets-sonarsource:
Can we rename: 
-`tests/testing/` to `tests/tools/`
- `tests/testing/framework/` to `tests/tools/comment-based/`

eslint-bridge/src/linting/eslint/index.ts
yassin-kammoun-sonarsource: What about relying on relative paths to the current folder, e.g. `'./foo'`, or `'./foo/bar'`? Should we use the absolute path or are we fine with this?

ilia-kebets-sonarsource: For current folder, I'm fine with relative paths, only if you go deeper (starting with `./`, not `../`). especially if you're deeply nested.

eslint-bridge/src/linting/eslint/index.ts
yassin-kammoun-sonarsource: I like it 👍

yassin-kammoun-sonarsource:
> Can we rename:
> * `tests/testing/` to `tests/tools/`

👍 

> * `tests/testing/framework/` to `tests/tools/comment-based/`

👍 I realize that's even better to move it to `tests/tools/testers/comment-based`

> * `tests/testing/sonar-runtime/` to `tests/tools/secondary-locations/`

I would rather completely rename the whole `sonar-runtime` concept everywhere into something more meaningful, and I would prefer to do that in another iteration. As I said in the review instructions, it's better to focus on the critical aspects of the bridge as a first step.

> * `tests/testing/sonar-runtime/rule/` to `tests/tools/secondary-locations/fixtures/`
 
Same as above

yassin-kammoun-sonarsource:
> what does DTO in `services/analysis/analyzers/**/dto.ts` stand for?
> what's the difference with `analysis.ts` in the same folder? It seems to me like just another interface/type holder file.

`dto` stands for [Data Transfer Object](https://en.wikipedia.org/wiki/Data_transfer_object), but I think I will just merge it with `analysis.ts` and use the latter instead. There is no need to split them.

eslint-bridge/src/services/analysis/analyzers/js/dto.ts
```diff
@@ -0,0 +1,91 @@
+/*
```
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: https://github.com/SonarSource/SonarJS/pull/3261#issuecomment-1206311277

yassin-kammoun-sonarsource:
> Can we rename `src/helpers/context.ts` and its associated methods to `config.ts` as it is indeed the eslint-bridge's configuration? Also, we could use something like [nconf](https://www.npmjs.com/package/nconf) for it. Also, can we move it to the root, instead of hiding it in `helpers/`
> 
> `linting/eslint/index.ts` also holds some config data in its `customRules` variable.
> 
> This might be a bigger change, I could tackle it in my external PR.

I am not entirely convinced we would gain anything by adding another dependency for such a small thing, but feel free to give it a try in your external pull request 👍

eslint-bridge/src/services/analysis/errors.ts
```diff
@@ -0,0 +1,79 @@
...
+ * The `GeneralError` value denotes a runtime error which is either
+ * unpredicatble or occurs rarely to deserve its own category.
+ */
+export enum AnalysisErrorCode {
+  Parsing = 'PARSING',
+  MissingTypeScript = 'MISSING_TYPESCRIPT',
+  UnsupportedTypeScript = 'UNSUPPORTED_TYPESCRIPT',
+  FailingTypeScript = 'FAILING_TYPESCRIPT',
+  GeneralError = 'GENERAL_ERROR',
+}
```
yassin-kammoun-sonarsource, line 49: @saberduck Do you think we could remove the `MissingTypeScript` and `UnsupportedTypeScript` values? I wonder if we still need them since we have been embedding TypeScript dependency for a while now.

eslint-bridge/src/linting/eslint/linter/issues/issue.ts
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: I would avoid mixing the two because they don't exactly share the same interface; some fields are optional in one and mandatory in the other. If we start mixing them, we will lose the information as to which one should be mandatory in which context.

eslint-bridge/src/services/analysis/errors.ts
saberduck: yes, we can remove them if they are not used

saberduck (approved):
thanks for taking the time to do this!

PR #3267, opened 2022-07-25
Execute ruling in parallel

saberduck:


ilia-kebets-sonarsource:
@saberduck: Looks good. You mentioned that running this locally brings your machine to a halt. Should we include a concurrency parameter that we could feed from the command line?

ilia-kebets-sonarsource (approved):
LGTM, I ran it locally and can see that it is using less cores than without `junit.jupiter.execution.parallel.config.dynamic.factor=0.5`.

PR #3271, opened 2022-07-26
Add decorator to ESLint core rule no-extra-semi

### LINKED ISSUE #3250: Add exception to S1116 for ASI
S1116 - extra semi-colons should be removed

- Should ignore lines starting with `;[]` and `;()`, since these are deliberate and common in codebases that omit semi-colons and rely on automatic semi-colon insertion

francoismora:
Fixes #3250


francoismora:
Recreating the branch as the diff contains unrelated changes.

PR #3276, opened 2022-07-27
Add rule S6252: Disabling versioning of S3 buckets is security-sensitive

### LINKED ISSUE #3207: Add rule S6252: Disabling versioning of S3 buckets is security-sensitive
## WHAT
The rule detects when an S3 bucket is not configured with versioning using the JavaScript AWS CDK  library. It does not cover custom policy statement definitions. This has been previously implemented for Python, see [SONARPY-1013](https://jira.sonarsource.com/browse/SONARPY-1013).

Specification ticket: [APPSEC-42](https://jira.sonarsource.com/browse/APPSEC-42)
[MMF-2812](https://jira.sonarsource.com/browse/MMF-2812)

## HOW

Detection logic
- Class/Construct: `aws-cdk-lib.aws_s3.Bucket` & argument: `props` 3rd positional argument
- Property: `versioned`
  - Safe values: `true`
  - Sensitive value: `false`
  - Default sensitive value: when not set `versioned` value defaults to false

The rule raises on the _Bucket_ construct class when any of the above conditions are met.
```javascript
new s3.Bucket(this, 'id', {
    bucketName: 'bucket'
}); // Sensitive: Omitting "versioned" disables S3 bucket versioning. Make sure it is safe here

------

new s3.Bucket(this, 'id', {
    bucketName: 'bucket',
    versioned: false // Sensitive: Make sure using unversioned S3 bucket is safe here
});

------
const versioned = false;

new s3.Bucket(this, 'id', {
      bucketName: 'bucket',
      versioned: versioned // Sensitive: Make sure using unversioned S3 bucket is safe here
});

```

## Code examples

https://github.com/SonarSource/security-expected-issues/pull/724

## Message and highlighting

See https://github.com/SonarSource/rspec/pull/1093

ilia-kebets-sonarsource:
Fixes #3207


ilia-kebets-sonarsource:
- python test cases: python-checks/src/test/resources/checks/cdk/s3BucketVersioning.py
- python rule implementation: python-checks/src/main/java/org/sonar/python/checks/cdk/S3BucketVersioningCheck.java

eslint-bridge/tests/rules/comment-based/aws-s3-bucket-versioning.js
```diff
@@ -0,0 +1,58 @@
+const cdk = require('aws-cdk-lib');
+const s3 = require('aws-cdk-lib/aws-s3');
+
+const noOptions = new s3.Bucket(this, 'id'); // Noncompliant {{Omitting the "versioned" argument disables S3 bucket versioning. Make sure it is safe here.}}
+//                    ^^^^^^^^^
+const otherImport = new cdk.aws_s3.Bucket(this, 'id') // Noncompliant {{Omitting the "versioned" argument disables S3 bucket versioning. Make sure it is safe here.}}
```
yassin-kammoun-sonarsource, line 6: I think it's okay if we assert the issue message only once. You could only keep the `// Noncompliant` part only for this type of issue.

eslint-bridge/tests/rules/comment-based/aws-s3-bucket-versioning.js
```diff
@@ -0,0 +1,58 @@
...
+//                      ^^^^^^^^^^^^^^^^^
+const noOtionsParam = new s3.Bucket(this, 'id', { // Noncompliant {{Omitting the "versioned" argument disables S3 bucket versioning. Make sure it is safe here.}}
```
yassin-kammoun-sonarsource, line 8: Typo `noOtionsParam `

yassin-kammoun-sonarsource (changes requested):

eslint-bridge/src/utils/utils-location.ts
```diff
@@ -27,7 +27,7 @@ export type LocationHolder = AST.Token | TSESTree.Node | estree.Node | { loc: AS
 export function toEncodedMessage(
   message: string,
   secondaryLocationsHolder: Array<LocationHolder> = [],
-  secondaryMessages?: (string | undefined)[],
+  secondaryMessages: (string | undefined)[] = [],
```
yassin-kammoun-sonarsource, line 30: Was there a need for that change? Is the semantics of the following line preserved?

https://github.com/SonarSource/SonarJS/blob/20a9f724a82ff2c1c87c0dd21dfcf40e1160ebdd/eslint-bridge/src/utils/utils-location.ts#L39

I haven't tested it so I am just asking to be sure.

yassin-kammoun-sonarsource (approved):
Looks good to me but we are missing a test file in the integration tests to validate the rule. 

Also, I am surprised that we are not raising issues on the following file:

https://github.com/SonarSource/SonarJS/blob/d65837cab786e6fbe19f8d4ac3fa9dbccdb19fc7/its/sources/file-for-rules/S6245.js#L1-L17

Maybe you need to pull again the main branch's recent changes.

eslint-bridge/src/utils/utils-location.ts
ilia-kebets-sonarsource: The goal was to avoid having to specify default params here:

https://github.com/SonarSource/SonarJS/pull/3276/files#diff-d1eefdd61ea1ad1ed8f3f7e0660b7314741d37aa772654ef915735fac335d3b7R46

ilia-kebets-sonarsource: It does not break as long as we have the same number of elements in `secondaryLocationsHolder` and `secondaryMessages`. However, as they are consumed together, shouldn't they be in the same data struct like?:

```javascript
secondaries = {
 locationHolder: ...,
 message: ....
}
```

yassin-kammoun-sonarsource: It's okay like this I guess.

PR #3279, opened 2022-07-27
Add exception to S3358 and JSX

### LINKED ISSUE #3244: Add exception to S3358 and JSX
S3358 - nested ternary

- Should ignore single ternary expressions assigned to JSX attributes when there is an enclosing ternary expression that contains two JSX expressions as the branches of the condition

francoismora:
Fixes #3244


saberduck (approved):
nice, many issues gone from ruling

PR #3281, opened 2022-07-28
 Fix FP in S1172 Unused function parameters should be removed

### LINKED ISSUE #3243: Fix FP in S1172  Unused function parameters should be removed
S1172  Unused function parameters should be removed

- Should ignore preceding arguments when one argument has been used. For example, given a signature (a, b, c), when c is being used, a and b must be ignored.

francoismora:
…unused argument appears before a used argument in the function declaration.

Fixes #3243


francoismora:
This PR removes a *lot* of noise mostly found in callback functions where the list of parameters is imposed by an API (iterator, XHR callbacks, etc.). Even though the noise could be avoided by prefixing unused parameters with an underscore, this is not possible sometimes as it may conflict with another naming convention (AngularJS uses $ prefix). However, we lose some true positives when the function is not a callback.

PR #3283, opened 2022-07-28
Improve S3 Bucket rule template: make rule metadata parameterizable

yassin-kammoun-sonarsource:
A rule's metadata needs to be parameterizable. Otherwise, if we were to force the support of secondary locations, rules that don't it would still have to use the secondary location API when reporting issues, that is, calling `toEncodedMessage`.


ilia-kebets-sonarsource (approved):
LGTM

PR #3286, opened 2022-07-31
Fix FP S1172: Ignore parameter properties with default value

aukevanleeuwen:
(TypeScript) ParameterProperties that had an assignment (default value),
were giving a false positive for the 'no-unused-function-argument' rule.

More information here: https://community.sonarsource.com/t/false-positive-s1172-for-constructor-parameter/64575/14

Fixes: #3293


aukevanleeuwen:
👋  hello, I'm having a bunch of false positives (IMHO) on TypeScript code. I figured I could fix the rule myself and I did in this PR. The build is probably failing on something that seems CI setup related:

```
npm config set registry "${ARTIFACTORY_URL}/api/npm/npm"
/tmp/scripts9566c74d10037c4d7bbb0407d1e2c649.sh: line 6: ARTIFACTORY_URL: unbound variable
```

Is this the right way to go about trying to get this fixed? I had a look at all of the issues that were reported between `expected` and `actual` by the LITS plugin, and moved the new state over to the `expected` directory.

francoismora:
Thanks, @aukevanleeuwen, for your contribution. Your change is in commit 1c65f266b444c15bf1b04511eaa225a25abfd498 merged after PR #3382 as required by the CI.

PR #3292, opened 2022-08-02
Add exceptions to the ESLint core rule no-empty-function (S1186).

### LINKED ISSUE #3241: Fix FP in S1186 - Unexpected empty function
Rule S1186 - Unexpected empty function

- Should ignore arrow functions when assigned as default parameter values
- Should ignore arrow functions when assigned as an attribute of a React defaultProps object
- Should ignore arrow functions when assigned as JSX attributes
- Should ignore arrow functions when assigned to variables, attributes, or properties named with the pattern onX, where X can be any string. For example, onLoaded, onScroll, onCustomEvent, etc.

We should consider ignoring arrow functions altogether.

francoismora:
Fixes #3241


eslint-bridge/src/rules/decorators/no-empty-function-decorator.ts
```diff
@@ -19,24 +19,83 @@
...
+  }
+
+  // Matches: function foo() {}
+  // But not: function onClose() {}
+  function isFunctionDeclaration(): boolean {
+    return functionNode.type === 'FunctionDeclaration' && !isCallbackIdentifier(functionNode.id);
+  }
+
+  // Matches: Class A { foo() {} }
+  // But not: Class A { onClose() {} }
```
saberduck, line 68: `class` should be lowercase
```suggestion
  // Matches: class A { foo() {} }
  // But not: class A { onClose() {} }
```

eslint-bridge/src/rules/decorators/no-empty-function-decorator.ts
```diff
@@ -19,24 +19,83 @@
...
+  }
+
+  return isFunctionDeclaration() || isMethodDefinition() || isVariableDeclarator();
+}
+
+function reportWithQuickFix(
+  context: Rule.RuleContext,
+  reportDescriptor: Rule.ReportDescriptor,
+  func: FunctionNodeType,
+): void {
```
saberduck, line 96: I would skip writing return type for functions which are not part of the API 
```suggestion
) {
```

saberduck (approved):
some minor suggestions

PR #3296, opened 2022-08-03
Ignore React components defined with lambda expressions (S100)

### LINKED ISSUE #3215: FP S100 (`function-name`): Add exception to S100 for React Functional Components implemented with arrow functions
This React component implemented with an arrow function will be reported:

```js
const Welcome = () => {
  const greeting = 'Hello, World!';

  // ...

  return (
    <div className="Welcome">
      <p>{greeting}</p>
    </div>
  );
}
```

Reported in community: https://community.sonarsource.com/t/typescript-s100-is-reporting-react-functional-components-arrow-functions/67522

francoismora:
Fixes #3215


eslint-bridge/src/rules/function-name.ts
```diff
@@ -25,8 +25,8 @@ import { TSESTree } from '@typescript-eslint/experimental-utils';
 import { last } from '../utils';
 
 interface FunctionKnowledge {
-  startsWithCapital: boolean;
   returnsJSX: boolean;
+  id: TSESTree.Node | null;
```
saberduck, line 29: it seems to me that we do nothing if id is `null`,  we might as well forbid putting `null` in the type

eslint-bridge/src/rules/function-name.ts
```diff
@@ -111,10 +99,29 @@ function isFunctionExpression(node: TSESTree.Node | null) {
...
+}
+
+function nameStartsWithCapital(node: TSESTree.Node | undefined | null) {
+  return node != null && node.type === 'Identifier' && node.name[0] === node.name[0].toUpperCase();
 }
 
-function nameStartsWithCapital(node: TSESTree.FunctionDeclaration) {
-  return node.id !== null && node.id.name[0] === node.id.name[0].toUpperCase();
+function createKnowledge(node: TSESTree.Node): FunctionKnowledge | null {
+  if (node.type === 'FunctionDeclaration') {
```
saberduck, line 111: you can check for `node.id == null` here and remove `| null` whenever `id` is used

saberduck (changes requested):

eslint-bridge/src/rules/function-name.ts
francoismora: Indeed you are right; the case where id is null can be propagated to the case where the knowledge base is null. Nice catch. Thanks for pointing this out.

eslint-bridge/src/rules/function-name.ts
francoismora: Done

saberduck (approved):

PR #3298, opened 2022-08-03
Update comment-based testing framework: fix secondary location encoding

yassin-kammoun-sonarsource:
When a rule enables the support of secondary locations with the internal `sonar-runtime` parameter, it forces the encoding of issue messages even if an issue does not always include secondary locations. On the other hand, the comment-based testing framework only considers local comments in test files when extracting issue expectations. Therefore, if an issue expectation does not include secondary locations, the extraction process considers it a regular issue and bypasses the encoding of the issue message. This prevents the comment-based testing framework from correctly testing issues without secondary locations reported by a rule enabling the support of secondary locations.

### DIFF
--- a/eslint-bridge/src/linter.ts
+++ b/eslint-bridge/src/linter.ts
@@ -319,7 +319,7 @@ export function getRuleConfig(ruleModule: ESLintRule.RuleModule | undefined, inp
   return options;
 }
 
-function hasSonarRuntimeOption(
+export function hasSonarRuntimeOption(
   ruleModule: ESLintRule.RuleModule | undefined,
   ruleId: string,
 ): boolean {
--- a/eslint-bridge/tests/testing-framework/assertions.ts
+++ b/eslint-bridge/tests/testing-framework/assertions.ts
@@ -41,22 +41,26 @@ function convertToTestCaseErrors(
   issue: LineIssues,
   usesSecondaryLocations: boolean,
 ): RuleTester.TestCaseError[] {
-  const toMessage = usesSecondaryLocations ? toEncodedMessage : message => message;
+  const encodeMessageIfNeeded = usesSecondaryLocations ? toEncodedMessage : message => message;
   const line = issue.line;
   const primary = issue.primaryLocation;
   const messages = [...issue.messages.values()];
   if (primary === null) {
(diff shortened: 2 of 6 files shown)


eslint-bridge/src/linter.ts
```diff
@@ -319,7 +319,7 @@ export function getRuleConfig(ruleModule: ESLintRule.RuleModule | undefined, inp
   return options;
 }
 
-function hasSonarRuntimeOption(
+export function hasSonarRuntimeOption(
```
ilia-kebets-sonarsource, line 322: can we rename this fuzzy name to `supportsSecondaryLocations`? This naming comment applies to the rest of the codebase.

yassin-kammoun-sonarsource: Probably something that should be done in the restructuring PR :)

eslint-bridge/tests/testing-framework/assertions.ts
```diff
@@ -24,31 +24,39 @@ import { FileIssues, LineIssues } from './issues';
...
+  expectedIssues.forEach(issue => errors.push(...convertToTestCaseErrors(issue, usesSonarRuntime)));
   return errors;
 }
 
-function convertToTestCaseErrors(issue: LineIssues): RuleTester.TestCaseError[] {
+function convertToTestCaseErrors(
+  issue: LineIssues,
+  usesSecondaryLocations: boolean,
+): RuleTester.TestCaseError[] {
+  const toMessage = usesSecondaryLocations ? toEncodedMessage : message => message;
```
ilia-kebets-sonarsource, line 44: rename: `toMessage` to `parseMessageIfNeeded`

ilia-kebets-sonarsource (approved):
I made 2 naming comments, otherwise LGTM

eslint-bridge/src/linter.ts
ilia-kebets-sonarsource: I'll add the note then.

eslint-bridge/tests/testing-framework/assertions.ts
yassin-kammoun-sonarsource: Not sure that's the right name as parsing has nothing to do here. I will go for `encodeMessageIfNeeded`.

PR #3306, opened 2022-08-08
Use eslint-plugin-sonarjs 0.15.0

saberduck:


yassin-kammoun-sonarsource (approved):
Already approving the PR, but we need to update the issue differencing in the ruling expectations.

PR #3316, opened 2022-08-11
Add check for eslint-bridge distribution filepath length

### LINKED ISSUE #3135: Eslint-bridge should not use too long paths
SonarLint is bundling eslint-bridge in Visual Studio. On Windows by default, there is a limit on the length of the path in the filesystem - 260 characters. Some of the dependencies of eslint-bridge are deeply nested, so extraction can fail.

This could be improved by using a bundler, such as ncc https://github.com/vercel/ncc

See also work by Andrea in https://github.com/SonarSource/SonarJS/issues/2520

ilia-kebets-sonarsource:
Fixes #3135

### DIFF
--- a/eslint-bridge/.gitignore
+++ b/eslint-bridge/.gitignore
@@ -2,4 +2,3 @@ lib/
 node_modules/
 coverage/
 test-report.xml
-bundle/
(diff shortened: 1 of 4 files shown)


ilia-kebets-sonarsource:
This fails the maven jobs as expected.

francoismora (commented):
LGTM

francoismora (approved):
LGTM

PR #3318, opened 2022-08-15
Add heartbeat keepalive request to eslint bridge

### LINKED ISSUE #3131: Eslint-bridge process should terminate gracefully when parent process dies
In SonarLint context it can happen that parent java process (which starts eslint-bridge Node.js process) dies unexpectetly. Currently eslint-bridge process keeps running as a zombie process. We should implement heartbeat mechanism and if client doesn't issue a request after certain timeout eslint-bridge process should terminate itself gracefully.

victor-diez-sonarsource:
Fixes #3131

### DIFF
--- a/eslint-bridge/src/helpers/timeout.ts
+++ b/eslint-bridge/src/helpers/timeout.ts
@@ -37,21 +37,30 @@ class Timeout {
 }
 
 function timeoutTimeoutMiddleware(server: http.Server, shutdownTimeout: number) {
+  let closed = false;
   const timeout = new Timeout(() => {
     server.close();
   }, shutdownTimeout);
   timeout.init();
   return {
     middleware(_request: express.Request, res: express.Response, next: express.NextFunction) {
-      timeout.cancel();
-      res.on('finish', function () {
-        timeout.init();
-      });
+      if (!closed) {
+        timeout.cancel();
+        res.on('finish', function () {
+          timeout.init();
+        });
+      }
       next();
     },
     heartBeatHandler(_req: express.Request, res: express.Response) {
-      timeout.init();
+      if (!closed) {
+        timeout.init();
+      }
       res.end();
+    },
+    close() {
+      closed = true;
+      timeout.cancel();
     },
   };
 }
--- a/eslint-bridge/src/helpers/timeout.ts
+++ b/eslint-bridge/src/helpers/timeout.ts
@@ -42,15 +42,12 @@ function timeoutTimeoutMiddleware(server: http.Server, shutdownTimeout: number) 
       server.close();
     }
   }, shutdownTimeout);
-  let requests = 0;
   timeout.init();
   return {
     middleware(_request: express.Request, res: express.Response, next: express.NextFunction) {
       timeout.cancel();
-      requests++;
       res.on('finish', function () {
-        requests--;
-        if (requests === 0) timeout.init();
+        timeout.init();
       });
       next();
     },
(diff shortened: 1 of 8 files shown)


eslint-bridge/src/helpers/timeout.ts
```diff
@@ -0,0 +1,60 @@
...
+  }
+  cancel() {
+    if (this.timeout) {
+      clearTimeout(this.timeout);
+      this.timeout = null;
+    }
+  }
+}
+
+function timeoutTimeoutMiddleware(server: http.Server, shutdownTimeout: number) {
```
saberduck, line 39: why the name has "timeout" twice? Could this be just `timeoutMiddleware`?

eslint-bridge/src/server.ts
```diff
@@ -38,21 +44,28 @@ const MAX_REQUEST_SIZE = '50mb';
...
  *
- * Altough a web server, the bridge is not exposed to the outside world
+ * Although a web server, the bridge is not exposed to the outside world
  * but rather exclusively communicate either with the JavaScript plugin
- * which embedds it or directly with SonarLint.
+ * which embeds it or directly with SonarLint.
  *
  * @param port the port to listen to
  * @param host the host to listen to
+ * @param shutdownTimeout timeout in ms to shut down the server since last heartbeat request
```
saberduck, line 53:
```suggestion
 * @param shutdownTimeout timeout in ms to shut down the server since the last request
```

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
```diff
@@ -105,6 +112,12 @@ public EslintBridgeServerImpl(NodeCommandBuilder nodeCommandBuilder, Bundle bund
     this.hostAddress = InetAddress.getLoopbackAddress().getHostAddress();
     this.deployLocation = tempFolder.newDir(DEPLOY_LOCATION).toPath();
     this.monitoring = monitoring;
+    this.heartbeatService = Executors.newSingleThreadScheduledExecutor();
+  }
+
+  void heartbeat() {
+    LOG.info("Pinging the server");
```
saberduck, line 119: this should not be on `INFO` level, that's too noisy, imo `DEBUG` is sufficient

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
```diff
@@ -135,6 +148,10 @@ void startServer(SensorContext context, List<Path> deployedBundles) throws IOExc
       throw new NodeCommandException("Failed to start server (" + timeoutSeconds + "s timeout)");
     } else {
       status = Status.STARTED;
+      if (heartbeatFuture == null) {
+        LOG.info("Starting heartbeat service");
+        heartbeatFuture = heartbeatService.scheduleAtFixedRate(this::heartbeat, 5, 5, TimeUnit.SECONDS);
```
saberduck, line 153: I think we should extract 5 second constant as a static field on the class level, so it's better documented

saberduck (changes requested):
I left some minor comments.

eslint-bridge/src/server.ts
saberduck: I would ommit "heartbeat" because we react on any request

PR #3320, opened 2022-08-16
Add specific code when the linter is not initialized.

### LINKED ISSUE #2627: eslint-bridge should return specific code when linter is not initialized
SonarLint in Visual Studio sometimes needs to recover from exceptional conditions - out of memory, etc.. They need to able to detect when calling `/init-linter` is necessary. Currently, if `/init-linter` is not called before analysis, we throw generic exception. We should change this to return specific error code which can be interpreted on SonarLint side.

francoismora:
Fixes #2627

### DIFF
--- a/eslint-bridge/src/linting/eslint/index.ts
+++ b/eslint-bridge/src/linting/eslint/index.ts
@@ -20,7 +20,6 @@
 
 import { debug, getContext } from 'helpers';
 import { CustomRule, LinterWrapper, RuleConfig } from './linter';
-import { AnalysisErrorCode, ServerError } from '../../services/analysis';
 
 export * from './linter';
 export * from './rules';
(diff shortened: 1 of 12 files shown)


francoismora:
This change aims to introduce a specific error code when the linter is not initialized. Previously the server returned:
```
➜  curl -X POST -H 'Content-Type: application/json' http://localhost:8080/analyze-yaml
{"parsingError":{"message":"Linter is undefined. Did you call /init-linter?","code":"GENERAL_ERROR"}}%
```
Now it returns:
```
➜  curl -X POST -H 'Content-Type: application/json' http://localhost:8080/analyze-yaml
{"parsingError":{"message":"Linter is undefined. Did you call /init-linter?","code":"LINTER_INITIALIZATION"}}%
```
The SLVS client is currently checking only the message value. With this change, they can check the code value.

saberduck (approved):

PR #3323, opened 2022-08-17
Fix S5860: FP when regex is accessed with the non-null assertion operator

### LINKED ISSUE #3236: Fix FP S5860 ('unused-named-groups'): Consider non-null assertions
S5860 is raising false positives when a matched group of a regular expression is accessed with the non-null assertion operator:

```typescript
const pattern = /(?<foo>\\w)/;
const matched = 'str'.matchAll(pattern);
if (matched) {
    matched.groups!.foo; // FP because of the '!' operator
}
```

victor-diez-sonarsource:
Fixes #3236


yassin-kammoun-sonarsource (approved):
LGTM!

PR #3344, opened 2022-08-25
Remove duplicate test

ilia-kebets-sonarsource:
There's literally the same test just above.


yassin-kammoun-sonarsource (approved):
Nice catch!

PR #3350, opened 2022-08-29
Fix 'RULE' metric type for performance monitoring

### LINKED ISSUE #3341: Fix 'RULE' metric type for performance monitoring
For some reason (refactoring in the analyzer or change in ESLint?) since recently (in the middle of 9.6 cycle?) 'RULE' metric type is not provided anymore.

vilchik-elena:
Fixes #3341

### DIFF
--- a/its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/TypeScriptRuleTest.java
+++ b/its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/TypeScriptRuleTest.java
@@ -27,6 +27,7 @@ import java.io.File;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
+import java.nio.file.Path;
 import java.nio.file.Paths;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
@@ -84,7 +85,7 @@ class TypeScriptRuleTest {
     orchestrator.getServer().associateProjectToQualityProfile(PROJECT_KEY, "js", "empty-profile");
     orchestrator.getServer().associateProjectToQualityProfile(PROJECT_KEY, "css", "empty-profile");
 
-    String perfMonitoringDir = "target/monitoring/" + PROJECT_KEY;
+    Path perfMonitoringDir = Path.of("target/monitoring/", PROJECT_KEY);
     SonarScanner build = getSonarScanner()
       .setProjectDir(PROJECT_DIR)
       .setProjectKey(PROJECT_KEY)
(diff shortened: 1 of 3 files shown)


vilchik-elena:
One condition is not covered but I think it's not worthy to make it covered

its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/TypeScriptRuleTest.java
```diff
@@ -92,10 +93,23 @@ void test() throws Exception {
       .setProperty("sonar.lits.dump.old", FileLocation.of("target/expected/ts/" + PROJECT_KEY).getFile().getAbsolutePath())
       .setProperty("sonar.lits.dump.new", FileLocation.of("target/actual/ts/" + PROJECT_KEY).getFile().getAbsolutePath())
       .setProperty("sonar.lits.differences", FileLocation.of("target/differences").getFile().getAbsolutePath())
+      .setProperty("sonar.javascript.monitoring", "true")
+      .setProperty("sonar.javascript.monitoring.path", FileLocation.of(perfMonitoringDir).getFile().getAbsolutePath())
```
saberduck, line 97: I prefer to use Java's API instead of orchestrator `FileLocation` , you can use `Path.of(...).toAbsolutePath` (I think the `FileLocation` usage was copy&pasted from other analyzers, and we should stop this bad habit :)

its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/TypeScriptRuleTest.java
```diff
@@ -92,10 +93,23 @@ void test() throws Exception {
...
       .setProperty("sonar.cpd.exclusions", "**/*");
 
     orchestrator.executeBuild(build);
 
     assertThat(new String(Files.readAllBytes(Paths.get("target/differences")), StandardCharsets.UTF_8)).isEmpty();
+    assertPerfMonitoringAvailable(perfMonitoringDir);
+  }
+
+  // asserting perf monitoring on TypeScript project as it creates all kinds of metrics
+  private void assertPerfMonitoringAvailable(String perfMonitoringDir) throws IOException {
```
saberduck, line 107: better to have `Path` as argument type

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
```diff
@@ -461,7 +463,7 @@ static class MonitoringOutputConsumer implements Consumer<String> {
 
     @Override
     public void accept(String s) {
-      if (HEADER.equals(s)) {
+      if (s.startsWith(HEADER_HEAD) && s.endsWith(HEADER_TAIL)) {
```
saberduck, line 466: wdyt about using regex here? seems like we could just have `Rule .+Time (ms) | Relative`

saberduck (changes requested):
some minor comments

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
vilchik-elena: I try to avoid using regex when it's possible :) but let me try)

saberduck (approved):

PR #3353, opened 2022-08-29
Update versions of sonar dependencies

francoismora:
Update sonarqube version to latest and introduce dependency to sonar-plugin-api pointing to latest version.

### DIFF
--- a/its/plugin/tests/pom.xml
+++ b/its/plugin/tests/pom.xml
@@ -31,12 +31,12 @@
       <artifactId>sonar-orchestrator</artifactId>
     </dependency>
     <dependency>
-      <groupId>org.sonarsource.sonarqube</groupId>
-      <artifactId>sonar-ws</artifactId>
-    </dependency>
-    <dependency>
       <groupId>org.sonarsource.sonarlint.core</groupId>
       <artifactId>sonarlint-core</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.sonarsource.sonarqube</groupId>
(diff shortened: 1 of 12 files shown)


vilchik-elena (approved):

vilchik-elena (changes requested):
Plugin test for SonarLint failed, not sure why. May be there sonarlint API version we use is somehow not compatible

its/plugin/tests/pom.xml
```diff
@@ -30,14 +30,14 @@
...
     </dependency>
-    <dependency>
-      <groupId>org.sonarsource.sonarqube</groupId>
-      <artifactId>sonar-ws</artifactId>
-    </dependency>
     <dependency>
       <groupId>org.sonarsource.sonarlint.core</groupId>
       <artifactId>sonarlint-core</artifactId>
     </dependency>
+    <dependency>
```
vilchik-elena, line 37: @francoismora could you share why this fixes the problem?

francoismora: A method has been removed in the latest version of the plugin API (`Loggers.setTarget()`) but was still available in the version included in sonarlint. Changing the classpath order fixed the issue, but we decided with @saberduck to upgrade sonarlint instead to have all components aligned to the latest version.

vilchik-elena: should we revert this change of order then?

vilchik-elena (approved):
LGTM, just wonder if we can revert the change of order

PR #3375, opened 2022-09-13
Replace forked yaml dependency with the original one

### LINKED ISSUE #3234: Replace forked `yaml` dependency with the original one
Depends on #3231 

As a reminder, we had to fork [yaml](https://github.com/eemeli/yaml) as it was not compatible with Node.js 12. Once we drop the support of this version, we can switch back to the original dependency.

yassin-kammoun-sonarsource:
Fixes #3234

### DIFF
--- a/eslint-bridge/package-lock.json
+++ b/eslint-bridge/package-lock.json
@@ -37,7 +37,7 @@
         "stylelint",
         "vue-eslint-parser",
         "typescript",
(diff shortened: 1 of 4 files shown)


ilia-kebets-sonarsource (approved):
LGTM. `yaml-node12`, you will be missed.

PR #3376, opened 2022-09-13
Support ES modules file extensions

ilia-kebets-sonarsource:
Fixes #3147


yassin-kammoun-sonarsource (approved):
LGTM. I don't think we need an integration test, but could you make sure we are indeed able to analyze files with these extensions with a local SonarQube instance? I would test a code snippet against a type-aware rule. Also, I would check for both JavaScript and TypeScript (there should be a TSConfig).

PR #3379, opened 2022-09-13
Log in debug a path to the file being analyzed

### LINKED ISSUE #3355: Log in debug a path to the file being analyzed
Sometimes it's hard to know which file made the analysis fail like [here](https://community.sonarsource.com/t/java-lang-illegalargumentexception-overlapping-symbol-declaration-and-reference-for-symbol-at-range/69996/3) and having a log would help a lot in the investigation

ilia-kebets-sonarsource:
Fixes #3355

### DIFF
--- a/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/CssRuleSensor.java
+++ b/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/CssRuleSensor.java
@@ -126,9 +126,9 @@ public class CssRuleSensor extends AbstractEslintSensor {
         LOG.debug("Skipping {} as it has not 'file' scheme", uri);
         return;
       }
+      LOG.debug("Analyzing file: " + inputFile.uri());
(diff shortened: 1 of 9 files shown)


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/CssRuleSensor.java
```diff
@@ -126,9 +126,9 @@ void analyzeFile(InputFile inputFile, SensorContext context, List<StylelintRule>
         LOG.debug("Skipping {} as it has not 'file' scheme", uri);
         return;
       }
+      LOG.debug("Analyzing file: " + inputFile.uri());
```
francoismora, line 129: Using a placeholder here, like in the class `AnalysisWithProgram.java`, makes sense, as there's a minor performance penalty in using string concatenation when the log level is not showing DEBUG messages.

francoismora (approved):
LGTM

PR #3381, opened 2022-09-14
Lower log level of heartbeat

### LINKED ISSUE #3370: Avoid too much logging for heartbeat
We should log on `TRACE` level or not at all

victor-diez-sonarsource:
Fixes #3370


ilia-kebets-sonarsource (approved):
LGTM

PR #3392, opened 2022-09-16
Add a cache reporter class

### LINKED ISSUE #3391: Add reporting logs with cache statistics
Cache logs are written with level DEBUG. When analyzing analyzer performance, we thus lack details about the caching activity. We need statistics written with level INFO at the end of the analysis.

francoismora:
Fixes #3391


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/Monitoring.java
```diff
@@ -94,16 +99,28 @@ void stopSensor() {
     }
     sensorMetric.duration = sensorMetric.clock.stop();
     metrics.add(sensorMetric);
+    cacheReporter.logReport();
```
vilchik-elena, line 102: we can't rely on this class as monitoring might not be enabled (which is by default)

francoismora:
Closed as replaced by #3395

PR #3404, opened 2022-09-21
Refactor eslint-bridge: asynchronous file reads and rule bundle loading

victor-diez-sonarsource:
Fixes #


eslint-bridge/src/linting/eslint/linter/bundle-loader.ts
```diff
@@ -0,0 +1,120 @@
...
+  for (const bundleId of rulesBundles) {
+    loaders[bundleId](linter);
+  }
+}
+
+/**
+ * Loaders for each of the predefined rules bundles. Each bundle comes with a
+ * different data structure (array/record/object), plus on some cases
+ * there are specifics that must be taken into account, like ignoring some
+ * rules from some bundles or decorating them in order to be compatible.
```
yassin-kammoun-sonarsource, line 47: `in order to be compatible` ?

eslint-bridge/src/linting/eslint/linter/bundle-loader.ts
```diff
@@ -0,0 +1,120 @@
...
+  /**
+   * Loads plugin rules
+   *
+   * Adds the rules from the Sonar ESLint plugin
+   */
+  pluginRules(linter: Linter) {
+    linter.defineRules(pluginRules);
+  },
+  /**
+   * Loads internal rules
```
yassin-kammoun-sonarsource, line 85:
```suggestion
   * Loads internal rules.
```

eslint-bridge/src/services/analysis/analysis.ts
```diff
@@ -40,7 +40,7 @@ export type Analysis = (input: AnalysisInput) => Promise<AnalysisOutput>;
  */
 export interface AnalysisInput {
   filePath: string;
-  fileContent: string | undefined;
```
yassin-kammoun-sonarsource, line 43: We should undo this change because this type denotes part of the contract between the plugin and the bridge. The file content can be undefined depending on the analysis context:

https://github.com/SonarSource/SonarJS/blob/b350ccb01cff555e914e3c52a5037a8215e7dc11/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/JavaScriptEslintBasedSensor.java#L113-L115

Therefore, this type should reflect the actual content of data sent to the bridge.

yassin-kammoun-sonarsource (commented):
Great job overall. I really liked the whole bundle loading thing as well as deciding early if we should read the input file or use the file content if provided, especially at a single place in the whole production code base. That would have avoided the oversight of not considering the input file content for YAML analysis.

However, I am still not convinced about introducing asynchronous file reads, and here is why:
- That's subjective and maybe just me not liking the impact of async-await on a code base, but those asynchronous reads are making existing unit tests more convoluted and less readable, and new unit tests more cumbersome to write.
- We could still keep asynchronous reads in production code, but I would prefer to have synchronous ones for test code and avoid async-await as much as possible in unit tests, otherwise we end up with weird code patterns like awaiting for an input as follows: `const result = analyzeYAML(await yamlInput({ filePath: join(fixturesPath, 'secondary.yaml') }));`.
- There are no improvements in terms of code structure or performance since we still rely on a sequential analysis protocol, so the _why_ for this change remains unclear to me. What are we trying to _refactor_ or _improve_ here? What's the problem we are trying to solve?
- Even if we are talking about baby steps for parallel analysis sometimes in an undefined future, the whole approach of how eslint-bridge handles analysis requests will likely change, and I don't see how asynchronous reads could help or make a difference as of now.

A lot of effort has been put into this now, I don't want to discard it, but I need to be convinced that it improves the code base in one way or another :). Therefore, before going further in the review, could you challenge the points I am raising and argue in favor of why we need this. Maybe I am just missing something obvious.

eslint-bridge/report.json
```diff
@@ -0,0 +1 @@
+[]
```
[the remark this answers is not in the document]
victor-diez-sonarsource: yes, removing

eslint-bridge/src/helpers/files.ts
```diff
@@ -18,37 +18,51 @@
...
+/**
+ * Async read of file contents from a file path
+ *
+ * The function gets rid of any Byte Order Marker (BOM)
+ * present in the file's header.
+ *
+ * @param filePath the path of a file
+ * @returns Promise which resolves with the content of the file
+ */
+export async function readFileAsync(filePath: string) {
```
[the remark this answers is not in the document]
victor-diez-sonarsource: yep, I guess one of the rebases did this... I'll remove readFileAsync

eslint-bridge/src/linting/eslint/linter/bundle-loader.ts
victor-diez-sonarsource: I meant the sanitizeTypeScriptESLintRule on the decorate function

yassin-kammoun-sonarsource (approved):

PR #3409, opened 2022-09-26
Fix FP S1172 ('no-unused-function-argument'): Ignore parameters referenced as JSX identifiers

### LINKED ISSUE #3117: Fix FP S1172 ('no-unused-function-argument'): Ignore parameters referenced as JSX identifiers
S1172 currently raises false positives when a parameter occurs only in the position of a JSX identifier. This happens because ESLint context does not bind JSX identifier references to parameter definitions, if any.

```javascript
function f(Tag) { // FP: `Tag` is flagged as unused
  return <Tag />
}
```

We need to resolve those references ourselves and filter out "unused" parameters if they are actually used as JSXIdentifier nodes.

https://community.sonarsource.com/t/cannot-suppress-jsx-false-positive-javascript-s1172-unusedfunctionargument-in-intellij-idea-ultimate/61474

yassin-kammoun-sonarsource:
Fixes #3117 

The false positive originally reported is no longer raised. Upgrading ESLint at some points probably included a bugfix in the scope analysis, but I was not able to identify which release fixed it. In any case, I added a test case to track any regression.

### DIFF
--- a/eslint-bridge/tests/linting/eslint/rules/no-unused-function-argument.test.ts
+++ b/eslint-bridge/tests/linting/eslint/rules/no-unused-function-argument.test.ts
@@ -22,7 +22,7 @@ import { isParameterProperty, rule } from 'linting/eslint/rules/no-unused-functi
 
 const ruleTester = new RuleTester({
   parser: require.resolve('@typescript-eslint/parser'),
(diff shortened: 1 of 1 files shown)


ilia-kebets-sonarsource (approved):
LGTM. Thanks for the test!

PR #3410, opened 2022-09-26
Filter out YAML files that are not AWS configurations for inline JavaScript

### LINKED ISSUE #3388: Filter out YAML files that are not AWS configurations for inline JavaScript
With the introduction of JavaScript code analysis inside YAML configuration files for AWS, the JavaScript analyzer tries to analyze all kinds of YAML files. This creates a dependency on Node.js for some projects, which is only necessary for the context of AWS for JavaScript.

Therefore, we should improve the detection of JavaScript code inside YAML, and we must avoid the unnecessary launching of a Node.js process to deploy ESLint bridge. Similar to the IaC and security analyzers, we need to check the YAML file header (`Transform: AWS::Serverless-2016-10-31`) early and look for any occurrences of Node.js runtime.

Reported in https://community.sonarsource.com/t/sonarcloud-bitbucket-timing-out/71471/2

ilia-kebets-sonarsource:
Fixes #3388


sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/YamlSensorTest.java
```diff
@@ -224,11 +232,15 @@ private static JavaScriptChecks checks(String... ruleKeys) {
     return new JavaScriptChecks(new CheckFactory(builder.build()));
   }
 
-  private static DefaultInputFile createInputFile(SensorContextTester context) {
+  private static DefaultInputFile createInputFile(SensorContextTester context, String contents) {
+    if (contents == null) {
+      contents = YamlSensor.SAM_TRANSFORM_FIELD;
+      contents += "\nif (cond)\ndoFoo(); \nelse \ndoFoo();";
+    }
```
yassin-kammoun-sonarsource, line 239: It's a good idea to pass the contents explicitly. However, using it as hack to inject the serverless header if the content is null is not so great. The tests are already convoluted enough, so let's try not to make it worse :)

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/YamlSensor.java
```diff
@@ -98,11 +102,30 @@ protected void analyzeFiles(List<InputFile> inputFiles) throws IOException {
...
+          return true;
+        }
+      }
+    } catch (IOException e) {
+      logger.error(String.format("Unable to read file: %s.", inputFile.uri()));
+      logger.error(e.getMessage());
+    }
+
+    return false;
+  }
```
yassin-kammoun-sonarsource, line 127: This check is not enough to prevent the plugin from starting a Node.js process needlessly.

To give an example, let's assume that a project to analyze includes a YAML file denoting Inline AWS Lambda for Python. Provided that the configuration file includes the Serverless header, the file will therefore not be ignored, a Node.js process will be started, ESLint bridge will be deployed, and the latter will try to parse the YAML file and stops once it figures out that the runtime is different from Node.js. The problem here is that we are forcing the users to have Node.js runtime for every project including a YAML file with the expected header.

As explained in the GitHub issue, and in addition to the Serverless header, we need to look for any occurrence of Node.js runtime to accept the input file. A case-insensitive regular expression looking for `Runtime: nodejs` should do the trick. Furthermore, we need an integration test that checks that the plugin doesn't try to start a Node.js process and deploy ESLint bridge for the scenario I described above.

yassin-kammoun-sonarsource (changes requested):

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/YamlSensorTest.java
ilia-kebets-sonarsource: I've refactored it into this, WDYT? https://github.com/SonarSource/SonarJS/pull/3410/files#diff-b8698d58ef815e880646a3f6a4fac2af87ce96d67f50809e6c1684ba531908b6R235

yassin-kammoun-sonarsource: 👍

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/YamlSensor.java
ilia-kebets-sonarsource: done

yassin-kammoun-sonarsource:
Reminder: we need to create a PR and update the integration tests on sonar-security to include the expected header, if not the case already. Also, we should merge both PRs on the same day, ideally at the same time to minimize a transient state where integration tests fail.

yassin-kammoun-sonarsource:
> Reminder: we need to create a PR and update the integration tests on sonar-security to include the expected header, if not the case already. Also, we should merge both PRs on the same day, ideally at the same time to minimize a transient state where integration tests fail.

Never mind this comment. The integration tests we introduced already include the header. No action should be required in the end.

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/YamlSensor.java
```diff
@@ -37,6 +40,8 @@
 public class YamlSensor extends AbstractEslintSensor {
 
   public static final String LANGUAGE = "yaml";
+  public static final String SAM_TRANSFORM_FIELD = "AWS::Serverless-2016-10-31";
+  public static final String NODEJS_RUNTIME_REGEX = "^\\s*Runtime:\\s*[\'\"]?nodejs\\S*[\'\"]?";
```
yassin-kammoun-sonarsource, line 44: The regex doesn't work if there is any trailing whitespace or a comment.

yassin-kammoun-sonarsource (changes requested):

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/YamlSensor.java
ilia-kebets-sonarsource: I believe it will work, since I'm didn't include any end anchor. I checked on https://regex101.com/ using:
- choose `Java 8`
- `^\s*Runtime:\s*[\'\"]?nodejs\S*[\'\"]?` (I still don't know why it requires single escape)
- `      Runtime: nodejs12.x  #`

yassin-kammoun-sonarsource: I checked with jshell (Java REPL), and it didn't seem to work:

```
% jshell
|  Welcome to JShell -- Version 11.0.16
|  For an introduction type: /help intro

jshell> String regex = "^\\s*Runtime:\\s*[\'\"]?nodejs\\S*[\'\"]?"
regex ==> "^\\s*Runtime:\\s*['\"]?nodejs\\S*['\"]?"

jshell> "Runtime: nodejs12.x #".matches(regex)
$2 ==> false

jshell> "Runtime: nodejs12.x ".matches(regex)
$4 ==> false

jshell> "Runtime: nodejs12.x".matches(regex)
$4 ==> true
```

ilia-kebets-sonarsource: I changed the implementation, but kept the regex, since I was getting the S5852 flag if the regex ended with `.*`
@yassin-kammoun-sonarsource WDYT?

yassin-kammoun-sonarsource: 👍

its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/YamlAnalysisTest.java
```diff
@@ -34,27 +35,46 @@
...
+      .setSourceEncoding("UTF-8")
+      .setSourceDirs(".")
+      .setDebugLogs(true)
+      .setProjectDir(TestUtils.projectDir("yaml-aws-lambda-skipped"));
+
+    OrchestratorStarter.setProfiles(projectKey, Map.of(
+      "yaml-aws-lambda-profile", "cloudformation",
+      "eslint-based-rules-profile", "js"));
+    BuildResult result = orchestrator.executeBuild(build);
+    assertThat(result.getLogsLines(log -> log.contains("Starting Node.js process"))).hasSize(0);
```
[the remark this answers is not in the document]
ilia-kebets-sonarsource: I'm getting `The main branch has no lines of code.` from SQ after running an analysis. It gets correctly ignored!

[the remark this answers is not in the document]
ilia-kebets-sonarsource: I can also see from the scanner logs that I'm not getting:

```
INFO: Sensor JavaScript inside YAML analysis [javascript]
INFO: 1 source file to be analyzed
INFO: 1/1 source file has been analyzed
INFO: Hit the cache for 0 out of 1
INFO: Miss the cache for 1 out of 1: ANALYSIS_MODE_INELIGIBLE [1/1]
INFO: Sensor JavaScript inside YAML analysis [javascript] (done) | time=4774ms
```

but

```
INFO: Sensor JavaScript inside YAML analysis [javascript]
INFO: No input files found for analysis
INFO: Hit the cache for 0 out of 0
INFO: Miss the cache for 0 out of 0
INFO: Sensor JavaScript inside YAML analysis [javascript] (done) | time=6ms
```

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/YamlSensor.java
```diff
@@ -98,11 +105,43 @@ protected void analyzeFiles(List<InputFile> inputFiles) throws IOException {
...
+        // Normally, we would be looking for an entry like "Transform: AWS::Serverless-2016-10-31", however, checking the whole entry could be
+        // problematic with whitespaces, so we will be looking just for the field value.
+        if (line.contains(SAM_TRANSFORM_FIELD)) {
+          hasAwsTransform = true;
+        }
+        Matcher lineMatch = regex.matcher(line);
+        // AWS SAM can contain code in other languages such as python.
+        if (lineMatch.find()) {
+          hasNodeJsRuntime = true;
+        }
```
yassin-kammoun-sonarsource, line 132:
```suggestion
        // We check early the runtime to avoid making Node.js a mandatory dependency on projects that include YAML configuration files for AWS,
        // and we consider only those which define Node.js as the runtime, which potentially embed JavaScript code.
        Matcher lineMatch = regex.matcher(line);
        if (lineMatch.find()) {
          hasNodeJsRuntime = true;
        }
```

its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/YamlAnalysisTest.java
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: Perfect! Thanks for double-checking.

yassin-kammoun-sonarsource (approved):
Looks good to me. Nice job!

PR #3411, opened 2022-09-27
Do not mention custom rules when deploying security frontend

### LINKED ISSUE #2625: Do not mention custom rules when deploying security frontend
Currently we print following logs when deploying security frontend custom rules:
```
INFO: Sensor JavaScript analysis [javascript]
INFO: Deploying custom rules bundle jar:file:/home/ganncamp/.sonar/cache/1c0d0d6d358b2ff25a27998c60079e2e/sonar-securityjsfrontend-plugin.jar!/js-vulnerabilities-rules-1.0.0.tgz to /home/ganncamp/testProjects/javascript/juice-shop/.scannerwork/.sonartmp/eslint-bridge-bundle/package/custom-rules1939007438647855735
INFO: 5 source files to be analyzed
INFO: 5/5 source files have been analyzed
```

This definitely might sound confusing for the user to know that some "custom" rules are provided by analyzer as they are not.

We should reword this technical details and/or hide them in debug logs.

ilia-kebets-sonarsource:
Fixes #2625


yassin-kammoun-sonarsource (changes requested):
PR is red.

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/RulesBundlesTest.java
```diff
@@ -56,6 +62,28 @@ void test_empty() {
...
+    }
+    assertThat(isSeen).isFalse();
+    for (String logLine: logTester.logs(LoggerLevel.DEBUG)) {
+      System.out.println(logLine);
+      if (logLine.contains(filename)) {
+        isSeen = true;
+      }
+    }
+    assertThat(isSeen).isTrue();
+  }
```
yassin-kammoun-sonarsource, line 86: What do you think about removing this whole test and adding the following lines

```java
assertThat(logTester.logs(LoggerLevel.DEBUG)).hasSize(1);
assertThat(logTester.logs(LoggerLevel.DEBUG).get(0)).startsWith("Deploying custom rules bundle");
```

at the end of the `void test() ...` unit test instead ?

ilia-kebets-sonarsource: done

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/RulesBundlesTest.java
```diff
@@ -56,6 +62,15 @@ void test_empty() {
     assertThat(rulesBundles.deploy(tempDir)).isEmpty();
   }
 
+  @Test
+  void test_deploy_should_log_filename_in_debug() {
```
yassin-kammoun-sonarsource, line 66:
```suggestion
  void test_deploy_should_log_deployment() {
```

yassin-kammoun-sonarsource (approved):
A minor thing to rename, but all good otherwise.

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/RulesBundlesTest.java
ilia-kebets-sonarsource: I have refactored the test to match the test name. I believe it is important to mention the log level.

yassin-kammoun-sonarsource: Sure but it still feels weird to have a test name that says that we log a filename while the functionality being tested is logging at debug level the deployment of a rule bundle. Ideally, we should have a single assertion about the whole logging thing, but since we don't have control over the temporary file created, I would keep the three assertions then: the two I suggested and the one you just added. Also, I would still rename the test, this time to `test_deploy_should_log_deployment_in_debug`.

This may be nitpicking but I would prefer the test name to reflect the functionality being tested, that is deploying, rather than a consequence of it. Does it make sense or am I just being annoying (I probably am :) )?

ilia-kebets-sonarsource: Not sure I understand, but I choose to sit this one out. WDYT about the last version?

yassin-kammoun-sonarsource: 👍

PR #3423, opened 2022-09-29
Update S4138 ('prefer-for-of'): Include warning about browser array-like collections

### LINKED ISSUE #3385: Fix FP S4138 ('prefer-for-of'): Check if node does have an iterator to use for-of
The rule S4138 can raise issues when `for...of` is not available for the iterable.

The [typescript-eslint rule implementation](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/rules/prefer-for-of.ts) does not check the type, so we could add a extend the [current decorator](https://github.com/SonarSource/SonarJS/blob/master/eslint-bridge/src/linting/eslint/rules/decorators/prefer-for-of-decorator.ts) to filter the flagged nodes for type.

Reported in https://community.sonarsource.com/t/fp-seeking-for-of-where-typescript-prohibits-it/67142

ilia-kebets-sonarsource:
Fixes #3385

### DIFF
--- a/eslint-bridge/src/linting/eslint/rules/decorators/prefer-for-of-decorator.ts
+++ b/eslint-bridge/src/linting/eslint/rules/decorators/prefer-for-of-decorator.ts
@@ -44,7 +44,8 @@ export function decoratePreferForOf(rule: Rule.RuleModule): Rule.RuleModule {
       suggest,
       messageId: undefined,
       message:
-        'Expected a `for-of` loop instead of a `for` loop with this simple iteration. In a browser environment, `NodeList` and other array-like collections should work by default. If you are using TypeScript and seeing a type error, make sure your configuration is correct.',
+        rule.meta!.messages!['preferForOf'] +
+        ' In a browser environment, `NodeList` and other array-like collections should work by default. If you are using TypeScript and seeing a type error, make sure your configuration is correct.',
     });
   });
 }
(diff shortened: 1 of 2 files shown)


ilia-kebets-sonarsource:
I tried using `!` to affirm that the `rule.meta.messages` props exist, but the QG kicked me down, so I went with optional chaining that fails on coverage.

eslint-bridge/src/linting/eslint/rules/decorators/prefer-for-of-decorator.ts
```diff
@@ -41,6 +42,10 @@ export function decoratePreferForOf(rule: Rule.RuleModule): Rule.RuleModule {
     context.report({
       ...reportDescriptor,
       suggest,
+      messageId: undefined,
+      message:
+        rule.meta?.messages?.preferForOf +
+        ' In a browser environment, `NodeList` and other array-like collections should work by default. If you are using TypeScript and seeing a type error, make sure your configuration is correct.',
```
yassin-kammoun-sonarsource, line 48: I don't think this kind of message should be part of the issue message. It should be moved in the rule description, that is, the RSPEC. Issue messages should be kept as short as possible as well as actionable.

yassin-kammoun-sonarsource (changes requested):
Let's close this PR without merging and open a PR in the RSPEC repository instead to update the rule description.

ilia-kebets-sonarsource:
Closing this as we have modified the message in rspec: https://github.com/SonarSource/rspec/pull/1303

PR #3429, opened 2022-09-30
Fix FP S4328: Ignore `data:` and `file:` imports

### LINKED ISSUE #3414: Fix FP S4328: Ignore `data:` and `file:` imports
#3413 introduced an exception to ignore `node:` imports for S4328. We should do the same for `data:` and `file:` imports.

yassin-kammoun-sonarsource:
Fixes #3414

### DIFF
--- a/eslint-bridge/tests/linting/eslint/rules/no-implicit-dependencies.test.ts
+++ b/eslint-bridge/tests/linting/eslint/rules/no-implicit-dependencies.test.ts
@@ -107,6 +107,16 @@ ruleTester.run('Dependencies should be explicit', rule, {
       filename,
       options,
     },
+    {
+      code: `import 'data:text/javascript,console.log("hello, world!");';`,
+      filename,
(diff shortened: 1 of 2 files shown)


eslint-bridge/tests/linting/eslint/rules/no-implicit-dependencies.test.ts
```diff
@@ -107,6 +107,16 @@ ruleTester.run('Dependencies should be explicit', rule, {
       filename,
       options,
     },
+    {
+      code: `import 'data:text/javascript,console.log("hello, world!");';`,
```
ilia-kebets-sonarsource, line 111: what the XXX is that? I understand that it is part of the [spec](https://nodejs.org/api/esm.html#urls), but should we really encourage this kind of feature?

ilia-kebets-sonarsource (approved):
LGTM

PR #3430, opened 2022-09-30
Fix FP S1848 `constructor-for-side-effects`: Ignore standard APIs that produce side effects

### LINKED ISSUE #3163: Fix FP S1848 `constructor-for-side-effects`: Ignore standard APIs that produce side effects
The rule S1848 raises an issue when instantiating a Notification object:

```javascript
new Notification('msg')
```

However, the [Notification API](https://developer.mozilla.org/en-US/docs/Web/API/notification) is meant to be used this way to send a message.

Reported in https://community.sonarsource.com/t/javascript-s1848-and-notification-api/63604

ilia-kebets-sonarsource:
Fixes #3163

### DIFF
--- a/eslint-bridge/src/linting/eslint/rules/constructor-for-side-effects.ts
+++ b/eslint-bridge/src/linting/eslint/rules/constructor-for-side-effects.ts
@@ -99,6 +99,9 @@ function isException(
   node: estree.Identifier | estree.MemberExpression,
   name: string,
 ) {
+  if (name === 'Notification') {
+    return true;
+  }
   let identifierFromModule: estree.Identifier;
   if (node.type === 'MemberExpression' && node.object.type === 'Identifier') {
     identifierFromModule = node.object;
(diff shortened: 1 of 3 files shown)


eslint-bridge/src/linting/eslint/rules/constructor-for-side-effects.ts
```diff
@@ -82,3 +90,38 @@ function reportIssue(
...
+  let { module } = getModuleAndCalledMethod(node, context);
+  // handle import
+  if (module === undefined) {
+    module =
+      getModuleNameOfImportedIdentifier(context, identifierFromModule) ||
+      getModuleNameOfIdentifier(context, identifierFromModule);
+  }
+
+  if (
+    name === 'Notification' ||
```
ilia-kebets-sonarsource, line 120: I'm not sure if there is a way to know that it is `Notification` from the browser, is there?

ilia-kebets-sonarsource: As discussed with @victor-diez-sonarsource we should capture the global `Notification`, but I'm not sure there is a clean way to perform this.

victor-diez-sonarsource (approved):

eslint-bridge/src/linting/eslint/rules/constructor-for-side-effects.ts
yassin-kammoun-sonarsource: Was this question answered or investigated? I am not sure it's a good idea to merge something when there are pending questions or leave them as it is if they were clarified in another communication channel.

ilia-kebets-sonarsource: As we didn't find an easy way to perform this, we decide to leave this check as simple as this and improve it if someone reports any FNs.

PR #3433, opened 2022-10-03
Fix FP S4043 (`no-misleading-array-reverse`): Consider optional chaining

### LINKED ISSUE #2513: Fix FP S4043 ('no-misleading-array-reverse'): Consider optional chaining
User reported under RSPEC https://jira.sonarsource.com/browse/RSPEC-4043 

Hi, I tried this in SonarJS, it seems `items?.sort()` reports while `items.sort()` does not which is unexpected.

What means this rule is not compatible with `optional chaining`.

victor-diez-sonarsource:
Fixes #2513


victor-diez-sonarsource:
agreed with @gabriel-vivas-sonarsource to cover both:
`a?.sort()`
and 
`a && a.sort()`

ilia-kebets-sonarsource (approved):
LGTM

PR #3438, opened 2022-10-03
FP S1105 (`brace-style`): reporting on closing brace

### LINKED ISSUE #2856: FP S1105 `brace-style`:  reporting on closing brace
Rule S1105 is about open brace only, while eslint's `brace-style` is reporting on both open and close braces.

So we are having following issues
```js
if (cond) { return foo(); } // 'Closing curly brace does not appear on the same line as the subsequent block. '
//                        ^
else { return bar(); }
```

Reported in https://community.sonarsource.com/t/brace-style-rule-s1105-flagging-single-line-if-else-pairs/30865

victor-diez-sonarsource:
Fixes #2856


ilia-kebets-sonarsource (approved):
LGTM, don't forget to deprecate this rule as per https://github.com/SonarSource/SonarJS/issues/3441

PR #3446, opened 2022-10-05
Support wildcard patterns for sonar.typescript.tsconfigPath(s)

### LINKED ISSUE #1954: Support wildcard patterns for sonar.typescript.tsconfigPath(s)
When using default name `tsconfig.json` we will recursively search for all `tsconfig.json` files in the project. However, if project uses different naming convention, property `sonar.typescript.tsconfigPath` allows to configure only single file.

We should support wildcard pattern in this property to allow using different naming convention, e.g.
`sonar.typescript.tsconfigPath=**/tsconfig.settings.json` 
We should also allow multiple files delimited by comma (similar to sonar.sources)
`sonar.typescript.tsconfigPath=first.tsconfig.json,another.tsconfig.json`

We can consider changing the property name to plural form to reflect the fact that it accepts multiple files.

yassin-kammoun-sonarsource:
Fixes #1954


yassin-kammoun-sonarsource:
Please let me know if the amount of logs is fine (or if we need more) as well as the logs level.

saberduck (approved):

PR #3447, opened 2022-10-05
Update S4325 (`no-unnecessary-type-assertion`): write test for FP so we know when typescript-eslint fixes it

### LINKED ISSUE #3235: Fix FP S4325 (`no-unnecessary-type-assertion`): `Object.values()` result
```ts
export function isPositive(x: number) {
  return x > 0;
}

export type PositiveNumberRecord<T extends string | number | symbol> =
  Record<T, number> & { __type: 'positiveNumberRecord' };

export const isPositiveNumberRecord = <T extends string | number | symbol>
  (value: Record<T, number> | null | undefined): value is PositiveNumberRecord<T> => {

    if (!value) {
      return false;
    }

    const values = Object.values(value) as number[];  // FP is here

    return values.every(isPositive);
};
```

Reported in https://community.sonarsource.com/t/sonarlint-says-cast-is-unnecessary/64340

ilia-kebets-sonarsource:
Fixes #3235


yassin-kammoun-sonarsource (commented):
Okay, now that I look at this and see that we had to go as far as decorating a rule just for the sake of testing, it feels weird to me to add a unit test to keep track of a problem that is on TypeScript ESLint side, or maybe even TypeScript. I think we should rather create an issue on TypeScript ESLint GitHub repository and close our issue with a comment that links to that other issue.

Does it make sense?

ilia-kebets-sonarsource:
@yassin-kammoun-sonarsource It sure does. I'll raise the question at tomorrow's standup.

ilia-kebets-sonarsource:
As the issue comes from typescript-eslint or typescript's TypeChecker, we will open an issue on https://github.com/typescript-eslint/typescript-eslint/issues

PR #3455, opened 2022-10-07
Fix FP S2189 ('no-infinite-loop'): Ignore condition variables called on or passed over

### LINKED ISSUE #2866: Fix FP S2189 ('no-infinite-loop'): Ignore condition variables called on or passed over
We are currently raising false positives on variables of loop conditions that could possibly be subject to mutation through function and methods calls:

```javascript
let foo = { /* ... */ };
while (foo) { // FP
  foo.doSomething(); // `doSomething` could modify `foo`
}
```

We should assume that such variables are always modified in order to kill the noise.

Reported in https://community.sonarsource.com/t/l3989-false-positive-in-loopsshouldnotbeinfinitecheck-should-support-guard-clause

yassin-kammoun-sonarsource:
Fixes #2866

I did a bit of refactoring before tackling the scope of the ticket. To help the review, I pushed two different commits: one for the refactoring and the other for the actual fix.


ilia-kebets-sonarsource (approved):
LGTM! Good job!

PR #3466, opened 2022-10-10
Document how to define secondary locations in comment-based framework

ilia-kebets-sonarsource:


victor-diez-sonarsource (commented):
can you add an explanation about secondary messages? what are they used for? if it's internal to our reporting or if ESLint has them?

victor-diez-sonarsource (approved):

PR #3472, opened 2022-10-12
Increase timeout so test won't fail in `build_win` CI check

ilia-kebets-sonarsource:
It fails often: https://gist.github.com/ilia-kebets-sonarsource/557cbdfa09d496c2dd28251b97336423#build_win


ilia-kebets-sonarsource:
@victor-diez-sonarsource Do you see a reason why the failure was mainly (or only) appearing in the Windows container? 
I believe the server was shutting down because the first call wasn't quick enough, that's why I increased the shutdown timeout from 200 to 500.

PR #3475, opened 2022-10-13
Add rule S6275: Using unencrypted EBS volumes is security-sensitive 

### LINKED ISSUE #3417: Rule S6275: Using unencrypted EBS volumes is security-sensitive
## WHAT
The rule detects when an EBS volume is not configured to enforce data encryption using the JavaScript AWS CDK  library. It does not cover custom policy statement definitions.

This has been previously implemented for Python see [SONARPY-1110](https://sonarsource.atlassian.net/browse/SONARPY-1110). 

Specification ticket: [APPSEC-163](https://jira.sonarsource.com/browse/APPSEC-163)
MMF: [MMF-2906](https://sonarsource.atlassian.net/browse/MMF-2906)

## HOW
- Class/Construct: `aws-cdk-lib.aws_ec2.Volume`
- Argument: `props` 3rd positional argument
- Property: `encrypted`
  - Safe values: `true`
  - Sensitive value: `false`
  - Default sensitive value: when not set `encrypted` value defaults to `false`

```typescript
import { Size } from 'aws-cdk-lib';
import { Volume } from 'aws-cdk-lib/aws-ec2';

// Default
const enforceEncryptionDefault = new Volume(this, 'Volume', {
      availabilityZone: 'us-west-1a',
      size: Size.gibibytes(1)
}); // Sensitive: Omitting 'encrypted' disables volumes encryption. Make sure it is safe here.

const enforceEncryptionFalse = new Volume(this, 'Volume', {
      availabilityZone: 'us-west-1a',
      size: Size.gibibytes(1),
      encrypted: false // Sensitive: Make sure that using unencrypted volumes is safe here.
});

const enforceEncryptionFalse = new Volume(this, 'Volume', {
      availabilityZone: 'us-west-1a',
      size: Size.gibibytes(1),
      encrypted: true // Compliant
});
```

## Code examples
https://github.com/SonarSource/appsec-poc/pull/29

## Message and highlighting

https://github.com/SonarSource/rspec/pull/1286

After implementation, please merge this https://github.com/SonarSource/rspec/pull/1286.

francoismora:
Fixes #3417


francoismora:
The work on this PR is still in progress and is meant to give quick feedback on the rule template API.

yassin-kammoun-sonarsource (commented):
This draft already looks good to me.

I am sharing below pointers to helper functions that you might find useful to cover more cases. You can see a concrete example of their use with my draft PR [here](https://github.com/SonarSource/SonarJS/pull/3476).

__getUniqueWriteUsageOrNode__

This helper tracks back the definition of an identifier and extracts the value assigned to it, in case it is a constant. That could help you support the following case:

```javascript
const maybeEncrypted = false;
/* ... */
new Volume(this, "Volume", { encrypted: maybeEncrypted });
```

In this case, the helper would look for the definition of `maybeEncrypted`, if any, and returns its value.

__getValueOfExpression__

This helper function does basically the same thing as `getUniqueWriteUsageOrNode` but also checks that the extracted value matches a given node type. That could help you support the following case:

```javascript
const props = { encrypted: false };
new Volume(this, "Volume", props);
```

In this case, the helper would look for the definition of `props` and make sure at the same time that the underlying node has a specific type, `ObjectExpression` in this case.

yassin-kammoun-sonarsource:
I forgot to mention that if the rule doesn't raise any issues on the ruling projects, you can drop a small test file [there](https://github.com/SonarSource/SonarJS/tree/master/its/sources/file-for-rules). We use this folder for rules that don't raise any issues on the current ruling projects. It often happens with hotspot rules.

francoismora:
The PR is now complete and identifier values are now resolved using the helper functions `isUndefined()`, `isLiteral()`, `isIdentifier()` and `getUniqueWriteUsage()`. The new version uses now an error as this helped simplify significantly the checker logic.

yassin-kammoun-sonarsource:
@francoismora Let's try to say that a PR is complete and ready for review once all its jobs are green. Otherwise, it could happen that some CI jobs, especially the ruling ones for new rules, could highlight a problem that requires implementation changes and consequently invalidates previous a PR approval.

francoismora:
A change was necessary to fix the ruling in the `getFullyQualifiedName()` function to support null in import definition source value triggered by the file `PgCollection.ts` from the project 'postgraphql'. None of the file imports put in isolation in unit tests triggered the bug, so I haven't been able to add a test to cover the fix.

yassin-kammoun-sonarsource (approved):
LGTM. I suggest we rename the ESLint rule key to so that it mentions encryption.

yassin-kammoun-sonarsource:
@francoismora Don't forget to merge the RSPEC PR as well.

PR #3482, opened 2022-10-17
Improve detection of bundled code

### LINKED ISSUE #3368: Ignore bundled code
We would currently analyze (and fail) on files such as 

https://github.com/SonarResearchScanning/odoo/blob/master/addons/web/static/lib/pdfjs/build/pdf.js 

we should revisit our detection of generated code and avoid such files automatically.

saberduck:
Fixes #3368


saberduck:
Closed because branch needed to be renamed

PR #3487, opened 2022-10-18
Fix FP S1128 (`unused-imports`): Vue.js imports used in templates

### LINKED ISSUE #3484: Fix FP S1128 (`unused-imports`): Vue.js imports used in templates
The script attribute `lang="ts"` is necessary for this FP to occur.

```javascript
<template>
  <Foo />
</template>

<script setup lang="ts">
import Foo from './Foo' // FP
</script>
```

Reported in:
- https://community.sonarsource.com/t/remove-this-unused-import-of-xxx-rule-is-not-detecting-them-in-template-vue-js/74330
- https://community.sonarsource.com/t/false-postive-javascript-s1128-with-vue-script-setup-lang-ts/73990
- https://community.sonarsource.com/t/vue3-composition-api-and-script-setup-syntax-reports-unused-imports-since-october-22/74614

ilia-kebets-sonarsource:
Fixes #3484

### DIFF
--- a/eslint-bridge/tests/tools/testers/comment-based/helpers/comments.ts
+++ b/eslint-bridge/tests/tools/testers/comment-based/helpers/comments.ts
@@ -29,11 +29,14 @@ export interface Comment {
   endColumn: number;
 }
 
-export function extractComments(fileContent: string): Comment[] {
-  const parsed = buildSourceCode(
-    { fileContent, filePath: '', fileType: null, tsConfigs: [] },
-    null,
-  );
+/**
+ *
+ * @param fileContent
+ * @param filePath
+ * @returns
+ */
+export function extractComments(fileContent: string, filePath: string): Comment[] {
+  const parsed = buildSourceCode({ fileContent, filePath, fileType: null, tsConfigs: [] }, null);
   let esTreeComments: estree.Comment[];
   if (parsed instanceof SourceCode) {
     esTreeComments = parsed.getAllComments();
--- a/eslint-bridge/tests/tools/testers/comment-based/helpers/file.ts
+++ b/eslint-bridge/tests/tools/testers/comment-based/helpers/file.ts
@@ -37,9 +37,10 @@ export class FileIssues {
   /**
    * Parses the file into its expected errors. Throws if error flags are not well formatted.
    * @param fileContent
+   * @param filePath
    */
-  constructor(fileContent: string) {
-    const comments = extractComments(fileContent);
+  constructor(fileContent: string, filePath: string) {
+    const comments = extractComments(fileContent, filePath);
     for (const comment of comments) {
       if (isNonCompliantLine(comment.value)) {
         extractLineIssues(this, comment);
--- a/eslint-bridge/tests/tools/testers/comment-based/launcher.test.ts
+++ b/eslint-bridge/tests/tools/testers/comment-based/launcher.test.ts
@@ -59,11 +59,15 @@ function runRuleTests(rules: Record<string, Rule.RuleModule>, ruleTester: RuleTe
     const filename = path.join(fixtures, testFile);
     const { ext, name } = path.parse(filename);
     const rule = name.toLowerCase();
-    if (['.js', '.jsx', '.ts', '.tsx'].includes(ext.toLowerCase()) && rules.hasOwnProperty(rule)) {
+    if (
+      ['.js', '.jsx', '.ts', '.tsx', '.vue'].includes(ext.toLowerCase()) &&
+      rules.hasOwnProperty(rule)
+    ) {
       describe(`Running comment-based tests for rule ${rule} ${ext}`, () => {
         const code = fs.readFileSync(filename, { encoding: 'utf8' }).replace(/\r?\n|\r/g, '\n');
         const { errors, output } = extractExpectations(
           code,
+          filename,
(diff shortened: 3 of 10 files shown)


ilia-kebets-sonarsource:
@yassin-kammoun-sonarsource I verified that putting back the code removed in that commit does indeed fix the FP: https://github.com/SonarSource/SonarJS/commit/03dc60ab105e0133e9d3b096038a38239a896e96

eslint-bridge/tests/tools/testers/comment-based/helpers/comments.ts
```diff
@@ -29,11 +29,14 @@ export interface Comment {
...
-  const parsed = buildSourceCode(
-    { fileContent, filePath: '', fileType: null, tsConfigs: [] },
-    null,
-  );
+/**
+ *
+ * @param fileContent
+ * @param filePath used to know if it's Vue or not
+ * @returns
+ */
```
yassin-kammoun-sonarsource, line 37:
```suggestion
```

yassin-kammoun-sonarsource: I don't think we need to duplicate this comment, especially because we can track back the change to the commit which introduced the change.

eslint-bridge/tests/tools/testers/comment-based/helpers/file.ts
```diff
@@ -37,9 +37,10 @@ export class FileIssues {
   /**
    * Parses the file into its expected errors. Throws if error flags are not well formatted.
    * @param fileContent
+   * @param filePath used to know if it's Vue or not
```
yassin-kammoun-sonarsource, line 40:
```suggestion
   * @param filePath
```

yassin-kammoun-sonarsource (changes requested):
Lookgs good to me. However, could you update the definition of `languageFromFilePath` and take into account `.vue` files:

https://github.com/SonarSource/SonarJS/blob/ff94357392bc2b742ce4cdfe0a0d6d7ce2d836be/eslint-bridge/tests/tools/testers/comment-based/launcher.test.ts#L110-L117

You will need to pass the `fileContent` from `parseForESLint` and use this [regular expression](https://github.com/SonarSource/SonarJS/blob/d6b6ca841ef300235775bd611107bdf64e4117ab/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/JavaScriptFilePredicate.java#L32) to make the decision on the language of such a file.

There won't be a major impact, but by doing so, the comment-based testing framework will be closer to a `.vue` file analysis in production.

eslint-bridge/tests/tools/testers/comment-based/helpers/comments.ts
ilia-kebets-sonarsource: ack

eslint-bridge/tests/tools/testers/comment-based/launcher.test.ts
```diff
@@ -105,11 +109,15 @@ export function parseForESLint(
 }
 
 /**
- * Returns the source code's language based on the file path.
+ * Returns the source code's language based on the file content and path.
  */
-function languageFromFilePath(filePath: string): Language {
+function languageFromFile(fileContent: string, filePath: string): Language {
+  const hasScriptTagWithLangTs = /\\<script[.\\s]*lang=['\"]ts['\"][.\\s]*\\>/;
```
yassin-kammoun-sonarsource, line 115: The regex doesn't seem to work. Because we are using a JavaScript regex literal here, the semantics is different from Java, and we can't copy-paste the regex because we don't need to escape characters.

Also, I realize now that the regex in the file predicate (as well as this one) does not support `<script setup...`. Let's fix the regex of `JavaScriptFilePredicate.java` in a different PR or have the change in this one but in a dedicated commit (which we should not squash when the time to merge comes).

Long story short, the regex should include `setup` and mark it as optional.

yassin-kammoun-sonarsource (changes requested):

ilia-kebets-sonarsource:
@yassin-kammoun-sonarsource LMK if you're fine with the refactoring of the regex. If yes, I'll rebase this branch into 3 commits:
- Support VueJS test files in comment-based framework
- Fix FP S1128 (`unused-imports`): Vue.js imports used in templates on TS script
- Fix VueJS typescript filter

You can check the pre-last commit for another variant of the regex, but I prefer the one I settled on:
- the last: `/<script[^>]+lang=['"]ts['"][^>]*>/`
- the pre-last: `/<script[\s\S]*?lang=['"]ts['"][\s\S]*?>/`

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/JavaScriptFilePredicate.java
```diff
@@ -29,7 +29,7 @@
 
 public class JavaScriptFilePredicate {
 
-  private static final String regex = "\\<script[.\\s]*lang=['\"]ts['\"][.\\s]*\\>";
+  private static final String regex = "\\<script[^>]+lang=['\"]ts['\"][^>]*>";
```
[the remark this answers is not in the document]
ilia-kebets-sonarsource: @yassin-kammoun-sonarsource I have added 3 more test cases to illustrate my changes.

PR #3491, opened 2022-10-19
Fix S6327 ('aws-sns-unencrypted-topics'): Report on the property value

yassin-kammoun-sonarsource:
I used to report on the whole property rather than the property value, which happens to be inconsistent with the implemented reporting of other rules.


francoismora (approved):
LGTM

PR #3509, opened 2022-10-26
Workaround for bug on WSL when discovering tsconfig

### LINKED ISSUE #3431: Fix discovery of tsconfig file for projects on WSL
The code that tries to locate the tsconfig file doesn't work on Windows when project files are on WSL.

See details here:
https://sonarsource.atlassian.net/browse/SLI-779

As a workaround, we could maybe replace the NIO `Files.walk` by a recursive call to `File.listFiles` (to be tested).

I don't think this is urgent to fix since there is a workaround to manually set the tsconfig path. Maybe SonarLint should always try to detect the tsconfig file on its side (using IDE virtual file system), and pass it to the analyzer?

saberduck:
Fixes #3431 

Tested manually. 

reproduced
2022-10-28 15_41_00-Window

fixed
2022-10-28 15_53_08-file ts - tstmp - Visual Studio Code

Note that this is not the way VS Code recommends to work with WSL. Proper way to work with wsl is to use Remote development extension and to open project within WSL with SonarLint installed inside the WSL workspace.

### DIFF
--- a/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/TsConfigProvider.java
+++ b/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/TsConfigProvider.java
@@ -26,6 +26,7 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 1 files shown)


ilia-kebets-sonarsource (approved):
So basically, you refactored Files.walk() with a good old BFS. Nice.

PR #3510, opened 2022-10-27
Update secrets for CI pipeline

saberduck:


saberduck:
user needs to be refreshed too

PR #3511, opened 2022-10-27
Add bundle version to `RulesBundle` interface and use it for cache key

### LINKED ISSUE #3386: Invalidate UCFG cache when analysis is done with another sonar-security version
As part of [MMF-2827](https://sonarsource.atlassian.net/browse/MMF-2827), the analyzer API allows to cache rule results and in order to support cache invalidation, the javascript analyzer now requires to have more details about the rule bundles provided by external projects like sonar-security.

The new rules bundle details should allow the javascript analyzer to identify the bundle coming from sonar-security and should have a version that can be used to invalidate cached results generated by older versions. 

The version of security plugin should be part of the cache key along with the SonarJS version.

HOW:
Add 2 more methods to `RulesBundle` interface: `version`, `id`. 

P.S. This ticket was not done as part of MMF-2827 as it matters for SonarCloud only (SonarQube has both security and SonarJS updated at the same time). And SonarCloud didn't yet implement caching.

vilchik-elena:
Fixes #3386

### DIFF
--- a/javascript-checks/src/main/java/org/sonar/plugins/javascript/api/RulesBundle.java
+++ b/javascript-checks/src/main/java/org/sonar/plugins/javascript/api/RulesBundle.java
@@ -32,12 +32,12 @@ public interface RulesBundle {
    */
   String bundlePath();
 
-  default String bundleId() {
-    return "n/a";
+  default String bundleKey() {
+    return null;
   }
 
   default String bundleVersion() {
-    return "n/a";
+    return null;
   }
 
 }
--- a/javascript-checks/src/main/java/org/sonar/plugins/javascript/api/RulesBundle.java
+++ b/javascript-checks/src/main/java/org/sonar/plugins/javascript/api/RulesBundle.java
@@ -32,6 +32,12 @@ public interface RulesBundle {
    */
   String bundlePath();
 
+  default String bundleId() {
+    return "n/a";
+  }
 
+  default String bundleVersion() {
+    return "n/a";
+  }
 
 }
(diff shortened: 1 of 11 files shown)


javascript-checks/src/main/java/org/sonar/plugins/javascript/api/RulesBundle.java
```diff
@@ -32,6 +32,12 @@ public interface RulesBundle {
    */
   String bundlePath();
 
+  default String bundleId() {
+    return "n/a";
```
vilchik-elena, line 36: use `null`

javascript-checks/src/main/java/org/sonar/plugins/javascript/api/RulesBundle.java
```diff
@@ -32,6 +32,12 @@ public interface RulesBundle {
    */
   String bundlePath();
 
+  default String bundleId() {
```
vilchik-elena, line 35: `key`

javascript-checks/src/main/java/org/sonar/plugins/javascript/api/RulesBundle.java
```diff
@@ -32,6 +32,12 @@ public interface RulesBundle {
    */
   String bundlePath();
 
+  default String bundleKey() {
+    return null;
```
vilchik-elena, line 36: this line is shown as uncovered, while there is RulesBundlesTest which hits it

saberduck (approved):

PR #3523, opened 2022-11-02
Rename test case branch to match new SonarQube default branch naming

francoismora:
Rename the test case branch to match the new SonarQube default branch naming.

### DIFF
--- a/its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/PRAnalysisTest.java
+++ b/its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/PRAnalysisTest.java
@@ -287,7 +287,7 @@ class PRAnalysisTest {
   }
 
   static class Master {
-    static final String BRANCH = "master";
+    static final String BRANCH = "main";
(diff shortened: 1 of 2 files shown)


yassin-kammoun-sonarsource (commented):
@francoismora  There are red jobs.

its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/PRAnalysisTest.java
```diff
@@ -287,7 +287,7 @@ public void close() {
   }
 
   static class Master {
```
yassin-kammoun-sonarsource, line 289: We could even rename that to `Main` since it's now git's new terminology, but that's just a suggestion.

yassin-kammoun-sonarsource (approved):

its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/PRAnalysisTest.java
francoismora: Renamed

PR #3499, opened 2022-10-25
Ignore bundled code

saberduck:
Fixes #3368

Bundled code is detected and filtered out automatically

PR #3527, opened 2022-11-02
Fix logging when bundled code is detected

saberduck:
Feature was introduced in #3499


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/filter/BundleAssessor.java
```diff
@@ -56,13 +60,13 @@ public boolean test(InputFile inputFile) {
   }
 
   boolean triggered() {
```
ilia-kebets-sonarsource, line 62: You can remove this function now that it is only checked internally.

ilia-kebets-sonarsource (changes requested):
A few minor changes proposed. Let me know if you have any questions.

ilia-kebets-sonarsource (approved):
LGTM. Good job!

PR #3532, opened 2022-11-03
Fix FQN helper

victor-diez-sonarsource:
Look for variable definition in the proper scope, instead of always in the lower scope


francoismora (approved):
LGTM

PR #3544, opened 2022-11-08
Add `.vs/` to .gitignore

pavel-mikula-sonarsource:
I'll try to update some YML templates.

This will make my life a bit easier


ilia-kebets-sonarsource (approved):
ACK, thanks!

PR #3551, opened 2022-11-09
Support analysis when extended tsconfig is not available

### LINKED ISSUE #3548: Use empty tsconfig when extended tsconfig cannot be loaded
https://docs.google.com/document/d/1MiXKEjCzymqdmSolAB-_QdD1JSJlIUHZhEQJQdrEoPk/edit

victor-diez-sonarsource:
Fixes #3548


eslint-bridge/tests/services/program/program.test.ts
```diff
@@ -40,22 +45,49 @@ describe('program', () => {
...
+
+    const config = createProgram(tsConfig);
+    const { programId, files, projectReferences } = config;
+
+    expect(programId).toBeDefined();
+    expect(files).toEqual(expect.arrayContaining([toUnixPath(path.join(fixtures, 'file.ts'))]));
+    expect(projectReferences).toEqual([]);
+  });
+
+  it('missing external tsconfig should be different than found external tsconfig', () => {
```
vilchik-elena, line 72: that's a weird test. Why one tsconfig should be the same as another one?
I would have 2 tests asserting that for missing tsconfig we have empty options and `missingTsconfig` equal true. And for found one I would assert the values (not just that they are defined)

its/plugin/projects/external-tsconfig-dependency-project/package-lock.json
```diff
@@ -0,0 +1,46 @@
+{
```
vilchik-elena, line 1: do we need this file?

its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/ExternalTSConfigDependencyTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+  @Test
+  void test() throws Exception {
+    SonarScanner build = getSonarScanner()
+      .setProjectKey(PROJECT)
+      .setSourceEncoding("UTF-8")
+      .setSourceDirs(".")
+      .setProjectDir(PROJECT_DIR)
+      // setting inclusions like this will exclude tsconfig.json files, which is what we want to test
+      .setProperty("sonar.inclusions", "**/*.ts");
```
vilchik-elena, line 51: I don't think we need this one

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServer.java
```diff
@@ -213,18 +213,29 @@ class TsProgram {
...
+
+    TsProgram(String programId, List<String> files, List<String> projectReferences, boolean missingTSConfig) {
+      this.programId = programId;
+      this.files = files;
+      this.projectReferences = projectReferences;
+      this.error = null;
+      this.missingTSConfig = missingTSConfig;
     }
 
     TsProgram(String error) {
```
vilchik-elena, line 234: you can call another constructor here

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServer.java
```diff
@@ -213,18 +213,29 @@ class TsProgram {
     final List<String> files;
     final List<String> projectReferences;
     final String error;
+    final boolean missingTSConfig;
 
     TsProgram(String programId, List<String> files, List<String> projectReferences) {
```
vilchik-elena, line 218: you can call another constructor here

sonar-javascript-plugin/src/test/resources/external-tsconfig/tsconfig.json
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": "@tsconfig/node14/tsconfig.json",
```
vilchik-elena, line 2: why do we need all these files if you mock the response?

eslint-bridge/src/services/program/program.ts
```diff
@@ -69,37 +69,28 @@ export function getProgramById(programId: string): ts.Program {
...
+      }
+      return true;
+    },
+    readFile: file => {
+      const fileContents = ts.sys.readFile(file);
+      if (file.endsWith('package.json') || fileContents) {
+        return fileContents;
+      }
+      missingTSConfig = true;
+      console.log(`WARN Could not find extended tsconfig: ${file}`);
```
vilchik-elena, line 91:
```suggestion
      console.log(`WARN Could not find extended tsconfig: ${file}; falling back to an empty configuration.`);
```

vilchik-elena (changes requested):

eslint-bridge/tests/services/program/program.test.ts
```diff
@@ -40,22 +45,49 @@ describe('program', () => {
     expect(projectReferences).toEqual([toUnixPath(reference)]);
   });
 
-  it('should fail creating a program with a syntactically incorrect tsconfig', async () => {
+  it('should fail creating a program with a syntactically incorrect tsconfig', () => {
     const tsConfig = path.join(__dirname, 'fixtures', 'tsconfig.syntax.json');
-    const error = await createProgram(tsConfig).catch(err => err);
-    expect(error).toBeInstanceOf(Error);
+    expect(() => createProgram(tsConfig)).toThrow();
```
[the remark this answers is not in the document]
victor-diez-sonarsource: because now createProgram does not return a Promise, we can just check it throws, instead of catching promise rejection

sonar-javascript-plugin/src/test/resources/external-tsconfig/tsconfig.json
victor-diez-sonarsource: I want to check it does not fail as it would before the change, when extended tsconfig which did not exist returned an error

eslint-bridge/tests/services/program/program.test.ts
victor-diez-sonarsource: because when empty config, typescript does not return empty options, just fills in some defaults. I just checked some instead of the whole list which may change as it depends on internal Typescript implementation. Also, some returned config does not  match fully what is in the tsconfig.json

sonar-javascript-plugin/src/test/resources/external-tsconfig/tsconfig.json
vilchik-elena: Ok, but at least we can remove package.json and package-lock

eslint-bridge/tests/services/program/program.test.ts
vilchik-elena: then assert that options are not equal
Also why not asserting values of `configMissing` and `configFound`

eslint-bridge/tests/services/program/program.test.ts
```diff
@@ -64,7 +135,7 @@ describe('program', () => {
       files.map(toUnixPath).filter(file => file.startsWith(toUnixPath(fixtures))),
     );
     expect(program.getProjectReferences().map(reference => reference.path)).toEqual(
-      projectReferences.map(toUnixPath),
+      projectReferences.map(p => p),
```
vilchik-elena, line 138: why is it noop now?

eslint-bridge/tests/services/program/program.test.ts
```diff
@@ -52,6 +59,70 @@ describe('program', () => {
     expect(error.message).toMatch(/^Unknown compiler option 'targetSomething'./);
   });
 
+  it('should still create a program when extended tsconfig does not exist', async () => {
+    const fixtures = path.join(__dirname, 'fixtures');
+    const tsConfig = path.join(fixtures, `tsconfig_missing.json`);
```
vilchik-elena, line 64: here and further you use backticks for now reason

eslint-bridge/tests/services/program/program.test.ts
```diff
@@ -52,6 +59,70 @@ describe('program', () => {
...
+    do {
+      searchFolder = path.join(nodeModulesFolder, 'node_modules', '@tsconfig', 'node_missing');
+      searchedFiles.push(path.join(searchFolder, 'tsconfig.json', 'package.json'));
+      searchedFiles.push(path.join(searchFolder, 'package.json'));
+      searchedFiles.push(path.join(searchFolder, 'tsconfig.json'));
+      searchedFiles.push(path.join(searchFolder, 'tsconfig.json', 'tsconfig.json'));
+      nodeModulesFolder = path.dirname(nodeModulesFolder);
+    } while (!isRootNodeModules(searchFolder));
+
+    expect(() => createProgramOptions(tsConfigMissing, configHost)).toThrow();
```
vilchik-elena, line 119: I don't like exposing host as parameter of `createProgramOptions` just to test here. I would copy the required logic from the main code to some test function.

eslint-bridge/src/services/program/program.ts
```diff
@@ -124,11 +123,38 @@ export async function createProgram(inputTSConfig: string): Promise<{
...
+ *          resolved files and project references
+ */
+export async function createProgram(tsConfig: string): Promise<{
+  programId: string;
+  files: string[];
+  projectReferences: string[];
+  missingTsConfig: boolean;
+}> {
+  if ((await fs.lstat(tsConfig)).isDirectory()) {
+    tsConfig = path.join(tsConfig, 'tsconfig.json');
```
vilchik-elena, line 154: we should not overwrite parameter value

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/TypeScriptSensorTest.java
```diff
@@ -372,6 +372,32 @@ void should_analyze_by_tsconfig() throws Exception {
     verify(eslintBridgeServerMock, times(3)).newTsConfig();
   }
 
+  @Test
+  void should_analyze_by_program_on_missing_extended_tsconfig() throws Exception {
```
vilchik-elena, line 376: can we assert that an analysis warning is created?

vilchik-elena (changes requested):

vilchik-elena (approved):

PR #3559, opened 2022-11-15
Avoid scanning twice tsconfig.json project references

### LINKED ISSUE #3556: Avoid scanning twice `tsconfig.json` project references
tsconfig returned references should include `tsconfig.json` paths to avoid duplicating the paths (and thus scanning twice) on Java workList

victor-diez-sonarsource:
Fixes #3556


eslint-bridge/src/services/program/program.ts
```diff
@@ -149,22 +148,36 @@ export async function createProgram(inputTsConfig: string): Promise<{
...
   const program = ts.createProgram(programOptions);
-  const maybeProjectReferences = program.getProjectReferences();
-  const projectReferences = maybeProjectReferences ? maybeProjectReferences.map(p => p.path) : [];
+  const projectReferences = program.getProjectReferences() || [];
+  const sanitizedProjectReferences: string[] = [];
+
+  for (const reference of projectReferences) {
+    const sanitizedReference = await addTsConfigIfMissing(reference.path);
+    if (!sanitizedReference) {
+      console.log(`WARN Skipping missing referenced tsconfig: ${reference.path}`);
```
vilchik-elena, line 164: I suggest in the logs we either use `tsconfig.json` or `TSConfig` and not just `tsconfig`

eslint-bridge/tests/services/program/program.test.ts
```diff
@@ -44,7 +44,19 @@ describe('program', () => {
         toUnixPath(path.join(reference, 'file.ts')),
       ]),
     );
-    expect(projectReferences).toEqual([toUnixPath(reference)]);
+    expect(projectReferences).toEqual([await addTsConfigIfMissing(reference)]);
```
vilchik-elena, line 47: I prefer avoid using `addTsConfigIfMissing` and use `path.join` instead to be explicit about the expectation

eslint-bridge/tests/services/program/program.test.ts
```diff
@@ -44,7 +44,19 @@ describe('program', () => {
...
+  });
+
+  it('should skip missing reference a program', async () => {
```
vilchik-elena, line 50: not sure about the grammar of this phrase :)

eslint-bridge/src/services/program/program.ts
```diff
@@ -149,22 +148,36 @@ export async function createProgram(inputTsConfig: string): Promise<{
   projectReferences: string[];
   missingTsConfig: boolean;
```
vilchik-elena, line 149: as now indeed it's ambiguous why we don't set this property to true for a missing reference, I suggest to add jsdoc to this function explaining that it's true when extended tsconfig is missing and we fallback to a default config.

eslint-bridge/tests/services/program/program.test.ts
```diff
@@ -151,9 +163,13 @@ describe('program', () => {
     expect(program.getRootFileNames()).toEqual(
       files.map(toUnixPath).filter(file => file.startsWith(toUnixPath(fixtures))),
     );
-    expect(program.getProjectReferences().map(reference => reference.path)).toEqual(
-      projectReferences,
-    );
+    expect(
```
vilchik-elena, line 166: I wonder why we even care about this assertion, I think we never rely on `program.getProjectReferences()`.
Even if we keep it, is it possible to simplify asyncs/awaits?

vilchik-elena (changes requested):

eslint-bridge/src/helpers/files.ts
```diff
@@ -74,3 +75,20 @@ export function stripBOM(str: string) {
...
+ */
+export async function addTsConfigIfMissing(tsConfig: string) {
+  try {
+    if ((await fs.lstat(tsConfig)).isDirectory()) {
+      tsConfig = path.join(tsConfig, 'tsconfig.json');
+    }
+
+    return tsConfig;
+  } catch {
+    return null;
```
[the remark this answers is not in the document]
victor-diez-sonarsource: yes, we can try catch here or on the function call. if not here, we will need to handle promise rejections, or try...catchs. What do you prefer?

eslint-bridge/src/services/program/program.ts
```diff
@@ -149,22 +148,36 @@ export async function createProgram(inputTsConfig: string): Promise<{
...
   const programOptions = createProgramOptions(tsConfig);
 
   const program = ts.createProgram(programOptions);
-  const maybeProjectReferences = program.getProjectReferences();
-  const projectReferences = maybeProjectReferences ? maybeProjectReferences.map(p => p.path) : [];
+  const projectReferences = program.getProjectReferences() || [];
+  const sanitizedProjectReferences: string[] = [];
+
+  for (const reference of projectReferences) {
+    const sanitizedReference = await addTsConfigIfMissing(reference.path);
```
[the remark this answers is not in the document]
victor-diez-sonarsource: agreed, we had this conversation, remember? :P

its/plugin/projects/referenced-tsconfigs/dir/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "files": ["file.ts"],
+  "references": [
+    {
+      "path": "..\\tsconfig.json"
```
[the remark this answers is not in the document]
victor-diez-sonarsource: checking windows paths in tsconfig.json. seems they are supported, so I wanted to test proper behaviour

eslint-bridge/tests/services/program/program.test.ts
victor-diez-sonarsource: ok, remove it. references are already tested in other tests

vilchik-elena (approved):

PR #3561, opened 2022-11-16
Traverse tsconfigs paths from child to parent

victor-diez-sonarsource:
Also, normalize tsconfig paths (slash/backslash) to avoid scanning paths more than once

Fixes #


vilchik-elena:
Closing this PR as it was an experimental one

PR #3564, opened 2022-11-16
Improve performance of S1854 `no-dead-store`

saberduck:
There are two main changes to the rule implementation

- avoid creating intermediate arrays
- limit recursion when looking up variables in the scope

It improves ± 30%, on some projects even ± 50%


ilia-kebets-sonarsource (approved):
LGTM. Did the cost of performance appear in the profiling results for the lva.ts methods?

saberduck:
> LGTM. Did the cost of performance appear in the profiling results for the lva.ts methods?

partially yes, but it was far from obvious to understand the profiling info

PR #3581, opened 2022-11-24
Fix dependency injection of JavaScriptProjectChecker for SonarCloud

francoismora:
SonarCloud is failing to load the SonarJS plugin because the JavaScript sensor depends on an interface that only has an implementation in SonarLint. Currently, the sensor receives null for the missing implementation class in our integration tests. But on SonarCloud, it raises an exception. The solution here is to provide a second constructor to the JavaScript sensor without this dependency. SonarCloud doesn't throw an exception if there's a constructor that it can invoke.

### DIFF
--- a/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/JavaScriptEslintBasedSensor.java
+++ b/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/JavaScriptEslintBasedSensor.java
@@ -53,6 +53,9 @@ public class JavaScriptEslintBasedSensor extends AbstractEslintSensor {
   private final JavaScriptProjectChecker javaScriptProjectChecker;
   private AnalysisMode analysisMode;
 
+  // This constructor is required to avoid an error in SonarCloud because there's no implementation available for the interface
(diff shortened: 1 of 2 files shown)


yassin-kammoun-sonarsource (changes requested):
The changes are pretty simple, but to be honest, I don't understand what they fix and why they fix it :)

Therefore, I suggest we do the following:

1. Let's document in the PR description comment the original problem (we can share the stack trace provided by SonarCloud), give a bit of context, and briefly explain its root cause.
2. Let's rename the PR title accordingly. The title should describe the problem fixed by the changes rather than what they are (not the best suggestion but maybe`Fix dependency injection of JavaScriptProjectChecker`).
3. If it is doable, let's add a unit test that reflects the initial problem and that validates the changes. In other words, the test should fail if we remove the constructor introduced by this PR.

francoismora:
I was thinking adding the test after this PR as the release of SonarCloud is blocked for the moment. WDYT?

yassin-kammoun-sonarsource:
> I was thinking adding the test after this PR as the release of SonarCloud is blocked for the moment. WDYT?

Since we will need to do a bugfix release, we'll still need this test before the release. So I think it should be part of this PR.

PR #3588, opened 2022-11-25
Fix QA due to changes introduced in SONARSEC-3437

saberduck:
* use `DEV` alias for sonar-security to discover such issues earlier
* adjust the asserts to use new `.ucfgs` files instead of `.ucfg`


francoismora (approved):
LGTM

PR #3593, opened 2022-11-30
Add rule S6479 (`no-array-index-key`): No array index for keys in JSX list components

### LINKED ISSUE #3573: Add rule S6479 (`no-array-index-key`): No array index for keys in JSX list components
RSPEC: https://github.com/SonarSource/rspec/pull/1420

ilia-kebets-sonarsource:
Fixes #3573


javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S6479.json
```diff
@@ -0,0 +1,22 @@
...
+  ],
+  "defaultSeverity": "Major",
+  "ruleSpecification": "RSPEC-6479",
+  "sqKey": "S6479",
+  "scope": "All",
+  "quickfix": "unknown",
+  "compatibleLanguages": [
+    "JAVASCRIPT",
+    "TYPESCRIPT"
+  ]
```
francoismora, line 21: Not in sync with the RSPEC PR but I guess it can wait for the release when metadata will be updated.

francoismora (approved):
If you have time, you could add a comment-based test, as it shows the message raised and its location.

ilia-kebets-sonarsource:
> If you have time, you could add a comment-based test, as it shows the message raised and its location.

As discussed, since the rule is loaded from the eslint-plugin-react as-is, the manual verification of the error message and location is enough.

PR #3598, opened 2022-12-01
Support JS in HTML

### LINKED ISSUE #3597: Support JS in HTML
https://sonarsource.atlassian.net/jira/software/c/projects/MMF/issues/MMF-2427

victor-diez-sonarsource:
Fixes #3597


victor-diez-sonarsource:
We tested both `eslint-html-parser` and `eslint-plugin-html`.

We managed to make `eslint-plugin-html`, but it required forking it, and the ESLint setup was too specific and using non-exposed methods to override certain behaviours.

`eslint-html-parser` created an AST which was accepted to create a SourceCode object, but then that instance did not seem usable in the Linter.

Instead of depending on already existing ESLint plugins, we chose to follow the same path as the YAML parser, this is using an existing HTML parser to extract the js code ourselves and treat it like standalone js.

PR #3599, opened 2022-12-02
Fix tsconfig.json generation to reuse file only if it exists

francoismora:


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/tsconfig/DefaultTsConfigProvider.java
```diff
@@ -38,6 +40,10 @@ private static String getBaseDir(SensorContext context) {
     return "/".equals(File.separator) ? projectBaseDir : projectBaseDir.replace(File.separator, "/");
   }
 
+  private static boolean isValidTsConfigFile(List<String> tsconfigs) {
```
saberduck, line 43: I find it really confusing that these utility methods are above the fields in the class, I prefer that we follow standard convention that `static` fields are first, then instance fields, constructor, methods

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/tsconfig/DefaultTsConfigProvider.java
```diff
@@ -38,6 +40,10 @@ private static String getBaseDir(SensorContext context) {
...
+    return tsconfigs.size() == 1 && tsconfigs.get(0) != null && Files.exists(Path.of(tsconfigs.get(0)));
+  }
+
   private static final Logger LOG = Loggers.get(DefaultTsConfigProvider.class);
 
   private static final Map<String, List<String>> defaultWildcardTsConfig = new ConcurrentHashMap<>();
```
saberduck, line 49: we should not use concurrent data structures if there is no concurrent access

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/TsConfigProviderTest.java
```diff
@@ -219,22 +221,35 @@ void should_create_wildcard_tsconfig() throws Exception {
   }
 
   @Test
-  void should_not_recreate_wildcart_tsconfig_in_sonarlint() throws Exception {
-    List<String> tsconfigs;
-    Path file;
+  void should_not_recreate_wildcart_tsconfig_in_sonarlint_if_not_necessary() throws IOException {
```
saberduck, line 224: typo - wildcard*

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/tsconfig/DefaultTsConfigProvider.java
```diff
@@ -38,6 +40,10 @@ private static String getBaseDir(SensorContext context) {
     return "/".equals(File.separator) ? projectBaseDir : projectBaseDir.replace(File.separator, "/");
```
saberduck, line 40: not related to this PR - I believe it would have been better to use `Path` instead of `String` whenever possible, we would avoid the need for such "normalization", and it's easier to understand signatures. (I also think it was a mistake when I introduced `List<String>` for tsconfigs). This is just a remark, you don't have to refactor it in this PR.

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/TsConfigProviderTest.java
```diff
@@ -219,22 +221,35 @@ void should_create_wildcard_tsconfig() throws Exception {
...
 
     var ctx = SensorContextTester.create(baseDir);
     ctx.setRuntime(SonarRuntimeImpl.forSonarLint(Version.create(4, 4)));
 
     var checker = mock(ProjectChecker.class);
     when(checker.isBeyondLimit()).thenReturn(false);
 
-    tsconfigs = TsConfigProvider.generateDefaultTsConfigFile(ctx, checker, TsConfigProviderTest::createTsConfigFile);
-    assertThat(tsconfigs).hasSize(1);
+    var tsConfigFileCreator = mock(TsConfigFileCreator.class);
```
saberduck, line 234: maybe we don't need to mock such simple interface and just implement it for the purpose of the test instead?

saberduck (approved):
some minor remarks

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/tsconfig/DefaultTsConfigProvider.java
francoismora: That's something we discussed with @yassin-kammoun-sonarsource, so I guess it would be useful to document it somewhere. Static methods are maybe a bit less obvious as it is sometimes useful to put them close to the static variables they handle.

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/tsconfig/DefaultTsConfigProvider.java
francoismora: As discussed, the guarantee that we are always in a single-threaded environment allows us to use a `HashMap` indeed.

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/tsconfig/DefaultTsConfigProvider.java
francoismora: Moved to the bottom as this is a private utility method.

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/TsConfigProviderTest.java
francoismora: Changed

Issue #3605, opened 2022-12-07
Fix FN caused by `node:` imports in several rules


egon-okerman-sonarsource:
## Description
Last year, Node added ["`node:` imports"](https://nodejs.org/api/esm.html#node-imports) as aliases for its built-in modules. The prefix prevents malicious NPM packages from overriding them in certain circumstances.

Currently, SonarJS rules only trigger for the regular built-in modules, not for the `node:` aliases of those modules.

## Detection Logic
For all of Node's built-in modules (a complete list is given [here](https://github.com/sindresorhus/builtin-modules/blob/main/builtin-modules.json)), `node:` imports should be treated the same as regular imports of these modules.

### Examples
This example is correctly detected currently:
```javascript
const https = require("https");

const req = https.request({
    secureProtocol: 'TLSv1_method', // Noncompliant (S4432): TLS1.0 is insecure
    hostname: 'google.com',
    port: 443,
    path: '/',
    method: 'GET'
}, (res) => {
    // Do something with output
});
```

The example below currently does not get detected:
```javascript
const https = require("node:https");

const req = https.request({
    secureProtocol: 'TLSv1_method', // Noncompliant (S4432): TLS1.0 is insecure
    hostname: 'google.com',
    port: 443,
    path: '/',
    method: 'GET'
}, (res) => {
    // Do something with output
});
```

## Affected Rules
I looked at the usages of [isCallToFQN()](https://github.com/SonarSource/SonarJS/blob/master/eslint-bridge/src/linting/eslint/rules/helpers/module.ts#L167) to see which rules are impacted. I found S2612, S4423, S4830, S5542, S5547 and S5667 to be impacted by this, but it is possible that more exist.

## RSPEC
No RSPEC changes are necessary at this time.

## References
Specification ticket: [APPSEC-260](https://sonarsource.atlassian.net/browse/APPSEC-260)
[Original Community topic](https://community.sonarsource.com/t/javascript-vulnerability-not-found-by-sonar-scanner/75886)

PR #3618, opened 2023-01-03
Fix FNs caused by `node:` imports in several rules using `getFullyQualifiedName()`

ilia-kebets-sonarsource:
Fixes https://github.com/SonarSource/SonarJS/issues/3605

### DIFF
--- a/eslint-bridge/tests/tools/sonar-runtime/rule/no-missing-sonar-runtime.ts
+++ b/eslint-bridge/tests/tools/sonar-runtime/rule/no-missing-sonar-runtime.ts
@@ -116,11 +116,8 @@ function getModuleNameOfImportedIdentifier(
     } else {
       maybeRequireCall = writeExpression;
     }
-    const fqn = getFullyQualifiedName(context, maybeRequireCall, []);
-    if (fqn === null) {
-      return null;
-    }
-    return fqn.split('.')[0];
+    const fqn = getFullyQualifiedName(context, maybeRequireCall);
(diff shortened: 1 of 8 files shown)


eslint-bridge/tests/tools/sonar-runtime/rule/no-missing-sonar-runtime.ts
```diff
@@ -116,7 +116,11 @@ function getModuleNameOfImportedIdentifier(
     } else {
       maybeRequireCall = writeExpression;
     }
-    return getModuleNameFromRequire(maybeRequireCall);
+    const fqn = getFullyQualifiedName(context, maybeRequireCall, []);
+    if (fqn === null) {
+      return null;
+    }
+    return fqn.split('.')[0];
```
francoismora, line 123: The function mixes `null` and `undefined` in its return value. It would be simpler to stick to `undefined` and this change would become:
```suggestion
    const fqn = getFullyQualifiedName(context, maybeRequireCall);
    return fqn?.split('.')[0];
```

francoismora (approved):
LGTM, just a small suggestion.

eslint-bridge/tests/tools/sonar-runtime/rule/no-missing-sonar-runtime.ts
ilia-kebets-sonarsource: Indeed, thanks for the suggestion!

PR #3622, opened 2023-01-04
Fix FP S6442 (`no-hook-setter-in-body`): Shouldn't raise when used conditionally

### LINKED ISSUE #3516: Fix FP in S6442 improve corner case and fix rule title
| Label | Description |
| --- | --- |
|Rule | S6442 "React's useState used directly in render function" |
| Reproduce | False positive on some edge cases where a derived state is intended. |

## Explanation

As[ pointed out](https://twitter.com/gigonake/status/1584901827165884418?s=20&t=GDxDIFW7PbZQCUXEM7-I7A) by a user, there are some [corner cases](https://beta.reactjs.org/apis/react/useState#storing-information-from-previous-renders) where you might legitimately want to call the state setter in the render, although inside a condition to avoid the infinite loop.

React discourages this pattern ("This pattern can be hard to understand and is usually best avoided"), but I guess it is a false positive nonetheless.

## Suggestion

I suggest we ignore cases where the state setter is inside a conditional.
For example:

```js

    if (enjoySharpKnives) {
        setLanguage(navigator.language);
    }
```

Perhaps we can go all the way and require an explicit return statement since this is what React is fundamentally doing internally.

```js
    if (enjoySharpKnives) {
        setLanguage(navigator.language);
        return null;
    }
```

## Corrections

The title of the rule says the opposite of what we want:
```diff
- React's useState hook should only be used in the render function or body of a component
+ React's useState hook should not be used directly in the render function or body of a component
```

While we are there, there are also extra semicolons in the code examples:
```diff
function ShowLanguage() {
    ...
-};
+}
```

saberduck:
Fixes #3516

### DIFF
--- a/eslint-bridge/src/linting/eslint/rules/no-hook-setter-in-body.ts
+++ b/eslint-bridge/src/linting/eslint/rules/no-hook-setter-in-body.ts
@@ -82,7 +82,7 @@ export const rule: Rule.RuleModule = {
       return scope !== null && searchUpperFunctionScope(context.getScope()) === scope;
     }
 
-    function isConditional(node: estree.Node): boolean {
+    function isInsideConditional(node: estree.Node): boolean {
       return (
         findFirstMatchingLocalAncestor(node as TSESTree.Node, n => n.type === 'IfStatement') !==
         undefined
@@ -129,7 +129,7 @@ export const rule: Rule.RuleModule = {
         if (
           !isInsideFunctionScope(reactComponentScope) ||
           setters.length === 0 ||
-          isConditional(node)
+          isInsideConditional(node)
         ) {
           return;
         }
--- a/eslint-bridge/src/linting/eslint/rules/no-hook-setter-in-body.ts
+++ b/eslint-bridge/src/linting/eslint/rules/no-hook-setter-in-body.ts
@@ -24,11 +24,13 @@ import Variable = ESLintScope.Variable;
 import Scope = ESLintScope.Scope;
 import * as estree from 'estree';
 import {
+  findFirstMatchingLocalAncestor,
   getFullyQualifiedName,
   getVariableFromName,
   isFunctionNode,
   isIdentifier,
 } from './helpers';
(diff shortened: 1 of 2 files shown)


eslint-bridge/src/linting/eslint/rules/no-hook-setter-in-body.ts
```diff
@@ -80,6 +82,13 @@ export const rule: Rule.RuleModule = {
       return scope !== null && searchUpperFunctionScope(context.getScope()) === scope;
     }
 
+    function isConditional(node: estree.Node): boolean {
```
ilia-kebets-sonarsource, line 85:
```suggestion
    function isInsideConditional(node: estree.Node): boolean {
```

eslint-bridge/src/linting/eslint/rules/no-hook-setter-in-body.ts
```diff
@@ -117,7 +126,11 @@ export const rule: Rule.RuleModule = {
 
       // Selector matching function calls like: setCount(1)
       'CallExpression[callee.type="Identifier"][arguments.length=1]'(node: estree.CallExpression) {
-        if (!isInsideFunctionScope(reactComponentScope) || setters.length === 0) {
+        if (
+          !isInsideFunctionScope(reactComponentScope) ||
+          setters.length === 0 ||
+          isConditional(node)
```
ilia-kebets-sonarsource, line 132:
```suggestion
          isInsideConditional(node)
```

ilia-kebets-sonarsource:
I proposed a small function refactor.

Should you also apply the RSPEC change that Gab mentioned in corrections in https://github.com/SonarSource/SonarJS/issues/3516?

ilia-kebets-sonarsource (approved):
Apart from my comments, LGTM.

I Fixed the PR title to reflect https://xtranet-sonarsource.atlassian.net/wiki/spaces/LANG/pages/2567241836/Commit+and+Tickets+Guidelines

PR #3577, opened 2022-11-23
Refactor handling of tsconfig.json files

francoismora:
Fixes #3568

PR #3626, opened 2023-01-06
Revert "Refactor handling of tsconfig.json files (#3577)"

### LINKED ISSUE #3616: Regression linked to using generated tsconfig with wildcard instead of list of files
Regression introduced in 0c6ce5500b344e411f2658d96bb691a75e4cfa9b

We fail to use typescript-eslint to parse some files and fall back to the babel parser without types available

In the analysis logs, we see something like this (forwarded from typescript-eslint)
```
13:54:50.808 INFO: The file does not match your project config: grid-packages/ag-grid-docs/documentation/doc-pages/component-cell-renderer/examples/get-cell-renderer/medalCellRenderer_reactFunctional.jsx.
13:54:50.808 INFO: The file must be included in at least one of the projects provided.
```

We still don't understand the nature of the problem, and we should probably simply revert the commit.

vilchik-elena:
This reverts commit 0c6ce5500b344e411f2658d96bb691a75e4cfa9b.

Fixes #3616


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/TsConfigProvider.java
```diff
@@ -0,0 +1,290 @@
...
+    }
+  }
+
+  static class WildcardTsConfigProvider extends GeneratedTsConfigFileProvider {
+    private static String getProjectRoot(SensorContext context) {
+      var projectBaseDir = context.fileSystem().baseDir().getAbsolutePath();
+      return "/".equals(File.separator) ? projectBaseDir : projectBaseDir.replace(File.separator, "/");
+    }
+
+    private static final Map<String, List<String>> defaultWildcardTsConfig = new ConcurrentHashMap<>();
```
francoismora, line 262: I know now that we don't need to worry about threads, so we can simply use `HashMap`.

francoismora (approved):
LGTM

PR #3635, opened 2023-01-10
Deprecate S2814 (`no-redeclare`) for TypeScript

### LINKED ISSUE #3580: S2814 might not be needed in TypeScript
| Label | Description |
| --- | --- |
| Rule | S2814 "Variables and functions should not be redeclared"  |
| Subject | FP on TypeScript `type` and `function` with same name |
| Community | https://community.sonarsource.com/t/typescript-file-produces-a-false-positive-already-defined-when-types-and-variable-share-same-name/76080 |

### Explanation
`type` and `function` have different name spaces in TypeScript, so we should not raise an issue for this case.
Notice that this also applies to variables, `namespace`, `interface`, and perhaps more.

See many FP in Peach (internal):
* https://peach.sonarsource.com/coding_rules?languages=ts&q=already&open=typescript%3AS2814

### Suggestion
The TypeScript compiler will already check for reused names and output an error.
So this rule does not make much sense in TypeScript, as compared to JavaScript.

Regardless of potential fixes to the FP, we should consider removing this rule for TypeScript.
Or at least remove it from the "Sonar Way" profile for TypeScript.

victor-diez-sonarsource:
Fixes #3580 

Note: we failed to add `Deprecated` section to the rule description

### DIFF
--- a/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/rules/TypeScriptRulesDefinition.java
+++ b/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/rules/TypeScriptRulesDefinition.java
@@ -21,6 +21,7 @@ package org.sonar.plugins.javascript.rules;
 
 import java.util.Collections;
 import org.sonar.api.SonarRuntime;
+import org.sonar.api.rule.RuleStatus;
 import org.sonar.api.server.rule.RulesDefinition;
 import org.sonar.javascript.checks.CheckList;
 import org.sonar.plugins.javascript.JavaScriptProfilesDefinition;
@@ -49,7 +50,10 @@ public class TypeScriptRulesDefinition implements RulesDefinition {
     NewRule commentRegularExpression = repository.rule("S124");
     commentRegularExpression.setTemplate(true);
 
+    // deprecated for Typescript: https://github.com/SonarSource/SonarJS/issues/3580
+    NewRule redeclaredVariables = repository.rule("S2814");
+    redeclaredVariables.setStatus(RuleStatus.DEPRECATED);
+
     repository.done();
   }
-
 }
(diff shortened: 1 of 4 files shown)


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/rules/TypeScriptRulesDefinition.java
```diff
@@ -49,7 +50,10 @@ public void define(Context context) {
     NewRule commentRegularExpression = repository.rule("S124");
     commentRegularExpression.setTemplate(true);
 
+    // deprecated for Typescript: https://github.com/SonarSource/SonarJS/issues/3580
+    NewRule redeclaredVariables = repository.rule("S2814");
```
vilchik-elena, line 54: for the variable name use eslint key or java class name

sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/JavaScriptProfilesDefinitionTest.java
```diff
@@ -95,6 +95,7 @@ void sonar_way_ts() {
     assertThat(profile.rules()).extracting("repoKey").containsOnly(CheckList.TS_REPOSITORY_KEY);
     assertThat(profile.rules().size()).isGreaterThan(100);
     assertThat(profile.rules()).extracting(BuiltInQualityProfilesDefinition.BuiltInActiveRule::ruleKey).contains("S5122");
+    assertThat(profile.rules()).extracting(BuiltInQualityProfilesDefinition.BuiltInActiveRule::ruleKey).doesNotContain("S2814");
```
vilchik-elena, line 98: Assert that JS profile has it

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/rules/TypeScriptRulesDefinition.java
```diff
@@ -49,7 +50,10 @@ public void define(Context context) {
...
+    redeclaredVariables.setStatus(RuleStatus.DEPRECATED);
```
vilchik-elena, line 55: can we unit test it?

vilchik-elena (changes requested):
And won't fix the issue in Next

vilchik-elena (approved):

PR #3638, opened 2023-01-10
Use more efficient encoding for CPD tokens

### LINKED ISSUE #3633: Use more efficient encoding for CPD tokens
The javascript analyzer serializes the CPD tokens in the cache using JSON, which uses too much cache memory. Other plugins like PHP use a more efficient serialization. The javascript analyzer should use it, to reduce the memory footprint of CPD tokens saved in the cache.

francoismora:
Fixes #3633

### DIFF
--- a/sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/JavaScriptEslintBasedSensorTest.java
+++ b/sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/JavaScriptEslintBasedSensorTest.java
@@ -70,13 +70,11 @@ import org.sonar.plugins.javascript.TestUtils;
 import org.sonar.plugins.javascript.eslint.EslintBridgeServer.AnalysisResponse;
 import org.sonar.plugins.javascript.eslint.EslintBridgeServer.JsAnalysisRequest;
 import org.sonar.plugins.javascript.eslint.cache.CacheTestUtils;
-import org.sonar.plugins.javascript.eslint.tsconfig.TsConfigFile;
 import org.sonarsource.nodejs.NodeCommandException;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.Mockito.clearInvocations;
 import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.mock;


sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/cache/CacheAnalysisSerialization.java
```diff
@@ -20,29 +20,32 @@
...
+public class CacheAnalysisSerialization extends CacheSerialization {
 
   private final UCFGFilesSerialization ucfgFileSerialization;
-  private final JsonSerialization<CpdData> cpdDataSerialization;
+  private final CpdSerialization cpdSerialization;
   private final JsonSerialization<FileMetadata> fileMetadataSerialization;
+  private final String pluginVersion;
 
-  CacheAnalysisSerialization(SensorContext context, CacheKey cacheKey) {
+  CacheAnalysisSerialization(SensorContext context, CacheKey cacheKey, @Nullable String pluginVersion) {
```
saberduck, line 38: I find it inconsistent that for UCFGs, we access `PluginInfo` statically from CacheKey (and it's part of the cachekey) and for CPD we are passing it as an argument and we do validation when reading data (it's not part of the key). Approach should be the same

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/cache/CacheStrategies.java
```diff
@@ -57,15 +57,15 @@ static String getLogMessage(CacheStrategy cacheStrategy, InputFile inputFile, @N
     return logBuilder.toString();
   }
 
-  public static CacheStrategy getStrategyFor(SensorContext context, InputFile inputFile) throws IOException {
+  public static CacheStrategy getStrategyFor(SensorContext context, InputFile inputFile, @Nullable String pluginVersion) throws IOException {
```
saberduck, line 60: it's unclear why should strategy depend on `pluginVersion`

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/cache/CpdData.java
```diff
@@ -20,18 +20,30 @@
 package org.sonar.plugins.javascript.eslint.cache;
 
 import java.util.List;
+import java.util.Objects;
+import javax.annotation.Nullable;
 import org.sonar.plugins.javascript.eslint.EslintBridgeServer;
 
-class CpdData {
+public class CpdData {
```
saberduck, line 27: this class shouldn't be public, we can just have `public static` utility class in the test package

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/cache/CpdData.java
```diff
@@ -20,18 +20,30 @@
...
 
   private final List<EslintBridgeServer.CpdToken> cpdTokens;
+  private final String pluginVersion;
 
-  CpdData(List<EslintBridgeServer.CpdToken> cpdTokens) {
+  public CpdData(List<EslintBridgeServer.CpdToken> cpdTokens, @Nullable String pluginVersion) {
```
saberduck, line 32: we will not need `pluginVersion` if we make it part of the key

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/cache/SerializationResult.java
```diff
@@ -0,0 +1,40 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.plugins.javascript.eslint.cache;
+
+public class SerializationResult {
```
saberduck, line 22: imo this could be inner static class for `CpdSerializer`, it's not needed elsewhere. The name should somehow reflect that it's about cpd, maybe `CpdBinaryFormat`

sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/cache/StringTable.java
```diff
@@ -0,0 +1,62 @@
...
+package org.sonar.plugins.javascript.eslint.cache;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import javax.annotation.Nullable;
+
+public class StringTable {
```
saberduck, line 29: can this be package private?

saberduck (changes requested):
I think we should put pluginversion as part of the key for cpd data, so it works the same way as with ucfgs. Otherwise there are just some minor comments

PR #3642, opened 2023-01-11
Fix recursion in FQN resolution

### LINKED ISSUE #3641: Fix FQN resolution
FQN resolution can exceed the call stack when resolving member/call expressions which are assigned to the same variable, ie:

```js
var1 = var1.trim();
```

We need to keep track of variables which have been already checked.

victor-diez-sonarsource:
Fixes #3641


francoismora (approved):
LGTM

PR #3644, opened 2023-01-11
Update json5 to 2.2.3

saberduck:


saberduck:
This autoclosed after rebase because same changes where committed on master

PR #3645, opened 2023-01-11
Update eslint-plugin-sonarjs to 0.18.0

saberduck:


eslint-bridge/package-lock.json
```diff
@@ -14407,9 +14407,9 @@
       "requires": {}
     },
     "eslint-plugin-sonarjs": {
```
vilchik-elena, line 14409: do you know what we have it twice?

saberduck: it's the internal structure of lock-file, all dependencies are like this. I don't know the reason behind it

PR #3651, opened 2023-01-12
Improve test coverage for file hash in cache

saberduck:
I am still not 100% happy with it because of extended usage of mocks, we would need to refactor the code to make it more testable


francoismora (approved):
LGTM

PR #3675, opened 2023-01-24
Move nodejs-utils under sonar-javascript-plugin

saberduck:
This is the first preparatory step in the larger effort to refactor the plugin structure. In the past `nodejs-utils` was a small library shared between `sonar-css` and `sonar-javascript`, however since CSS was moved under SonarJS, this maven module / library is not needed anymore.

Reorganization of java packages will follow later.


ilia-kebets-sonarsource (approved):
LGTM!

PR #3679, opened 2023-01-30
Fix FP S128 (sonar-max-lines-per-function): Improve React functional components detection

### LINKED ISSUE #3678: Fix FP S128 (`sonar-max-lines-per-function`): Improve React functional components detection
FP reported in [community](https://community.sonarsource.com/t/react-functional-component-being-picked-up-as-a-function/81679
).

The JSX returns should not only detect JSXElement, but any kind of JSX (JSXFragment is not considered). The rule `function-name` considers any return node whose type starts with `JSX`, so we should probably use same logic.

Also, only `FunctionDeclaration` was considered, probably we should also add `FunctionExpression`.

victor-diez-sonarsource:
Fixes #3678


yassin-kammoun-sonarsource (approved):
LGTM!

PR #3490, opened 2022-10-19
 Add rule S6330: Using unencrypted SQS queues is security-sensitive 

francoismora:
Fixes #3434

PR #3682, opened 2023-02-02
Update eslint dependencies to latest version

### LINKED ISSUE #3681: Update eslint dependencies to latest version
The current version of eslint used is 8.26.0 and version 8.33.0 is available. In version 8.27.0 the PR [#16458](https://github.com/eslint/eslint/pull/16458) updated some core rules to use the new flat configuration and this is causing regressions on our side when they were decorated (like `comma-dangle`). The decoration logic should be changed to support this new behavior.

francoismora:
The function `sanitizeTypeScriptESLintRule()` causes an error when the decorated eslint rule uses the new flat configuration properties because it replaces the context with an intercepting context that is missing the new properties. However, the intercepting context is not changing the semantic of the eslint context. Therefore, it can be removed and this fixes the problem.

Also, the `estlint-plugin-react` plugin brings changes that improve the ruling:
1. Improve false-negative of 'jsx-key' S6477 with ([#3490](https://github.com/jsx-eslint/eslint-plugin-react/pull/3490))
2. Fix false-positive of 'jsx-no-constructed-context-values' S6481 when the issue is not in a React component.

Fixes #3681


eslint-bridge/src/linting/eslint/linter/decoration/sanitize.ts
```diff
@@ -41,62 +41,18 @@ export function sanitizeTypeScriptESLintRule(rule: Rule.RuleModule): Rule.RuleMo
...
-      };
       /**
        * Overrides the rule behaviour if it requires TypeScript's type checker
        * but type information is missing.
        */
       if (
         rule.meta?.docs &&
         (rule.meta.docs as any).requiresTypeChecking === true &&
-        interceptingContext.parserServices.hasFullTypeInformation !== true
+        originalContext.parserServices.hasFullTypeInformation !== true
```
yassin-kammoun-sonarsource, line 51: I think we can probably just rename `originalContext` to `context` since we are no longer creating an intercepting context.

yassin-kammoun-sonarsource (approved):

eslint-bridge/src/linting/eslint/linter/decoration/sanitize.ts
francoismora: Makes sense indeed. Thanks.

PR #3685, opened 2023-02-06
Fix FP S5860 (`unused-named-groups`): Consider accesses via the index syntax

### LINKED ISSUE #3683: Fix FP S5860 (`unused-named-groups`): Consider accesses via the index syntax
The rule currently raises false positives when the groups of a regular expression that matches an input are accessed via the index syntax:

```typescript
const date = '02/23';

const datePattern = /(?<month>\d{2})\/(?<year>\d{2})/;
const dateMatched = date.match(datePattern);

if (dateMatched !== null) {
  checkValidity(dateMatched.groups['month'], dateMatched.groups['year']);
}
```

Reported in https://community.sonarsource.com/t/typescript-s5860-not-considering-access-via-index-syntax/82684

yassin-kammoun-sonarsource:
Fixes #3683

### DIFF
--- a/eslint-bridge/src/linting/eslint/rules/helpers/ast.ts
+++ b/eslint-bridge/src/linting/eslint/rules/helpers/ast.ts
@@ -564,6 +564,12 @@ export function isDotNotation(
   return node.type === 'MemberExpression' && !node.computed && node.property.type === 'Identifier';
 }
 
+export function isIndexNotation(
+  node: estree.Node,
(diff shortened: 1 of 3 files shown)


victor-diez-sonarsource (changes requested):
Can you add a test for indices.groups['foo']? otherwise LGTM

victor-diez-sonarsource (approved):

PR #3687, opened 2023-02-07
Update S4423 (`weak-ssl`): Change issue message when omitting tlsSecurityPolicy

### LINKED ISSUE #3629: Modify rule S4423: Change issue message for noncompliant default cases (AWS CDK)
# What
Rule S4423 identifies weak TLS protocols being used in code. Implementation ticket https://github.com/SonarSource/SonarJS/issues/3453 specifies this for AWS CDK resources.

In most cases, when an issue is raised, the issue message is nice and clean:

>  Change this code to enforce TLS 1.2 or above.

However, when the issue is present in a piece of code that does not specify a security parameter and the default behavior is noncompliant, the message contains a duplicate.

> Omitting "tlsSecurityPolicy" enables a deprecated version of TLS. Set it to enforce TLS 1.2 or above. Change this code to enforce TLS 1.2 or above.

# Example
image

# How
In this case, the issue message should be:

> Omitting "tlsSecurityPolicy" enables a deprecated version of TLS. Set it to enforce TLS 1.2 or above.

The duplicated part "Change this code to enforce TLS 1.2 or above." should be removed.

yassin-kammoun-sonarsource:
Fixes #3629

### DIFF
--- a/eslint-bridge/src/linting/eslint/rules/weak-ssl.aws.ts
+++ b/eslint-bridge/src/linting/eslint/rules/weak-ssl.aws.ts
@@ -75,7 +75,7 @@ export const rule: Rule.RuleModule = AwsCdkTemplate(
         enforceTLS12: 'Change this code to enforce TLS 1.2 or above.',
         AWSApiGateway: 'Change this code to enforce TLS 1.2 or above.',
         AWSOpenElasticSearch:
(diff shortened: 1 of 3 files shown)


francoismora (approved):
LGTM

PR #3695, opened 2023-02-13
ESLint rules evaluation setup

ilia-kebets-sonarsource:
you also need to delete:
- `its/ruling/src/test/expected/js`
- `its/ruling/src/test/expected/ts`

The rule IDs were chosen among: [sonar way profile rules](https://github.com/SonarSource/SonarJS/blob/master/javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/Sonar_way_profile.json)


ilia-kebets-sonarsource:
closing as I'll keep the code reference here: https://github.com/ilia-kebets-sonarsource/SonarJS/pull/2

PR #3721, opened 2023-02-21
Fix FN S2138 (`no-undefined-assignment`): raise issue when assigning undefined to a property.

### LINKED ISSUE #3715: Fix FN S2138 (`no-undefined-assignment`): We don't flag when assigning value to an object's property
Reproducer:

```javascript
const a = { foo: undefined };
```

philnash:
Fixes #3715


ilia-kebets-sonarsource:
@philnash The implementation looks good. The [ruling tests](https://github.com/SonarSource/SonarJS/blob/master/docs/DEV.md#ruling-tests) are failing, you should have a look at them and update the expected issues. You can have a look the the docs I linked, but it might be a bit obscure. I can take care of it if it's unclear.

ilia-kebets-sonarsource (approved):
LGTM

PR #3738, opened 2023-02-28
Fix typo in 'sonar.javascript.globals' default value 'OpenLayers'

francoismora:


yassin-kammoun-sonarsource (approved):
Nice catch! And thanks for the test case!

PR #3741, opened 2023-03-01
Improve S2137 (`no-globals-shadowing`): Check shadowing of global builtins

### LINKED ISSUE #3708: Update S2137 (`no-globals-shadowing`): cover cases from ESLint's `no-global-assign`
https://eslint.org/docs/latest/rules/no-global-assign

francoismora:
Fixes #3708


src/linting/eslint/rules/no-globals-shadowing.ts
```diff
@@ -52,46 +53,66 @@ export const rule: Rule.RuleModule = {
       },
       VariableDeclaration(node: estree.Node) {
         (node as estree.VariableDeclaration).declarations.forEach(decl => {
-          reportBadUsage(decl.id, getDeclarationIssue('variable'), context);
+          reportBadUsage(decl.id, getDeclarationIssue('variable'), context, decl.init != null);
```
yassin-kammoun-sonarsource, line 56: Just to make sure I understand, we exclude the case of `var Number` because we assume that it will be assigned sometime later like `Number = 42`. This way, we avoid reporting twice on `Number`, right?

yassin-kammoun-sonarsource (approved):
LGTM (just left a question for clarification)

PR #3744, opened 2023-03-02
Improve S1472 (`call-argument-line`): Template tag should be on the same line as the literal

### LINKED ISSUE #3705: Update S1472 (`call-argument-line`): cover scope of ESLint's `no-unexpected-multiline`
https://eslint.org/docs/latest/rules/no-unexpected-multiline

francoismora:
Fixes #3705

### DIFF
--- a/src/linting/eslint/rules/call-argument-line.ts
+++ b/src/linting/eslint/rules/call-argument-line.ts
@@ -49,9 +49,9 @@ export const rule: Rule.RuleModule = {
             const { end } = sourceCode.getLastToken(call)!.loc;
             if (end.line !== start.line) {
               //If arguments span multiple lines, we only report the first one
-              reportIssue(start, calleeLastLine, context);
+              reportIssue('moveArguments', start, calleeLastLine, context);
             } else {
-              reportIssue({ start, end }, calleeLastLine, context);
+              reportIssue('moveArguments', { start, end }, calleeLastLine, context);
             }
           }
         }
@@ -67,7 +67,7 @@ export const rule: Rule.RuleModule = {
               column: quasi.loc.start.column + 1,
             },
           };
-          reportIssue(loc, tokenBefore.loc.start.line, context, 'moveTemplateLiteral');
+          reportIssue('moveTemplateLiteral', loc, tokenBefore.loc.start.line, context);
         }
       },
     };
@@ -83,10 +83,10 @@ function isNotClosingParen(token: AST.Token) {
 }
 
 function reportIssue(
+  messageId: string,
   loc: { start: Position; end: Position } | Position,
   line: number,
   context: Rule.RuleContext,
-  messageId = 'moveArguments',
 ) {
   context.report({
     messageId,
--- a/src/linting/eslint/rules/call-argument-line.ts
+++ b/src/linting/eslint/rules/call-argument-line.ts
@@ -19,29 +19,32 @@
  */
 // https://sonarsource.github.io/rspec/#/rspec/S1472/javascript
 
-import { Rule } from 'eslint';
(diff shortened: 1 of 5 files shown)


src/linting/eslint/rules/call-argument-line.ts
```diff
@@ -19,29 +19,32 @@
...
     return {
       CallExpression: (node: estree.Node) => {
         const call = node as estree.CallExpression;
         if (call.callee.type !== 'CallExpression' && call.arguments.length === 1) {
-          const sourceCode = context.getSourceCode();
           const parenthesis = sourceCode.getLastTokenBetween(
             call.callee,
             call.arguments[0],
-            token => token.type === 'Punctuator' && token.value === ')',
+            isClosingParen,
```
yassin-kammoun-sonarsource, line 43: Could you comment on this PR's comment why this change? We already discussed it, but let's record here the reason :)

src/linting/eslint/rules/call-argument-line.ts
```diff
@@ -53,17 +56,40 @@ export const rule: Rule.RuleModule = {
...
+function isNotClosingParen(token: AST.Token) {
+  return !isClosingParen(token);
+}
+
 function reportIssue(
-  loc: { start: estree.Position; end: estree.Position } | estree.Position,
+  loc: { start: Position; end: Position } | Position,
   line: number,
   context: Rule.RuleContext,
+  messageId = 'moveArguments',
```
yassin-kammoun-sonarsource, line 89: Just a suggestion, but let's remove this default value and be explicit about the message id when calling this function. Probably just nitpicking, but what do you think?

src/linting/eslint/rules/call-argument-line.ts
```diff
@@ -19,29 +19,32 @@
...
           const parenthesis = sourceCode.getLastTokenBetween(
             call.callee,
             call.arguments[0],
-            token => token.type === 'Punctuator' && token.value === ')',
+            isClosingParen,
           );
           const calleeLastLine = (parenthesis ? parenthesis : sourceCode.getLastToken(call.callee))!
             .loc.end.line;
-          const { start } = sourceCode.getTokenAfter(call.callee)!.loc;
+          const { start } = sourceCode.getTokenAfter(call.callee, isNotClosingParen)!.loc;
```
yassin-kammoun-sonarsource, line 47: There is a code smell here when using the exclamation mark. Let's resolve the code smell as "won't fix" since we know for a fact that there is a token after the function's callee.

yassin-kammoun-sonarsource (approved):
LGTM! Maybe we should wait for the weird issue on ruling to be fixed before merging. Your call...

src/linting/eslint/rules/call-argument-line.ts
yassin-kammoun-sonarsource: You can mention this here...

francoismora: To cover more cases, the logic determining the caller and argument lines has to use the same 'closing parent' token filter. The rule searches for closing parens to find the caller line and for anything except a closing paren to find the arguments line.
The rule now defines two new functions, `isClosingParen()` and `isNotClosingParen()` for that purpose.

src/linting/eslint/rules/call-argument-line.ts
francoismora: The default value's purpose was only to ease the reviewer's job by minimizing the changes. I will remove it.

francoismora:
> LGTM! Maybe we should wait for the weird issue on ruling to be fixed before merging. Your call...

Indeed, it's better to wait for the moment.

PR #3747, opened 2023-03-06
Remove node_modules from root of the project when running ruling locally

saberduck:
Fixes #


saberduck:
> LGTM, but isn't there a rmdir instead of looping through files? should be much faster I would say

unfortunately not. But it would have been the same wrt. speed.

PR #3748, opened 2023-03-06
Update S1534 (`sonar-no-dupe-keys`): Merge with ESLint `no-dupe-class-members`

### LINKED ISSUE #3699: Update S1534 (`sonar-no-dupe-keys`): Merge with ESLint `no-dupe-class-members`
https://eslint.org/docs/latest/rules/no-dupe-class-members

We want to support the cases from `no-dupe-class-members` in S1534

yassin-kammoun-sonarsource:
Fixes #3699


src/linting/eslint/rules/sonar-no-dupe-keys.ts
```diff
@@ -21,21 +21,31 @@
 
 import { Rule } from 'eslint';
 import { eslintRules } from 'linting/eslint/rules/core';
+import { rules as typescriptRules } from '@typescript-eslint/eslint-plugin';
```
yassin-kammoun-sonarsource, line 24: Note that we need to use explicitly TypeScript ESLint's implementation of the rule to reduce noise and avoid reporting duplicate members denoting method overloads (see https://typescript-eslint.io/rules/no-dupe-class-members).

francoismora (approved):
LGTM

PR #3749, opened 2023-03-07
Add rule S2639: Empty character classes should not be used

### LINKED ISSUE #3701: Add rule S2639: Empty character classes should not be used
Use https://eslint.org/docs/latest/rules/no-empty-character-class

yassin-kammoun-sonarsource:
Fixes #3701


francoismora (approved):
LGTM

PR #3758, opened 2023-03-10
JS analyzer in HTML should ignore non-javascript types

### LINKED ISSUE #3757: JS analyzer in HTML should ignore non-javascript types
Reported in community: https://community.sonarsource.com/t/jsx-error-appearing-in-angular-html-files/85243

We should ignore non-javascript types

[Reference](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type)
[Reference](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types#textjavascript)

victor-diez-sonarsource:
Fixes #3757


tests/parsing/embedded/html/parser/parse.test.ts
```diff
@@ -56,4 +56,11 @@ describe('parseHtml', () => {
     const embeddedJSs = parseHTML(fileContent);
     expect(embeddedJSs).toHaveLength(0);
   });
+
+  it('should ignore non js script tags', async () => {
```
ilia-kebets-sonarsource, line 60:
```suggestion
  it('should ignore non-js script tags', async () => {
```

ilia-kebets-sonarsource (approved):
Small text suggestion otherwise LGTM. Good job!

PR #3759, opened 2023-03-13
Update rule S4275 (no-accessor-field-mismatch): cover cases from ESLint's getter-return

francoismora:
Fixes #3702 
RSPEC: SonarSource/RSPEC#1643

### DIFF
--- a/tests/linting/eslint/rules/no-accessor-field-mismatch.test.ts
+++ b/tests/linting/eslint/rules/no-accessor-field-mismatch.test.ts
@@ -218,8 +218,47 @@ ruleTester.run('Getters and setters should access the expected fields', rule, {
           ...theRest
         };`,
     },
+    'Object.defineProperty()',
+    'Object.defineProperty(o)',
+    'Object.defineProperty(o, "b")',
+    'Object.defineProperty(o, "b", props)',
+    'Object.defineProperty(o, "b", { get() { return a; } })',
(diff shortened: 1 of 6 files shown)


tests/linting/eslint/rules/no-accessor-field-mismatch.test.ts
```diff
@@ -44,38 +44,54 @@ function invalid(code: string) {
...
+
 ruleTester.run('Getters and setters should access the expected fields', rule, {
   valid: [
     {
       code: `
         class OK {
           private x: string;
           private _y = "hello";
-        
+
```
yassin-kammoun-sonarsource, line 70: I am surprised to see these differences. Was it the pre-commit hook that did that?

yassin-kammoun-sonarsource (approved):
LGTM! Nice work! We might need to update the RSPEC description in case you make some changes in the corresponding RPSEC PR.

tests/linting/eslint/rules/no-accessor-field-mismatch.test.ts
francoismora: I think it comes from the IDE comment/uncomment operations I used when debugging unit tests.

yassin-kammoun-sonarsource: 👍

PR #3762, opened 2023-03-14
Add rule S6523 (`no-unsafe-optional-chaining`): No optional chaining where undefined is forbidden

### LINKED ISSUE #3698: Add rule for `no-unsafe-optional-chaining`
https://eslint.org/docs/latest/rules/no-unsafe-optional-chaining

- Not covered
- Makes sense?
  - Disallow usage of opt chaining where an undefined evaluation would break code
  - Seems to be useful
  - No issue raised on ruling though, as it’s a new language feature, so during implementation we should add more code co
- Scope takes care of lots of cases
- We should use the [disallowArithmeticOperators](https://eslint.org/docs/latest/rules/no-unsafe-optional-chaining#disallowarithmeticoperators) option
- Note that rule S2259 (currently having a non-efficient implementation) is a generic case, in the future we might deprecate the new rule in the favour of S2259 
- Decision: we add this as a NEW rule

ilia-kebets-sonarsource:
Fixes #3698


francoismora (approved):
LGTM

PR #3763, opened 2023-03-15
Fix FP S2137 (`no-globals-shadowing`): Exclude Object prototype properties and deprecated functions

### LINKED ISSUE #3708: Update S2137 (`no-globals-shadowing`): cover cases from ESLint's `no-global-assign`

francoismora:
Fixes #3708

This PR excludes:
1. Object prototype properties like `toString()` or `valueOf()`, inherited by the global object, because they are not helpful, and it's common practice to redefine functions with the same names to protect core API against redefinitions.
2. Deprecated functions like `escape()` or `unescape()` as they are not available in all environments


src/linting/eslint/rules/no-globals-shadowing.ts
```diff
@@ -143,3 +145,31 @@ function reportBadUsage(
     }
   }
 }
+
+function isBadNameUsage(name: string, isWrite: boolean) {
```
yassin-kammoun-sonarsource, line 149:
```suggestion
function isGlobalShadowing(name: string, isWrite: boolean) {
```

src/linting/eslint/rules/no-globals-shadowing.ts
```diff
@@ -143,3 +145,31 @@ function reportBadUsage(
...
+
+function isBuiltInName(name: string) {
+  return globalsByLibraries.builtin.includes(name);
+}
+
+function isBadUndefinedUsage(isWrite: boolean, name: string) {
+  return isWrite && name === 'undefined';
+}
+
+function isBadNameException(name: string) {
```
yassin-kammoun-sonarsource, line 165:
```suggestion
function isShadowingException(name: string) {
```

yassin-kammoun-sonarsource (approved):
Looks good to me. I added two minor renaming suggestions to have something more meaningful than "bad name", but I can live with this so feel free to ignore them.

PR #3768, opened 2023-03-17
Update S5868 (`sonar-no-misleading-character-class`): Add checks for more Unicode character sequences

### LINKED ISSUE #3726: Update S5868 (`sonar-no-misleading-character-class`): cover cases from ESLint's implementation
1. new RegExp("\\u001F", flags)
2. var regex = /(?<\u{1d49c}>.)\x1f/
3. var regex = RegExp('\x1f')
4. var regex = new RegExp('FOO\x1fFOO\x1f')
5. var regex = new RegExp('\x1fFOO\x00')
6. var regex = new RegExp('\x1f\x1e')

francoismora:
Fixes #3726


src/linting/eslint/rules/sonar-no-misleading-character-class.ts
```diff
@@ -46,23 +61,245 @@ export const rule: Rule.RuleModule = createRegExpRule(context => {
...
+      index !== 0 &&
+      isCombiningCharacter(character.value) &&
+      !isCombiningCharacter(characters[index - 1].value)
+    ) {
+      const combinedChar = characters[index - 1].raw + characters[index].raw;
+      const message = `Move this Unicode combined character '${combinedChar}' outside of [...]`;
+      context.reportRegExpNode({ regexpNode: characters[index], node: context.node, message });
+      reported = true;
+    }
+    return reported;
```
ilia-kebets-sonarsource, line 96: why use `reported`? why not just `return true` at line 94 and `return false` at 96?

src/linting/eslint/rules/sonar-no-misleading-character-class.ts
```diff
@@ -46,23 +61,245 @@ export const rule: Rule.RuleModule = createRegExpRule(context => {
...
+  }
+
+  function checkSurrogatePairTailCharacter(
+    character: Character,
+    index: number,
+    characters: Character[],
+  ) {
+    let reported = false;
```
ilia-kebets-sonarsource, line 104: same for reported.

src/linting/eslint/rules/sonar-no-misleading-character-class.ts
```diff
@@ -46,23 +61,245 @@ export const rule: Rule.RuleModule = createRegExpRule(context => {
...
+        node: context.node,
+        message,
+        suggest,
+      });
+      reported = true;
+    }
+    return reported;
+  }
+
+  function getPattern(node: estree.Node): string | null | undefined {
```
ilia-kebets-sonarsource, line 131: Can't you reuse `getPatternFromNode()` from `extract.ts`?

src/linting/eslint/rules/sonar-no-misleading-character-class.ts
```diff
@@ -46,23 +61,245 @@ export const rule: Rule.RuleModule = createRegExpRule(context => {
...
+      return getSimpleRawStringValue(node);
+    } else {
+      return null;
+    }
+  }
+
+  function addUnicodeFlag(fixer: Rule.RuleFixer, node: estree.Node) {
+    if (isRegexLiteral(node)) {
+      return insertTextAfter(fixer, node, 'u');
+    } else if (!isRegExpConstructorPattern(node)) {
```
ilia-kebets-sonarsource, line 150: since you return, just use `if` instead of `else if` for all cases of this function?

src/linting/eslint/rules/sonar-no-misleading-character-class.ts
```diff
@@ -46,23 +61,245 @@ export const rule: Rule.RuleModule = createRegExpRule(context => {
...
+    // for instance a zero-width-sequence containing a modified emoji.
+    for (const check of characterChecks) {
+      if (check(character, index, characters)) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  function checkCombinedCharacter(character: Character, index: number, characters: Character[]) {
```
ilia-kebets-sonarsource, line 84: for all `characterChecks` functions, why not have the signature:

```
function checkXXX(character, previousCharacter) { 
// ... 
}
```

tests/linting/eslint/rules/sonar-no-misleading-character-class.test.ts
```diff
@@ -19,10 +19,30 @@
...
-const ruleTester = new RuleTester({ parserOptions: { ecmaVersion: 2018, sourceType: 'module' } });
+const ruleTester = new RuleTester({ parserOptions: { ecmaVersion: 2015 } });
 
 const combiningClass = c => `Move this Unicode combined character '${c}' outside of [...]`;
 
+function surrogatePair(c, output?, start?: number, end?: number) {
+  const error = {
+    message: `Move this Unicode surrogate pair '${c}' outside of [...] or use 'u' flag`,
+  };
+  if (output) {
```
ilia-kebets-sonarsource, line 30: `output != null` like below?

src/linting/eslint/rules/sonar-no-misleading-character-class.ts
francoismora: Returning a variable allows giving a name which I find interesting here as it helps understand that the function returns `true` if there's an issue instead of when the character is valid.
Instead, I could rename the function to `reportIfCombinedCharacter()` if you think it's more precise.

ilia-kebets-sonarsource: indeed, I like this "variable for readability" idea.

src/linting/eslint/rules/sonar-no-misleading-character-class.ts
ilia-kebets-sonarsource: I see that the implementation is different, but they seem to target the same goal.

ilia-kebets-sonarsource (changes requested):
See comments for styling change suggestions. Otherwise fantastic job!

src/linting/eslint/rules/sonar-no-misleading-character-class.ts
francoismora: That's a good point actually, Thanks. This method is used for quick fixes of surrogate pairs which support fewer cases than the check itself. It means we raise issues without quick fixes for cases like: `var r = new RegExp("[" + "👍" + "]")` because the pattern is a binary expression. Theoretically, we could extend the scope of quick fixes by reusing `getPatternFromNode()` but this impacts the logic of `addUnicodeFlag()` function which needs to find the argument after the pattern. I will take a look.

src/linting/eslint/rules/sonar-no-misleading-character-class.ts
francoismora: The goal was to share a switch-like structure between `getPattern()` and `addUnicodeFlag()` as both functions have a common logic. But this part will change if we reuse `getPatternFromNode()`.

src/linting/eslint/rules/sonar-no-misleading-character-class.ts
francoismora: For `checkZeroWidthJoinerCharacter()` we need the character before and after, so we can't generalize this to all checks.
To simplify the checks signatures, we could wrap them with one of the following utility functions. WDYT?
```typescript
  function checkFollowingCharacter(check: (previous: Character, character: Character) => boolean) {
    return (character: Character, index: number, characters: Character[]) =>
      index !== 0 && check(characters[index - 1], character);
  }

  function checkJoiningCharacter(
    check: (previous: Character, character: Character, next: Character) => boolean,
  ) {
    return (character: Character, index: number, characters: Character[]) =>
      index !== 0 &&
      index !== characters.length - 1 &&
      check(characters[index - 1], character, characters[index + 1]);
  }

```

src/linting/eslint/rules/sonar-no-misleading-character-class.ts
francoismora: I have updated the code to reuse `getPatternFromNode()` as suggested. This indeed improves the rule, thanks for this great idea.

src/linting/eslint/rules/sonar-no-misleading-character-class.ts
francoismora: Following the last change, the code, now, simply uses if with returns.

src/linting/eslint/rules/sonar-no-misleading-character-class.ts
ilia-kebets-sonarsource: Ah, nevermind then. I think repetition is better than some obscure modularization. Thanks for investigating it.

ilia-kebets-sonarsource: My main initial goal was to simplify the index accesses like `characters[index - 1]`, but if it's not doable simply, I prefer to drop it.

src/linting/eslint/rules/sonar-no-misleading-character-class.ts
```diff
@@ -46,23 +61,227 @@ export const rule: Rule.RuleModule = createRegExpRule(context => {
...
+    characters: Character[],
+  ) {
+    let reported = false;
+    if (
+      index !== 0 &&
+      isEmojiModifier(character.value) &&
+      !isEmojiModifier(characters[index - 1].value)
+    ) {
+      const modifiedEmoji = characters[index - 1].raw + characters[index].raw;
+      const message = `Move this Unicode modified Emoji '${modifiedEmoji}' outside of [...]`;
```
ilia-kebets-sonarsource, line 167: WDYT?

```suggestion
      const message = `Move this Unicode modified Emoji '${modifiedEmoji}' outside of the character class`;
```

src/linting/eslint/rules/sonar-no-misleading-character-class.ts
```diff
@@ -46,23 +61,227 @@ export const rule: Rule.RuleModule = createRegExpRule(context => {
...
+    characters: Character[],
+  ) {
+    let reported = false;
+    if (
+      index !== 0 &&
+      isRegionalIndicator(character.value) &&
+      isRegionalIndicator(characters[index - 1].value)
+    ) {
+      const regionalIndicator = characters[index - 1].raw + characters[index].raw;
+      const message = `Move this Unicode regional indicator '${regionalIndicator}' outside of [...]`;
```
ilia-kebets-sonarsource, line 186:
```suggestion
      const message = `Move this Unicode regional indicator '${regionalIndicator}' outside of the character class`;
```

src/linting/eslint/rules/sonar-no-misleading-character-class.ts
```diff
@@ -46,23 +61,227 @@ export const rule: Rule.RuleModule = createRegExpRule(context => {
...
+      index !== 0 &&
+      index !== characters.length - 1 &&
+      isZeroWidthJoiner(character.value) &&
+      !isZeroWidthJoiner(characters[index - 1].value) &&
+      !isZeroWidthJoiner(characters[index + 1].value)
+    ) {
+      // It's practically difficult to determine the full joined character sequence
+      // as it may join more than 2 elements that consist of characters or modified Emojis
+      // see: https://unicode.org/emoji/charts/emoji-zwj-sequences.html
+      const message = 'Move this Unicode joined character sequence outside of [...]';
```
ilia-kebets-sonarsource, line 209:
```suggestion
      const message = 'Move this Unicode joined character sequence outside of the character class';
```

ilia-kebets-sonarsource (approved):
Let me know what you decide regarding the issue message. Otherwise, LGTM!

src/linting/eslint/rules/sonar-no-misleading-character-class.ts
francoismora: Changed

francoismora:
> Let me know what you decide regarding the issue message. Otherwise, LGTM!

I applied the suggested change.

PR #3774, opened 2023-03-20
Add rule S2486: Exceptions should not be ignored

### LINKED ISSUE #3772: Add rule S2486: Exceptions should not be ignored
The first case of https://github.com/SonarSource/SonarJS/issues/3746 makes more sense in S2486: `try{}catch(err){}`

yassin-kammoun-sonarsource:
Fixes #3772


francoismora (approved):
LGTM

PR #3784, opened 2023-03-23
Move tsconfigs logic to Node

victor-diez-sonarsource:


src/helpers/lru.ts
```diff
@@ -0,0 +1,46 @@
...
+export class LRU<T> {
+  private readonly max: number;
+  private readonly cache: T[];
+  constructor(max = 2) {
+    this.max = max;
+    this.cache = [];
+  }
+
+  get() {
+    return this.cache;
```
ilia-kebets-sonarsource, line 29: aren't reads supposed to count as usage?

src/helpers/tsconfigs.ts
```diff
@@ -0,0 +1,95 @@
...
+ */
+import fs from 'fs';
+import path from 'path';
+import { getContext } from './context';
+import { readFileSync, toUnixPath } from './files';
+import * as console from 'console';
+import { defaultCache } from 'services/program';
+
+export type TSConfigs = Map<string, TSConfig>;
+export const projectTSConfigs: TSConfigs = new Map<string, TSConfig>();
```
ilia-kebets-sonarsource, line 28: somehow specify that it's a singleton

tests/helpers/tsconfigs.test.ts
```diff
@@ -0,0 +1,102 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import { projectTSConfigs, setContext, toUnixPath, tsConfigLookup, updateTsConfigs } from 'helpers';
+import path from 'path';
+import console from 'console';
+
+describe('TSConfigs', () => {
```
ilia-kebets-sonarsource, line 25: specify the function you're testing like `tsConfigLookup()` or `updateTsConfigs()`

src/helpers/lru.ts
victor-diez-sonarsource: The GC only collects WeakRefs. LRU is used to keep strong refs and avoid GC. I can explain this on a call, but no worries I'll further comment on the code to clarify it better

src/helpers/tsconfigs.ts
```diff
@@ -0,0 +1,95 @@
...
+      projectTSConfigs.set(filename, {
+        filename,
+        contents,
+      });
+    }
+  }
+}
+
+export function updateTsConfigs(tsconfigs: string[], force = false) {
+  if (!getContext().sonarlint && !force) {
```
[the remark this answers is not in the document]
victor-diez-sonarsource: yes, it's just that negating the condition and moving the whole code inside increments the cognitive complexity of the function. :(

src/helpers/cache.ts
```diff
@@ -0,0 +1,50 @@
...
+ */
+import { LRU } from './lru';
+import ts from 'typescript';
+import { TSConfig } from './tsconfigs';
+
+export type ProgramResult = {
+  tsConfig: TSConfig;
+  files: string[];
+  projectReferences: string[];
+  missingTsConfig: boolean;
```
ilia-kebets-sonarsource, line 28:
```suggestion
  isMissingTsConfig: boolean;
```

src/helpers/cache.ts
```diff
@@ -0,0 +1,50 @@
...
+  program: WeakRef<ts.Program>;
+  fallbackProgram?: boolean;
```
ilia-kebets-sonarsource, line 30:
```suggestion
  isFallbackProgram?: boolean;
```

src/helpers/tsconfigs.ts
```diff
@@ -0,0 +1,95 @@
...
+
+export type TSConfigs = Map<string, TSConfig>;
+export const projectTSConfigs: TSConfigs = new Map<string, TSConfig>();
+export interface TSConfig {
+  filename: string;
+  contents: string;
+  fallbackTSConfig?: boolean;
+}
+
+export function tsConfigLookup(dir?: string) {
```
ilia-kebets-sonarsource, line 35: WDYT about changing this from DFS to BFS, and look for `index.(js|ts)` and `main.(js|ts)` files before iterating on all as we discussed the other day?

src/helpers/tsconfigs.ts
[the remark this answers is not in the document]
ilia-kebets-sonarsource: nevermind indeed.

src/helpers/tsconfigs.ts
victor-diez-sonarsource: I would prefer to leave the ordering aside of the lookup. I think ordering algorithm should be implemented [here](https://github.com/SonarSource/SonarJS/blob/e8c595efe085a262b18068674589f64f083262c6/src/services/program/program.ts#L50)

src/parsing/jsts/builders/build.ts
```diff
@@ -35,22 +35,79 @@ import { Language } from 'parsing/jsts';
...
+     */
+    throw moduleError;
   }
 }
 
-function shouldTryTypeScriptParser() {
-  const context = getContext();
-  return context ? context.shouldUseTypeScriptParserForJS : true;
+function shouldUseWatchProgram(file: string): boolean {
+  return getContext()?.sonarlint || file.toLowerCase().endsWith('.vue');
```
ilia-kebets-sonarsource, line 108: you can use `isVueFile` computed before as function param

src/helpers/cache.ts
victor-diez-sonarsource: I would leave this by now as it's used currently from the Java part.

saberduck (approved):

PR #3787, opened 2023-03-24
Fix Cirrus Mend scan command

francoismora:
Change the Cirrus Mend scan command to build the NPM package before verifying the Java plugin.


saberduck:
done in #3775

PR #3825, opened 2023-03-29
Add rule S6551 (`no-base-to-string`): Objects and classes converted or coerced to strings should define a "toString()" method

### LINKED ISSUE #3791: Add rule for `no-base-to-string` and `restrict-template-expressions`
https://typescript-eslint.io/rules/no-base-to-string/
https://typescript-eslint.io/rules/restrict-template-expressions

- Not covered
- Makes sense?
  - Disallow usage of `toString()` on non-primitives that don't provide a `toString()` implementation.
- No option, no quick fix
- Decision: we add this as a NEW rule 
- Decorate the rule to reduce false positives with generic types or implement it
- no-base-to-string seems to be a superset of restrict-template-expressions.

RSPEC PR: https://github.com/SonarSource/rspec/pull/1697

ilia-kebets-sonarsource:
Fixes #3791


yassin-kammoun-sonarsource:
> @yassin-kammoun-sonarsource It does flag generic types actually like:
> 
> ```ts
> static toStrings<T>(array: T[]): string[] {
>         return this.map(array, function (item) {
>             if (item === undefined || item === null || !item.toString) {
>                 return null;
>             } else {
>                 return item.toString(); // noncompliant
>             }
>         });
>     }
> ```
> 
> The case that was avoided is:
> 
> ```ts
> function test<TWithNoConstraint>(arg: T) {
>           return \`arg = \${arg}\`;
>         }
> ```
> 
> Should I decorate it to avoid generics?
> 
> Also, there is another case that I'm unsure, it's for `something | string`. WDYT?

Yes, I think that we should decorate it and avoid generics. That's actually what I suggested in the evaluation of the rule.

yassin-kammoun-sonarsource:
> Also, there is another case that I'm unsure, it's for `something | string`. WDYT?

I am not too sure here. If `something` doesn't override `toString`, that's definitely an issue as far as I understand. If we are able to detect that, that would be cool, let's give it a try and see if this pattern occurs often in ruling.

ilia-kebets-sonarsource:
> > Also, there is another case that I'm unsure, it's for `something | string`. WDYT?
> 
> I am not too sure here. If `something` doesn't override `toString`, that's definitely an issue as far as I understand. If we are able to detect that, that would be cool, let's give it a try and see if this pattern occurs often in ruling.

This pattern does occur in the ruling tests, and it is indeed caught.

yassin-kammoun-sonarsource:
I would b

> > > Also, there is another case that I'm unsure, it's for `something | string`. WDYT?
> > 
> > 
> > I am not too sure here. If `something` doesn't override `toString`, that's definitely an issue as far as I understand. If we are able to detect that, that would be cool, let's give it a try and see if this pattern occurs often in ruling.
> 
> This pattern does occur in the ruling tests, and it is indeed caught.

My gut feeling is that we should be conservative and ignore these issues. I am afraid that it would be quite noisy in a JavaScript codebase. If a variable is assigned values of different types and we are not able to discriminate the constituent type of the union, I think we would raise many false positives.

What do you think ?

ilia-kebets-sonarsource:
@yassin-kammoun-sonarsource I've brought the changes!

sonar-plugin/javascript-checks/src/main/java/org/sonar/javascript/checks/NoBaseToStringCheck.java
```diff
@@ -0,0 +1,34 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.javascript.checks;
+
+import org.sonar.check.Rule;
+import org.sonar.plugins.javascript.api.EslintBasedCheck;
+import org.sonar.plugins.javascript.api.TypeScriptRule;
+
+@TypeScriptRule
```
yassin-kammoun-sonarsource, line 26: Why is the rule not enabled for JavaScript ?

src/linting/eslint/rules/decorators/no-base-to-string-decorator.ts
```diff
@@ -0,0 +1,44 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+import { TSESTree } from '@typescript-eslint/experimental-utils';
```
yassin-kammoun-sonarsource, line 20:
```suggestion
// https://sonarsource.github.io/rspec/#/rspec/S6551/javascript

import { TSESTree } from '@typescript-eslint/experimental-utils';
```

src/linting/eslint/rules/decorators/no-base-to-string-decorator.ts
```diff
@@ -0,0 +1,44 @@
...
+  // maybe call sanitizeTSrule()
+  return interceptReport(rule, (context, reportDescriptor) => {
+    if ('node' in reportDescriptor) {
+      const services = context.parserServices;
+      const hasTypeInformation = isRequiredParserServices(services);
+      if (hasTypeInformation && isGenericType(reportDescriptor.node as TSESTree.Node, services)) {
+        // we skip
+      } else {
+        context.report(reportDescriptor);
+      }
```
yassin-kammoun-sonarsource, line 35: For your information, TypeScript ESLint rules are always sanitized before decoration. See:

https://github.com/SonarSource/SonarJS/blob/a146a0d4d4df335f101d52d0c1ce21ee733eddec/src/linting/eslint/linter/decoration/decorate.ts#L56-L75

Therefore, we can not only drop the comment but also the whole check about the availability of the type checker. If type information is missing, the (sanitized) rule will never raise any issues.

src/linting/eslint/rules/decorators/no-base-to-string-decorator.ts
```diff
@@ -0,0 +1,44 @@
...
+    }
+  });
+}
+
+function isGenericType(node: TSESTree.Node, services: RequiredParserServices) {
+  const checker = services.program.getTypeChecker();
+  const type = checker.getTypeAtLocation(services.esTreeNodeToTSNodeMap.get(node));
```
yassin-kammoun-sonarsource, line 42: We can reuse an existing helper function for that:

https://github.com/SonarSource/SonarJS/blob/a146a0d4d4df335f101d52d0c1ce21ee733eddec/src/linting/eslint/rules/helpers/type.ts#L87-L90

sonar-plugin/javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S6551.json
```diff
@@ -0,0 +1,21 @@
...
+    "object"
+  ],
+  "defaultSeverity": "Minor",
+  "ruleSpecification": "RSPEC-6551",
+  "sqKey": "S6551",
+  "scope": "All",
+  "quickfix": "infeasible",
+  "compatibleLanguages": [
+    "TYPESCRIPT"
+  ]
```
yassin-kammoun-sonarsource, line 20: If we decide to enable the rule for JavaScript, we will need to add it here as well.

yassin-kammoun-sonarsource (changes requested):

yassin-kammoun-sonarsource (approved):
LGTM!

PR #3827, opened 2023-03-29
Add rule S6550 (`prefer-literal-enum-member`): All enum members should be literals

francoismora:
The rule is about enum, so it targets TypeScript only. Also, bitwise expressions are not allowed for the moment.

Fixes #2811

RSPEC: SonarSource/RSPEC#1696


yassin-kammoun-sonarsource (approved):
LGTM!

PR #3828, opened 2023-03-29
Fix "mvn-copy" script

ilia-kebets-sonarsource:


saberduck:
@ilia-kebets-sonarsource we use `mkdirp` package to be compatible with Windows, I am not sure this would work otherwise.

ilia-kebets-sonarsource:
Thanks for the heads up, @saberduck. Closing this.

PR #3832, opened 2023-03-30
Update S4139 (`no-for-in-array`): Cover cases from TypeScript ESLint implementation

### LINKED ISSUE #3796: Update S4139 (`no-for-in-array`): Cover cases from TypeScript ESLint implementation
https://typescript-eslint.io/rules/no-for-in-array/

- We should improve our implementation by taking inspiration from how TypeScript ESLint’s rule detects an iterable.

Cases that should be covered:
```typescript
const fn = (arr: number[] | string[]) => {
  for (const x in arr) {
    console.log(x);
  }
};

const fn = <T extends any[]>(arr: T) => {
  for (const x in arr) {
    console.log(x);
  }
};
```

francoismora:
This PR reuses the logic from `no-alphabetical-sort` which is now shared in `type.ts` to cover cases where the array is a union type or a generic parameter extending an array type as in:

```typescript
const fn = (arr: number[] | string[]) => {
  for (const x in arr) {
    console.log(x);
  }
};
```

and 

```typescript
const fn = <T extends any[]>(arr: T) => {
  for (const x in arr) {
    console.log(x);
  }
};
```

Fixes #3796

### DIFF
--- a/src/linting/eslint/rules/helpers/type.ts
+++ b/src/linting/eslint/rules/helpers/type.ts
@@ -65,6 +65,10 @@ export function isUnion(node: estree.Node, services: RequiredParserServices) {
   return type.isUnion();
 }
 
+export function getUnionTypes(type: ts.Type): ts.Type[] {
+  return type.isUnion() ? type.types : [type];
+}
+
 export function isUndefinedOrNull(node: estree.Node, services: RequiredParserServices) {
   const checker = services.program.getTypeChecker();
   const typ = checker.getTypeAtLocation(services.esTreeNodeToTSNodeMap.get(node as TSESTree.Node));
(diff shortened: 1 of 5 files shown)


src/linting/eslint/rules/helpers/type.ts
```diff
@@ -105,3 +109,29 @@ export function getSignatureFromCallee(node: estree.Node, services: RequiredPars
     services.esTreeNodeToTSNodeMap.get(node as TSESTree.Node) as ts.CallLikeExpression,
   );
 }
+
+export function isArrayLikeType(type: ts.Type, services: RequiredParserServices) {
```
ilia-kebets-sonarsource, line 113: Could you add some JS doc to describe what is an ArrayLike?

ilia-kebets-sonarsource (approved):
See my small request for JS doc, otherwise LGTM!

src/linting/eslint/rules/helpers/type.ts
francoismora: Added

PR #3838, opened 2023-04-03
Add rule S6564: Redundant type aliases should not be used

### LINKED ISSUE #3800: Add rule S6564: Redundant type aliases should not be used
https://typescript-eslint.io/rules/no-type-alias

 - Covered?
   - Not covered
   - Opposite of https://sonarsource.github.io/rspec/#/rspec/S4323
 - Makes sense?
   - Disallow type aliases of primitive types like `undefined`, `string`, etc.
 - Option for aliasing primitive types should be turned on, No quickfix
 - Implement aliasing primitive types only: string, number, boolean, symbol, BigInt [JavaScript primitives](https://developer.mozilla.org/en-US/docs/Glossary/Primitive), [TypeScript everyday types](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html).

yassin-kammoun-sonarsource:
Fixes #3800


src/linting/eslint/rules/redundant-type-aliases.ts
```diff
@@ -0,0 +1,78 @@
...
+
+export const rule: Rule.RuleModule = {
+  meta: {
+    messages: {
+      redundantTypeAlias:
+        'Remove this redundant type alias and replace its occurrences with "{{type}}".',
+    },
+  },
+  create(context: Rule.RuleContext) {
+    function isAliasType(node: TSESTree.TypeNode) {
```
ilia-kebets-sonarsource, line 49: maybe move this to utils? could be useful elsewhere

ilia-kebets-sonarsource (changes requested):
see comment and fix CI please

src/linting/eslint/rules/redundant-type-aliases.ts
yassin-kammoun-sonarsource: Done

PR #3840, opened 2023-04-04
Add rule S2094 (`no-empty-class-check`): Classes should not be empty

### LINKED ISSUE #3794: Add rule for `no-extraneous-class`
https://typescript-eslint.io/rules/no-extraneous-class

- Not covered
- Makes sense?
  - Disallow classes with only static variables/methods used as namespaces
 - Options available, no quickfix
 - We might need to pay particular attention to empty decorated classes and exclude them
 - We should enable the rule for the main scope: classes with only static functions
 - We should enable the rule to create a JS/TS variant of https://sonarsource.github.io/rspec/#/rspec/S2094 for empty classes

RSPEC PR: https://github.com/SonarSource/rspec/pull/1720

ilia-kebets-sonarsource:
Fixes #3794


yassin-kammoun-sonarsource (approved):
LGTM. We might need to regenerate the rule description in case some changes are made in the corresponding RSPEC PR.

PR #3845, opened 2023-04-05
Add rule S6572: Enum member values should initialized

### LINKED ISSUE #3809: Add rule for `prefer-enum-initializers`
https://typescript-eslint.io/rules/prefer-enum-initializers

 - Covered?
   - Not covered
 - Makes sense?
   - Require each enum member value to be explicitly initialized
 - No options, no quickfix
 - Decorate or implement it ourselves. We should raise an issue if and only if there are some enum values initialized but not all like:

```typescript
enum Status {
 Open = 1,
 Close,
}
```

yassin-kammoun-sonarsource:
Fixes #3809


src/linting/eslint/rules/decorators/prefer-enum-initializers-decorator.ts
```diff
@@ -0,0 +1,55 @@
...
+      context.report(descriptor);
+    }
+  });
+}
+
+function anyInitialized(enumDecl: TSESTree.TSEnumDeclaration) {
+  return enumDecl.members.some(m => m.initializer !== undefined);
+}
+
+function numericalOrder(enumDecl: TSESTree.TSEnumDeclaration) {
```
ilia-kebets-sonarsource, line 49: You must check that the other members are initialized, right?

ilia-kebets-sonarsource (changes requested):
see comment

src/linting/eslint/rules/decorators/prefer-enum-initializers-decorator.ts
yassin-kammoun-sonarsource: Yes, or rather, check they are not initialized.

ilia-kebets-sonarsource (approved):
LGTM!

PR #3848, opened 2023-04-11
Add rule S6578: Enum values should be unique

### LINKED ISSUE #3793: Add rule S6578: Enum values should be unique
https://typescript-eslint.io/rules/no-duplicate-enum-values/

- Not covered
- Makes sense?
  - Disallow defining an enum with multiple members initialized to the same value.
- No options, no quick fix
- Might be interesting to consider a different rule for TypeScript similar as https://sonarsource.github.io/rspec/#/rspec/S2753

yassin-kammoun-sonarsource:
Fixes #3793


ilia-kebets-sonarsource (approved):
LGTM!

PR #3852, opened 2023-04-13
Add rule S6590: "as const" assertions should be preferred

### LINKED ISSUE #3808: Add rule S6590: "as const" assertions should be preferred
https://typescript-eslint.io/rules/prefer-as-const

 - Covered?
   - Not covered
 - Makes sense?
   - Enforce the use of as const over literal type

yassin-kammoun-sonarsource:
Fixes #3808


ilia-kebets-sonarsource (approved):
LGTM!

PR #3853, opened 2023-04-13
Add rule S6592 (`promise-function-async`): Use async when returning a promise

### LINKED ISSUE #3817: Add rule for `promise-function-async`
https://typescript-eslint.io/rules/promise-function-async

 - Covered?
   - Not covered
 - Makes sense?
   - Require any function or method that returns a Promise to be marked [async](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function#description). Shows explicit intention about functions returning promises. Educates developers about promises

RSPEC PR: https://github.com/SonarSource/rspec/pull/1754

ilia-kebets-sonarsource:
Fixes #3817


ilia-kebets-sonarsource:
there are FPs in ruling tests like:
- `file-for-rules: S3854.js #82` has an FP because it returns a function of type undef

yassin-kammoun-sonarsource (commented):
As I suggested in the review of the RSPEC PR, I think we can provide quickfixes for simple cases like function declarations and functions expressions, where we add the `async` keyword before the `function` keyword.

ilia-kebets-sonarsource:
2237 issues on peach, I believe we can drop this rule.

ilia-kebets-sonarsource:
closing as per https://github.com/SonarSource/SonarJS/issues/3817

PR #3857, opened 2023-04-18
Add rule S6598: Function types should be preferred

### LINKED ISSUE #3810: Add rule S6598: Function types should be preferred
https://typescript-eslint.io/rules/prefer-function-type

 - Covered?
   - Not covered
 - Makes sense?
   - Enforce using function types instead of interfaces with call signatures
 - Enable it if not too noisy Maybe add an exception for interfaces with more than a function signature.

yassin-kammoun-sonarsource:
Fixes #3810


ilia-kebets-sonarsource (approved):
LGTM!

PR #3868, opened 2023-04-26
Update S6571 (`no-redundant-type-constituents`): Limit unknown detection to the `unknown` type annotation

yassin-kammoun-sonarsource:
I noticed that we already reduced the scope of the rule for `any` to simple cases. I did the same for the case of `unknown` since we are facing the same issue as François did when he first implemented the rule and limited the scope.


ilia-kebets-sonarsource (approved):
LGTM. Smooth fix!

PR #3870, opened 2023-04-26
Fix FP S6544 (`no-misused-promises`): properties that contain a function returning a promise

ilia-kebets-sonarsource:


ilia-kebets-sonarsource:
actually, the cases I reviewed on Peach are TPs, the fields of objects where a promise-returning function is assigned were all typed for something else.

PR #3872, opened 2023-04-26
Use Node to resolve tsconfig

saberduck:


victor-diez-sonarsource (approved):

stevendarby:
@saberduck @victor-diez-sonarsource does it look like our issue here could be caused by recent changes around resolving tsconfig? https://community.sonarsource.com/t/typescript-s4328-suddenly-detecting-internal-modules/92846/3

PR #3873, opened 2023-04-27
Updates S2871 (no-alphabetical-sort) to include toSorted function.

### LINKED ISSUE #3864: Update S2871 (`no-alphabetical-sort`) to include `toSorted`
[`Array.prototype.toSorted`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted) is a new array method with the same behaviour as `Array.prototype.sort` but returning a copy of the original array instead of mutating. Since the sorting behaviour is the same, the `toSorted` method should also be covered by this rule.

philnash:
Fixes #3864


tests/linting/eslint/rules/no-alphabetical-sort.test.ts
```diff
@@ -339,3 +339,321 @@ ruleTester.run(`A compare function should be provided when using "Array.prototyp
...
+      arrayOfNumbers.toSorted((n, m) => n - m);
+      `,
+      },
+      {
+        code: `unknownArrayType.toSorted();`,
+      },
+      {
+        code: `
+      function f(a: any[]) {
+        a.toSorted(undefined);
```
ilia-kebets-sonarsource, line 360: Could you adapt these examples so they make sense with `toSorted()` not making the change _in place_?

```suggestion
        return a.toSorted(undefined);
```

ilia-kebets-sonarsource (changes requested):
Great improvement!

As I mentioned in my other comment, could you adapt the unit tests so they make more sense?
And for rules like this one that does not raise any issue in the integration ruling tests, we add a file ourselves in [file-for-rules](https://github.com/SonarSource/SonarJS/tree/master/its/sources/file-for-rules). You can simply add a `S2871.js` file with one of the invalid unit test cases and run the JavaScriptRulingTest and update the expected values.

Could you also open a PR in RSPEC to update the rule description to include `Array.toSorted()`?

philnash:
Ok, I added a [S2871.js file](https://github.com/SonarSource/SonarJS/pull/3873/files#diff-c8753034a5c4644809f866a3c1dd77413c6f2b68438d97c4faba37f727f037f8) but nothing happened in the rulings for me. And I don't know why.

Am I running the tests wrong? Did I need to add anything else? I am quite confused.

ilia-kebets-sonarsource:
> Ok, I added a [S2871.js file](https://github.com/SonarSource/SonarJS/pull/3873/files#diff-c8753034a5c4644809f866a3c1dd77413c6f2b68438d97c4faba37f727f037f8) but nothing happened in the rulings for me. And I don't know why.
> 
> Am I running the tests wrong? Did I need to add anything else? I am quite confused.

You should follow [these steps and update the integration ruling tests](https://github.com/SonarSource/SonarJS/blob/master/docs/DEV.md#ruling-tests).
1. run the ruling tests for JS locally, which will fail because there are new issues raised:
```bash
cd its/ruling
mvn verify -Dtest=JavaScriptRulingTest -Dmaven.test.redirectTestOutputToFile=false
```
2. update the expected test results
```bash
# in its/ruling
cp -R target/actual/js/ src/test/expected/js
```

Let me know if this answers your question.

ilia-kebets-sonarsource (changes requested):
see comment

philnash:
I don't know why, but last week I could not make the ruling test fail. First time I ran it today, fail. I copied across the updated rulings and this should be good now!

ilia-kebets-sonarsource (approved):
LGTM, don't forget to update the RSPEC repo.

philnash:
https://github.com/SonarSource/rspec/pull/1831

PR #3892, opened 2023-05-04
Analyze files without tsconfig also when using AnalysisWithWatchProgram

### LINKED ISSUE #3883: Regression on security benchmark
We are missing JS issues on codeql-js benchmark

saberduck:
- `ProjectTsConfigs` need to be initialized lazily
- Analyze files without tsconfig also when using analysis with watch program

Fixes #3883

### DIFF
--- a/src/services/program/program.ts
+++ b/src/services/program/program.ts
@@ -46,7 +46,14 @@ import { ProgramCache } from 'helpers/cache';
 import { JsTsAnalysisInput } from 'services/analysis';
 
 export const programCache = new ProgramCache();
-const projectTSConfigs = new ProjectTSConfigs();
+let projectTSConfigs: ProjectTSConfigs;
+
+function getDefaultTSConfigs() {
+  if (!projectTSConfigs) {
+    projectTSConfigs = new ProjectTSConfigs();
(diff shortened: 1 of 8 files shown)


src/services/program/program.ts
```diff
@@ -58,8 +65,11 @@ const projectTSConfigs = new ProjectTSConfigs();
 export function getProgramForFile(
   input: JsTsAnalysisInput,
   cache = programCache,
-  tsconfigs = projectTSConfigs,
+  tsconfigs?: ProjectTSConfigs,
 ): ts.Program {
+  if (!tsconfigs) {
+    tsconfigs = getDefaultTSConfigs();
```
victor-diez-sonarsource, line 71: I understand the change and I agree with it, but AFAIK this is not being used in master right? This should only apply on the new PR enabling the new codepath. We can discuss this on monday.

victor-diez-sonarsource (approved):

src/services/program/program.ts
saberduck: it is, because `program.ts` is imported in routing modules, so the constructor was invoked.

PR #3926, opened 2023-05-16
Fix FP S1128 (`unused-import`): Ignore imported symbols used as Vue.js directives

### LINKED ISSUE #3790: Fix FP S1128 (`unused-import`): Ignore imported symbols used as Vue.js directives
In the context of Vue.js Single File Components, the rule S1128 (`unused-import`) ignores imported symbols that are used as tags in the `template` section of such components. However, the rule is currently reporting false positives when imported symbols are used as tag attributes in the very same section.

```vue
<script lang="ts" setup>
  import { FooBar } from "./FooBar.vue"; // FP - FooBar is incorrectly flagged as unused
</script>
<template>
  <input v-foo-bar />
</template>
```

Reported in https://community.sonarsource.com/t/s1128-fp-with-use-vue3-directive-in-composition-api/86079

yassin-kammoun-sonarsource:
Fixes #3790

### DIFF
--- a/src/linting/eslint/rules/unused-import.ts
+++ b/src/linting/eslint/rules/unused-import.ts
@@ -138,6 +138,13 @@ export const rule: Rule.RuleModule = {
               vueIdentifiers.add(toPascalCase(rawName));
             }
           },
+          VDirectiveKey: (node: AST.VDirectiveKey) => {
+            const {
+              name: { name },
+            } = node;
+            vueIdentifiers.add(toCamelCase(name));
+            vueIdentifiers.add(toPascalCase(name));
+          },
           Identifier: (node: AST.ESLintIdentifier) => {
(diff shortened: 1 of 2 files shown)


ilia-kebets-sonarsource:
Would it fix the issue from this guy? https://github.com/SonarSource/SonarJS/issues/3484#issuecomment-1541969937

yassin-kammoun-sonarsource:
> Would it fix the issue from this guy? [#3484 (comment)](https://github.com/SonarSource/SonarJS/issues/3484#issuecomment-1541969937)

We are considering multiple kinds of imports, but it's hard to tell because the user doesn't say how the imported symbol is used.

ilia-kebets-sonarsource (approved):
LGTM!

PR #3927, opened 2023-05-17
Add option to test with already running Node process

victor-diez-sonarsource:


sonar-plugin/sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImplTest.java
```diff
@@ -380,6 +375,36 @@ void test_lazy_start() throws Exception {
     assertThat(logTester.logs(DEBUG)).contains(alreadyStarted);
   }
 
+  @Test
+  void test_use_existing_node() throws Exception {
+    String useExisting = "Will use existing Node.js process in port 60000";
```
saberduck, line 380: use `var`

sonar-plugin/sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImplTest.java
```diff
@@ -380,6 +375,36 @@ void test_lazy_start() throws Exception {
...
+    String alreadyStarted = "eslint-bridge server is up, no need to start.";
+    String wrongPortRange =
+      "Provided port for existing Node.js process is not valid (should be in the 1 to 65535 range).";
+    String wrongPortValue = "Provided port for existing Node.js process is not valid.";
+
+    eslintBridgeServer = createEslintBridgeServer("startServer.js");
+    EslintBridgeServerImpl eslintBridgeServerMock = spy(eslintBridgeServer);
```
saberduck, line 387: use `var`

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
```diff
@@ -261,6 +267,24 @@ public void startServerLazily(SensorContext context) throws IOException {
       // required for SonarLint context to avoid restarting already failed server
       throw new ServerAlreadyFailedException();
     }
+    // if SONARJS_EXISTING_NODE_PROCESS_PORT is set, use existing node process
+    if (port == 0 && nodePortIsProvided()) {
+      try {
+        port = nodeAlreadyRunningPort();
+        if (port <= 0 || port >= 65535) {
```
saberduck, line 274: I think we don't need these checks, it's developer option. If you think we need them, then they should be moved inside the function which reads the env variable

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
```diff
@@ -454,6 +478,14 @@ private URI url(String endpoint) {
     }
   }
 
+  boolean nodePortIsProvided() {
+    return System.getenv(SONARJS_EXISTING_NODE_PROCESS_PORT) != null;
+  }
+
+  int nodeAlreadyRunningPort() throws NumberFormatException {
+    return Integer.parseInt(System.getenv(SONARJS_EXISTING_NODE_PROCESS_PORT));
```
saberduck, line 486: catch the `NumberFormatException` and re-throw as  `IllegalStateException` , you can also do the range checks here and just throw `IllegalStateException` if value is invalid

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
```diff
@@ -454,6 +478,14 @@ private URI url(String endpoint) {
     }
   }
 
+  boolean nodePortIsProvided() {
+    return System.getenv(SONARJS_EXISTING_NODE_PROCESS_PORT) != null;
```
saberduck, line 482: I am not sure we need two methods to read the same variable, you could maybe use `Optional<Integer>` instead, ie `Optional.ofNullable(System.getenv(...))`

saberduck (changes requested):

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
```diff
@@ -454,6 +477,27 @@ private URI url(String endpoint) {
     }
   }
 
+  int nodeAlreadyRunningPort() throws IllegalStateException {
```
saberduck, line 480: you don't need to declare `IllegalStateException` because it inherits from `RuntimeException`

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
```diff
@@ -454,6 +477,27 @@ private URI url(String endpoint) {
...
+    int port;
+    try {
+      port =
+        Optional
+          .ofNullable(getenv(SONARJS_EXISTING_NODE_PROCESS_PORT))
+          .map(Integer::parseInt)
+          .orElse(0);
+      if (port < 0 || port > 65535) {
+        throw new NumberFormatException();
```
saberduck, line 489: you should not throw this exception, you should throw `IllegalStateException` , also add some message as parameter to make troubleshooting easier

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
```diff
@@ -454,6 +477,27 @@ private URI url(String endpoint) {
...
+      }
+    } catch (NumberFormatException nfe) {
+      throw new IllegalStateException();
```
saberduck, line 492: You should include `nfe` as parameter to provide cause of exception

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
```diff
@@ -454,6 +477,27 @@ private URI url(String endpoint) {
...
+    }
+    return port;
+  }
+
+  public String getenv(String name) {
```
saberduck, line 497: why do we need this function?

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
```diff
@@ -261,6 +268,20 @@ public void startServerLazily(SensorContext context) throws IOException {
       // required for SonarLint context to avoid restarting already failed server
       throw new ServerAlreadyFailedException();
     }
+    try {
+      int providedPort = nodeAlreadyRunningPort();
```
saberduck, line 272: use `var`

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
```diff
@@ -454,6 +477,27 @@ private URI url(String endpoint) {
     }
   }
 
+  int nodeAlreadyRunningPort() throws IllegalStateException {
+    int port;
```
saberduck, line 481: why do you need to split declaration and assignment? seems like unnecessary line

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
```diff
@@ -454,6 +477,27 @@ private URI url(String endpoint) {
...
+          .ofNullable(getenv(SONARJS_EXISTING_NODE_PROCESS_PORT))
+          .map(Integer::parseInt)
+          .orElse(0);
+      if (port < 0 || port > 65535) {
+        throw new NumberFormatException();
+      }
+    } catch (NumberFormatException nfe) {
+      throw new IllegalStateException();
+    }
+    return port;
```
saberduck, line 494: you can return inside `try` body

saberduck (changes requested):

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
```diff
@@ -261,6 +268,20 @@ public void startServerLazily(SensorContext context) throws IOException {
...
     }
+    try {
+      var providedPort = nodeAlreadyRunningPort();
+      // if SONARJS_EXISTING_NODE_PROCESS_PORT is set, use existing node process
+      if (providedPort != 0) {
+        port = providedPort;
+        serverHasStarted();
+        LOG.info("Will use existing Node.js process in port " + port);
+      }
+    } catch (IllegalStateException e) {
```
saberduck, line 279: we should not catch this, just let it fail

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/EslintBridgeServerImpl.java
saberduck: what about this one?

saberduck (approved):

PR #3929, opened 2023-05-17
Update S4043 (`no-misleading-array-reverse`): Add quickfix support

### LINKED ISSUE #3865: Update S4043 (`no-misleading-array-reverse`) to suggest using `toReversed` or `toSorted`
Rule S4043 currently suggests that you copy an array before using `reverse` or `sort` if you are assigning the result, with this  example code:

```js
var reversed = [...a].reverse();  // de-structure and create a new array, so reverse doesn't impact 'a'
a.reverse();

c.sort(); // this sorts array in place
```

The new methods [`toReversed`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toReversed) and [`toSorted`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted) perform a copy and then the reverse or sort, so are good options to use in place. For example:

```js
var reversed = a.toReversed(); // Use ES2023's `toReversed` to create a new array and reverse
var altReversed = [...a].reverse();  // spread the contents of the array into a new array then reverse, so reverse doesn't impact 'a'
a.reverse(); // just reverse the array in place

var sorted = c.toSorted(); // Use ES2023's `toSorted` to create a new array and sort
var altSorted = [...c].sort(); // spread the contents of the array into a new array, then sort
c.sort(); // this sorts array in place
```

yassin-kammoun-sonarsource:
Fixes #3865

### DIFF
--- a/sonar-plugin/javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S4043.html
+++ b/sonar-plugin/javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S4043.html
@@ -6,14 +6,17 @@ which is likely not what was intended.</p>
 value is altered.</p>
 <h3>Noncompliant code example</h3>
 <pre>
-var b = a.reverse(); // Noncompliant
-var d = c.sort(); // Noncompliant
+const reversed = a.reverse(); // Noncompliant
+const sorted = b.sort(); // Noncompliant
 </pre>
 <h3>Compliant solution</h3>
 <pre>
-var b = [...a].reverse();  // de-structure and create a new array, so reverse doesn't impact 'a'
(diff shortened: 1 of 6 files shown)


alexander-kamushkin-sonarsource (approved):
Maybe it's worth adding a test to check if quickfix works when there is a comparator argument to the `sort()` method.

There is also a `toSpliced()` method, which is not covered, so maybe another rule?

yassin-kammoun-sonarsource:
> There is also a `toSpliced()` method, which is not covered, so maybe another rule?

I created https://github.com/SonarSource/SonarJS/issues/3933 to extend the current scope of the rule with `splice` and `toSpliced` methods.

PR #3941, opened 2023-05-24
Improve S3796 (`array-callback-without-return`): cover new array methods, typed arrays and computed literals

### LINKED ISSUE #3885: Update S3796 (`array-callback-without-return`): cover new array functions, typed arrays and bracket notation
Extend the implementation of S3796 to support:

### Bracket notation

`array.[“every”]`

### New Array functions

1. findLast
2. findLastIndex
3. toSorted
4. flatMap

### Cover typed arrays

The current rule only supports Array, we want to support [typed arrays](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Typed_arrays) as well

RSPEC PR: https://github.com/SonarSource/rspec/pull/1946

ilia-kebets-sonarsource:
Fixes #3885


src/linting/eslint/rules/helpers/type.ts
```diff
@@ -31,6 +31,36 @@ export function isArray(node: estree.Node, services: RequiredParserServices) {
...
+ * Checks if the provided node is a JS typed array like "BigInt64Array". See TYPED_ARRAY_TYPES
+ *
+ * @param node
+ * @param services
+ * @returns
+ */
+export function isTypedArray(node: estree.Node, services: RequiredParserServices) {
+  const type = getTypeFromTreeNode(node, services);
+  return TYPED_ARRAY_TYPES.includes(type?.symbol.name);
+}
```
yassin-kammoun-sonarsource, line 62: I wonder whether we should just merge all of these with `isArray`. A typed array is an array after all. Could you maybe check whether a merge is possible without impacting negatively rules that use `isArray`? If there is such an impact, never mind my comment then.

src/linting/eslint/rules/array-callback-without-return.ts
```diff
@@ -98,6 +107,18 @@ export const rule: Rule.RuleModule = {
...
+  if (isIdentifier(property) && !isComputed) {
+    return property.name;
+    // we want arr["some"]
+  } else if (isStringLiteral(property)) {
+    return property.value;
+  } else {
+    return null;
+  }
+}
+
```
yassin-kammoun-sonarsource, line 121: Just a suggestion, but we could reuse `isDotNotation` and `isIndexNotation` and write instead:

```typescript
function extractPropName(memberExpr: estree.MemberExpression) {
  // we want arr.some(), not arr[some]()
  if (isDotNotation(memberExpr)) {
    return memberExpr.property.name;
    // we want arr["some"]
  } else if (isIndexNotation(memberExpr)) {
    return memberExpr.property.value;
  } else {
    return null;
  }
}
```

src/linting/eslint/rules/array-callback-without-return.ts
```diff
@@ -37,12 +40,16 @@ const methodsWithCallback = [
...
+  'findLast',
   'findIndex',
+  'findLastIndex',
   'map',
+  'flatMap',
   'reduce',
   'reduceRight',
   'some',
   'sort',
+  'toSorted',
```
yassin-kammoun-sonarsource, line 52: Let's update the RSPEC and mention these new covered methods.

yassin-kammoun-sonarsource:
Ah, and seems like ruling is failing for JavaScript.

src/linting/eslint/rules/helpers/type.ts
ilia-kebets-sonarsource: I would be careful with mixing these because of:


>However, typed arrays are not to be confused with normal arrays, as calling [Array.isArray()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray) on a typed array returns false. Moreover, not all methods available for normal arrays are supported by typed arrays (e.g. push and pop).

from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Typed_arrays

However, that's exactly what we're doing here...

yassin-kammoun-sonarsource: Thanks for the quote!

ilia-kebets-sonarsource: OK, so we're not doing it for all.

src/linting/eslint/rules/array-callback-without-return.ts
ilia-kebets-sonarsource: done.

src/linting/eslint/rules/array-callback-without-return.ts
ilia-kebets-sonarsource: done https://github.com/SonarSource/rspec/pull/1946

yassin-kammoun-sonarsource (approved):
LGTM

PR #3962, opened 2023-06-02
Rely completely on typescript-eslint for projects with Vue.js files

victor-diez-sonarsource:


sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/ContextUtils.java
```diff
@@ -52,6 +55,17 @@ boolean shouldSendFileContent(InputFile file) {
     return isSonarLint() || !StandardCharsets.UTF_8.equals(file.charset());
   }
 
+  static boolean canUseWildcardForTypeChecking(SensorContext context, int filesCount) {
```
saberduck, line 58: this function should not be static, it should use `context` instance field and not take `context` parameter. The point of this class is to wrap context and provide additional methods around it. You will need to change some signatures to adapt, but it's rather straightforward

sonar-plugin/sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/JsTsSensorTest.java
```diff
@@ -366,6 +370,54 @@ void stop_analysis_if_server_is_not_responding() throws Exception {
       .isEqualTo("eslint-bridge server is not answering");
   }
 
+  @Test
+  void should_use_wildcard_value_for_typechecking_in_sonarlint() throws Exception {
+    SensorContextTester ctx = createSensorContext(tempDir);
```
saberduck, line 375: use `var` for all types

sonar-plugin/sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/eslint/JsTsSensorTest.java
```diff
@@ -366,6 +370,54 @@ void stop_analysis_if_server_is_not_responding() throws Exception {
...
+    verify(eslintBridgeServerMock, times(2)).analyzeTypeScript(captor.capture());
+    assertThat(captor.getAllValues())
+      .extracting(c -> c.createWildcardTSConfig)
+      .isEqualTo(List.of(true, true));
+  }
+
+  @Test
+  void should_use_wildcard_value_for_typechecking_in_sonarqube_with_vue_below_limit()
+    throws Exception {
+    SensorContextTester ctx = createSensorContext(tempDir);
```
saberduck, line 391: Use `var`

saberduck (changes requested):

saberduck (approved):

PR #3963, opened 2023-06-05
Add rule S6643: Prototypes of builtin objects should not be modified

### LINKED ISSUE #3893: Add rule S6643: Prototypes of builtin objects should not be modified
ESLint's rule: https://eslint.org/docs/latest/rules/no-extend-native

- [ ] Use ESLint's implementation instead of our own:
- [ ] Update the RSPEC to put it back in the sonar way
- [ ] If possible, improve the location to not include the whole right side of the assignment, because it's a lot when it's a function definition

yassin-kammoun-sonarsource:
Fixes #3893


alexander-kamushkin-sonarsource:
Are we doing the right thing here as this ESLint rule does a very different thing than our current implementation? The closest equivalent to current S2424 is [no-global-assign](https://eslint.org/docs/latest/rules/no-global-assign), or [no-redeclare](https://eslint.org/docs/latest/rules/no-redeclare) / [no-shadow](https://eslint.org/docs/latest/rules/no-shadow) with globals objects lists. But this [no-extend-native](https://eslint.org/docs/latest/rules/no-extend-native) rule is not covering assign/shadow at all, it is only about modifying the prototypes?

yassin-kammoun-sonarsource:
> Are we doing the right thing here as this ESLint rule does a very different thing than our current implementation? The closest equivalent to current [S2424](https://sonarsource.github.io/rspec/#/rspec/S2424/javascript) is [no-global-assign](https://eslint.org/docs/latest/rules/no-global-assign), or [no-redeclare](https://eslint.org/docs/latest/rules/no-redeclare) / [no-shadow](https://eslint.org/docs/latest/rules/no-shadow) with globals objects lists. But this [no-extend-native](https://eslint.org/docs/latest/rules/no-extend-native) rule is not covering assign/shadow at all, it is only about modifying the prototypes?

As discussed, [no-extend-native](https://eslint.org/docs/latest/rules/no-extend-native) is doing something completely unrelated. Therefore, and as agreed, I created a dedicated new rule for the code pattern of extending JavaScript builtin objects.

Thanks for pointing that out 👍

PR #4002, opened 2023-06-29
Fix performance regression in projects with many tsconfigs

### LINKED ISSUE #3988: Performance regression in projects with many tsconfigs
Due to changes in the way we analyse projects we've noticed performance regression on projects with many `tsconfig.json` files. We will revert changes introduced in #3872

saberduck:
Fixes #3988


saberduck:
Replaced with #4015

PR #4010, opened 2023-07-06
Do not rely on external browserlists files

victor-diez-sonarsource:
As mentioned in https://babeljs.io/docs/babel-preset-env#browserslist-integration:

> By default @babel/preset-env will use [browserslist config sources](https://github.com/ai/browserslist#queries) unless either the [targets](https://babeljs.io/docs/babel-preset-env#targets) or [ignoreBrowserslistConfig](https://babeljs.io/docs/babel-preset-env#ignorebrowserslistconfig) options are set.

Babel will walk the tree looking for browserlist files. Some external browserlists will make babel fail, as Python squad saw with django-cms peach analysis.

This will avoid babel to look for those files.

### DIFF
--- a/src/parsing/jsts/parsers/options.ts
+++ b/src/parsing/jsts/parsers/options.ts
@@ -74,7 +74,7 @@ function babelParserOptions(options: Linter.ParserOptions) {
   const babelOptions = {
     targets: 'defaults',
     presets: [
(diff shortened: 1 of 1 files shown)


yassin-kammoun-sonarsource (approved):
Thanks for looking into it!

PR #4012, opened 2023-07-07
Update "submit review" and "request review" actions to work for projectsV2

ilia-kebets-sonarsource:
When merged, we would need to change the branches to @V1 like:

```yaml
      - uses: sonarsource/gh-action-lt-backlog/MoveCardToReview@V1
```

instead of 

```yaml
      - uses: sonarsource/gh-action-lt-backlog/MoveCardToReview@ use-until-merged
```

### DIFF
--- a/.github/workflows/RequestReview.yml
+++ b/.github/workflows/RequestReview.yml
@@ -9,9 +9,12 @@ jobs:
     name: Move card to review
     runs-on: ubuntu-latest
     # PRs from forks don't have required token authorization
-    if: github.event.pull_request.head.repo.full_name == github.repository
+    if: |
+        github.event.pull_request.head.repo.full_name == github.repository
+        && github.event.review.author_association != 'NONE'
     steps:
-      - uses: sonarsource/gh-action-lt-backlog/MoveCardToReview@v1
+      - uses: sonarsource/gh-action-lt-backlog/MoveCardToReviewV2@use-until-merged
         with:
           github-token: ${{secrets.GITHUB_TOKEN}}
-          column-id: 16320793     # Kanban "In review" column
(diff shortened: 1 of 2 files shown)


victor-diez-sonarsource (approved):

.github/workflows/RequestReview.yml
```diff
@@ -8,10 +8,20 @@ jobs:
   MoveCardToReview_job:
     name: Move card to review
     runs-on: ubuntu-latest
+    permissions:
+      id-token: write
     # PRs from forks don't have required token authorization
-    if: github.event.pull_request.head.repo.full_name == github.repository
+    if: |
+        github.event.pull_request.head.repo.full_name == github.repository
+        && github.event.review.author_association != 'NONE'
```
yassin-kammoun-sonarsource, line 16: What's the meaning of this extra condition ? What's an author association ?

ilia-kebets-sonarsource: it means that the reviewer has at least some rights on the repo: https://docs.github.com/en/webhooks-and-events/events/issue-event-types#properties-for-commented

yassin-kammoun-sonarsource: 👍

yassin-kammoun-sonarsource: Do you plan to document this kind of detail somewhere?

ilia-kebets-sonarsource: These actions are made to be copy-pasted from their templates, and a few parameters replaced like this one from here: https://github.com/SonarSource/gh-action-lt-backlog/tree/master/MoveCardAfterReview#projects-v2

So we don't deem the documentation necessary.

PR #4028, opened 2023-07-19
Update rule S3758 (`values-not-covertable-to-numbers`) to allow ES2020 BigInt

timicienio:
# Description 

Currently, [S3758](https://sonarsource.github.io/rspec/#/rspec/S3758) only allow comparisons between two NumberLike types, but ES2020 added BigInt ([ref](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt)), which can validly use the `comparisonOperators` defined in the implementation, but is incompatible since they are not `NumberLike` (instead, they are `BigIntLike`).

This PR modified the implementation to allow the operands to be either `NumberLike` or `BigIntLike`, and related test cases are also added.

This is the first time that I contributed to this project, plz let me know if any improvements can be made to this PR!

### DIFF
--- a/src/linting/eslint/rules/values-not-convertible-to-numbers.ts
+++ b/src/linting/eslint/rules/values-not-convertible-to-numbers.ts
@@ -98,7 +98,10 @@ function isConvertibleToNumber(typ: ts.Type, checker: ts.TypeChecker) {
   const valueOfSignatures = getValueOfSignatures(typ, checker);
   return (
     valueOfSignatures.length === 0 ||
-    valueOfSignatures.some(signature => isNumberLike(signature.getReturnType()))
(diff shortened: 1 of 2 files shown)


victor-diez-sonarsource:
Hi @timicienio,

thank you for your contribution. Sorry but right now external users cannot create PRs. We usually recommend to provide this feedback in out [community forum](https://community.sonarsource.com/)

In any case, the changes make total sense so I created a PR applying [the diff from your commit](https://github.com/SonarSource/SonarJS/commit/de3caca4e15d04730982121faede27b8b37b158b.diff). Let's see the CI results and merge if it passes.

Thanks again,
Victor

PR #4033, opened 2023-07-21
Add rule S6676 (`no-useless-call`): Calls to .call() and .apply() methods should not be redundant

### LINKED ISSUE #3905: Create rule S6676 (`no-useless-call`): Calls to .call() and .apply() methods should not be redundant
ESLint definition: https://eslint.org/docs/latest/rules/no-useless-call

Investigate effort to add quickfix that removes the `.call` and the useless `thisArg` like:

```javascript
foo.call(foo, 12)
```
into
```javascript
foo(12)
```

alexander-kamushkin-sonarsource:
Fixes #3905


victor-diez-sonarsource (approved):
very nice suite of tests, LGTM!

PR #4050, opened 2023-08-07
Fix typedoc - new folder locations

alexander-kamushkin-sonarsource:
Fix typedoc after moving folders


yassin-kammoun-sonarsource (approved):
For your information, it didn't fail on your previous PR because the corresponding GitHub action is triggered only when a commit is pushed on master (see [typedoc.yaml](https://github.com/SonarSource/SonarJS/blob/master/.github/workflows/typedoc.yml)).

PR #4051, opened 2023-08-07
Try if c6id.4xlarge instance improves win qa

alexander-kamushkin-sonarsource:
use c6id.4xlarge instance


alexander-kamushkin-sonarsource:
More CPUs did not help -  we have two tests running 19 minutes each (PRAnalysisTest, TypeCheckerConfigTest). They run from start to finish and determine the overall run time

PR #4052, opened 2023-08-08
Use bigger machine and unite the two win QA jobs

saberduck:


saberduck:
This didn't improve performance as expected

PR #4055, opened 2023-08-09
Technical direction - one folder per rule

alexander-kamushkin-sonarsource:
[One folder per rule](https://xtranet-sonarsource.atlassian.net/wiki/spaces/LANG/pages/2853273808/SonarJS+technical+direction#3.-All-rule-files-in-one-folder)


packages/jsts/src/rules/tools/index.ts
```diff
@@ -0,0 +1,20 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+export * from '../../../tests/tools';
```
ilia-kebets-sonarsource, line 20: why do you need a shortcut to test tools here?

ilia-kebets-sonarsource:
Good job, As discussed with @yassin-kammoun-sonarsource, it would be nice to move `packages/jsts/tests/linter/fixtures/wrapper/quickfixes/*` files under the rule folder, but it can be done in a later commit.

Having the Sonar rule id Sxxxx in the filename allows to quickly open its files in the IDE.

packages/jsts/src/rules/tools/index.ts
alexander-kamushkin-sonarsource: There are >100 rules referring to it, so in case we change something we don't have to update 100s files.

ilia-kebets-sonarsource (approved):
LGTM!

PR #4064, opened 2023-08-16
Fix FP for S1874 (`deprecation`): only flag symbols if all their declarations are deprecated

### LINKED ISSUE #4008: Fix FP S1874: (`deprecation`): Should not report when at least one function signature is not deprecated

ilia-kebets-sonarsource:
Fixes #4008

### DIFF
--- a/packages/jsts/src/rules/S1874/cb.fixture.ts
+++ b/packages/jsts/src/rules/S1874/cb.fixture.ts
@@ -228,3 +228,6 @@ const a = {
     this.y // OK
 	},
 };
+
+// edge case: contains a declaration with no body and no jsDoc that we should not consider.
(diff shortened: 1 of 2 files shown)


ilia-kebets-sonarsource:
I put this on hold, I can't write logic that does work well for multiple declarations without JSdoc like this one:

https://github.com/SonarSource/SonarJS/pull/4064/files#diff-78f344a014f5edc1d54b57d5082482fb9259219013dffdd48de61eb794ccd6dcR188-R193

Screenshot 2023-08-17 at 14 29 23

and for global properties that are deprecated but somehow the TS type checker finds them additional declarations without a body and jsDoc like for `window.pageXOffset`.

Screenshot 2023-08-17 at 15 37 55

yassin-kammoun-sonarsource:
Solved in #4122.

PR #4068, opened 2023-08-17
Modify rule S107: Do not count TypeScript parameter properties

### LINKED ISSUE #4014: Fix FP S107: Do not count TypeScript parameter properties
Currently, we have an exception for constructors with only parameter properties, but if we provide a few parameters among those and the total amount exceeds the limit, it gets flagged. We decide to extend the exception: when mixing both parameter properties and parameters, only count the parameters against the limit.

Issue raised in https://community.sonarsource.com/t/about-typescript-s107-dependency-injection-and-non-property-parameters/95186

yassin-kammoun-sonarsource:
Fixes #4014

### DIFF
--- a/packages/jsts/src/rules/S107/rule.ts
+++ b/packages/jsts/src/rules/S107/rule.ts
@@ -45,6 +45,7 @@ export const rule: Rule.RuleModule = {
     const ruleDecoration: Rule.RuleModule = interceptReport(
       eslintMaxParams,
       function (context: Rule.RuleContext, descriptor: Rule.ReportDescriptor) {
+        const maxParams = context.options[0] as number;
         if ('node' in descriptor) {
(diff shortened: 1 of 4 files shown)


ilia-kebets-sonarsource (approved):
Small change, otherwise LGTM!

packages/jsts/src/rules/S107/rule.ts
```diff
@@ -45,6 +45,7 @@ export const rule: Rule.RuleModule = {
     const ruleDecoration: Rule.RuleModule = interceptReport(
       eslintMaxParams,
       function (context: Rule.RuleContext, descriptor: Rule.ReportDescriptor) {
+        const maxParams = context.options[0] as number;
```
ilia-kebets-sonarsource, line 48: Can you remove the duplicate value at line 122?

PR #4069, opened 2023-08-17
Fix FP S4123 (`no-invalid-await`): do not raise issue for `unknown` type

### LINKED ISSUE #3966: Fix FP S4123 (`no-invalid-await`): do not raise issue for `unknown` type
Our current implementation does not cover `unknown` type, for which the issue should not be raised (same as `any`).
Reported in [community post](https://community.sonarsource.com/t/redundant-await-on-a-non-promise-can-give-a-false-positive/91594)

```ts
export class NoErrorThrownError extends Error {}

export class TestUtils {
    /**
     * A method to be able to get round eslint-jest
     * Avoid calling `expect` conditionally`  jest/no-conditional-expect
     *
     * @param call the function to call that an error is expected to be thrown.
     * @returns The error that was thrown or NoErrorThrownError if no error was thrown.
     */
    public static getError = async <TError>(
        call: () => PromiseLike<unknown> | unknown
    ): Promise<TError> => {
        try {
            // Do NOT remove the await here, it is needed as the call maybe awaitable.
            // The tests for this function proves it.
            await call();
            throw new NoErrorThrownError();
        } catch (error) {
            return error as TError;
        }
    };
}
`

alexander-kamushkin-sonarsource:
Fixes #3966

### DIFF
--- a/packages/jsts/src/rules/S4123/rule.ts
+++ b/packages/jsts/src/rules/S4123/rule.ts
@@ -39,7 +39,12 @@ export const rule: Rule.RuleModule = {
             (node as estree.AwaitExpression).argument,
             services,
           );
-          if (!hasThenMethod(awaitedType) && !isAny(awaitedType) && !isUnion(awaitedType)) {
+          if (
+            !hasThenMethod(awaitedType) &&
+            !isAny(awaitedType) &&
+            !isUnknown(awaitedType) &&
+            !isUnion(awaitedType)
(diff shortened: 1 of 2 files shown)


yassin-kammoun-sonarsource (commented):
Could you add a test case with the more complicated code snippet from #3966?

yassin-kammoun-sonarsource (approved):

PR #4070, opened 2023-08-17
Replace all occurrences of ESLint bridge with bridge

### LINKED ISSUE #2867: Replace all occurrences of ESLint bridge with bridge
We should stop using `eslint-bridge` (mostly in logs but also in class names), as this name poorly reflects what is behind this component (Node-based component of analyzer). Since #2854 it's also launching Stylelint for CSS and it's not only bridge as it's used to be, it contains a lot of logic, mainly rule implementation (based on ESLint API).

yassin-kammoun-sonarsource:
Fixes #2867


ilia-kebets-sonarsource (approved):
LGTM!

PR #4072, opened 2023-08-17
Update sonar-plugin-api

saberduck:


sonar-plugin/sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/lcov/CoverageSensorTest.java
```diff
@@ -181,39 +182,38 @@ void should_ignore_and_log_warning_for_invalid_line() {
...
       File.separator + "reports" + File.separator + "report_with_unresolved_path.lcov";
-    assertThat(logTester.logs(LoggerLevel.WARN))
+    assertThat(logTester.logs(Level.WARN))
       .contains(
         "Could not resolve 2 file paths in [" + moduleBaseDir.getAbsolutePath() + fileName + "]"
       )
       .contains(
         "First unresolved path: unresolved/file1.js (Run in DEBUG mode to get full list of unresolved paths)"
       );
-    assertThat(logTester.logs(LoggerLevel.DEBUG))
```
ilia-kebets-sonarsource, line 198: Is this assertion not relevant anymore?

ilia-kebets-sonarsource (approved):
LGTM, just 2 small questions about removed assertions.

sonar-plugin/sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/lcov/CoverageSensorTest.java
saberduck: This was removed, because new logging system will only catch logs on the level which is configured (`INFO` here), and there is conditional behavior in the logging which is being tested. There is `debug` level test just below

sonar-plugin/javascript-checks/src/test/java/org/sonar/javascript/checks/CheckListTest.java
```diff
@@ -64,21 +62,16 @@ void test() {
...
       assertThat(
-        getClass()
-          .getResource("/org/sonar/l10n/javascript/rules/javascript/" + rule.getKey() + ".html")
+        getClass().getResource("/org/sonar/l10n/javascript/rules/javascript/" + key + ".html")
       )
-        .overridingErrorMessage("No description for " + rule.getKey())
+        .overridingErrorMessage("No description for " + key)
         .isNotNull();
-
-      assertThat(rule.getDescription())
```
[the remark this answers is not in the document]
saberduck: This was testing that we don't use the deprecated method of adding descriptions to the test via annotation on the Java class. Since the way we load metadata has changed (`AnnotationRuleParser` was removed), this test didn't make sense anymore

PR #4074, opened 2023-08-18
Disable ucfg rule for JS embedded in HTML files

### LINKED ISSUE #4036: Do not generate UCFGs for JavaScript code inside HTML
Generating correct UCFGs for HTML would require a job similar to [what was done for YAML in sonar-security](https://github.com/SonarSource/sonar-security/commit/14251a6e51d210d268fa71abbac40e4996d03227).

We are wrongly generating UCFGs for JavaScript code embedded inside HTML files. This is due to the fact that the analysis of JavaScript code inside HTML is reusing the same logic as the analysis of JavaScript code inside YAML. Since we haven't implemented yet the support for analyzing injection vulnerabilities of JavaScript code inside HTML, we are needlessly generating UCFGs which not only waste resources on the security analyzer but also fill the JavaScript analyzer's cache for nothing. 

Incidentally, these useless generated UCFGs for JS inside HTML cause failures when reading back these UCFGs from the analyzer's cache. The root cause of these failures is a separate issue to be investigated further.

https://community.sonarsource.com/t/getting-error-failure-when-reading-cache-entry-with-github-actions-but-scan-completes-as-success/95769/3

ilia-kebets-sonarsource:
Fixes #4036

### DIFF
--- a/its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/HtmlSecurityTest.java
+++ b/its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/HtmlSecurityTest.java
@@ -1,0 +1,106 @@
+/*
+ * SonarQube JavaScript Plugin
+ * Copyright (C) 2012-2024 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package com.sonar.javascript.it.plugin;
+
+import static com.sonar.javascript.it.plugin.OrchestratorStarter.JAVASCRIPT_PLUGIN_LOCATION;
+import static com.sonar.javascript.it.plugin.OrchestratorStarter.getSonarScanner;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import com.sonar.javascript.it.plugin.assertj.BuildResultAssert;
+import com.sonar.orchestrator.Orchestrator;
+import com.sonar.orchestrator.build.SonarScanner;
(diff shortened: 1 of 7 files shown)


yassin-kammoun-sonarsource (changes requested):
I think we need an integration test for this change.

ilia-kebets-sonarsource:
Dropping this, because we haven't found the source of the issue that was initially raised in community.

its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/HtmlSecurityTest.java
```diff
@@ -0,0 +1,106 @@
...
+
+    var result = orchestrator.executeBuild(getScanner(projectPath, projectKey));
+    assertThat(result.isSuccess()).isTrue();
+
+    var stream = Files.find(
+      projectPath.toPath().resolve(".scannerwork"),
+      3,
+      BuildResultAssert::isUcfgFile
+    );
+    assertThat(stream.toList().isEmpty()).isTrue();
```
yassin-kammoun-sonarsource, line 64: Can we get rid of `isTrue` and move `isEmpty`?

its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/HtmlSecurityTest.java
```diff
@@ -0,0 +1,106 @@
...
+      .addPlugin(JAVASCRIPT_PLUGIN_LOCATION)
+      .setEdition(Edition.DEVELOPER)
+      .activateLicense()
+      .addPlugin(MavenLocation.of("com.sonarsource.security", "sonar-security-plugin", "DEV"))
+      .addPlugin(
+        MavenLocation.of("com.sonarsource.security", "sonar-security-js-frontend-plugin", "DEV")
+      )
+      .addPlugin(
+        MavenLocation.of("org.sonarsource.config", "sonar-config-plugin", "LATEST_RELEASE")
+      )
```
yassin-kammoun-sonarsource, line 82: I think we can remove this because it was introduced for YAML language detection.

its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/YamlSecurityTest.java
```diff
@@ -0,0 +1,105 @@
...
+
+    var result = orchestrator.executeBuild(getScanner(projectPath, projectKey));
+    assertThat(result.isSuccess()).isTrue();
+
+    var stream = Files.find(
+      projectPath.toPath().resolve(".scannerwork"),
+      3,
+      BuildResultAssert::isUcfgFile
+    );
+    assertThat(stream.toList().size() == 1).isTrue();
```
yassin-kammoun-sonarsource, line 64: To be verified, but I think there should be an assertion method `hasSize`.

yassin-kammoun-sonarsource (approved):
Great job writing those integration tests.
I left some suggestions about a few assertions.
Looks good to me otherwise!

PR #4075, opened 2023-08-18
Do not analyse JSON files returned on TypeScript program creation

### LINKED ISSUE #4054: Do not analyse JSON files returned on TypeScript program creation
If a TSConfig instructs the TypeScript compiler to resolve JSON files with [resolveJsonModule](https://www.typescriptlang.org/tsconfig#resolveJsonModule) because those are imported in the codebase, they will be considered during the compilation process. As a result, they will be scanned as well by the JS/TS analyzer and issues will be raised accordingly since a JSON file is nothing more than a JavaScript object. Analyzing JSON files is of little interest. Therefore, we should exclude those from the list of files returned on TypeScript programs creation.

Reported in https://community.sonarsource.com/t/json-file-error-expected-an-assignment-or-function-call-and-instead-saw-an-expression/96301/10

yassin-kammoun-sonarsource:
Fixes #4054


yassin-kammoun-sonarsource:
Checked with a local SonarQube instance on the community user repository 👍

ilia-kebets-sonarsource (approved):
LGTM!

Issue #4047, opened 2023-08-04
Fix FP S1128 ('unused-import'): Consider all names of Vue.js elements


yassin-kammoun-sonarsource:
Despite several attempts to reduce the noise of unused imports for Vue Single File Components, the rule still reports false positives for some family of elements. In this particular case, it is reporting one-word lowercase names of elements. It is only considering names matching either snake-case or pascal-case case sensitivity types. We should just avoid these cumbersome checks and just pick any string naming a Vue.js element, that is, a node of type `VElement`. 

Reported in https://community.sonarsource.com/t/remove-this-unused-import-of-xxx-there-seems-to-be-something-wrong-with-this-rule/96184

PR #4077, opened 2023-08-21
Fix FP S3827 (`no-reference-error`): Ignore Vue3 compiler macros

### LINKED ISSUE #3780: Fix FP S3827 (`no-reference-error`): Ignore Vue3 `defineProps()`, `defineEmits()` and `defineExpose()`
When using Vue3 with JavaScript, in a `<script setup>` section, Vue provides compiler macros to configure the component: `defineProps()`, `defineEmits()`, and `defineExpose()`.
Rule S3827 should check the AST to ignore these functions if they are inside a Vue file (that's to say, inside a `VElement` with a `VStartTag` having a `VAttribute` named `setup`)

Reported in https://community.sonarsource.com/t/on-the-vue3-project-defineprops-and-defineemits-should-be-automatically-available-in-script-setup/84965

alexander-kamushkin-sonarsource:
Fixes #3780

### DIFF
--- a/packages/jsts/src/rules/S3827/rule.ts
+++ b/packages/jsts/src/rules/S3827/rule.ts
@@ -19,10 +19,19 @@
  */
 // https://sonarsource.github.io/rspec/#/rspec/S3827/javascript
 import { TSESTree } from '@typescript-eslint/experimental-utils';
+import { AST } from 'vue-eslint-parser';
 import { Rule } from 'eslint';
 import * as estree from 'estree';
 import { findFirstMatchingAncestor, toEncodedMessage } from '../helpers';
 import { SONAR_RUNTIME } from '../../linter/parameters';
+
+const vueMacroNames = new Set([
+  'defineProps',
+  'defineEmits',
+  'defineExpose',
+  'defineOptions',
+  'defineSlots',
+]);
 
 export const rule: Rule.RuleModule = {
   meta: {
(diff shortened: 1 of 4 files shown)


packages/jsts/src/rules/S3827/rule.ts
```diff
@@ -83,3 +97,31 @@ function hasTypeOfOperator(node: TSESTree.Node) {
   const parent = node.parent;
   return parent?.type === 'UnaryExpression' && parent.operator === 'typeof';
 }
+
+function isVueMacro(identifier: TSESTree.Identifier, block: AST.ESLintProgram) {
```
yassin-kammoun-sonarsource, line 101: FYI

```typescript
const services = context.parserServices;
const fragment = services.getDocumentFragment();
const script = fragment.children.find(child => child.type === 'VElement' && child.name === 'script');
```

yassin-kammoun-sonarsource (approved):

packages/jsts/src/rules/S3827/rule.ts
alexander-kamushkin-sonarsource: Moved `isInsideVueSetupScript` to helpers because it is also needed in S1128 (#4047)

yassin-kammoun-sonarsource (approved):
Thanks for extracting the helper function. LGTM!

PR #4081, opened 2023-08-21
Update orchestrator

saberduck:


its/ruling/pom.xml
```diff
@@ -28,8 +28,7 @@
     </dependency>
     <dependency>
       <groupId>org.sonarsource.orchestrator</groupId>
-      <artifactId>sonar-orchestrator</artifactId>
-      <version>${sonar-orchestrator.version}</version>
```
victor-diez-sonarsource, line 32: no need for version? if not specified it takes parent version I'm guessing?

victor-diez-sonarsource (approved):

its/ruling/pom.xml
saberduck: yes, this is inherited from parent, this was redundant

PR #4089, opened 2023-08-23
Fix references to srs and tests folders in main pom.xml

alexander-kamushkin-sonarsource:
replace src, tests with packages


ilia-kebets-sonarsource (approved):
LGTM!

PR #4108, opened 2023-08-25
Fix S6535 (unnecessary-character-escapes): Disambiguate conflicting message ids

### LINKED ISSUE #3987: Fix S6535 (`unnecessary-character-escapes`): Disambiguate conflicting message ids
When using a [`Regexp()` constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/RegExp) and providing a string as parameter, you need to escape special characters like character classes with 2 backslashes `\\`.
Using strings is necessary when you want to make the regex pattern dynamic like:

```js
new RegExp(`\d ${foo}`);
```

It can also happen, that the user started writing his regex in a literal, and copied it to a constructor when making it dynamic. In this case, the rule will flag the backslash with the message `Unnecessary escape character: \d.`
This can be confusing if you were indeed wishing to escape it correctly. There a quickfix in the `no-useless-escape` ESLint implementation that we use that proposes to add the missing backslash, but the error message could be improved to hint to it.

We could also make it clearer in the RSPEC.

The quickfix message in SL for VS code seems to fail to interpolate the variable name:

Screenshot 2023-06-27 at 10 55 05

### Special cases

We should take into account cases like that:

```js
const pattern="\d";
const regex = new Regexp(pattern);
```

yassin-kammoun-sonarsource:
Fixes #3987 

S6535 implementation consists of merging ESLint rules `no-useless-escape` and `no-nonoctal-decimal-escape`. However, both share a common message id `escapeBackslash` with a different description for their respective suggestions. Because this wasn't noticed at the time of merging the rules, the message id `escapeBackslash` from `no-nonoctal-decimal-escape` used to overwrite the one from `no-useless-escape`. This is why not only the suggestion messages were sometimes confusing but also why the data interpolation of some suggestions didn't occur.

### DIFF
--- a/packages/jsts/src/rules/S6535/rule.ts
+++ b/packages/jsts/src/rules/S6535/rule.ts
@@ -35,14 +35,14 @@ const noUselessEscapeRule = eslintRules['no-useless-escape'];
 const noNonoctalDecimalEscapeRule = eslintRules['no-nonoctal-decimal-escape'];
 
 /**
- * We replace the message id 'escapeBackslash' of 'no-nonoctal-decimal-escape' with 'otherEscapeBackslash'.
+ * We replace the message id 'escapeBackslash' of 'no-nonoctal-decimal-escape' with 'nonOctalEscapeBacklash'.
  */
-noNonoctalDecimalEscapeRule.meta!.messages!['otherEscapeBackslash'] =
+noNonoctalDecimalEscapeRule.meta!.messages!['nonOctalEscapeBacklash'] =
   noNonoctalDecimalEscapeRule.meta!.messages!['escapeBackslash'];
 delete noNonoctalDecimalEscapeRule.meta!.messages!['escapeBackslash'];
 
 /**
- * We decorate 'no-nonoctal-decimal-escape' to map suggestions with the message id 'escapeBackslash' to 'otherEscapeBackslash'.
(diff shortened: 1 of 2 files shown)


packages/jsts/src/rules/S6535/rule.ts
```diff
@@ -21,24 +21,58 @@
...
+ * To prevent one overwritting the other, we need to decorate one and map the conflicting message
+ * id to a different one when intercepting a report.
+ *
+ * Here we arbitrarily choose to decorate 'no-nonoctal-decimal-escape'.
+ */
 const noUselessEscapeRule = eslintRules['no-useless-escape'];
 const noNonoctalDecimalEscapeRule = eslintRules['no-nonoctal-decimal-escape'];
 
+/**
+ * We replace the message id 'escapeBackslash' of 'no-nonoctal-decimal-escape' with 'otherEscapeBackslash'.
```
ilia-kebets-sonarsource, line 38: Maybe rename it `nonOctalEscapeBacklash`?

ilia-kebets-sonarsource (approved):
LGTM, but one suggestion.

packages/jsts/src/rules/S6535/rule.ts
yassin-kammoun-sonarsource: 👍

PR #4110, opened 2023-08-25
Add Node.js 20 as a possible recommended version

### LINKED ISSUE #4013: Update deprecation message to support Node v20
We should update the deprecation message to support v20 and we should deprecate v16.
We should keep support for v14 and v16, so we will have two deprecated versions as a result.

We should also add v20 to the CI axis to run integration tests against it.

yassin-kammoun-sonarsource:
Fixes #4013


sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/eslint/NodeDeprecationWarning.java
```diff
@@ -56,8 +56,8 @@ public class NodeDeprecationWarning {
    */
   static final Version MIN_SUPPORTED_NODE_VERSION = Version.create(14, 17, 0);
   static final int MIN_RECOMMENDED_NODE_VERSION = 16;
-  static final List<Integer> RECOMMENDED_NODE_VERSIONS = Arrays.asList(16, 18);
-  static final List<Integer> ALL_RECOMMENDED_NODE_VERSIONS = Arrays.asList(14, 16, 18);
+  static final List<Integer> RECOMMENDED_NODE_VERSIONS = Arrays.asList(16, 18, 20);
+  static final List<Integer> ALL_RECOMMENDED_NODE_VERSIONS = Arrays.asList(14, 16, 18, 20);
```
yassin-kammoun-sonarsource, line 60: I wonder whether we should get rid of these lists and rework the deprecation logic as well as the logs so that we only test whether the actual node version is an even number greater or equal than the current value of `MIN_RECOMMENDED_NODE_VERSION`. This means we would no longer tell explicitly which versions are recommended. Users would need to read the analyzer documentation to figure out which one, which will redirect them to Node.js LTS page.

Maybe it's not worth the effort since we will soon embed Node.js runtime. What do you think?

saberduck: I think it's not worth the effort with our current plans

saberduck (approved):

saberduck:
@yassin-kammoun-sonarsource you need to rebase, I am not sure what changed

yassin-kammoun-sonarsource:
> @yassin-kammoun-sonarsource you need to rebase, I am not sure what changed

I renamed a package in a previous commit. Nothing serious.

PR #4119, opened 2023-08-28
Fix S2699 (`assertions-in-tests`): infinite loop

ilia-kebets-sonarsource:


packages/jsts/src/rules/S2699/rule.ts
```diff
@@ -31,12 +31,13 @@ import { Sinon } from '../helpers/sinon';
  */
 export const rule: Rule.RuleModule = {
   create(context: Rule.RuleContext) {
+    const visitedNodes = new Map();
```
yassin-kammoun-sonarsource, line 34: Is there a reason you preferred using a map over a set?

ilia-kebets-sonarsource: no, haven't thought of the set. Great idea!

yassin-kammoun-sonarsource (approved):

PR #4122, opened 2023-08-30
Improve S1874 (`deprecation`): Report deprecations from TypeScript compiler

### LINKED ISSUE #4008: Fix FP S1874: (`deprecation`): Should not report when at least one function signature is not deprecated

yassin-kammoun-sonarsource:
Fixes #4008 

Building from #4064, it turns out quite tricky to cover every single case of deprecated symbols that are used. In fact, after looking at TypeScript's codebase, there are many places where usages of deprecated APIs are detected and reported. This is due to several edge cases. Instead of replicating the logic for every single one, we propose to extract and propagate deprecation suggestions emitted by TypeScript's type checker. There are minor functional differences compared to our current implementation, like the fact that TypeScript reports on import statements. However, with these changes, we behave precisely like TypeScript compilers, thus reducing false positives and negatives.

### DIFF
--- a/packages/jsts/src/rules/S1874/cb.fixture.ts
+++ b/packages/jsts/src/rules/S1874/cb.fixture.ts
@@ -9,7 +9,7 @@ function foo(strings: TemplateStringsArray, ...values: any[]): string {
 foo``;
 foo`${'foo'}`;
 foo`${42}`;
-foo`${[1, 2, 3]}`; // Noncompliant FN
+foo`${[1, 2, 3]}`; // Noncompliant
 
 /* separator */
 
@@ -18,7 +18,7 @@ interface X {
     foo(pp: string): number;
 }
 function bar(x: X, p: string) {
-    x.foo(p); // Noncompliant {{'foo' is deprecated. use bar instead}}
+    x.foo(p); // Noncompliant {{The signature '(pp: string): number' of 'x.foo' is deprecated.}}
 //    ^^^
 }
 
@@ -81,7 +81,7 @@ let deprecatedVar;
 const const1 = 1,
     const2 = 2;
 
-h(const1 + const2); // Noncompliant
+h(const1 + const2); // Noncompliant 2
 export default const1;// Noncompliant
 
 /* separator */
@@ -93,7 +93,7 @@ function fn() { }
 
 fn<number>();
 fn(1); // Noncompliant
-h(fn); // Noncompliant
+h(fn);
 
 /* separator */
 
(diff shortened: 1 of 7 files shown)


packages/jsts/src/rules/S1874/cb.fixture.ts
```diff
@@ -142,9 +142,9 @@ deprecatedCallSignature2(); // Noncompliant
 /* separator */
 
 import * as allDeprecations from './cb.fixture.deprecations';
-z(allDeprecations.deprecatedFunction);// Noncompliant
+z(allDeprecations.deprecatedFunction);
 
-import defaultImport, {deprecatedFunction, anotherDeprecatedFunction as aliasForDeprecated, notDeprecated1, notDeprecated2} from './cb.fixture.deprecations';
+import defaultImport, {deprecatedFunction, anotherDeprecatedFunction as aliasForDeprecated, notDeprecated1, notDeprecated2} from './cb.fixture.deprecations'; // Noncompliant 2
```
yassin-kammoun-sonarsource, line 147: TypeScript reports on imported deprecated symbols, which we didn't previously.

packages/jsts/src/rules/S1874/cb.fixture.ts
```diff
@@ -81,7 +81,7 @@ let deprecatedVar;
 const const1 = 1,
     const2 = 2;
 
-h(const1 + const2); // Noncompliant
+h(const1 + const2); // Noncompliant 2
```
yassin-kammoun-sonarsource, line 84: We had a false negative before on `const2`.

packages/jsts/src/rules/S1874/cb.fixture.ts
```diff
@@ -93,7 +93,7 @@ function fn() { }
 
 fn<number>();
 fn(1); // Noncompliant
-h(fn); // Noncompliant
+h(fn);
```
yassin-kammoun-sonarsource, line 96: TypeScript doesn't report deprecated symbols used as references, like here.

packages/jsts/src/rules/S1874/cb.fixture.ts
```diff
@@ -155,8 +155,32 @@ noDeprecated2(); // OK
...
+/** @deprecated */ function allDeprecated(a: string): string;
+/** @deprecated */ function allDeprecated(a: number): number;
+function allDeprecated(a: number | string): number | string {
+    if (typeof a === 'number') return a + 1;
+    return a;
+}
+
+allDeprecated('yolo'); // Noncompliant
+allDeprecated(42); // Noncompliant
+allDeprecated; // OK
```
yassin-kammoun-sonarsource, line 186: Taken from https://github.com/SonarSource/SonarJS/pull/4064

victor-diez-sonarsource (approved):
Really good job!

PR #4123, opened 2023-08-31
Fix FP S4123 (`no-invalid-await`): Improve thenable type detection

### LINKED ISSUE #4113: Fix FP S4123 (`no-invalid-await`): don't flag when function returns a thenable object
When using `await foo()` and `foo()` returns a thenable object, a FP issue is raised by rule S4123.

Reproducer:

```ts
class Foo {
  then: Promise<Bar>;
}
function qux(): Foo {}
const baz = await qux();
```

Issue reported on community ticket: https://community.sonarsource.com/t/potential-false-positive-typescript-rule-s4123-with-mongoose-6-x/97482

yassin-kammoun-sonarsource:
Fixes #4113

### DIFF
--- a/packages/jsts/src/rules/S4123/rule.ts
+++ b/packages/jsts/src/rules/S4123/rule.ts
@@ -40,7 +40,7 @@ export const rule: Rule.RuleModule = {
             services,
           );
           if (
-            !hasThenMethod(awaitedType) &&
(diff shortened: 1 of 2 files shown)


victor-diez-sonarsource (approved):
LGTM!

PR #4120, opened 2023-08-29
Use embedded NodeJS runtime in the Plugin

ilia-kebets-sonarsource:

PR #4127, opened 2023-09-04
Bundle node

ilia-kebets-sonarsource:


ilia-kebets-sonarsource:
this PR replaces https://github.com/SonarSource/SonarJS/pull/4120.
Since we're doing cascaded branches and PRs, we shouldn't use squash and merge between them as it will break the diff.

PR #4146, opened 2023-09-05
Add rule S6747 (`no-unknown-property`): JSX elements should not use unknown properties and attributes

### LINKED ISSUE #4144: Create rule for `react/no-unknown-property​`: Disallow usage of unknown DOM property
- [Reference](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-unknown-property.md)
- [Implementation](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/lib/rules/no-unknown-property.js)
- [Tests](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/tests/lib/rules/no-unknown-property.js)

has quickfix

yassin-kammoun-sonarsource:
Fixes #4144


packages/jsts/tests/linter/wrapper.test.ts
```diff
@@ -17,11 +17,11 @@
  * along with this program; if not, write to the Free Software Foundation,
```
yassin-kammoun-sonarsource, line 17: The changes on this file were required to also consider `.jsx` files triggering quick fixes.

sonar-plugin/javascript-checks/src/main/java/org/sonar/javascript/checks/NoUnknownPropertyCheck.java
```diff
@@ -0,0 +1,66 @@
...
+import org.sonar.plugins.javascript.api.EslintBasedCheck;
+import org.sonar.plugins.javascript.api.JavaScriptRule;
+import org.sonar.plugins.javascript.api.TypeScriptRule;
+
+@JavaScriptRule
+@TypeScriptRule
+@Rule(key = "S6747")
+public class NoUnknownPropertyCheck implements EslintBasedCheck {
+
+  @RuleProperty(
```
yassin-kammoun-sonarsource, line 37: In general, I would avoid as much as possible to include properties to a rule. However, if we don't provide this feature, I am afraid too many users will complain about false positives.

Please let me know if that makes sense to you, or if we should drop that property :)

alexander-kamushkin-sonarsource: Yes, I agree that it is necessary - that was my first thought when I was reading the rule source.

alexander-kamushkin-sonarsource (approved):
LGTM

PR #4147, opened 2023-09-06
Add rule S6749 (`jsx-no-useless-fragment`): Redundant React fragments should be removed

### LINKED ISSUE #4135: Create rule for `react/jsx-no-useless-fragment​`: Disallow unnecessary fragments
- [Reference](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-useless-fragment.md)
- [Implementation](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/lib/rules/jsx-no-useless-fragment.js)
- [Tests](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/tests/lib/rules/jsx-no-useless-fragment.js)

yassin-kammoun-sonarsource:
Fixes #4135


alexander-kamushkin-sonarsource (changes requested):
I guess we need explicitly activate the quickfix, correct?

PR #4150, opened 2023-09-07
[MMF-2861] Embed Node.js runtime in analyzer

saberduck:


pom.xml
```diff
@@ -229,6 +230,13 @@
         <scope>test</scope>
       </dependency>
 
+      <!-- Tooling dependencies -->
```
ilia-kebets-sonarsource, line 233: remove this comment: it's used in the plugin for decompression as well

sonar-plugin/sonar-javascript-plugin/pom.xml
```diff
@@ -169,7 +305,7 @@
               <rules>
                 <requireFilesSize>
                   <minsize>19000000</minsize>
-                  <maxsize>26000000</maxsize>
+                  <maxsize>160000000</maxsize>
```
ilia-kebets-sonarsource, line 308: we should probably change that back to something meaningful, and maybe update the minimum value.

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/EmbeddedNode.java
```diff
@@ -0,0 +1,269 @@
...
+  }
+
+  /**
+   * @return a path to `DEPLOY_LOCATION` from the given `baseDir`
+   */
+  private static Path runtimeCachePathFrom(String baseDir) {
+    return Path.of(baseDir).resolve(DEPLOY_LOCATION);
+  }
+
+  public boolean isAvailable() {
```
ilia-kebets-sonarsource, line 142: is this method still necessary?

saberduck: yes, for example when we don't find the runtime inside the jar this will return false

sonar-plugin/sonar-javascript-plugin/pom.xml
saberduck: ok, I will push a fix for this

ilia-kebets-sonarsource (approved):
approving again to attempt to trigger "promote" check

PR #4155, opened 2023-09-12
Create rule S6757 (`react/no-this-in-sfc​​`): `this` should not be used in functional components

### LINKED ISSUE #4133: Create rule S6757 (`react/no-this-in-sfc​​`): `this` should not be used in functional components
- [Reference](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-this-in-sfc.md)
- [Implementation](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/lib/rules/no-this-in-sfc.js)
- [Tests](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/tests/lib/rules/no-this-in-sfc.js)

alexander-kamushkin-sonarsource:
Fixes #4133


its/ruling/src/test/expected/js/javascript-test-sources/javascript-S6757.json
```diff
@@ -0,0 +1,19 @@
+{
```
yassin-kammoun-sonarsource, line 1: Since these all seem like false positives, let's add a true positive code example in https://github.com/SonarSource/SonarJS/tree/master/its/sources/file-for-rules.

alexander-kamushkin-sonarsource: Done

yassin-kammoun-sonarsource (approved):

PR #4157, opened 2023-09-12
Add rule S6759 (`prefer-read-only-props`): React props should be read-only

### LINKED ISSUE #4130: Create rule for `react/prefer-read-only-props` Enforce that props are read-only
- [Reference](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/prefer-read-only-props.md)
- [Implementation](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/lib/rules/prefer-read-only-props.js)
- [Tests](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/tests/lib/rules/prefer-read-only-props.js)

has quickfix

yassin-kammoun-sonarsource:
Fixes #4130


package.json
```diff
@@ -69,7 +69,7 @@
     "builtin-modules": "3.3.0",
     "bytes": "3.1.2",
     "eslint": "8.47.0",
-    "eslint-plugin-react": "7.32.2",
+    "eslint-plugin-react": "7.33.2",
```
yassin-kammoun-sonarsource, line 72: A recent version includes TypeScript support for `prefer-read-only-props` (see https://github.com/jsx-eslint/eslint-plugin-react/commit/9c5ac98c506de10d48044740ee4f6bb48886f67e)

alexander-kamushkin-sonarsource (approved):
LGTM

PR #4161, opened 2023-09-12
Cache node runtimes downloads in CI and download from artifactory if possible

ilia-kebets-sonarsource:


tools/fetch-node/node-distros.mjs
```diff
@@ -9,16 +9,19 @@ export default [
   {
     id: 'win-x64',
     url: `https://nodejs.org/dist/${NODE_VERSION}/node-${NODE_VERSION}-win-x64.zip`,
+    artifactoryUrl: `https://repox.jfrog.io/artifactory/nodejs-dist/${NODE_VERSION}/node-${NODE_VERSION}-win-x64.zip`,
```
saberduck, line 12: wdyt about having the url in a variable instead repeating 3 times?

PR #4162, opened 2023-09-13
Build artifacts

saberduck:


tools/fetch-node/scripts/directories.mjs
```diff
@@ -43,5 +43,5 @@ export const TARGET_DIR = path.join(
   'sonar-plugin',
   'sonar-javascript-plugin',
   'target',
-  'classes',
+  'node',
```
ilia-kebets-sonarsource, line 46: don't you need to adapt the plugin code to fetch it in the new folder?

ilia-kebets-sonarsource (approved):
small question, otherwise LGTM

PR #4164, opened 2023-09-13
Implement authentication to download runtimes from Artifactory

ilia-kebets-sonarsource:


tools/fetch-node/scripts/fetch-node.mjs
```diff
@@ -201,3 +221,29 @@ function removeExtension(filename) {
...
+  }
+
+  function retrieveForDevMachine() {
+    const npmrcFile = path.join(os.homedir(), '.npmrc');
+    if (!fs.existsSync(npmrcFile)) {
+      console.log(`NPM RC file ${npmrcFile} does not exist.`);
+      return;
+    }
+    const npmrcContent = fs.readFileSync(npmrcFile, 'utf-8');
+    const secondLine = npmrcContent?.split('\n')[1];
```
ilia-kebets-sonarsource, line 246:
```suggestion
    const secondLine = npmrcContent?.split('\n')?[1];
```

tools/fetch-node/scripts/fetch-node.mjs
```diff
@@ -201,3 +221,29 @@ function removeExtension(filename) {
...
+    return secondLine?.split('authToken=')[1];
```
ilia-kebets-sonarsource, line 247:
```suggestion
    return secondLine?.split('authToken=')?[1];
```

ilia-kebets-sonarsource: actually we catch it so not important

saberduck (approved):

PR #4182, opened 2023-09-19
Fix FP S6759 (`prefer-read-only-props`): Restrict the scope to functional components

### LINKED ISSUE #4175: Fix FP S6759 (`prefer-read-only-props`): Restrict the scope to functional components
As reported [here](https://discuss.sonarsource.com/t/rule-typescript-s6759-produces-a-lot-of-false-positives/15817), S6759 (currently borrowing the implementation of [prefer-read-only-props](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/prefer-read-only-props.md)) raises false positives on React class components. The scope of the rule should be restricted to React functional components only. The rule should report on such components that neither use TypeScript's utility type [Readonly](https://www.typescriptlang.org/docs/handbook/utility-types.html) nor pure read-only props interfaces/type aliases. The corresponding RSPEC should be updated accordingly.

yassin-kammoun-sonarsource:
Fixes #4175


packages/jsts/src/rules/S6759/rule.ts
```diff
@@ -0,0 +1,193 @@
...
+          typeAnnotation?: TSESTree.TSTypeAnnotation;
+        };
+        if (!typeAnnotation) {
+          return;
+        }
+
+        /* Read-only props */
+        if (!isReadOnly(props, services)) {
+          context.report({
+            node: props,
```
yassin-kammoun-sonarsource, line 83: Not sure whether we should report on the whole parameter or only on the type annotation. WDYT ?

alexander-kamushkin-sonarsource: IMHO whole parameter is ok

alexander-kamushkin-sonarsource (approved):
LGTM!

PR #4185, opened 2023-09-20
Use Node 18 for build

saberduck:


saberduck:
Dropped. I will do this on master later

PR #4188, opened 2023-09-20
Improve coverage & fix code smells

saberduck:
Fixes #

### DIFF
--- a/sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/EmbeddedNode.java
+++ b/sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/EmbeddedNode.java
@@ -52,10 +52,10 @@ public class EmbeddedNode {
   public static final String VERSION_FILENAME = "version.txt";
   private static final long EXTRACTION_LOCK_WAIT_TIME_MILLIS = 10000;
   private static final Logger LOG = Loggers.get(EmbeddedNode.class);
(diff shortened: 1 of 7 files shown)


sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/EmbeddedNode.java
```diff
@@ -151,7 +151,7 @@ public boolean isAvailable() {
    */
   public void deploy() throws IOException {
     LOG.debug("Detected os: {} arch: {} platform: {}", env.getOsName(), env.getOsArch(), platform);
-    if (platform == UNSUPPORTED || isAvailable) {
```
ilia-kebets-sonarsource, line 154: should we remove this variable altogether?

ilia-kebets-sonarsource (approved):
LGTM, we can do this change in the feature branch: https://github.com/SonarSource/SonarJS/pull/4188#discussion_r1332517616

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/EmbeddedNode.java
saberduck: `isAvailable` is needed to know that deployment was successful and embeddednode can be used. We rely on it in `NodeCommandBuilder`

PR #4189, opened 2023-09-21
Fix windows build issues

### LINKED ISSUE #3997: Fix windows build issues
Current build scripts require admin privileges on windows, rely on unix-type line ending and require presence of unix command-line tools. We should fix this for the benefit of the future newjoiners and our reputation as an open source project.
Investigate/rewrite failing tests. Replace unix utils with node packages.

alexander-kamushkin-sonarsource:
Removes unix commands from the build script (see #3997)


ilia-kebets-sonarsource (approved):
LGTM!

PR #4193, opened 2023-09-21
Fix FP S6749 (`jsx-no-useless-fragment`): Ignore empty React fragments

### LINKED ISSUE #4192: Fix FP S6749 (`jsx-no-useless-fragment`): Ignore empty React fragments
Empty React fragments might be conditionally returned. As a result, it makes sense not to consider those as redundant and avoid reporting them.

yassin-kammoun-sonarsource:
Fixes #4192

### DIFF
--- a/packages/jsts/src/rules/S6749/decorator.ts
+++ b/packages/jsts/src/rules/S6749/decorator.ts
@@ -1,0 +1,46 @@
+/*
+ * SonarQube JavaScript Plugin
+ * Copyright (C) 2011-2023 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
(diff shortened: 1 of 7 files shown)


packages/jsts/src/rules/S6749/decorator.ts
```diff
@@ -0,0 +1,46 @@
...
+  };
+  return interceptReport(rule, (context, descriptor) => {
+    const { messageId, node, ...rest } = descriptor as any;
+
+    /* Ignore empty fragments */
+    if (node.type === 'JSXFragment' && node.children.length === 0) {
+      return;
+    }
+
+    context.report({ messageId, node, ...rest });
```
alexander-kamushkin-sonarsource, line 44: Looks like you are not changing anything in the descriptor, should it be `context.report(descriptor);` ?

alexander-kamushkin-sonarsource (approved):

packages/jsts/src/rules/S6749/decorator.ts
yassin-kammoun-sonarsource: Indeed 👍

PR #4198, opened 2023-09-22
Reduce the technical debt on the overall code

yassin-kammoun-sonarsource:


sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/AnalysisWithProgram.java
```diff
@@ -75,12 +73,11 @@ void analyzeFiles(List<InputFile> inputFiles, List<String> tsConfigs) throws IOE
           continue;
         }
         monitoring.startProgram(tsConfig);
-        PROFILER.startInfo("Creating TypeScript program");
```
victor-diez-sonarsource, line 78: aren't we losing the profiler information?

victor-diez-sonarsource (approved):
just some minor comments, for you to judge

packages/jsts/src/rules/helpers/decorators/interceptor.ts
```diff
@@ -65,11 +65,11 @@ export function interceptReport(
           physicalFilename: originalContext.physicalFilename,
 
           getCwd(): string {
-            return originalContext.getCwd();
+            return originalContext.cwd;
```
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: Just tried but it's not possible in the end. Since the interceptor context has type`Rule.RuleContext`, we still need to keep those methods. They are deprecated, not dropped :/

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/AnalysisWithProgram.java
yassin-kammoun-sonarsource: Indeed. I put it back with a handmade implementation. Could you review this commit [5d6d19a](https://github.com/SonarSource/SonarJS/pull/4198/commits/5d6d19a9b6ce23123acd1e958a03213514520155)?

yassin-kammoun-sonarsource: Never mind. It just doesn't work with SonarLint. I will just revert everything related to Profiler.

PR #4200, opened 2023-09-25
Update rule count

yassin-kammoun-sonarsource:
The updated numbers are based on those on Peach, i.e., the current master. Those from [rules.sonarsource.com](rules.sonarsource.com) will be updated once SonarJS is released.


ilia-kebets-sonarsource (approved):
I didn't count them but I trust that you did.

PR #4201, opened 2023-09-26
Move css rule to a separate folder

alexander-kamushkin-sonarsource:
Remaining item for 'each rule in a separate folder' topic


ilia-kebets-sonarsource (approved):
LGTM. As discussed, check with a local SQ instance that the scope changes do not miss files.

PR #4206, opened 2023-09-27
Optimize CI: run Sonar analysis separately from build task

ilia-kebets-sonarsource:


.cirrus/analyze.bash
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+# inspired by https://github.com/SonarSource/re-ci-images/blob/b9f43ffec926c0828a279aefc5a94cb5060bfee5/docker/bin/src/regular_mvn_build_deploy_analyze#L51-L61
+
+mvn sonar:sonar \
+      -Pcoverage,deploy-sonarsource,release,sign \
```
ilia-kebets-sonarsource, line 6: I'm not sure about which of these profiles are useful here

.cirrus.yml
```diff
@@ -80,27 +80,49 @@ build_task:
...
     PGP_PASSPHRASE: VAULT[development/kv/data/sign data.passphrase]
   <<: *MAVEN_CACHE
   <<: *RUNTIME_CACHE
+  build_and_deploy_script:
+    - source cirrus-env BUILD
+    - npm config set registry "${ARTIFACTORY_URL}/api/npm/npm"
+    - npm run build
+    - ./.cirrus/prepare_deploy_analyze.bash
+    # we've already run the tests above
+    - mvn deploy -DskipTests -Pcoverage,deploy-sonarsource,release,sign -B -e -V "$@"
```
ilia-kebets-sonarsource, line 96: I'm not sure about which of these profiles are useful here

.cirrus/analyze.bash
```diff
@@ -0,0 +1,15 @@
...
+      -Dmaven.test.redirectTestOutputToFile=false \
```
ilia-kebets-sonarsource, line 7: I believe we can remove this since we don't run tests during `sonar:sonar`

.cirrus/prepare_deploy_analyze.bash
ilia-kebets-sonarsource, line 1: I've copied bits from `SonarSource/re-ci-images/` to this script. 
- Would you know a nice way to maintain this?
- Should I modularize code from `SonarSource/re-ci-images/` instead?

.cirrus.yml
```diff
@@ -80,27 +80,49 @@ build_task:
...
     folder: ${HOME}/.sonar/cache
     fingerprint_script:
       - echo ${CIRRUS_OS}
       - curl --silent ${SONAR_HOST_URL}/deploy/plugins/index.txt | sort
-  build_script:
-    - source cirrus-env BUILD
+  build_and_analyze_script:
+    - source cirrus-env CI
     - npm config set registry "${ARTIFACTORY_URL}/api/npm/npm"
     - npm run build
```
ilia-kebets-sonarsource, line 123: instead of running tests twice, should we cache the coverage results between build and analyze tasks?

.cirrus.yml
saberduck: I think you can ommit `coverage` , because it's only needed for analysis

.cirrus.yml
saberduck: not sure if this is worth the complexity

.cirrus/analyze.bash
saberduck: you probably don't need `deploy-sonarsource,release,sign`

.cirrus/analyze.bash
saberduck: I think it might cause more issues when we don't run the tests

.cirrus/prepare_deploy_analyze.bash
saberduck: imo it's ok like this.

saberduck (approved):

PR #4213, opened 2023-09-28
Remove unneeded direct dependency `decompress-targz`

saberduck:
Mend reports [CVE-2020-12265](https://www.mend.io/vulnerability-database/CVE-2020-12265) for `decompress-targz` because it has "vulnerable" version 4.1.1 . `decompress-targz` is not needed directly, it's included transitively by `decompress` which has "non-vulnerable" version `4.2.1`


ilia-kebets-sonarsource (approved):
LGTM

PR #4218, opened 2023-09-28
Replace `decompress` lib with `compressing` in fetch-node script because of vulnerability

ilia-kebets-sonarsource:


ilia-kebets-sonarsource:
closing this. It works for fetch-node, but doesn't for [check-distribution-length.js](https://github.com/SonarSource/SonarJS/blob/fdf1762d994be64845cdcbd3ac36066f622b6ea9/tools/check-distribution-filepath-length.js#L21-L45)

Issue #4221, opened 2023-09-29
Add support for Sass syntax


yassin-kammoun-sonarsource:
The CSS analyzer does not support the [Sass](https://sass-lang.com/) syntax. Currently, if we try to analyze such syntax from `.sass` files or from the script section (`<script lang="sass">`) of Vue.js single file components, the analysis fails with the following error:

```
INFO: Sensor CSS Rules [javascript]
INFO: X source files to be analyzed
ERROR: Error: Cannot find module 'postcss-sass'
ERROR: Require stack:
ERROR: - /some/project/.scannerwork/.sonartmp/bridge-bundle/package/node_modules/postcss-syntax/get-syntax.js
ERROR: - /some/project/.scannerwork/.sonartmp/bridge-bundle/package/node_modules/postcss-syntax/parser.js
ERROR: - /some/project/.scannerwork/.sonartmp/bridge-bundle/package/node_modules/postcss-syntax/parse.js
ERROR: - /some/project/.scannerwork/.sonartmp/bridge-bundle/package/node_modules/postcss-syntax/index.js
ERROR: - /some/project/.scannerwork/.sonartmp/bridge-bundle/package/node_modules/stylelint/lib/getPostcssResult.js
ERROR: - /some/project/.scannerwork/.sonartmp/bridge-bundle/package/node_modules/stylelint/lib/lintSource.js
ERROR: - /some/project/.scannerwork/.sonartmp/bridge-bundle/package/node_modules/stylelint/lib/postcssPlugin.js
ERROR: - /some/project/.scannerwork/.sonartmp/bridge-bundle/package/node_modules/stylelint/lib/index.js
ERROR: - /some/project/.scannerwork/.sonartmp/bridge-bundle/package/lib/css/src/linter/wrapper.js
ERROR: - /some/project/.scannerwork/.sonartmp/bridge-bundle/package/lib/css/src/linter/index.js
ERROR: - /some/project/.scannerwork/.sonartmp/bridge-bundle/package/lib/css/src/analysis/analyzer.js
ERROR: - /some/project/.scannerwork/.sonartmp/bridge-bundle/package/lib/css/src/analysis/index.js
ERROR: - /some/project/.scannerwork/.sonartmp/bridge-bundle/package/lib/css/src/index.js
ERROR: - /some/project/.scannerwork/.sonartmp/bridge-bundle/package/lib/bridge/src/router/on-analyze-css.js
[...]

PR #4229, opened 2023-10-02
Add PostCSS JSX syntax support for Stylelint

ilia-kebets-sonarsource:
Fixes `peachee-java-kotlin` slang project `solidus`


yassin-kammoun-sonarsource (changes requested):
Let's add an integration test making sure that we are able to analyze files that use that particular syntax so that we know for a fact that we are no longer crashing. Also, while you are at it, could you also fix this related ticket https://github.com/SonarSource/SonarJS/issues/4221?

ilia-kebets-sonarsource:
closing this because we decided https://github.com/SonarSource/SonarJS/issues/4230

PR #4261, opened 2023-10-10
Delegate analysis-related HTTP requests to a worker thread

### LINKED ISSUE #4253: Delegate analysis-related HTTP requests to a worker thread

yassin-kammoun-sonarsource:
Fixes #4253


packages/shared/src/helpers/context.ts
```diff
@@ -57,6 +55,5 @@ export function getContext(): Context {
  * @param ctx the new global context
  */
 export function setContext(ctx: Context) {
-  debug(`Context set with ${JSON.stringify(ctx)}`);
```
yassin-kammoun-sonarsource, line 60: I decided to remove the log when setting the "context" for the following reasons:

- This iwas logged twice: once in the main thread from `bin/server` and once in the worker thread `worker.js`
- The log included a stringified object that looks ugly and doesn't seem worth printing
- This is redundant with what's logged when we launch the Node.js command from `NodeCommand.java`
- We are logging enough (too many) things already during analysis

Please let me know if you think we should keep this (and eventually improve it)

packages/bridge/tests/router.test.ts
```diff
@@ -66,10 +67,10 @@ describe('router', () => {
   });
 
   it('should route /analyze-js requests', async () => {
-    initializeLinter([
+    await requestInitLinter(server, [
```
yassin-kammoun-sonarsource, line 70: This unit test (and those that follow) was initializing the linter in the main thread by directly calling the initialization function. Since this test suite is testing the server routes with HTTP requests (through the worker), we need to initialize the linter on the worker with an initial HTTP request.

packages/bridge/src/worker.js
```diff
@@ -0,0 +1,181 @@
...
+  writeTSConfigFile,
+} = require('@sonar/jsts');
+const { readFile, setContext } = require('@sonar/shared/helpers');
+const { analyzeCSS } = require('@sonar/css');
+const { analyzeHTML } = require('@sonar/html');
+const { analyzeYAML } = require('@sonar/yaml');
+const { APIError, ErrorCode } = require('@sonar/shared/errors');
+
+/**
+ * Delegate the handling of an HTTP request to a worker thread
```
yassin-kammoun-sonarsource, line 40: Please let me know if reading the code is enough to understand the "protocol" between the main thread and the worker. Should I document it here explicitly?

packages/bridge/src/worker.js
```diff
@@ -0,0 +1,181 @@
...
+exports.delegate = function (worker, type) {
+  return async (request, response, next) => {
+    worker.once('message', message => {
+      switch (message.type) {
+        case 'success':
+          if (typeof message.result === 'object') {
+            response.json(message.result);
+          } else {
+            response.send(message.result);
+          }
```
yassin-kammoun-sonarsource, line 51: We could avoid this condition by having the worker stringifying `message.result` upstream and making the main thread send the response as plain text with `response.send` in all cases. However, we would lose the `Content-Type` set by `response.json`.

Is it fine like this? Do you think we should avoid serializing the result twice in JSON?

packages/bridge/src/worker.js
```diff
@@ -0,0 +1,181 @@
...
+const { readFile, setContext } = require('@sonar/shared/helpers');
+const { analyzeCSS } = require('@sonar/css');
+const { analyzeHTML } = require('@sonar/html');
+const { analyzeYAML } = require('@sonar/yaml');
+const { APIError, ErrorCode } = require('@sonar/shared/errors');
+
+/**
+ * Delegate the handling of an HTTP request to a worker thread
+ */
+exports.delegate = function (worker, type) {
```
yassin-kammoun-sonarsource, line 42: Feel free to suggest a better function name.

its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/HtmlAnalysisTest.java
```diff
@@ -76,7 +76,7 @@ void should_raise_issues_in_html_files() throws IOException {
         tuple(4, "javascript:S3923"),
         tuple(7, "javascript:S3834")
       );
-    assertPerfMonitoringAvailable(perfMonitoringDir);
+    // assertPerfMonitoringAvailable(perfMonitoringDir);
```
yassin-kammoun-sonarsource, line 79: Will be investigated separately.

packages/shared/src/helpers/context.ts
yassin-kammoun-sonarsource, line 1: I decided to remove the log when setting the "context" for the following reasons:

- This was logged twice: once in the main thread from `bin/server` and once in the worker thread `worker.js`
- The log included a stringified object that looks ugly and doesn't seem worth printing
- This is redundant with what's logged when we launch the Node.js command from `NodeCommand.java`
- We are logging enough (too many) things already during analysis

Please let me know if you think we should keep this (and eventually improve it)

packages/bridge/tests/router.test.ts
yassin-kammoun-sonarsource, line 1: This test file includes unit tests that were initializing the linter in the main thread by directly calling the initialization function. Since this test suite is testing the server routes with HTTP requests (through the worker), we need to initialize the linter on the worker with an initial HTTP request.

packages/bridge/src/worker.js
yassin-kammoun-sonarsource, line 1: Please let me know if the "protocol" between the main thread and the worker is easy to understand with a single read or if I should document it.

packages/bridge/src/errors/middleware.ts
yassin-kammoun-sonarsource, line 1: Changes here are the consequence of forwarding exceptions from the worker thread to the main thread. Will be clearer while reading the worker thread's implementation.

packages/bridge/src/worker.js
```diff
@@ -0,0 +1,181 @@
...
+ * Code executed by the worker thread
+ */
+if (parentPort) {
+  setContext(workerData.context);
+
+  const parentThread = parentPort;
+  parentThread.on('message', async message => {
+    try {
+      const { type, data } = message;
+      if (data?.filePath && !data.fileContent) {
```
saberduck, line 72: I find it a bit confusing that we do this for all messages - evenif I understand that `filePath` will be undefined. I think it will be a bit cleaner to have this logic only with the endpoints where it actually matters

packages/shared/src/helpers/context.ts
saberduck: ok, makes sense

saberduck (approved):

packages/bridge/src/worker.js
yassin-kammoun-sonarsource: 👍

PR #4268, opened 2023-10-12
Add rule S6821 (`jsx-a11y/aria-role`): DOM elements with ARIA roles should have a valid non-abstract role

### LINKED ISSUE #4238: Create rule S6821 (`jsx-a11y/aria-role`): DOM elements with ARIA roles should have a valid non-abstract role
[plugin rule](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/aria-role.md#jsx-a11yaria-role)

https://github.com/SonarSource/rspec/pull/3268

ilia-kebets-sonarsource:
Fixes #4238


its/ruling/src/test/expected/jsts/fireact/javascript-S6821.json
```diff
@@ -0,0 +1,15 @@
+{
+"fireact:src/App.js": [
```
ilia-kebets-sonarsource, line 2: this and the one below have nothing to do with ARIA, I think they simply use `role` as attribute name for their thing. This could be noisy on Peach.

alexander-kamushkin-sonarsource (approved):
LGTM

PR #4294, opened 2023-10-17
Report heap usage status with `sonar.javascript.node.debugMemory`

### LINKED ISSUE #4256: Report heap usage status conditionally
- Enabled conditionally with a property (too noisy in the logs otherwise)
- Reporting at some significant points or periodically
- Register on Node.js garbage collection hooks
- Warn critical heap usage status
- Purely on Node.js

saberduck:
Fixes #4256

### DIFF
--- a/bin/server
+++ b/bin/server
@@ -22,7 +22,10 @@ const shouldUseTypeScriptParserForJS = process.argv[5] !== 'false';
 const sonarlint = process.argv[6] === 'true';
 const debugMemory = process.argv[7] === 'true';
 
-const bundles = process.argv[8]?.split(path.delimiter) ?? [];
+let bundles = [];
+if (process.argv[8]) {
+  bundles = process.argv[8].split(path.delimiter);
+}
 
 context.setContext({ workDir, shouldUseTypeScriptParserForJS, sonarlint, debugMemory, bundles });
 server.start(Number.parseInt(port), host).catch(() => {});
--- a/bin/server
+++ b/bin/server
@@ -20,11 +20,12 @@ const host = process.argv[3];
 const workDir = process.argv[4];
 const shouldUseTypeScriptParserForJS = process.argv[5] !== 'false';
 const sonarlint = process.argv[6] === 'true';
+const debugMemory = process.argv[7] === 'true';
 
 let bundles = [];
-if (process.argv[7]) {
-  bundles = process.argv[7].split(path.delimiter);
+if (process.argv[8]) {
+  bundles = process.argv[8].split(path.delimiter);
 }
 
-context.setContext({ workDir, shouldUseTypeScriptParserForJS, sonarlint, bundles });
+context.setContext({ workDir, shouldUseTypeScriptParserForJS, sonarlint, debugMemory, bundles });
 server.start(Number.parseInt(port), host).catch(() => {});
--- a/bin/server
+++ b/bin/server
@@ -22,10 +22,7 @@ const shouldUseTypeScriptParserForJS = process.argv[5] !== 'false';
 const sonarlint = process.argv[6] === 'true';
 const debugMemory = process.argv[7] === 'true';
 
-let bundles = [];
-if (process.argv[8]) {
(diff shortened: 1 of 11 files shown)


packages/bridge/src/worker.js
```diff
@@ -109,6 +110,7 @@ if (parentPort) {
 
         case 'on-create-program': {
           const { tsConfig } = data;
+          logHeapStatistics();
```
yassin-kammoun-sonarsource, line 113: Just to be sure that it's intentional, logs will only be printed when analyzing with programs, not with watch programs, right?

We might face a weird situation where SonarLint users see the log about out-of-memory suggesting to enable the property `sonar.javascript.node.debugMemory`. However, even if they try to enable it, they won't see heap status.

Maybe the following line should be executed conditionally, when `getContext().sonarlint === false`:

https://github.com/SonarSource/SonarJS/blob/0ef574916ab3967f136589cc02453bdeeba861d1/packages/bridge/src/memory.ts#L70-L72

What do you think?

packages/bridge/src/memory.ts
```diff
@@ -83,3 +85,26 @@ export function logMemoryError(err: any) {
...
   }
 }
+
+export function registerGarbageCollectionObserver() {
+  const obs = new PerformanceObserver(items => {
+    items
+      .getEntries()
+      .filter(
+        item =>
+          (item.detail as NodeGCPerformanceDetail).kind === constants.NODE_PERFORMANCE_GC_MAJOR,
```
yassin-kammoun-sonarsource, line 95: Should we use optional chaining here as a precaution since `detail` can be nullable (`NodeGCPerformanceDetail | unknown | undefined`)? Or are you certain that it will never be null?

```suggestion
          (item.detail as NodeGCPerformanceDetail)?.kind === constants.NODE_PERFORMANCE_GC_MAJOR,
```

packages/bridge/tests/server.test.ts
```diff
@@ -146,7 +146,7 @@ describe('server', () => {
     await new Promise(r => setTimeout(r, 600));
     expect(server.listening).toBeFalsy();
 
-    expect(console.log).toHaveBeenCalledWith('DEBUG The bridge server shutted down');
+    expect(console.log).toHaveBeenCalledWith('DEBUG The bridge server shut down');
```
yassin-kammoun-sonarsource, line 149: Thanks!

bin/server
```diff
@@ -20,11 +20,12 @@ const host = process.argv[3];
 const workDir = process.argv[4];
 const shouldUseTypeScriptParserForJS = process.argv[5] !== 'false';
 const sonarlint = process.argv[6] === 'true';
+const debugMemory = process.argv[7] === 'true';
 
 let bundles = [];
```
yassin-kammoun-sonarsource, line 25: Maybe we could get rid of the if-statement and replace these three lines with:

```javascript
const bundles = process.argv[8]?.split(path.delimiter) ?? [];
```

packages/shared/src/helpers/context.ts
```diff
@@ -30,6 +30,7 @@ export interface Context {
   workDir: string;
   shouldUseTypeScriptParserForJS: boolean;
   sonarlint: boolean;
+  debugMemory?: boolean;
```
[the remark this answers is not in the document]
saberduck: exactly, it's created in many places.

packages/bridge/src/worker.js
saberduck: yes, you are right, however, it's quite a corner case, I would not overcomplicate this

packages/shared/src/helpers/context.ts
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: 👍

packages/bridge/src/worker.js
yassin-kammoun-sonarsource: 👍

yassin-kammoun-sonarsource (approved):
Nice work!

PR #4309, opened 2023-10-24
Tag RSPECs of JS/TS rules relying on type information

### LINKED ISSUE #4190: Tag RSPECs of JS/TS rules relying on type information
Users currently have no way of knowing which JavaScript rules rely on type information. This information matters for SonarLint users when type-checking of JavaScript files is deactivated due to project size. By tagging these rules, users would then be able to identify which rules are impacted when type information is missing.

yassin-kammoun-sonarsource:
Fixes #4190


ilia-kebets-sonarsource (approved):
The rules with the tag have been reviewed in the RSPEC repo. I'm only verifying the DEV.md

Issue #4187, opened 2023-09-20
Fix project size calculation for JavaScript type-checking in SonarLint context


yassin-kammoun-sonarsource:
For performance reasons, the type-checking of JavaScript files in SonarLint context is limited according to the project size, that is, the number of files supposed to be analyzed. However, it currently does not rely on the [file-filtering logic of the analyzer](https://github.com/SonarSource/SonarJS/blob/master/sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/filter/JavaScriptExclusionsFileFilter.java). As a result, files from `node_modules` (among other things) are incorrectly included in the counting.

Reported in https://discuss.sonarsource.com/t/sonarlint-not-showing-javascript-issues-when-certain-dependancies-are-included/15800

PR #4311, opened 2023-10-24
Improve logging of JavaScript files type-checking in SonarLint context

### LINKED ISSUE #4186: Improve logging of JavaScript files type-checking in SonarLint context
Depends on #4190

The type-checking of JavaScript files in SonarLint context is deactivated when the project size exceeds a given threshold. Currently, users are warned about this deactivation with the following log:

```
Project type checking for JavaScript files deactivated as project has too many files (maximum is 20000 files)
````

The log should emphasize the consequence of type-checking deactivation: JavaScript rules using type information might not behave as expected.

Furthermore, since the log mentions the value of the threshold, it should also mention the actual project size so that users know to what value they could increase the threshold if necessary.

### PR DESCRIPTION
**Pull request by yassin-kammoun-sonarsource**
Fixes #4186

### DIFF
--- a/sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/SonarLintJavaScriptProjectChecker.java
+++ b/sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/SonarLintJavaScriptProjectChecker.java
@@ -56,43 +56,40 @@ public class SonarLintJavaScriptProjectChecker implements JavaScriptProjectCheck
 
   private void checkLimit(SensorContext context) {
     try {
-      var start = Instant.now();
-      var maxFilesForTypeChecking = getMaxFilesForTypeChecking(context);
-      long cappedFileCount = countFiles(context, maxFilesForTypeChecking);
+      var typeCheckingLimit = getTypeCheckingLimit(context);
+      var projectSize = countProjectSize(context);
 
-      beyondLimit = cappedFileCount >= maxFilesForTypeChecking;
+      beyondLimit = projectSize >= typeCheckingLimit;
       if (!beyondLimit) {
-        LOG.debug(
-          "Project type checking for JavaScript files activated as project size is below limit (total number of files is {}, maximum is {})",
-          cappedFileCount,
-          maxFilesForTypeChecking
-        );
+        LOG.info("Turning on type-checking of JavaScript files");
       } else {
         // TypeScript type checking mechanism creates performance issues for large projects. Analyzing a file can take more than a minute in
         // SonarLint, and it can even lead to runtime errors due to Node.js being out of memory during the process.
-        LOG.debug(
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/SonarLintJavaScriptProjectChecker.java [hunk 1]
```diff
@@ -56,51 +56,47 @@ public void checkOnce(SensorContext context) {
...
+          "Please be aware that this could potentially impact the performance of the analysis"
         );
-        // We can't inform the user of the actual number of files as the performance impact may be too high for large projects.
-        LOG.debug("Update \"{}\" to set a different limit.", MAX_FILES_PROPERTY);
       }
-
-      LOG.debug(
-        "Project checker duration took: {}ms",
-        Duration.between(start, Instant.now()).toMillis()
-      );
```
yassin-kammoun-sonarsource, line 84: Didn't see the point of logging this.

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/SonarLintJavaScriptProjectChecker.java
```diff
@@ -56,51 +56,47 @@ public void checkOnce(SensorContext context) {
...
-  private static long countFiles(SensorContext context, int maxFilesForTypeChecking) {
+  private static long countProjectSize(SensorContext context) {
     var isPluginFile = Pattern.compile("\\.(js|cjs|mjs|jsx|ts|cts|mts|tsx|vue)$").asPredicate();
 
     try (var files = walkProjectFiles(context)) {
       return files
         .filter(Files::isRegularFile)
         .map(path -> path.getFileName().toString())
         .filter(isPluginFile)
-        .limit(maxFilesForTypeChecking)
```
victor-diez-sonarsource, line 103: why removing the limit? if I understand correctly, the limit would avoid continue counting once the threshold has been reached, so if the project size is 1M files, this limit protects us from walking the whole project tree

victor-diez-sonarsource (approved):
LGTM, just a doubt about removing `limit` and thus not short-circuiting the stream after threshold has been reached, I think we should leave it

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/SonarLintJavaScriptProjectChecker.java
yassin-kammoun-sonarsource: Yeah, I am not entirely sure about this change. When I was investigating some issues from a customer report, I realized that we were computing the project size incorrectly (see https://github.com/SonarSource/SonarJS/issues/4187). However, it took me quite some time to realize that, Knowing the number of files would have helped in the investigation.

I will revert the change.

PR #4315, opened 2023-10-25
Include copying of Node.js runtimes to the plugin in `build-fast` script

ilia-kebets-sonarsource:


yassin-kammoun-sonarsource (approved):
Tested and worked great! Thanks.

PR #4316, opened 2023-10-25
Ignore file size limits for CSS

### LINKED ISSUE #4234: Fix confusion over `sonar.javascript.maxFileSize` being applied on CSS files
The `sonar.javascript.maxFileSize` option is applied at the analyzer level for JS, TS, and CSS.

At least 1 customer was confused that his big CSS file was not analyzed until he changed the aforementioned option which is labelled for JS.

We need to investigate if the file size limit is necessary for CSS, by analyzing a big file for example.
- If yes, we introduce a dedicated option like `sonar.css.maxFileSize`
- otherwise, we don't apply this limitation for CSS files

~~We can also improve the current behaviour by logging clearly when a file is excluded from the analysis because of its size.~~

This issue was raised in https://community.sonarsource.com/t/problems-parsing-css-file/101595

ilia-kebets-sonarsource:
Fixes #4234


sonar-plugin/sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/filter/JavaScriptExclusionsFileFilterTest.java
```diff
@@ -180,6 +180,9 @@ void should_exclude_huge_files() {
       assertThat(filter.accept(inputFile("name.java", syntheticJavaFileContent(size, "Foo"))))
         .withFailMessage("Wrong result for size " + size)
```
ilia-kebets-sonarsource, line 181: I don't understand why this case expects a fail message, since `filter.accept(...)` is also expected to return `true`

sonar-plugin/sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/filter/JavaScriptExclusionsFileFilterTest.java
```diff
@@ -180,6 +180,9 @@ void should_exclude_huge_files() {
...
         .isTrue();
+
+      // we ignore size limits for CSS files
+      assertThat(filter.accept(inputFile("name.css", syntheticJsFileContent(size)))).isTrue();
```
ilia-kebets-sonarsource, line 185: I reused `syntheticJsFileContent()`, because the file content doesn't matter here.

sonar-plugin/sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/filter/JavaScriptExclusionsFileFilterTest.java
saberduck: I think that only provides message when assertions fails

saberduck:
I would keep assesors agnostic of the language of the file. 
I believe this logic should be in the filter itself and not in the assessor, ie here https://github.com/SonarSource/SonarJS/blob/ac74f3d875397f0b62e272352af4baf2b8ba7bcc/sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/filter/JavaScriptExclusionsFileFilter.java#L61

We should create different set of assesors for CSS and JS/TS files, and simply not include size assesor in the CSS files. Keep it mind that order of "assessing" matters -> we should do cheap exclusions based on file properties before touching the content of the file.

saberduck (changes requested):
see comment above

ilia-kebets-sonarsource:
I was wondering if it is problematic that the filter scope is [only covering CssLanguage here](https://github.com/SonarSource/SonarJS/pull/4316/files#diff-28dde6e2d26aa37a519a3ee653f370200e5705d1dd012a3a92e1a0631d47e828R69) contrary to how the [CSSRuleSensor filters it](https://github.com/SonarSource/SonarJS/blob/f600f32768137b1433ee6df6c16df147eba2b213/sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/CssRuleSensor.java#L211-L226).

saberduck:
> I was wondering if it is problematic that the filter scope is [only covering CssLanguage here](https://github.com/SonarSource/SonarJS/pull/4316/files#diff-28dde6e2d26aa37a519a3ee653f370200e5705d1dd012a3a92e1a0631d47e828R69) contrary to how the [CSSRuleSensor filters it](https://github.com/SonarSource/SonarJS/blob/f600f32768137b1433ee6df6c16df147eba2b213/sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/CssRuleSensor.java#L211-L226).

IMO, it's okay like this. We don't need to filter `html` or other files. We can adjust if we discover there is an issue with such setup.

saberduck (approved):

PR #4317, opened 2023-10-25
Fix project size calculation for JavaScript type-checking in SonarLint context

### LINKED ISSUE #4187: Fix project size calculation for JavaScript type-checking in SonarLint context

yassin-kammoun-sonarsource:
Fixes #4187

### DIFF
--- a/sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/sonarlint/SonarLintTypeCheckingFilter.java
+++ b/sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/sonarlint/SonarLintTypeCheckingFilter.java
@@ -1,0 +1,116 @@
+/*
+ * SonarQube JavaScript Plugin
+ * Copyright (C) 2011-2023 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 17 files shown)


sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/sonarlint/SonarLintTypeCheckingFilter.java
```diff
@@ -0,0 +1,125 @@
...
+
+    @Override
+    public boolean test(Path path) {
+      return extensions.stream().anyMatch(ext -> path.toString().endsWith(ext));
+    }
+  }
+
+  static class PathFilter implements Predicate<Path> {
+
+    private static final String[] EXCLUSIONS_DEFAULT_VALUE = new String[] {
```
vdiez, line 86: probably we should move this to `JavaScriptPlugin` class, as it's also used in PathAccessor?

vdiez (approved):

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/sonarlint/SonarLintTypeCheckingFilter.java
yassin-kammoun-sonarsource: Good point!

PR #4320, opened 2023-10-26
Fix FP S3800 (`function-return-type`): Make the exception of returning at least one `this` instead of all `this`

### LINKED ISSUE #4251: Fix FP S3800 (`function-return-type`): Make the exception of returning "this" work
The rule reports functions with multiple return statements of different types. Also, it is supposed to make exceptions to functions whose one of its return statements returns `this`. However, the exception is not working, and a false positive is raised on the following snippet:

```javascript
String.prototype.foo = function()
{
    if (this.length == 10)
        return this.substring(5);
    
    return this;
}
```

Reported in https://community.sonarsource.com/t/javascript-s3800-fp-when-working-on-prototype-function/102118

ilia-kebets-sonarsource:
Fixes #4251

### DIFF
--- a/its/ruling/src/test/expected/jsts/ace/javascript-S3800.json
+++ b/its/ruling/src/test/expected/jsts/ace/javascript-S3800.json
@@ -11,7 +11,6 @@
 9484
 ],
 "ace:lib/ace/mode/javascript/jshint.js": [
-7902,
 18292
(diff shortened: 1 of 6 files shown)


yassin-kammoun-sonarsource (approved):
LGTM!

PR #4321, opened 2023-10-27
Filter out files from node_modules on program creation

### LINKED ISSUE #4297: Filter out files from node_modules on program creation
There are files returned on program creation that are not part of the project. To raise awareness about those, we log which files will be ignored since they don't belong to any program. However, some of those files should just be ignored completely without any logs:

```
DEBUG: File not part of the project: '/tmp/project/node_modules/typescript/lib/lib.d.ts'
DEBUG: File not part of the project: '/tmp/project/node_modules/typescript/lib/lib.es5.d.ts'
DEBUG: File not part of the project: '/tmp/project/node_modules/typescript/lib/lib.dom.d.ts'
```

yassin-kammoun-sonarsource:
Fixes #4297


vdiez (approved):
LGTM!!

PR #4326, opened 2023-10-30
Fix FP S4023 (`no-empty-interface`): Ignore when extending external interfaces

### LINKED ISSUE #2089: Fix FP S4023 (`no-empty-interface`): Ignore when extending external interfaces
S4023 is derived from typescript-eslint/no-empty-interface. That rule has the ability to "silence warnings about extending a single interface without adding additional members". For example in the following code:

```ts
interface Array<T> extends Ember.ArrayPrototypeExtensions<T> { }
```

This should be allowed, or at least configurable.

See also: https://github.com/typescript-eslint/typescript-eslint/issues/1640

yassin-kammoun-sonarsource:
Fixes #2089


ilia-kebets-sonarsource (approved):
LGTM!

PR #4336, opened 2023-10-30
Fix `fetch-node` scripts for fresh machine, especially Windows

ilia-kebets-sonarsource:
prerequisite for https://github.com/SonarSource/SonarJS/pull/4333

### DIFF
--- a/tools/fetch-node/src/main/java/org/sonarsource/javascript/XZ.java
+++ b/tools/fetch-node/src/main/java/org/sonarsource/javascript/XZ.java
@@ -24,6 +24,10 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import org.tukaani.xz.LZMA2Options;
 import org.tukaani.xz.XZOutputStream;
 
@@ -35,10 +39,20 @@ public class XZ {
     if (args.length == 0) {
       throw new IllegalArgumentException("Please provide at least 1 filename to compress using XZ");
     }
+    List<String> filenames = Collections.emptyList();
     try {
-      compress(args);
+      filenames =
+        Stream
+          .of(args)
+          .map(filename -> filename.replaceAll("[\\\\/]+", "/"))
+          .collect(Collectors.toList());
+
+      compress(filenames);
     } catch (IOException e) {
-      throw new IllegalStateException("Error while compressing " + Arrays.toString(args), e);
+      throw new IllegalStateException(
+        "Error while compressing " + Arrays.toString(filenames.toArray()),
+        e
+      );
     }
   }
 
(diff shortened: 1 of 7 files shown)


ilia-kebets-sonarsource:
That's the paths failing on Windows: https://github.com/SonarSource/SonarJS/blob/be2ebd92bba88464a59611f8c8eca6f54be67a33/tools/fetch-node/pom.xml#L49-L52

```
Caused by: java.io.FileNotFoundException: File C:\Windows\SystemTemp\cirrus-ci-build\tools\fetch-node/downloads/runtimes/darwin-arm64/node does not exist.
```

tools/fetch-node/src/main/java/org/sonarsource/javascript/XZ.java
```diff
@@ -36,7 +39,12 @@ public static void main(String[] args) {
       throw new IllegalArgumentException("Please provide at least 1 filename to compress using XZ");
     }
     try {
-      compress(args);
+      List<String> filenames = Stream
```
Wohops, line 42: Looking at this, I'm not sure to understand why you need to do a first pass out of this. Why not simply rename each file in the loop of the "compress" method, since anyway you loop on everything?

tools/fetch-node/src/main/java/org/sonarsource/javascript/XZ.java
```diff
@@ -48,10 +56,10 @@ public static void main(String[] args) {
    * @param filenames
    * @throws IOException
    */
-  public static void compress(String[] filenames) throws IOException {
+  public static void compress(List<String> filenames) throws IOException {
     for (var filename : filenames) {
-      System.out.println("Compressing " + filename);
       var file = Path.of(filename);
```
Wohops, line 61: Instead of looping twice on the files, I would simply add here:
``` 
String normalizedFilename = filename.replaceAll("[\\\\/]+", "/"));
var file = Path.of(normalizedFilename);
```
Or even simpler
```
var file = Path.of(filename.replaceAll("[\\\\/]+", "/"));
```

Wohops (commented):
Some java observations.

tools/fetch-node/src/main/java/org/sonarsource/javascript/XZ.java
ilia-kebets-sonarsource: we forgot to use it, but we made the change here so the error message below is coherent: https://github.com/SonarSource/SonarJS/pull/4336/files#diff-6ccd72addd938c293a5bec03b66365bc8bae4a6fd6f96ebd51d92a8765fb2eb9R49

tools/fetch-node/src/main/java/org/sonarsource/javascript/XZ.java
ilia-kebets-sonarsource: We will never have >10 files here, so readability is more important than performance, hence we opted to separate the operations on 2 lines.

vdiez (approved):

PR #4337, opened 2023-10-31
Fix regex-related rules: Skip issues with inconsistent locations

### LINKED ISSUE #4258: getRegexpRange() returns incorrect results when regexp contains unescaped newline characters
The error may lead to analysis exceptions when inserting new issues: `java.lang.IllegalArgumentException: Start pointer [...] should be before end pointer [...]` - #4207

vdiez:
Fixes #4258

### DIFF
--- a/packages/jsts/src/rules/S5843/rule.ts
+++ b/packages/jsts/src/rules/S5843/rule.ts
@@ -266,12 +266,14 @@ class ComplexityCalculator {
       message += ` (incl ${increment - 1} for nesting)`;
     }
     const loc = getRegexpLocation(this.regexPart, node, this.context, offset);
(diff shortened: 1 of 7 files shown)


yassin-kammoun-sonarsource (approved):
LGTM!

Did you check locally that the original issue no longer makes the analysis crash with this fix?

yassin-kammoun-sonarsource:
FYI, I renamed the PR title.

vdiez:
Yes, it's precisely the test I added [here](https://github.com/SonarSource/SonarJS/blob/e8c3027713a597bcb119917f7679dbd0ca54a581/packages/jsts/src/rules/S5843/unit.test.ts#L120)

PR #4338, opened 2023-10-31
Create rule CSS:S125: Sections of code should not be commented out

### LINKED ISSUE #4227: Create rule CSS:S125: Sections of code should not be commented out
Feedback from community: https://community.sonarsource.com/t/css-rspec-125-sections-of-code-should-not-be-commented-out/101692

Probably just checking for some basic regex `/[\w-]+:.*;/` can be enough

yassin-kammoun-sonarsource:
Fixes #4227 

Relying solely on regular expressions to detect commented-out CSS code wasn't reliable. All issues raised on ruling were false positives. Therefore, I adopted the same approach with took with the implementation for JavaScript:

1. Filter comments likely to contain commented code using regular expressions
2. Check for syntactically valid code using a parsing function


packages/css/src/rules/S125/rule.ts
```diff
@@ -0,0 +1,69 @@
...
+            node: comment,
+          });
+        } catch {
+          /* syntax error */
+        }
+      }
+    });
+  };
+
+  function isLikelyCss(text: string) {
```
vdiez, line 48: would be nice to have an example of what each regex is trying to match, next to the description

vdiez (approved):
amazing job! very complete

packages/css/src/rules/S125/rule.ts
yassin-kammoun-sonarsource: 👍

PR #4344, opened 2023-10-31
Remove `node:` protocol from import

### LINKED ISSUE #4343: Avoid using `node:` protocol for imports
`node:` protocol for imports is supported only since Node.js v14.18 and we are still supposed to support v14.17

saberduck:
Fixes #4343


ilia-kebets-sonarsource (approved):
LGTM!

PR #4350, opened 2023-11-01
Add a sanity check for rules that should be mapped

yassin-kammoun-sonarsource:


packages/jsts/tests/rules/index.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    //        checking if the rule key is in the mapping.
+    const ruleFolder = path.join(__dirname, '../../src/rules');
+    const sonarKeys = fs.readdirSync(ruleFolder).filter(name => /^S\d+/.test(name));
+    const mappedRules = Object.values(mapping);
+    const missing = [];
+    for (const sonarKey of sonarKeys) {
+      const { rule } = require(path.join(ruleFolder, sonarKey));
+      if (!mappedRules.some(mapped => mapped === rule)) {
+        missing.push(sonarKey);
+      }
```
ilia-kebets-sonarsource, line 39: why wouldn't this be enough?

```suggestion
      if (!mappedRules[sonarKey]) {
        missing.push(sonarKey);
      }
```

packages/jsts/tests/rules/index.test.ts
```diff
@@ -0,0 +1,43 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import { rules as mapping } from '../../src/rules';
+import fs from 'fs';
+import path from 'path';
+
+describe('index', () => {
+  it('should map keys to rules definitions', () => {
+    // FIXME: This test runs with a time complexity of O(n^2) where n is the number of rules.
```
ilia-kebets-sonarsource, line 27: Isn't it O(n)?
- gather `sonarKeys`: n
- gather `mappedRules`: n
- find missing: n

packages/jsts/tests/rules/index.test.ts
ilia-kebets-sonarsource: The check we want to implement is that we don't forget to add the rule redefinition/decoration/whatever, so it's the case where we:
1. write it, and its units tests in `packages/src/ rules/Sxxxx/`
2. forget to load it in `packages/src/rules/index.ts`

No?

ilia-kebets-sonarsource (changes requested):
I think you can simplify it as per my comments. I'm open to discuss it if you think we need a more thorough check, but I don't see what it brings, except checking that require works.

packages/jsts/tests/rules/index.test.ts
yassin-kammoun-sonarsource: As I was saying in my comment, we are not able to do what you are suggesting because the keys in the mapping aren't Sonar keys:

https://github.com/SonarSource/SonarJS/blob/0b48c18aaad06886ecb1733d4ed55e5fc5d57a77/packages/jsts/src/rules/index.ts#L300-L302

We don't have `rules['S2376'] = S2376; ` but `rules['accessor-pairs'] = S2376;` instead. This is why we need to check the other way around.

packages/jsts/tests/rules/index.test.ts
yassin-kammoun-sonarsource: The current implementation is O(n^2) since we iterate with the for-loop O(n) over all the rules available in the folder `rules/SXXX`, and for each of them, we look for the corresponding definition with `mappedRules.some(mapped => mapped === rule))` O(n).

ilia-kebets-sonarsource: indeed

packages/jsts/tests/rules/index.test.ts
ilia-kebets-sonarsource: thanks for the explanation

ilia-kebets-sonarsource (approved):
LGTM!

packages/jsts/tests/rules/index.test.ts
yassin-kammoun-sonarsource: You're welcome :)

PR #4351, opened 2023-11-02
Improve the time complexity of the sanity check for mapped rules

yassin-kammoun-sonarsource:


ilia-kebets-sonarsource (approved):
Locally they have the same performance, I believe because the bottleneck is the synchronous I/O during the rules load, but it might be better if we one day hit 10k rules or more.

PR #4363, opened 2023-11-07
Load `package.json` on init-linter

vdiez:


packages/jsts/tests/dependencies/index.test.ts
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import path from 'path';
+import { toUnixPath } from '@sonar/shared/helpers';
+
+const { initPackageJsons, getAllPackageJsons, getNearestPackageJson } = require('@sonar/jsts');
```
yassin-kammoun-sonarsource, line 24: Why do you need to use `require` instead of an `import` statement like above?

package.json
```diff
@@ -102,6 +102,7 @@
     "scslre": "0.2.0",
     "stylelint": "15.10.0",
     "tmp": "0.2.1",
+    "type-fest": "4.6.0",
```
yassin-kammoun-sonarsource, line 105: We don't need to include this dependency in the bundled dependencies because this is only about type definitions, right?

packages/jsts/src/dependencies/package-json/index.ts
```diff
@@ -0,0 +1,39 @@
+/*
```
yassin-kammoun-sonarsource, line 1: The `dependencies` folder seems a bit extra. Maybe we could just have the `package-json` folder. Maybe I am missing something. Was there a reason you found it necessary to add this extra layer?

packages/jsts/src/dependencies/package-json/index.ts
```diff
@@ -0,0 +1,39 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+import { PackageJsons, PackageJson } from './project-package-json';
+
+const PackageJsonsByBaseDir = new PackageJsons();
+
+async function initPackageJsons(baseDir?: string) {
```
yassin-kammoun-sonarsource, line 25: Why `baseDir` is optional? Can it be missing in some case? Maybe in SonarLint context?

packages/jsts/src/dependencies/package-json/project-package-json.ts
```diff
@@ -0,0 +1,80 @@
...
+const PACKAGE_JSON = 'package.json';
+const IGNORED_DIRS = ['node_modules', '.scannerwork'];
+
+export interface PackageJson {
+  filename: string;
+  contents: PJ;
+}
+
+export class PackageJsons {
+  public db: Map<string, PackageJson>;
```
yassin-kammoun-sonarsource, line 34:
```suggestion
  readonly db: Map<string, PackageJson>;
```

yassin-kammoun-sonarsource: Fields are public by default (https://www.typescriptlang.org/docs/handbook/classes.html#public-by-default)

packages/jsts/src/dependencies/package-json/project-package-json.ts
```diff
@@ -0,0 +1,80 @@
...
+  constructor() {
+    this.db = new Map();
+  }
```
yassin-kammoun-sonarsource, line 37: We could get rid of the constructor and instantiate `this.db` at the declaration-level.

packages/jsts/src/parsers/parse.ts
```diff
@@ -20,21 +20,30 @@
...
+export function parseForESLint(
+  code: string,
+  parse: ParseFunction,
+  options: {},
+  packageJson?: PackageJson,
+): SourceCode {
   try {
     const result = parse(code, options);
+    const parserServices = result.services || {};
+    parserServices.packageJson = packageJson;
```
yassin-kammoun-sonarsource, line 42: I think it would be nice to have a unit test with a vue file that makes sure that `packageJson` is also available despite the parsing delegation.

packages/jsts/tests/analysis/fixtures/package-json/rule-package-json/package.json
```diff
@@ -0,0 +1,5 @@
+{
```
yassin-kammoun-sonarsource, line 1: Is this still used?

packages/jsts/tests/analysis/fixtures/package-json/rule-package-json/rules.js
```diff
@@ -0,0 +1,19 @@
+exports.rules = [
```
yassin-kammoun-sonarsource, line 1: Is this still used?

packages/jsts/src/dependencies/package-json/project-package-json.ts
```diff
@@ -0,0 +1,80 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+import fs from 'node:fs/promises';
+import path from 'path';
+import { toUnixPath, debug } from '@sonar/shared/helpers';
+import { PackageJson as PJ } from 'type-fest';
+
+const PACKAGE_JSON = 'package.json';
+const IGNORED_DIRS = ['node_modules', '.scannerwork'];
```
yassin-kammoun-sonarsource, line 26: I wonder whether the Java part should pass to the bridge the value of `JS_EXCLUSIONS_KEY` and `TS_EXCLUSIONS_KEY` to avoid traversing the file system where we shouldn't. WDYT?

packages/jsts/src/dependencies/package-json/project-package-json.ts
```diff
@@ -0,0 +1,80 @@
...
+      debug(`ERROR: Failed package.json file search: ${e}`);
+    }
+  }
+
+  /**
+   * Given a filename, find the nearest package.json
+   *
+   * @param file source file for which we need a package.json
+   */
+  getPackageJsonForFile(file: string) {
```
yassin-kammoun-sonarsource, line 69: Should we try using this function with the rule S4328? Maybe a separate PR?

packages/jsts/tests/dependencies/index.test.ts
vdiez: intellij did that :(

packages/jsts/tests/analysis/fixtures/package-json/rule-package-json/package.json
vdiez: nope, good catch

packages/jsts/src/dependencies/package-json/project-package-json.ts
vdiez: Yes, rules that can be improved by this should be on separate PRs IMHO

packages/jsts/tests/analysis/fixtures/package-json/rule-package-json/rules.js
vdiez: nope, good catch

packages/jsts/src/dependencies/package-json/index.ts
vdiez: just to be backwards compatible, have not really tested the other sensors and I don't want to crash if baseDir is not passed, not properly read, or no package.json files are found.

package.json
vdiez: correct

packages/jsts/src/dependencies/package-json/index.ts
yassin-kammoun-sonarsource: ok

packages/jsts/src/dependencies/package-json/project-package-json.ts
yassin-kammoun-sonarsource: 👍

vdiez: Created #4367

packages/jsts/tests/analysis/analyzer.test.ts
```diff
@@ -893,4 +895,38 @@ describe('analyzeJSTS', () => {
...
+    await initPackageJsons(baseDir);
+    const log = `DEBUG package.json found: ${toUnixPath(packageJson)}`;
+    expect(console.log).toHaveBeenCalledWith(log);
+
+    const linter = new Linter();
+    linter.defineRule('custom-rule-file', {
+      create(context) {
+        return {
+          CallExpression() {
+            console.log('detected call expression');
```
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: We can get rid of this log as well as the mock of console.log and its assertion.

packages/jsts/src/dependencies/package-json/project-package-json.ts
```diff
@@ -0,0 +1,77 @@
...
+  async packageJsonLookup(dir: string) {
+    dir = path.posix.normalize(toUnixPath(dir));
+    try {
+      const files = await fs.readdir(dir, { withFileTypes: true });
+      for (const file of files) {
+        const filename = path.posix.join(dir, file.name);
+        if (file.isDirectory() && !IGNORED_DIRS.includes(file.name)) {
+          await this.packageJsonLookup(filename);
+        } else if (file.name.toLowerCase() === PACKAGE_JSON && !file.isDirectory()) {
+          debug(`package.json found: ${filename}`);
```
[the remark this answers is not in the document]
vdiez: most times should be just one package.json, so that would be one line instead of 2. not sure about it

packages/jsts/src/dependencies/package-json/project-package-json.ts
```diff
@@ -0,0 +1,77 @@
...
+          const contents = JSON.parse(await fs.readFile(filename, 'utf-8'));
+          this.db.set(dir, { filename, contents });
+        }
+      }
+    } catch (e) {
+      debug(`ERROR: Failed package.json file search: ${e}`);
+    }
```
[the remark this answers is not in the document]
vdiez: I think is enough as well. I prefer not to log too much

packages/jsts/src/dependencies/package-json/project-package-json.ts
```diff
@@ -0,0 +1,84 @@
...
+      }
+    }
+  }
+  /**
+   * Given a filename, find the nearest package.json
+   *
+   * @param file source file for which we need a package.json
+   */
+  getPackageJsonForFile(file: string) {
+    let currentDir = path.posix.dirname(path.posix.normalize(toUnixPath(file)));
```
yassin-kammoun-sonarsource, line 74: I think we can add a tiny optimisation here and return early when the map is empty:

```suggestion
    if (this.db.size === 0) {
      return null;
    }
    let currentDir = path.posix.dirname(path.posix.normalize(toUnixPath(file)));
```

By having this early return, we would avoid unnecessarily looping from the file path until the root.

packages/jsts/src/dependencies/package-json/project-package-json.ts
```diff
@@ -0,0 +1,84 @@
...
+    for (const file of files) {
+      const filename = path.posix.join(dir, file.name);
+      if (ignoredPatterns.some(pattern => pattern.match(filename))) {
+        continue; // is ignored pattern
+      }
+      if (file.isDirectory()) {
+        await this.walkDirectory(filename, ignoredPatterns);
+      } else if (file.name.toLowerCase() === PACKAGE_JSON && !file.isDirectory()) {
+        debug(`Found package.json: ${filename}`);
+        const contents = JSON.parse(await fs.readFile(filename, 'utf-8'));
```
yassin-kammoun-sonarsource, line 63: We have a helper function `readFile` from `@sonar/shared/helpers` that takes into account BOM files. I am not sure this could happen with package.json files.

Feel free to ignore this suggestion.

packages/jsts/tests/analysis/analyzer.test.ts
```diff
@@ -893,4 +895,51 @@ describe('analyzeJSTS', () => {
...
+
+    const issues = linter.verify(
+      sourceCode,
+      { rules: { 'custom-rule-file': 'error' } },
+      { filename: filePath, allowInlineConfig: false },
+    );
+    expect(issues).toHaveLength(1);
+    expect(issues[0].message).toEqual('call');
+
+    const vueFilePath = path.join(baseDir, 'custom.js');
```
yassin-kammoun-sonarsource, line 932: I think you meant

```suggestion
    const vueFilePath = path.join(baseDir, 'code.vue');
```

packages/jsts/src/dependencies/package-json/project-package-json.ts
```diff
@@ -0,0 +1,96 @@
...
+    }
+    let currentDir = path.posix.dirname(path.posix.normalize(toUnixPath(file)));
+    do {
+      const packageJson = this.db.get(currentDir);
+      if (packageJson) {
+        return packageJson;
+      }
+      currentDir = path.posix.dirname(currentDir);
+    } while (currentDir !== path.posix.dirname(currentDir));
+    return null;
```
yassin-kammoun-sonarsource, line 94: We are missing coverage on this line.

packages/jsts/src/dependencies/package-json/project-package-json.ts
```diff
@@ -0,0 +1,96 @@
...
+   * @param exclusions glob patterns to ignore while walking the tree
+   */
+  async searchPackageJsonFiles(dir: string, exclusions: string[]) {
+    try {
+      const patterns = exclusions
+        .concat(IGNORED_PATTERNS)
+        .map(exclusion => new Minimatch(exclusion));
+      await this.walkDirectory(path.posix.normalize(toUnixPath(dir)), patterns);
+    } catch (e) {
+      error(`Error while searching for package.json files: ${e}`);
```
yassin-kammoun-sonarsource, line 53: We are missing coverage on this line.

yassin-kammoun-sonarsource (approved):
LGTM! Great job!

packages/jsts/src/dependencies/package-json/project-package-json.ts
```diff
@@ -0,0 +1,96 @@
...
+    }
+  }
+
+  async walkDirectory(dir: string, ignoredPatterns: Minimatch[]) {
+    const files = await fs.readdir(dir, { withFileTypes: true });
```
saberduck, line 58: why do you need to `await` the result here, if function is `async`?

packages/jsts/src/dependencies/package-json/project-package-json.ts
```diff
@@ -0,0 +1,96 @@
...
+    for (const file of files) {
+      const filename = path.posix.join(dir, file.name);
+      if (ignoredPatterns.some(pattern => pattern.match(filename))) {
+        continue; // is ignored pattern
+      }
+      if (file.isDirectory()) {
+        await this.walkDirectory(filename, ignoredPatterns);
```
saberduck, line 65: why `await`?

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/AbstractBridgeSensor.java
```diff
@@ -126,4 +132,18 @@ protected boolean shouldAnalyzeWithProgram(List<InputFile> inputFiles) {
     LOG.debug("Will use AnalysisWithProgram");
     return true;
   }
+
+  protected List<String> getExcludedPaths() {
+    var configuration = this.context.config();
+    var excludedPatterns = Arrays.asList(JavaScriptPlugin.EXCLUSIONS_DEFAULT_VALUE);
+    var jsExcludedPatterns = configuration.get(JavaScriptPlugin.JS_EXCLUSIONS_KEY).isPresent()
```
saberduck, line 139: you don't need to do this, if you call `getStringArray` it will return `[]` if the property is not defined

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/AbstractBridgeSensor.java
```diff
@@ -126,4 +132,18 @@ protected boolean shouldAnalyzeWithProgram(List<InputFile> inputFiles) {
     LOG.debug("Will use AnalysisWithProgram");
     return true;
   }
+
+  protected List<String> getExcludedPaths() {
+    var configuration = this.context.config();
+    var excludedPatterns = Arrays.asList(JavaScriptPlugin.EXCLUSIONS_DEFAULT_VALUE);
```
saberduck, line 138: we don't need this variable

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/AbstractBridgeSensor.java
```diff
@@ -126,4 +132,18 @@ protected boolean shouldAnalyzeWithProgram(List<InputFile> inputFiles) {
     LOG.debug("Will use AnalysisWithProgram");
     return true;
   }
+
+  protected List<String> getExcludedPaths() {
```
saberduck, line 136: we should share this logic with `PathAssesor`, by creating utility class with static function, otherwise we will forget to keep the two places in sync

saberduck (commented):
@vdiez pls see the comment about sharing the exclusion logic

packages/jsts/src/dependencies/package-json/project-package-json.ts
vdiez: because otherwise files is a promise and would not be able to iterate

packages/jsts/src/dependencies/package-json/project-package-json.ts
vdiez: not needed but to keep order of execution

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/AbstractBridgeSensor.java
vdiez: I wanted to do this, but had doubts about the how-to

PR #4371, opened 2023-11-13
Modify S1874: Merge with `react/no-deprecated`

### LINKED ISSUE #3658: Extend S1874 with `react/no-deprecated`: Disallow usage of deprecated methods
[Reference](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-deprecated.md)

[Implementation](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/lib/rules/no-deprecated.js)

[RSPEC PR](https://github.com/SonarSource/rspec/pull/3184)

vdiez:
Fixes #3658


packages/jsts/src/rules/S1874/cb.test.ts
```diff
@@ -20,9 +20,11 @@
 import { check } from '../tools';
 import { rule } from './';
 import path from 'path';
+import { searchPackageJsonFiles } from '@sonar/jsts';
 
 const sonarId = path.basename(__dirname);
 
 describe(`Rule ${sonarId}`, () => {
+  searchPackageJsonFiles(__dirname, []);
```
yassin-kammoun-sonarsource, line 28: Just a suggestion, but we could call in the `beforeAll` hook and then clear the database in the `afterAll` hook.

yassin-kammoun-sonarsource (approved):

packages/jsts/src/rules/S1874/cb.test.ts
vdiez: not sure we can do that... comment base will then clear before each noncompliant, right? and we can use same DB for all

vdiez: oh, you mean beforeAll and not beforeEach, sorry yes you're right

also, just checked and indeed it just creates one test, not multiple, so beforeEach would also work

yassin-kammoun-sonarsource: 👍

PR #4375, opened 2023-11-14
Return all package.json files instead of just nearest

vdiez:


packages/jsts/src/parsers/parse.ts
```diff
@@ -20,26 +20,26 @@
...
   parse: ParseFunction,
   options: {},
-  packageJson?: PackageJson,
+  packageJsons?: PackageJson[],
 ): SourceCode {
   try {
     const result = parse(code, options);
     const parserServices = result.services || {};
-    parserServices.packageJson = packageJson;
+    parserServices.packageJsons = packageJsons;
```
yassin-kammoun-sonarsource, line 42: Instead of injecting the package.json themselves, why don't we just expose `getNearestPackageJsons` and `getAllPackageJsons`, and leave the rules decide what to do with them?

yassin-kammoun-sonarsource (commented):
Could you clarify why you are introducing this change?

packages/jsts/src/parsers/parse.ts
vdiez: I like the idea, let me implement that

yassin-kammoun-sonarsource: 👍

yassin-kammoun-sonarsource (approved):
LGTM

PR #4386, opened 2023-11-14
BUILD-4131 use GitHub token from vault instead of sonartech api token

sebastienvermeille:
# BUILD-4131 use GitHub token from vault instead of sonartech api token

## Changes
* Adapt the  dogfood pipeline
  That way this secret can be rotated easily as it comes from Vault

## Depends on
* https://github.com/SonarSource/re-terraform-aws-vault/pull/823

### DIFF
--- a/.github/workflows/dogfood.yml
+++ b/.github/workflows/dogfood.yml
@@ -11,13 +11,19 @@ jobs:
     runs-on: ubuntu-latest
     name: Update dogfood branch
     steps:
-    - name: git octopus step      
-      env:
-        GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
-      id: dogfood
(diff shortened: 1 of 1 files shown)


.github/workflows/dogfood.yml
```diff
@@ -11,13 +11,19 @@ jobs:
     runs-on: ubuntu-latest
     name: Update dogfood branch
     steps:
```
matemoln, line 13:
```suggestion
    permissions:
      id-token: write # required for SonarSource/vault-action-wrapper
    steps:
```

matemoln (changes requested):

matemoln (approved):

yassin-kammoun-sonarsource (approved):

PR #4388, opened 2023-11-15
Create rule S6847 (`jsx-a11y/no-noninteractive-element-interactions`): Non-interactive elements shouldn't have event handlers

### LINKED ISSUE #4285: Create rule S6847 (`jsx-a11y/no-noninteractive-element-interactions`)
[plugin rule](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/no-noninteractive-element-interactions.md#jsx-a11yno-noninteractive-element-interactions)

saberduck:
Fixes #4285


its/sources/jsts/custom/S6847.js
ilia-kebets-sonarsource, line 1: could we remove this one since we have occurrences in the ruling projects?

ilia-kebets-sonarsource (approved):
I propose to remove the custom ruling file since the new rule raises issues in the other projects.
This would require removing the expected issues it raised.

PR #4392, opened 2023-11-16
Improve S1874 (`deprecation`): Remove references from React `no-deprecated` messages

vdiez:


yassin-kammoun-sonarsource (commented):
Let's update a comment-based fixture and add an assertion with the reworked message.

yassin-kammoun-sonarsource (approved):
LGTM!

PR #4398, opened 2023-11-16
Fix message in S6844

saberduck:


packages/jsts/src/rules/S6844/rule.ts
```diff
@@ -0,0 +1,31 @@
...
+
+import { rules as jsxA11yRules } from 'eslint-plugin-jsx-a11y';
+import { interceptReport } from '../helpers';
+
+const anchorIsValid = jsxA11yRules['anchor-is-valid'];
+
+export const rule = interceptReport(anchorIsValid, (context, reportDescriptor) => {
+  const msg = (reportDescriptor as any)!.message;
+  (reportDescriptor as any)!.message = msg.substring(0, msg.indexOf(' Learn'));
+  context.report(reportDescriptor);
```
yassin-kammoun-sonarsource, line 30:
```suggestion
  const descriptor = reportDescriptor as any;
  const { message } = descriptor;
  descriptor.message = message.substring(0, message.indexOf(' Learn'));
  context.report(descriptor);
```

yassin-kammoun-sonarsource: Another suggestion using a regex, but I am not sure it's any better than calling `substring` and `indexOf`:

```suggestion
  const descriptor = reportDescriptor as any;
  const { message } = descriptor;
  descriptor.message = message.replace(/ Learn.*/, '');
  context.report(descriptor);
```

yassin-kammoun-sonarsource: In case you ignore my first two suggestions, we can at least get rid of the non-null assertion operator `(!)` since we cast to `any` anyway:

```suggestion
  const msg = (reportDescriptor as any).message;
  (reportDescriptor as any).message = msg.substring(0, msg.indexOf(' Learn'));
  context.report(reportDescriptor);
```

yassin-kammoun-sonarsource (approved):
I left suggestions you are free to ignore.
LGTM!

PR #4405, opened 2023-11-20
Hardcode the RSPEC id for each cb test file

### LINKED ISSUE #4403: Hardcode the RSPEC id for each cb test file
The [vscode-jest-runner extension](https://marketplace.visualstudio.com/items?itemName=firsttris.vscode-jest-runner) we use in the squad to debug our rules doesn't support dynamic test names as we have now:

```js
describe(`Rule ${sonarId}`, () => {
```

We could replace them with hardcoded names when we generate this file using the `npm new-rule ...` script using [the template](https://github.com/SonarSource/SonarJS/blob/1640038df5b0f08d3d9314b3f2e17b4af84daab4/tools/resources/rule.cbtest_ts#L26)

```js
describe(`Rule S1234`, () => {
```

Currently, we must change it by hand when we need to use the debugger. 

This issue was introduced in this extension after version v0.4.60, but since the latest MacOS update, it doesn't work even on older versions.

yassin-kammoun-sonarsource:
Fixes #4403

### DIFF
--- a/packages/jsts/src/rules/S1082/cb.test.ts
+++ b/packages/jsts/src/rules/S1082/cb.test.ts
@@ -23,6 +23,6 @@ import path from 'path';
 
 const sonarId = path.basename(__dirname);
 
-describe(`Rule ${sonarId}`, () => {
+describe('Rule S1082', () => {
   check(sonarId, rule, __dirname);
 });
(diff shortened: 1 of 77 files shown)


ilia-kebets-sonarsource (approved):
LGTM!

PR #4413, opened 2023-11-22
Remove cache for `.sonar` dir

### LINKED ISSUE #4402: [ci] Remove usage of `deploy/plugins/index.txt` for cache
Remove usage of `deploy/plugins/index.txt` in CI pipeline.

We should test the impact of the cache, perhaps it's not needed anymore.

If the cache is still needed, we should replace usage of `deploy/plugins/index.txt` with supported API `https://next.sonarqube.com/sonarqube/web_api/api/plugins?query=plugins%2Finstalled` 

We also use `deploy/plugins/index.txt` in https://github.com/SonarSource/peachee-js/blob/js-ts-css-html/.cirrus.yml#L35, it should be removed/replaced in the same way

saberduck:
Fixes #4402 

Before:
```
[INFO] User cache: /home/sonarsource/.sonar/cache
[INFO] Load/download plugins
[INFO] Load plugins index
[INFO] Load plugins index (done) | time=84ms
[INFO] Load/download plugins (done) | time=3826ms
```

After:
```
[INFO] Load/download plugins
[INFO] Load plugins index
[INFO] Load plugins index (done) | time=81ms
[INFO] Load/download plugins (done) | time=5605ms
```

Cache is not functioning anymore, as feature was already removed. Anyway, it's not worth to optimize 5 second download, because fetching the cache would take similar amount of time.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -134,11 +134,6 @@ analyze_task:
     ARTIFACTORY_DEPLOY_ACCESS_TOKEN: VAULT[development/artifactory/token/${CIRRUS_REPO_OWNER}-${CIRRUS_REPO_NAME}-qa-deployer access_token]
   <<: *MAVEN_CACHE
   <<: *RUNTIME_CACHE
-  sonar_cache:
-    folder: ${HOME}/.sonar/cache
-    fingerprint_script:
-      - echo ${CIRRUS_OS}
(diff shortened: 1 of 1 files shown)


vdiez (approved):
LGTM! 

It would have been nice to know how often the cache was reused. With so many plugins I doubt there were many days (outside of weekends) where it was actually the same artifacts for all of them.

PR #4430, opened 2023-11-27
Use `tar` instead of `decompress` library

vdiez:


tools/fetch-node/scripts/fetch-node.mjs
```diff
@@ -164,12 +164,14 @@ async function extractFile(file, dir) {
   } else if (file.endsWith('.tar.gz')) {
     // decompress tar gz doesn't support overwrites
     deleteFolderIfExists(removeExtension(file));
```
ilia-kebets-sonarsource, line 166: could you check if that is still needed with the new lib?

ilia-kebets-sonarsource (approved):
LGTM and Good Job!

Could you check that these change work locally? I'm pretty sure fetch-node() is not used in the CI because of the cache.

I have left 1 suggestion in the code, otherwise LGTM!

PR #4442, opened 2023-11-29
Create rule S2187: Test files should contain at least one test case

### LINKED ISSUE #4428: Create rule S2187: Test files should contain at least one test case
[S2187](https://sonarsource.github.io/rspec/#/rspec/S2187)

### Detection

- Detect if files with `.test` or `.spec` extension or that imports jest/chai.js have at least a test
- Also check it and describe without callbacks

yassin-kammoun-sonarsource:
Fixes #4428


packages/jsts/src/rules/S2187/rule.ts
```diff
@@ -0,0 +1,124 @@
...
+  'test.each',
+  'test.failing',
+  'test.failing.each',
+  'test.only.failing',
+  'test.skip.failing',
+  'test.only',
+  'test.only.each',
+  'test.skip',
+  'test.skip.each',
+  'test.todo',
```
ilia-kebets-sonarsource, line 60: - Does it support other types of permutations, like `it.concurrent.each` <=> `it.each.concurrent`?
- Wouldn't it be easier to simply test the first part of these expressions, like: `it`, `test`, `fit`, `xit`?

packages/jsts/src/rules/S2187/rule.ts
```diff
@@ -0,0 +1,124 @@
...
+  'test',
+  'test.skip',
+  'test.todo',
+  'test.only',
+]);
+
+export const rule: Rule.RuleModule = {
+  meta: {
+    messages: {
+      missingTest: 'Add some tests to this file.',
```
ilia-kebets-sonarsource, line 82:
```suggestion
      missingTest: 'Add some tests to this file or delete it.',
```

ilia-kebets-sonarsource: this would require to adapt the invalid unit test case(s)

ilia-kebets-sonarsource (approved):
small questions and suggestions, otherwise LGTM!

packages/jsts/src/rules/S2187/rule.ts
yassin-kammoun-sonarsource: > Does it support other types of permutations, like it.concurrent.each <=> it.each.concurrent?

Nope, the API is pretty clear on each framework website.

> Wouldn't it be easier to simply test the first part of these expressions, like: it, test, fit, xit?

I prefer the checks to be as accurate as possible.

packages/jsts/src/rules/S2187/rule.ts
yassin-kammoun-sonarsource: 👍

PR #4452, opened 2023-12-01
Upgrade typescript-eslint

saberduck:


packages/jsts/src/rules/S3796/rule.ts
```diff
@@ -57,7 +58,7 @@ function hasCallBackWithoutReturn(argument: estree.Node, services: RequiredParse
   const type = checker.getTypeAtLocation(
     services.esTreeNodeToTSNodeMap.get(argument as TSESTree.Node),
   );
-  const signatures = type.getCallSignatures();
+  const signatures: Signature[] = type.getCallSignatures();
```
saberduck, line 61: unclear to me why this is needed, looks like regression in TS type inference

packages/jsts/src/rules/helpers/regex/ast.ts
```diff
@@ -42,7 +43,10 @@ export function isStringReplaceCall(call: estree.CallExpression, services: Parse
   );
 }
 
-export function isStringRegexMethodCall(call: estree.CallExpression, services: ParserServices) {
+export function isStringRegexMethodCall(
+  call: estree.CallExpression,
+  services: RequiredParserServices,
```
ilia-kebets-sonarsource, line 48: I think you wanted to use `ParserServices` here

ilia-kebets-sonarsource:
Here are the suspect cases I encountered:

S3863: Imports from the same module should be merged
- proposal: add secondary locations
	- useful especially when there is more than one secondary: like ag-grid > src/ts/columnController/balancedColumnTreeBuilder.ts lines 11 - 14
- desktop > app/src/ui/index.tsx, line 40: error msg missing
	- and error flagging "import" and the imported path
	- and same error flagged twice at line 37 and 44
- courselit > apps/web/components/public/base-layout/component-scaffold.tsx
	- same as above for lines 12 and 18
- maybe add a coding example for cases where you need to combine the export of properties and some default value like : import React, { Component } from React

S2201: Return values from functions without side effects should not be ignored
- all new errors have missing error messages: in JS and TS
- for cases where map() could be replaced by .forEach(), we could adapt the error message: 
	- desktop > app/src/ui/changes/changes-list.tsx: lines 574, 584
	- courselit > apps/web/components/admin/design/layout-manager/compatible-components-map.ts: lines 15, 18
	- courselit > apps/web/components/admin/media/index.tsx: line 115
	- eigen > src/app/Components/ArtworkFilter/Filters/WaysToBuyOptions.tsx: line 67
	- eigen > src/app/Scenes/SellWithArtsy/SubmitArtwork/ArtworkDetails/ArtworkDetailsForm.tests.tsx: line 58
	- eigen > src/app/utils/renderMarkdown.tests.tsx line 161
	- etc...
- same for filter:
	- router > src/activation.js line 146

S1082: Mouse events should have corresponding keyboard events
- missing error message:
	- desktop > app/src/ui/lib/list/list-row.tsx line 96
- double errors on same location:
	- Joust > ts/components/Tooltipper.tsx: line 62
	- Joust > ts/components/game/EntityInPlay.tsx: line 134

S4325: Redundant casts and non-null assertions should be avoided
- Missing error messages:
	- ag-grid > src/ts/entities/column.ts: line 227, 243
	- ag-grid > src/ts/rendering/renderedCell.ts: line 201
	- ant-design > site/theme/template/IconDisplay/index.tsx: line 87

S3981: Collection size and array length comparisons should make sense
- it might be nice to have a secondary location to know what condition or declaration verifies the redundant condition, but lots of work:
	- es5-shim > es5-shim.js: line 1084

S2260: JS parser failure
- jshint > tests/unit/fixtures/emptystmt.js: line 1 ("var;")
	- interesting that this is introduced now. I don't know what to think

saberduck:
Changes for S6582 come from the fact that @typescript-eslint provides `services.program` also in cases when it was not available before. The rule was doing early exit when `services.program` was not available.  There were also some improvements to the rule logic itself, https://github.com/typescript-eslint/typescript-eslint/commit/02a37c4c79d9b83998b7ee1376be43b06e12b3a0

jest-resolver.js
```diff
@@ -0,0 +1,41 @@
...
+  // Call the defaultResolver, so we leverage its cache, error handling, etc.
+  return options.defaultResolver(path, {
+    ...options,
+    // Use packageFilter to process parsed `package.json` before the resolution (see https://www.npmjs.com/package/resolve#resolveid-opts-cb)
+    packageFilter: pkg => {
+      if (pkg.name === '@typescript-eslint/parser') {
+        /**
+         * `@typescript-eslint/parser` uses more modern "exports" field in its package.json, which is not yet supported by jest default
+         * resolver (see https://github.com/browserify/resolve/issues/222, https://github.com/jestjs/jest/issues/9771).
+         * This is a workaround to make it work.
```
ilia-kebets-sonarsource, line 30:
```suggestion
         * `@typescript-eslint/parser` uses a more modern "exports" field in its package.json, which is not yet supported by the default Jest
         * resolver (see https://github.com/browserify/resolve/issues/222, https://github.com/jestjs/jest/issues/9771).
         * This is a workaround to make it work.
```

saberduck: I pushed this fix to have good formatting

packages/jsts/src/rules/S2234/rule.ts
```diff
@@ -21,9 +21,8 @@
 
 import { Rule } from 'eslint';
 import * as estree from 'estree';
-import { TSESTree } from '@typescript-eslint/experimental-utils';
+import { TSESTree } from '@typescript-eslint/utils';
 import { EncodedMessage } from 'eslint-plugin-sonarjs/lib/utils/locations';
-import { isIdentifier } from 'eslint-plugin-sonarjs/lib/utils/nodes';
```
ilia-kebets-sonarsource, line 26: is there a motivation not to use the [utils from SonarJS](https://github.com/SonarSource/SonarJS/blob/4c2d5b064a684b600aac5e4d0f2a47afb451e604/packages/jsts/src/rules/helpers/ast.ts#L54)

ilia-kebets-sonarsource, line 26: is there a reason not to use the [utils from SonarJS](https://github.com/SonarSource/SonarJS/blob/4c2d5b064a684b600aac5e4d0f2a47afb451e604/packages/jsts/src/rules/helpers/ast.ts#L54)

ilia-kebets-sonarsource (approved):
LGTM!

packages/jsts/src/rules/S2234/rule.ts
saberduck: not really, we could

PR #4470, opened 2023-12-15
Avoid NPE when performance data is not collected

saberduck:


vdiez (approved):
LGTM, but we should still find the root cause why the rule now fails.

PR #4473, opened 2023-12-20
Fix rule S6647 `no-useless-constructor`

saberduck:


packages/jsts/src/rules/S6647/unit.test.ts
```diff
@@ -21,69 +21,48 @@ import { RuleTester } from 'eslint';
 import { rule } from './';
 
 const ruleTester = new RuleTester({
```
vdiez, line 23: I would leave original test with typescript-eslint parser, and additionally add babel unit test, so we can see the differences. Ideally even use original rule with typescript-eslint parser, so we can track it in case original implementation changes.

I don't know if it's possible to add a test also that using babel with original rule should fail. In case babel supports new AST, that test would alert us that having the rule on our side is not needed any more.

vdiez (changes requested):

packages/jsts/tests/builders/build.test.ts
```diff
@@ -141,7 +141,6 @@ describe('buildSourceCode', () => {
     } = buildSourceCode(await jsTsInput({ filePath }), 'js');
 
     expect(stmt.type).toEqual('FunctionDeclaration');
-    expect(console.log).not.toHaveBeenCalled();
```
vdiez, line 144: why has this been removed?

packages/jsts/src/rules/S6647/unit.test.ts
saberduck: original parser is tested with comment based test, having additional test seems redundant

saberduck: I don't think testing original rule is worth the trouble. let's keep it simple

packages/jsts/tests/builders/build.test.ts
saberduck: currently there is output in the log. I don't think we should assert debug level logs

vdiez (approved):

PR #4474, opened 2023-12-20
Bridge should return error status on unexpected errors

saberduck:
Bridge should return error status on unexpected errors and plugin should fail interrupting the analysis.


ilia-kebets-sonarsource (approved):
LGTM!

PR #4477, opened 2023-12-21
Fix failure handling of unexpected errors

saberduck:
Middleware needs to set the status before body. Added logging of the response body on Java side


ilia-kebets-sonarsource (approved):
so `response.json().status()` didn't work?

saberduck:
See #4478

Issue #4478, opened 2023-12-21
Plugin should detect failures from bridge


saberduck:
Currently, in case of an unknown failure, the bridge will return a response with an `error` field. This is not handled on the Java side and the analysis simply continues as if nothing happened.

The bridge should return an HTTP error code (500) and this should be detected on the Java side and handled accordingly.

ilia-kebets-sonarsource:
this would help from more specification:
- a reproducer would be nice
- a proposed solution would be nice.

Wohops:
Migrated to Jira. New ticket number: JS-45

PR #4479, opened 2023-12-21
Revert "Bridge should return error status on unexpected errors (#4474)"

saberduck:
This reverts commit cd866c37d595fe19633067022dc7b15d5fef04d3.

I am reverting the change because it's more complicated than originally though. We will do the work in issue #4478


ilia-kebets-sonarsource (approved):
LGTM

PR #4491, opened 2024-01-03
Automate rules data generation

ilia-kebets-sonarsource:
Automate the rules data, instead of doing it uglyly manually and integrate the automation into scripts and workflow.


packages/jsts/src/rules/S1472/rule.ts
```diff
@@ -77,7 +77,7 @@ export const rule: Rule.RuleModule = {
 
 function getCallee(call: estree.CallExpression) {
   const node = call as TSESTree.CallExpression;
-  return (node.typeParameters ?? node.callee) as estree.Node;
```
saberduck, line 80: I would prefer to have this change in another PR if possible

sonar-plugin/sonar-javascript-plugin/pom.xml
```diff
@@ -120,6 +120,26 @@
         </configuration>
       </plugin>
 
+      <plugin>
```
saberduck, line 123: is this still needed? can't you reuse the one which is also used in production `jar`?

saberduck (approved):
I left some comments, pls have a look

PR #4492, opened 2024-01-04
some changes

### LINKED ISSUE #4008: Fix FP S1874: (`deprecation`): Should not report when at least one function signature is not deprecated

ilia-kebets-sonarsource:
Fixes #4008


ilia-kebets-sonarsource:
this was used for a 1h with a dev session.

PR #4498, opened 2024-01-15
Fast ruling: find coverage issue

ilia-kebets-sonarsource:
using dichotomy to debug where I lost the coverage data for the Sonar analysis


ilia-kebets-sonarsource:
closing, as this is only used to show that reverting the license update commit brings coverage back.

PR #4502, opened 2024-01-17
Allow minimatch to match hidden folders

### LINKED ISSUE #4490: Analysis failing because of resources hoard when `node_modules/` is present with lots of dependencies
It seems that our current implementation of dependencies inspection is quite resource hungry and could be optimized. When there are many of them, the analysis can timeout and fail as reported in [this community ticket](https://community.sonarsource.com/t/the-bridge-server-is-unresponsive-error-while-scanning-typescript-project/105494/29).

The current workaround is to analyze the code without its dependencies by simply deleting the `node_modules/` folder before running the analysis.

However, we must optimize the dependencies inspection element. Some fixes can be:
- respect the `sonar.exclusions` for the dependencies folder traversal
- limit the dependencies search:
  - by depth: hardcoded or parametrizable
  - by some heuristic: ignore transitive ones

vdiez:
Fixes #4490


yassin-kammoun-sonarsource (commented):
Would it be tricky to write a unit test for this?

packages/jsts/src/dependencies/package-json/project-package-json.ts
```diff
@@ -44,10 +44,11 @@ export class PackageJsons {
    * @param exclusions glob patterns to ignore while walking the tree
    */
   async searchPackageJsonFiles(dir: string, exclusions: string[]) {
+    console.log(exclusions);
```
yassin-kammoun-sonarsource, line 47: Do you really want to log this or is it a debugging leftover?

vdiez: good catch

vdiez:
changed the unit test to cover hidden paths. I tested it without the `dot: true` and it failed to ignore the pattern indeed.

yassin-kammoun-sonarsource (approved):
Build is failing, probably due to prettier.

vdiez:
it's a 404 on `npm ci`, I guess I'll retry later

PR #4515, opened 2024-01-23
fix repox node20

saberduck:
- Fix repox access
- test
- try use global env
- update docker image
- BUILD-4416 fix artifactory authentication
- fix ruling
- fix docker
- fix docker
- remove unused dockerfile
- debug npm version
- try another env variable
- try Node 20 for build


saberduck:
Just a test, see #4512

PR #4536, opened 2024-01-29
Setup to reproduce issue

### LINKED ISSUE #4516: Fix FP S4123 (`no-invalid-await`): Don't flag functions returning promises and message
When calling `getTypeFromTreeNode()` on the node, the returned values has a `.symbol.escapedName` property that is equal to `Promise`

Reproducer:

```js
/**
 * @return {string}
 */
async function foo() {
  await fetch('https://unpkg.com/tailwindcss/dist/tailwind.min.css');
  return 'bar';
}

async function baz() {
  await foo(); // FP
}
```

Improve the message or the rule description: add `make sure that your function is returning a promise`

Proposal: Add `ts.TypeFlags.Void` to the white list

ilia-kebets-sonarsource:
Fixes #4516


ilia-kebets-sonarsource:
closing as we have found a way to reproduce it.

PR #4539, opened 2024-01-30
Fix FP S2871 (`no-alphabetical-sort`): Rework the rule to emit a dedicated message for arrays of strings

### LINKED ISSUE #4531: Fix FP S2871 (`no-alphabetical-sort`): Add exception for string arrays
Do not raise the issue when the array is an array of strings:

```js
const names = ['foo', 'bar'];
names.sort(); // compliant
```

ericmorand-sonarsource:
Fixes #4531

### DIFF
--- a/packages/jsts/src/rules/S2871/rule.ts
+++ b/packages/jsts/src/rules/S2871/rule.ts
@@ -54,7 +54,7 @@ export const rule: Rule.RuleModule = {
       provideCompareFunction:
         'Provide a compare function to avoid sorting elements alphabetically.',
       provideCompareFunctionForArrayOfStrings:
-        'Provide a compare function that depends on String.localeCompare, to reliably sort elements alphabetically.',
+        'Provide a compare function that depends on "String.localeCompare", to reliably sort elements alphabetically.',
       suggestNumericOrder: 'Add a comparator function to sort in ascending order',
       suggestLanguageSensitiveOrder:
         'Add a comparator function to sort in ascending language-sensitive order',
--- a/packages/jsts/src/rules/S2871/rule.ts
+++ b/packages/jsts/src/rules/S2871/rule.ts
@@ -82,9 +82,9 @@ export const rule: Rule.RuleModule = {
             typeArgument = type;
           }
 
-          function isTypeAString(candidate: ts.Type): boolean {
+          const isTypeAString = (candidate: ts.Type): boolean => {
             return candidate.flags === 4;
-          }
+          };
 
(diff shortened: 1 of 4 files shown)


packages/jsts/src/rules/S2871/rule.ts
```diff
@@ -53,6 +53,8 @@ export const rule: Rule.RuleModule = {
     messages: {
       provideCompareFunction:
         'Provide a compare function to avoid sorting elements alphabetically.',
+      provideCompareFunctionForArrayOfStrings:
+        'Provide a compare function that depends on String.localeCompare, to reliably sort elements alphabetically.',
```
yassin-kammoun-sonarsource, line 57: This looks great!

yassin-kammoun-sonarsource:
```suggestion
        'Provide a compare function that depends on "String.localeCompare", to reliably sort elements alphabetically.',
```

packages/jsts/src/rules/S2871/unit.test.ts
```diff
@@ -606,6 +606,8 @@ ruleTester.run(
         code: 'const array = ["foo", "bar"]; const sortedArray = array.toSorted();',
         errors: [
           {
+            message:
+              'Provide a compare function that depends on String.localeCompare, to reliably sort elements alphabetically.',
```
yassin-kammoun-sonarsource, line 610: Might need to apply this suggestion if you applied the one I suggested above.

```suggestion
              'Provide a compare function that depends on "String.localeCompare", to reliably sort elements alphabetically.',
```

sonar-plugin/javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S2871.html
```diff
@@ -31,6 +31,16 @@ <h2>Why is this an issue?</h2>
...
+<h3>Exceptions</h3>
+<p>The rule ignores arrays that only contain strings, or are declared as such.</p>
+<pre>
+const names = ['foo', 'bar'];
+names.sort(); // Compliant
+</pre>
+<pre>
+const names: string[] = [];
+names.sort(); // Compliant
+</pre>
```
yassin-kammoun-sonarsource, line 43: I guess the exception no longer holds and we can get rid of this from the RSPEC, right?

yassin-kammoun-sonarsource:
We could also update the PR's title since it's no longer about adding an exception for string arrays but rather having a specific message for string arrays.

ilia-kebets-sonarsource (approved):
LGTM!

PR #4540, opened 2024-01-31
Fix FP S6594 (`sonar-prefer-regexp-exec`): Ignore match count use case

### LINKED ISSUE #4532: Fix FP S6594 (`sonar-prefer-regexp-exec`): Ignore match count use case
The rule should not suggest using `Regex.exec` if the returned value of `String.match` is eventually used to get the number of matches. It only involves getting the `.length` of the value, while a loop is needed in the case of `Regex.exec`.

```js
const regex = new RegExp(pattern);
const matches = 'foo'.match(regex); // Currently false positive, should be compliant

/* ... */

matches.length;

/* or */

matches?.length;
```

yassin-kammoun-sonarsource:
Fixes #4532

### DIFF
--- a/packages/jsts/src/rules/S6594/cb.fixture.js
+++ b/packages/jsts/src/rules/S6594/cb.fixture.js
@@ -6,3 +6,14 @@ foo.match(/foo/);
 //    ^^^^^
 // fix@qf1 {{Replace with "RegExp.exec()"}}
 // edit@qf1 [[ec=19]] {{RegExp(/bar/).exec('foo');}}
+
+const m1 = 'foo'.match(/bar/);
+if (m1.length > 0) {
+  console.log(m1[0]);
+}
+
+let m2;
+m2 = 'foo'.match(/bar/);
+if (m2?.length > 0) {
+  console.log(m2[0]);
+}
--- a/packages/jsts/src/rules/S6594/cb.fixture.js
+++ b/packages/jsts/src/rules/S6594/cb.fixture.js
@@ -13,7 +13,7 @@ if (m1.length > 0) {
 }
 
 let m2;
-m2 = 'foo'.match(/bar/);
+m2 = 'foo'.match(/bar/); // Compliant: we're checking `m2?.length` below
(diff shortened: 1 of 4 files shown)


packages/jsts/src/rules/S6594/cb.fixture.js
```diff
@@ -6,3 +6,14 @@ foo.match(/foo/);
 //    ^^^^^
 // fix@qf1 {{Replace with "RegExp.exec()"}}
 // edit@qf1 [[ec=19]] {{RegExp(/bar/).exec('foo');}}
+
+const m1 = 'foo'.match(/bar/);
```
ilia-kebets-sonarsource, line 10:
```suggestion
const m1 = 'foo'.match(/bar/); // Compliant: we're checking `m1.length` below
```

packages/jsts/src/rules/S6594/rule.ts
```diff
@@ -66,5 +83,23 @@ export const rule: Rule.RuleModule = {
           });
         },
     };
+
+    /**
+     * Extracts the left-hand side variable of expressions
+     * like `x` in `const x = <node>` or `x` in `x = <node>`.
+     */
+    function getLhsVariable(node: estree.Node) {
```
ilia-kebets-sonarsource, line 91: isn't it the same as https://sonarsource.github.io/SonarJS/typedoc/functions/getLhsVariable.html ?

ilia-kebets-sonarsource (approved):
See my comments, otherwise LGTM!

packages/jsts/src/rules/S6594/rule.ts
yassin-kammoun-sonarsource: Arf, I forgot to replace this one with my implementation.

yassin-kammoun-sonarsource: Nevermind, I think I will keep my duplicate because the function signature is a bit different, and I am not sure I would to merge both and impact other rules.

PR #4542, opened 2024-01-31
Replace deprecated "CallExpression#typeParameters" with "typeArguments"

yassin-kammoun-sonarsource:
Running fast ruling emits a huge amount of the following deprecation warning:

```
(node:24845) DeprecationWarning: The 'typeParameters' property is deprecated on CallExpression nodes.
Use 'typeArguments' instead. See https://typescript-eslint.io/linting/troubleshooting#the-key-property-is-deprecated-on-type-nodes-use-key-instead-warnings.
```

We can safely replace "CallExpression#typeParameters" with "CallExpression# typeArguments".


ilia-kebets-sonarsource (approved):
LGTM

PR #4546, opened 2024-02-01
Fix FP S6647 (`no-useless-constructor`): Ignore decorated classes and classes inheriting protected constructors

### LINKED ISSUE #4522: Fix FP S6647 (`no-useless-constructor`): Ignore decorated classes and classes inheriting protected constructors
1. False positive when a child class inherits from a parent whose constructor is protected or private; a constructor in the child class is required, otherwise there is a compilation error

```js
class Alpha {
   protected constructor() {
   }
}


class Beta extends Alpha {
   constructor() {
       super()
   }
}
```

2. We should ignore classes inheriting from external dependencies like Angular as removing the empty constructor could obstruct framework-specific runtime behaviors or mechanisms like dependency injection. The same applies to class that are decorated.

yassin-kammoun-sonarsource:
Fixes #4522


ericmorand-sonarsource (approved):
Looks good to me

PR #4548, opened 2024-02-01
Fix FP S3800 (`function-return-type`): Exception for sanitation functions

### LINKED ISSUE #4527: Fix FP S3800 (`function-return-type`): Exceptions for React components and sanitation functions
Add exception for

- React components, ignore when returning `null` or `false`
- sanitation functions: when 2 return types: string and boolean

ericmorand-sonarsource:
Fixes #4527

### DIFF
--- a/its/ruling/src/test/expected/jsts/vuetify/javascript-S3800.json
+++ b/its/ruling/src/test/expected/jsts/vuetify/javascript-S3800.json
@@ -1,37 +1,5 @@
 {
 "vuetify:packages/docs/src/examples/v-date-picker/event-events.vue": [
 44
-],
-"vuetify:packages/docs/src/examples/v-file-input/prop-validation.vue": [
-15
-],
-"vuetify:packages/docs/src/examples/v-form/misc-vee-validate.vue": [
-59,
-64,
-69,
-74,
-79
-],
-"vuetify:packages/docs/src/examples/v-form/prop-fast-fail.vue": [
-26,
-34
-],
-"vuetify:packages/docs/src/examples/v-form/rules-required.vue": [
-19
-],
-"vuetify:packages/docs/src/examples/v-form/usage.vue": [
-54,
-59,
-67,
-72
-],
-"vuetify:packages/docs/src/examples/v-input/prop-rules.vue": [
-11
-],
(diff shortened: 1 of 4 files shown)


its/ruling/src/test/expected/jsts/vuetify/javascript-S3800.json
```diff
@@ -1,37 +1,5 @@
 {
 "vuetify:packages/docs/src/examples/v-date-picker/event-events.vue": [
 44
-],
```
ericmorand-sonarsource, line 4: For information, I checked all the removed lines and can confirm that they are not triggering the rule anymore - they all belong to a sanitation function.

packages/jsts/src/rules/S3800/unit.test.ts
```diff
@@ -468,13 +491,6 @@ ruleTesterTs.run(`Functions should always return the same type [ts]`, rule, {
         },
       ],
     },
-    {
```
ericmorand-sonarsource, line 471: This test was moved from the invalid list to the valid one.

packages/jsts/src/rules/S3800/unit.test.ts
```diff
@@ -533,6 +549,14 @@ ruleTesterTs.run(`Functions should always return the same type [ts]`, rule, {
         },
       ],
     },
+    {
+      code: `
```
ericmorand-sonarsource, line 553: I added this test to guarantee that only functions with `true` and `string` return types are valid.

its/ruling/src/test/expected/jsts/vuetify/javascript-S3800.json
yassin-kammoun-sonarsource: 👍

packages/jsts/src/rules/S3800/unit.test.ts
yassin-kammoun-sonarsource: 👍

packages/jsts/src/rules/S3800/unit.test.ts
yassin-kammoun-sonarsource: 👍

yassin-kammoun-sonarsource (approved):
LGTM!

PR #4552, opened 2024-02-02
Fix S6661 (`prefer-object-spread`): Disable for projects that don't support the object spread syntax and improve rule description

### LINKED ISSUE #4518: Fix S6661 (`prefer-object-spread`): Disable for projects that don't support the object spread syntax and improve rule description
Improve message:
- using spread allows to be more explicit about the fact that you are mutating an object
- when using an empty first argument to avoid mutation, just use the spread operator

If possible, get ES version and disable rule in case it's old.

The extra case was moved to [its own issue](https://github.com/SonarSource/SonarJS/issues/4551) because it was not part of the feedback collected from SC.

ilia-kebets-sonarsource:
Fixes #4518

I checked that ruling issues were only removed for a project which minimal node version was below 8.3.0

I've organized the commits so you can rebase them without squashing.


packages/jsts/src/rules/S6661/unit.test.ts
```diff
@@ -71,3 +75,22 @@ const b = { ...foo, ...bar};`,
     },
   ],
 });
+
+clearPackageJsons();
+const project = path.join(__dirname, 'fixtures', 'unsupported-node');
+loadPackageJsons(project, []);
+const filename = path.join(project, 'file.js');
```
yassin-kammoun-sonarsource, line 82: Because this test relies on the state of local package.json files, I would replace this ESLint-based test with two comment-based tests like it is done for [S5973](https://github.com/SonarSource/SonarJS/tree/master/packages/jsts/src/rules/S5973/fixtures) does.

yassin-kammoun-sonarsource (changes requested):

packages/jsts/src/rules/S6661/unit.test.ts
yassin-kammoun-sonarsource: Never mind, a comment-based test requires a noncompliant snippet at least. That wouldn't be possible with an unsupported Node.js version. Let's keep it like this.

packages/jsts/src/project-metadata/index.ts
```diff
@@ -0,0 +1,73 @@
...
+  return isSupportedNodeVersion(filename, minVersions.node);
+}
+
+/**
+ * Check if the versions are valid semver
+ */
+function validateVersions(versions: MinimumVersions) {
+  for (const [ref, version] of Object.entries(versions)) {
+    if (semver.valid(version) === null) {
+      throw new Error(`Invalid semver version: "${version}" for "${ref}"`);
```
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: Clarified, all is good.

yassin-kammoun-sonarsource (approved):

PR #4559, opened 2024-02-08
Add rule S2004: Functions should not be nested too deeply

### LINKED ISSUE #3501: Add rule S2004: Functions should not be nested too deeply
As part of the change on #3500, we adopt a new rule to offset the changes in S3776.

yassin-kammoun-sonarsource:
Fixes #3501


yassin-kammoun-sonarsource:
[RSPEC PR](https://github.com/SonarSource/rspec/pull/3623)

PR #4565, opened 2024-02-12
Fix FP S6606 (`prefer-nullish-coalescing`): When the first argument s boolean | undefined

### LINKED ISSUE #3936: Fix FP S6606 (`prefer-nullish-coalescing`): When the first argument is `boolean | undefined`
The issue happens with Tsonfig's `strict` mode ON.

Reproducer:

```typescript
function foo(bar: boolean | undefined, baz: string) {
  return bar || baz; // FP
}
```

In case `bar` is false, the function will return `false || baz` which will be `baz`
But if we apply the proposed fix, `false ?? baz` will evaluate to `false` which is different!

This issue was notified by dogfooding our rule.

Add link to https://www.typescriptlang.org/tsconfig#strictNullChecks in RSPEC

ericmorand-sonarsource:
Also when first argument is `string | undefined` and `number | undefined`

Fixes #3936

### DIFF
--- a/its/ruling/src/test/expected/jsts/ant-design/typescript-S6606.json
+++ b/its/ruling/src/test/expected/jsts/ant-design/typescript-S6606.json
@@ -2,9 +2,6 @@
 "ant-design:components/_util/statusUtils.tsx": [
 23
 ],
-"ant-design:components/affix/index.tsx": [
-80
-],
 "ant-design:components/alert/ErrorBoundary.tsx": [
 34
 ],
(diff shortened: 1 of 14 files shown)


yassin-kammoun-sonarsource (approved):

yassin-kammoun-sonarsource:
You will need to declare the rule explicitly in `packages/jsts/src/rules/index.ts` now that it is decorated.

ericmorand-sonarsource:
@yassin-kammoun-sonarsource , can you please review this PR once more? I removed the `string` and `number` exceptions and it involved a few changes in the its tests.

yassin-kammoun-sonarsource (approved):
All good!

PR #4567, opened 2024-02-12
Check format with `prettier` before build

vdiez:
Until now, we were formatting the source code during build task, so we were not able to detect unformatted code. This adds back the format check before building the plugin

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -117,7 +117,7 @@ build_task:
   build_and_deploy_script:
     - source cirrus-env BUILD
     - npm ci
-    - npm run check-format
+    - prettier --list-different .
     - npm run build
     - regular_mvn_build_deploy_analyze -Dmaven.test.skip=true -Dmaven.install.skip=true -Dsonar.skip=true
   cleanup_before_cache_script: cleanup_maven_repository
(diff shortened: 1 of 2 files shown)


.cirrus.yml
```diff
@@ -116,6 +116,8 @@ build_task:
   <<: *NPMRC_SCRIPT_DEFINITION
   build_and_deploy_script:
     - source cirrus-env BUILD
+    - npm run bbf
```
saberduck, line 119: why do we need to run the build before?

saberduck: also I would avoid using shortcut in CI script, we can "afford" to have a long name here for clarity

vdiez: true, we don't... I will replace by `npm ci`

.cirrus.yml
```diff
@@ -116,6 +116,8 @@ build_task:
   <<: *NPMRC_SCRIPT_DEFINITION
   build_and_deploy_script:
     - source cirrus-env BUILD
+    - npm ci
+    - npx prettier --list-different .
```
saberduck, line 120: why removing `check-format`? imo I would prefer to be consistent and only use `package.json` scripts here

vdiez: that's what I understood by: 
> also I would avoid using shortcut in CI script, we can "afford" to have a long name here for clarity

saberduck (approved):

PR #4570, opened 2024-02-12
Improve S2004 (`no-nested-functions`): Increase default threshold to four nested functions

yassin-kammoun-sonarsource:
https://discuss.sonarsource.com/t/feedback-on-ts-s2004/17494/2


ilia-kebets-sonarsource (approved):
LGTM!

PR #4583, opened 2024-02-27
Use more memory for the CI: 10cores/40GB and set the Jest maxWorkers to 8 (#cores-2)

saberduck:
When using the % in the Jest `maxWorkers` parameters, it computes the number of cores from an unknown number, **not the one we provide to the Cirrus node**!

So we decided for 4GB of RAM per core as that's the default RAM per node process and set the jest maxWorkers to: `#cores-2 = 8`


ilia-kebets-sonarsource (approved):
LGTM!

PR #4594, opened 2024-03-10
Add reserved domain patterns to EXCEPTION_TOP_HOSTS list

wozitto:
SonarSource Community Forum Topic: 
https://community.sonarsource.com/t/adding-reserved-domain-patterns-to-sonarqubes-security-exception-list/110822/1

This PR addresses the need to align our security exception list with the reserved domain names outlined in [RFC 2606](https://www.rfc-editor.org/rfc/rfc2606.html).
By incorporating these changes, we aim to reduce false positives in security checks and ensure that our practices are consistent with established internet standards.

Changes:
- Updated the EXCEPTION_TOP_HOSTS array to include the .test and .example top-level domains (TLDs) as exceptions.
- Added test cases to validate that URLs containing these TLDs are correctly handled as non-sensitive and do not trigger security warnings.

### DIFF
--- a/packages/jsts/src/rules/S5332/rule.lib.ts
+++ b/packages/jsts/src/rules/S5332/rule.lib.ts
@@ -48,7 +48,7 @@ const EXCEPTION_FULL_HOSTS = [
   'graphml.graphdrawing.org',
   'json-schema.org',
 ];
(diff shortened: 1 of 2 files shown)


vdiez:
Thank you for providing this PR. I'll close this. Cherry picked your changes in #4597

PR #4603, opened 2024-03-20
Improve synchronization between Node.js parent process and worker thread

### LINKED ISSUE #4580: Improve synchronization between Node.js parent process and worker thread
Two points to improve:
* Output on worker_thread on process exit does not propagate to parent (ie. ESLint TIMING data)
* When worker thread fails with error event, parent closes server, but current request leaves response object hanging. Response should be ended before closing server.

vdiez:
Fixes #4580


packages/bridge/tests/router.test.ts
```diff
@@ -41,12 +40,14 @@ describe('router', () => {
...
-    close = promisify(server.close.bind(server));
+    const { server: serverInstance, serverClosed } = await start(port, '127.0.0.1', 60 * 60 * 1000);
+    server = serverInstance;
+    closePromise = serverClosed;
   });
 
   afterEach(async () => {
-    await close();
+    await request(server, '/close', 'POST');
+    await closePromise;
```
zglicz, line 50: Just thinking about it from the birds eye, I **would** want the `/close` request to be blocking and not to require knowing that you need to wait for a promise. Having said that, google also agrees its a _hard_ problem with jest and for the most part ppl advise putting in a timeout to allow to close the connections before jest exits..... This solution is better.

packages/bridge/tests/server.test.ts
```diff
@@ -143,10 +155,9 @@ describe('server', () => {
     expect(server.listening).toBeTruthy();
     await request(server, '/status', 'GET');
 
-    await new Promise(r => setTimeout(r, 600));
-    expect(server.listening).toBeFalsy();
-
+    await serverClosed;
```
zglicz, line 158: Explained offline, but please add a comment why this is the better way to solve this, as this is not obvious.

packages/tsconfig.json
```diff
@@ -3,7 +3,7 @@
     "target": "ES2021",
     "module": "nodenext",
     "moduleResolution": "nodenext",
-    "lib": ["ES2021"],
+    "lib": ["ES2021", "dom"],
```
zglicz, line 6: why is this needed?

zglicz (approved):
Only some questions

packages/tsconfig.json
vdiez: fetch API is only available in dom lib

Issue #4481, opened 2023-12-22
Upgrade Stylelint to v16


saberduck:

ilia-kebets-sonarsource:
related to https://github.com/SonarSource/SonarJS/issues/4600

zglicz:
As explained in the PR, we will not upgrade Stylelint to v16, as we require CommonJS module and that is incompatible with Stylelint.v16 - https://stylelint.io/migration-guide/to-16#deprecated-commonjs-api 

Hopefully, with the PR and once SonarJS moves to ESM, this will just be a version bump. We should keep this ticket alive and tackle it later.

PR #4608, opened 2024-03-21
Update usage of Stylelint, although without moving to v16

zglicz:
Improves #4481

After an initial investigation, turns out Styleint requires esmodules and is incompatible with CommonJs. Let us for now improve the usage of the plugins, as this would also be enforced in the new version of Stylelint - https://github.com/stylelint/stylelint/blob/46f5eaa7c464b56df1a37e516e21d0c64d5058e5/types/stylelint/index.d.ts#L252


.gitignore
```diff
@@ -22,6 +22,9 @@ sonarjs-*
 # Visual Studio
 .vs/
 
+# VS Code
```
zglicz, line 25: Removing VSCode, as if one wants to run Sonar with Connected-Mode the .vscode/settings.json would be different.

.vscode/settings.json
```diff
@@ -1,9 +0,0 @@
-{
-  "[typescript]": {
-    "editor.defaultFormatter": "esbenp.prettier-vscode",
-    "editor.formatOnSave": true
-  },
-  "sonarlint.connectedMode.project": {
```
zglicz, line 6: there is a key "connectionId" that links to a sonarqube instance

packages/css/src/linter/wrapper.ts
```diff
@@ -56,18 +48,6 @@ export class LinterWrapper {
...
-
-  /**
-   * Defines the wrapper's rules
-   *
-   * Besides Stylelint rules, the linter wrapper includes all the rules that
-   * are implemented internally.
-   */
-  private defineRules() {
-    for (const key in rules) {
-      stylelint.rules[key] = rules[key];
```
zglicz, line 68: here we change the internals of styleline, which is impossible in the next version.

packages/css/src/rules/S125/rule.ts
```diff
@@ -22,7 +22,7 @@
 import * as stylelint from 'stylelint';
 import { parse } from 'postcss';
 
-const ruleName = 'no-commented-code';
+const ruleName = 'sonar/no-commented-code';
```
zglicz, line 25: Just leaving a comment, that calling styelint.createPlugin doesn't actually register a plugin it only returns an object in a proper format. To correctly introduce a plugin, we need to add ```plugins``` as an option to the linter (as we now do in wrapper.ts)

vdiez (approved):
LGTM!

PR #4616, opened 2024-03-25
Skip embedded Node.js runtime deployment if `sonar.nodejs.executable` is set

### LINKED ISSUE #4400: Do not deploy embedded Node.js runtime if `sonar.nodejs.executable` is set
The embedded runtime is deployed [here in BridgeServer#startLazily()](https://github.com/SonarSource/SonarJS/blob/f992fc19ca354d7324273e0167e03d1d2d1d9d45/sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/BridgeServerImpl.java#L299), while the logic that uses the `sonar.nodejs.executable` is only accessed [afterwards, in BridgeServer#startServer()](https://github.com/SonarSource/SonarJS/blob/f992fc19ca354d7324273e0167e03d1d2d1d9d45/sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/BridgeServerImpl.java#L304).

Related to https://github.com/SonarSource/SonarJS/issues/4401

Reported in community:

https://community.sonarsource.com/t/org-sonar-plugins-javascript-nodejs-nodecommandexception-failed-to-start-server/99434/19?u=victor.diez

Also, we should not fail hard if the extraction of the runtime fails, but failover to the host runtime if present.

ilia-kebets-sonarsource:
Fixes #4400

### DIFF
--- a/sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/BridgeServerImpl.java
+++ b/sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/BridgeServerImpl.java
@@ -47,6 +47,7 @@ import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 import org.sonar.api.SonarProduct;
 import org.sonar.api.batch.sensor.SensorContext;
+import org.sonar.api.config.Configuration;
 import org.sonar.api.utils.TempFolder;
 import org.sonar.api.utils.log.Logger;
 import org.sonar.api.utils.log.Loggers;
@@ -162,9 +163,9 @@ public class BridgeServerImpl implements BridgeServer {
    *
    * @throws IOException
    */
-  void deploy() throws IOException {
+  void deploy(Configuration configuration) throws IOException {
     bundle.deploy(temporaryDeployLocation);
-    embeddedNode.deploy();
+    embeddedNode.deploy(configuration);
   }
 
   void startServer(SensorContext context, List<Path> deployedBundles) throws IOException {
@@ -285,7 +286,7 @@ public class BridgeServerImpl implements BridgeServer {
         status = Status.FAILED;
         throw new ServerAlreadyFailedException();
       }
-      deploy();
+      deploy(context.config());
       List<Path> deployedBundles = rulesBundles.deploy(temporaryDeployLocation.resolve("package"));
       rulesBundles
         .getUcfgRulesBundle()
--- a/sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/BridgeServerImpl.java
+++ b/sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/BridgeServerImpl.java
@@ -539,6 +539,13 @@ public class BridgeServerImpl implements BridgeServer {
     }
   }
 
+  /**
+   * used for tests only
+   */
+  public String getNodeCommandString() {
+    return nodeCommand.toString();
+  }
+
   @Override
   public void start() {
     // Server is started lazily from the org.sonar.plugins.javascript.eslint.EslintBasedRulesSensor
--- a/sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/BridgeServerImpl.java
+++ b/sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/BridgeServerImpl.java
@@ -21,6 +21,7 @@ package org.sonar.plugins.javascript.bridge;
 
 import static java.util.Collections.emptyList;
 import static org.sonar.plugins.javascript.bridge.NetUtils.findOpenPort;
+import static org.sonar.plugins.javascript.nodejs.NodeCommandBuilderImpl.NODE_EXECUTABLE_PROPERTY;
 
 import com.google.gson.Gson;
 import com.google.gson.JsonSyntaxException;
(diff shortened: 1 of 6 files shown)


sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/BridgeServerImpl.java
```diff
@@ -162,9 +163,9 @@ int getTimeoutSeconds() {
    *
    * @throws IOException
    */
-  void deploy() throws IOException {
+  void deploy(Configuration configuration) throws IOException {
     bundle.deploy(temporaryDeployLocation);
-    embeddedNode.deploy();
+    embeddedNode.deploy(configuration);
```
saberduck, line 168: could we move the condition check here? I don't like adding more sonar API dependencies in `EmbeddedNode`. I think `EmbeddedNode` shouldn't handle logic which is external to it.

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/EmbeddedNode.java
```diff
@@ -35,6 +36,8 @@
 import java.nio.file.Path;
 import java.util.Locale;
 import java.util.Set;
+import javax.annotation.Nullable;
+import org.sonar.api.config.Configuration;
```
saberduck, line 40: it would be better if this class doesn't depend on too many components from sonar API

saberduck (changes requested):
I think it would be better to handle the condition in BridgeServer and do not push dependencies into EmbeddedNode

sonar-plugin/sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/bridge/BridgeServerImplTest.java
```diff
@@ -750,6 +724,25 @@ void test_ucfg_bundle_version() throws Exception {
       .contains("Security Frontend version is available: [some_bundle_version]");
   }
 
+  @Test
+  void should_not_deploy_runtime_if_sonar_nodejs_executable_is_set() throws Exception {
+    var existingDoesntMatterScript = "logging.js";
+    bridgeServer = createBridgeServer(existingDoesntMatterScript);
+    bridgeServer.startServer(context, emptyList());
+    var nodeCommand = Arrays
```
ilia-kebets-sonarsource, line 732: I considered adding a method to BridgeServerImpl to retrieve the computed node path.

sonar-plugin/sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/bridge/BridgeServerImplTest.java
```diff
@@ -750,6 +724,25 @@ void test_ucfg_bundle_version() throws Exception {
...
+      .filter(s -> s.contains("node"))
+      .collect(Collectors.joining());
+    bridgeServer.stop();
+    context.setSettings(new MapSettings().setProperty(NODE_EXECUTABLE_PROPERTY, nodeCommand));
+    bridgeServer.startServerLazily(context);
+
+    assertThat(logTester.logs(INFO))
+      .contains(
+        "'" + NODE_EXECUTABLE_PROPERTY + "' is set. Skipping embedded Node.js runtime deployment."
+      );
```
ilia-kebets-sonarsource, line 743: There is currently no getter for a bridgeServer's embeddedNode instance, I could add one so we can assert a file non existence.

sonar-plugin/sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/bridge/BridgeServerImplTest.java
```diff
@@ -750,6 +723,20 @@ void test_ucfg_bundle_version() throws Exception {
...
+    bridgeServer = createBridgeServer(existingDoesntMatterScript);
+    context.setSettings(new MapSettings().setProperty(NODE_EXECUTABLE_PROPERTY, "whatever"));
+    assertThatThrownBy(() -> bridgeServer.startServerLazily(context))
+      .isInstanceOf(NodeCommandException.class);
+
+    assertThat(logTester.logs(INFO))
+      .contains(
+        "'" + NODE_EXECUTABLE_PROPERTY + "' is set. Skipping embedded Node.js runtime deployment."
+      );
+  }
```
ilia-kebets-sonarsource, line 738: Before this, my idea was to run bridgeServer.startLazily() twice. First to extract a node runtime that would be used by the second call to it.
This requires quite a lot of boilerplate code since the helper functions in BridgeServerImplTest.java have hardcoded  an unsupported platform for the BridgeServer EmbeddedNode instance.

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/EmbeddedNode.java
```diff
@@ -170,6 +170,7 @@ public void deploy() throws IOException {
     if (platform == UNSUPPORTED) {
       return;
     }
+
```
saberduck, line 173: unneeded change?

saberduck (approved):

saberduck:
> Huh, surprising that the code review still requires me specifically to review even if there is someone else who accepted. I'm used to only need to have a single accept, without counting how many request changes were there before, sorry for the wait.

I think we could configure it, maybe add a point to the weekly meeting to discuss it

PR #4617, opened 2024-03-25
Do not count LOCs for CSS files indexed as test files

### LINKED ISSUE #4615: Do not count LOCs for CSS files indexed as test files
Right now, the analyzer is counting LOCs for all CSS files, without checking if the file has been indexed as source or test. We should only count LOCs when a file has been indexed as source.

vdiez:
Fixes #4615


saberduck (approved):
good catch!

PR #4618, opened 2024-03-25
Add additional logs when embedded Node.js is not available

### LINKED ISSUE #4563: Improve logs when Node.js embedded is not usable in host
We should catch exception and better explain the user the workaround to provide local Node.js installation (either in the PATH or use `sonar.nodejs.executable`).

Decide about environment NODE_OPTIONS. Ignore it?

The goal would be to have nice logs that let users know what’s happening and how to fix it

Community: https://community.sonarsource.com/t/facing-issue-with-sonarcloud-compilation/108308/29

zglicz:
Fixes #4563 

As described in the task, we add the following logs:
1) Debug information, that the user platform is not supported
2) When the version of Node.js is unsupported, mention the possibility of setting it via the sonar.nodejs.executable property

Testing:
1) Augmented unit tests to verify that the logs get triggered
2) Manual test
- Build SonarJS
- Download SonarQube and copy the produced .jar into SonarQube extensions folder
- Change local Node version to 16
- Run Scanner on small JS project and observe in logs:
- 15:12:24.947 DEBUG: Your platform is unsupported for embedded Node.js. Falling back to host Node.js.
- 15:12:28.081 ERROR: Error during SonarScanner execution
java.lang.IllegalStateException: Error while running Node.js. A supported version of Node.js is required for running the analysis of JS/TS files. Please make sure a supported version of Node.js is available in the PATH or an executable path is provided via 'sonar.nodejs.executable' property. Alternatively, you can exclude JS/TS files from your analysis using the 'sonar.exclusions' configuration property. See the docs for configuring the analysis environment: https://docs.sonarsource.com/sonarqube/latest/analyzing-source-code/languages/javascript-typescript-css/

### DIFF
--- a/sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/AbstractBridgeSensor.java
+++ b/sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/AbstractBridgeSensor.java
@@ -19,6 +19,8 @@
  */
 package org.sonar.plugins.javascript.bridge;
 
+import static org.sonar.plugins.javascript.nodejs.NodeCommandBuilderImpl.NODE_EXECUTABLE_PROPERTY;
+
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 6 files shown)


vdiez (approved):
LGTM!

PR #4621, opened 2024-03-25
Create rule S6958: No literal call

### LINKED ISSUE #4509: Create rule S6958: No literal call
https://github.com/SonarSource/eslint-plugin-sonarjs/issues/437

ericmorand-sonarsource:
Fixes #4509


packages/jsts/src/rules/S6958/unit.test.ts
```diff
@@ -0,0 +1,258 @@
...
+        {
+          message: 'Literal should not be used as function.',
+          line: 1,
+          column: 1,
+          endColumn: 8,
+        },
+      ],
+    },
+    {
+      code: '`hello```;',
```
zglicz, line 137: doesn't typescript catch this? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

image

packages/jsts/src/rules/S6958/unit.test.ts
```diff
@@ -0,0 +1,258 @@
...
+    { code: '(() => 0)``;' },
+  ],
+  invalid: [
+    {
+      code: 'true();',
+      errors: [
+        {
+          message: 'Literal should not be used as function.',
+          line: 1,
+          column: 1,
```
zglicz, line 43: A part of me looking at these repeated lines of code would love to add a helper to simplify these messages, which would help if these ever need to be updated and do:
a) use a constant for these messages - only 2 messages being repeated
b) automagically finding the last () or `` occurrence

or maybe its just me 😉

packages/jsts/src/rules/S6958/rule.ts
```diff
@@ -0,0 +1,52 @@
...
+export const rule: Rule.RuleModule = {
+  meta: {
+    messages: {
+      asFunction: 'Literal should not be used as function.',
+      asTagFunction: 'Literal should not be used as tag function.',
+    },
+  },
+
+  create(context) {
+    const pattern = /^(?:(?:Array|Class|Object)Expression|(?:Template)?Literal)$/u;
```
zglicz, line 32: eh, all languages have different regex-looking operators. Just to see if I understand, the types we are interested in are:
Expression
ArrayExpression
ClassExpression
ObjectExpression
Literal
TemplateLiteral

For the sake of readability, I would just list these in a set and check if the type is any one of them. You could even reference the actual constant from eslint like: TSESTree.AST_NODE_TYPES.ArrayExpression

zglicz (commented):
Really cool! It's my first time seeing a new rule getting added. 
Will leave to other folks to Approve/Request changes, if there is something else to add or missing.

packages/jsts/src/rules/S6958/unit.test.ts
ilia-kebets-sonarsource: I think you can use [messageId](https://eslint.org/docs/latest/integrate/nodejs-api#testing-errors-with-messageid).

packages/jsts/src/rules/S6958/unit.test.ts
ilia-kebets-sonarsource: It might still be worth it for JS files

sonar-plugin/javascript-checks/src/test/java/org/sonar/javascript/checks/NoLiteralCallCheckTest.java
ilia-kebets-sonarsource, line 1: That is unnecessary. AFAIK, We only use unit tests for these check classes when we provide configuration to a rule, to verify that the resulting JSON serialization works as expected as done [here](https://github.com/SonarSource/SonarJS/blob/master/sonar-plugin/javascript-checks/src/test/java/org/sonar/javascript/checks/EmptyBlockCheckTest.java).

ilia-kebets-sonarsource: - If you use an unexistent ruleId, you get a failure `CheckListTest.test:77 No description for S6999`
- if 2 checks use the same ruleId, you get a failure:
```
[ERROR]   CheckListTest.test:88
Found duplicate(s):
 <["S1874"]>
 ```
 - If you use a non-existent eslintKey, you get a failure in `TypeScriptRulesDefinitionTest.compatibleLanguagesInJson`
 - If you have 2 checks using the same eslintKey, you get:
 ```
[ERROR] Failures:
[ERROR]   CheckListTest.test_eslint_key:109
Found duplicate(s):
 <["no-literal-call"]>
```

packages/jsts/src/rules/S6958/rule.ts
ilia-kebets-sonarsource: I have ignored the proposition to reference constants as the types of AST nodes are quite stable.

ilia-kebets-sonarsource:
closing and continuing in https://github.com/SonarSource/SonarJS/pull/4638

PR #4626, opened 2024-03-26
Create rule S6959: "Array.reduce()" calls should include an initial value

### LINKED ISSUE #4591: Create rule S6959: "Array.reduce()" calls should include an initial value
Array.reduce() should include an initial value to avoid runtime errors, especially if the array is empty. When calling Array.reduce() without providing an initial value and if the array is empty, JavaScript throws a TypeError. This error occurs because there are no elements in the array to iterate over and no initial value to serve as a starting point for the reduction operation. By including an initial value, developers can ensure that the reduction process has a defined starting point even if the array is empty.

__Noncompliant__

```javascript
function sum(numbers) {
  return numbers.reduce((acc, current) => acc + current); // Noncompliant
}
console.log(sum([1, 2, 3, 4, 5])); // Prints 15
console.log(sum([])); // TypeError: Reduce of empty array with no initial value
```
__Compliant__

```javascript
function sum(numbers) {
  return numbers.reduce((acc, current) => acc + current, 0);
}
console.log(sum([1, 2, 3, 4, 5])); // Prints 15
console.log(sum([])); // Prints 0
```

Suggested by @erichirlemann.

yassin-kammoun-sonarsource:
Fixes #4591


packages/jsts/src/rules/S6959/cb.fixture.ts
```diff
@@ -0,0 +1,13 @@
+function noncompliant(xs: number[]) {
+    return xs.reduce((acc, x) => acc + x); // Noncompliant {{Add an initial value to this "reduce()" call.}}
+}
+
+function compliant(xs: number[]) {
+    return xs.reduce((acc, x) => acc + x, 0); // Compliant
+}
+
+function coverage(x: any) {
```
zglicz, line 9: what is this for?

sonar-plugin/javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S6959.json
```diff
@@ -0,0 +1,27 @@
+{
+  "title": "\"Array.reduce()\" calls should include an initial value",
+  "type": "BUG",
+  "status": "ready",
+  "remediation": {
+    "func": "Constant\/Issue",
+    "constantCost": "5min"
```
zglicz, line 7: heh, that's generous if not intimidating

zglicz (approved):
Please add an invalid test case before merging

sonar-plugin/javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S6959.json
yassin-kammoun-sonarsource: Good point! Could you actually comment on this [PR](https://github.com/SonarSource/rspec/pull/3818)?

Each rule has a description, and a rule description is stored in the [RSPEC](https://github.com/SonarSource/rspec) GitHub repository, a company-wide database for all Sonar rule descriptions. New descriptions or description updates are first introduced in this repository through pull requests and eventually merged. To bring back those changes in a given analyzer, we then use an internal tool called `rule-api` that fetches any changes for any rule of the analyzer or changes for a specific rule.

You will get familiar with this tool tomorrow when starting the release process of the analyzer.

packages/jsts/src/rules/S6959/cb.fixture.ts
yassin-kammoun-sonarsource: This file as a whole is not executed at all but rather denotes a test file with actual unit tests. The file uses a homemade testing framework that we call [comment-based testing framework](https://github.com/SonarSource/SonarJS/blob/master/docs/DEV.md#comment-based-testing). A test file contains special comments indicating that a rule should raise an issue at specific locations. For instance, this is what is meant here:

https://github.com/SonarSource/SonarJS/blob/5f2dbfe2f032b15e8a3f331c14a41eaa8820fe34/packages/jsts/src/rules/S6959/cb.fixture.ts#L1-L3

The comment `// Noncompliant` indicates that an issue should be raised at line 2 and whatever is inside `{{...}}` is the expected message of the reported issue. We could assert even more things like code fixes, starting and ending columns, and secondary locations.

Coming back to your question, I guess you should understand by now what's the purpose of the function `coverage`. No issue will be raised here, the rule will visit those method invocations for the sake of increasing the code coverage of the rule implementation, this line in other words:

https://github.com/SonarSource/SonarJS/blob/5f2dbfe2f032b15e8a3f331c14a41eaa8820fe34/packages/jsts/src/rules/S6959/rule.ts#L39

packages/jsts/src/rules/S6959/unit.test.ts
```diff
@@ -0,0 +1,31 @@
...
+import { rule } from './';
+
+const ruleTester = new RuleTester({ parserOptions: { ecmaVersion: 2018 } });
+ruleTester.run('"Array.reduce()" calls should include an initial value', rule, {
+  valid: [
+    {
+      code: 'xs.reduce((acc, x) => acc + x);',
+    },
+  ],
+  invalid: [],
```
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: If you come back to the rule implementation, you might have noticed this weird check:

https://github.com/SonarSource/SonarJS/blob/5f2dbfe2f032b15e8a3f331c14a41eaa8820fe34/packages/jsts/src/rules/S6959/rule.ts#L32-L35

While the function `isRequiredParserServices` isn't well-named, it actually tests that type-checking is available. If not, the rule turns into a noop. In order to test this behavior, this file uses ESLint rule tester for code coverage purposes. We currently don't have any other way to test this edge case (or at least, we haven't invested time on a better solution so far). That's why there is intentionally no invalid test case here.

packages/jsts/src/rules/S6959/cb.fixture.ts
zglicz: So this is just to increase code-coverage? That seems counter-productive and actually adds confusion where it shouldn't.

packages/jsts/src/rules/S6959/unit.test.ts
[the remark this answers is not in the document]
zglicz: I'm reading and re-reading and I don't understand. 
In this test, is there type-checking available?
I assume no, and thanks to that this rule becomes a no-op, as the code sample that is provided is missing the initial value.

Why isn't there typechecking in this test then? What should be the parserOptions to include type-checking? Or am I completely off-base here?

zglicz (commented):
Thank you for the explanations. It's now making a bit more sense, but still I left a few questions.

packages/jsts/src/rules/S6959/unit.test.ts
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: When we parse JS/TS code, we rely on [TypeScript ESLint parser](https://typescript-eslint.io/packages/parser/) which is able to parse both syntaxes. In addition, TypeScript ESLint wraps some utilities from TypeScript compiler, including type-checking. However, to benefit from type-checking, you need to configure the parsing options with a TSConfig somewhere.

By default, ESLint's rule tester uses Esprima (I think) to parse the code being tested. If you want to have type-checking available while using this rule tester, you first need to manually configure the parser to use as follows:

https://github.com/SonarSource/SonarJS/blob/911655ef461fc9aaa93b9daeb0a526729a80f998/packages/jsts/src/rules/S117/unit.test.ts#L23-L26

However, this is not enough as it only allows us to parse TypeScript syntax as well. To enable type-checking, you need to tweak ESLint's rule tester with a TSConfig. I am not going to show how because we want to avoid that and prefer using the comment-based testing framework which always makes type-checking available.

> I'm reading and re-reading and I don't understand. In this test, is there type-checking available?

No, type-checking is not available because we use ESLint's rule tester without customization.

> I assume no, and thanks to that this rule becomes a no-op, as the code sample that is provided is missing the initial value.

Exactly!

> Why isn't there typechecking in this test then? What should be the parserOptions to include type-checking? Or am I completely off-base here?

1. Configure the parser options to use TypeScript ESLint parser
2. Configure ESLint's rule tester with a TSConfig

packages/jsts/src/rules/S6959/cb.fixture.ts
yassin-kammoun-sonarsource: I don't necessarily agree that's confusing since the test file explicitly separates what's noncompliant from what's compliant and what's for coverage. Maybe I am missing something to make it less confusion.

Feel free to suggest anything that comes to your mind :)

packages/jsts/src/rules/S6959/rule.ts
```diff
@@ -0,0 +1,60 @@
...
+  create(context: Rule.RuleContext) {
+    const services = context.parserServices;
+
+    function isArray(node: estree.Node) {
+      if (isRequiredParserServices(services)) {
+        return isArrayType(node, services);
+      } else {
+        return isArrayExpression(getUniqueWriteUsageOrNode(context, node));
+      }
+    }
```
yassin-kammoun-sonarsource, line 47: @zglicz I updated the array detection so that it tries to determine syntactically if a node denotes an array in case type-checking is missing for whatever reason.

PR #4628, opened 2024-03-26
Resolve issue #4459: Fix FP S2699 (`Tests should include assertions`): global `expect()` aren't detected

### LINKED ISSUE #4459: Fix FP S2699 (`Tests should include assertions`): global `expect()` aren't detected
Assertion functions such as `expect()` are not detected when available globally.

This issue was raised in https://community.sonarsource.com/t/false-positive-s2699-add-at-least-one-assertion-to-this-test-case-with-repro/103847

Reproducer:

[link](https://github.com/HugoMercierYuc/sonar-jest-sinon/blob/de90e77a560bdfd99323f2bcbf1d654e47f94273/src/app/app.component.spec.ts#L1-L34)

```ts
import { TestBed } from '@angular/core/testing';
import { AppComponent } from './app.component';
import { NxWelcomeComponent } from './nx-welcome.component';
import { RouterTestingModule } from '@angular/router/testing';
import { spy } from 'sinon';

describe('AppComponent', () => {
  beforeEach(async () => {
    await TestBed.configureTestingModule({
      imports: [AppComponent, NxWelcomeComponent, RouterTestingModule],
    }).compileComponents();
  });

  it('should render title', () => {
    const fixture = TestBed.createComponent(AppComponent);
    fixture.detectChanges();
    const compiled = fixture.nativeElement as HTMLElement;
    expect(compiled.querySelector('h1')?.textContent).toContain('Welcome org');
  });

  it(`should have as title 'org'`, () => {
    const fixture = TestBed.createComponent(AppComponent);
    const app = fixture.componentInstance;
    expect(app.title).toEqual('org');
  });

  it('should not call action', () => {
    const fixture = TestBed.createComponent(AppComponent);
    const app = fixture.componentInstance;
    const spyFn = spy(app, 'action');
    fixture.detectChanges();
    expect(spyFn.calledOnce).toBeFalsy();
  });
});
```

ericmorand-sonarsource:
Fixes #4459

### DIFF
--- a/packages/jsts/src/rules/S2699/vitest.fixture.js
+++ b/packages/jsts/src/rules/S2699/vitest.fixture.js
@@ -19,6 +19,10 @@ describe('vitest test cases', () => {
     check();
   });
 
+  it('recognizes global expect as an assertion', () => {
+    expect(5).toEqual(5);
+  });
+
   function check() {
     expect(1).toEqual(2);
   }
--- a/packages/jsts/src/rules/S2699/rule.ts
+++ b/packages/jsts/src/rules/S2699/rule.ts
@@ -90,7 +90,14 @@ class TestCaseAssertionVisitor {
       return;
     }
     if (isFunctionCall(node)) {
-      const functionDef = resolveFunction(this.context, node.callee);
+      const { callee } = node;
+
+      if (callee.name === 'expect') {
+        this.hasAssertions = true;
+        return;
+      }
+
+      const functionDef = resolveFunction(this.context, callee);
       if (functionDef) {
         this.visit(context, functionDef.body, visitedNodes);
       }


vdiez (approved):

packages/jsts/src/rules/S2699/vitest.fixture.js
```diff
@@ -19,6 +19,10 @@ describe('vitest test cases', () => {
     check();
   });
 
+  it('recognizes global expect as an assertion', () => {
```
zglicz, line 22: qq: this test is exactly the same as the one in lines 10-12. Whats the change here? Why is it marked as compliant when it seems it should only become compliant after your update?

ericmorand-sonarsource: That's a very good point. The test I added does not prove anything since `expect` is imported in this file and my test intended to prove that `expect` was recognized as an assertion even when it is _not_ imported.

I'll add a dedicated test. Thanks for noticing.

PR #4636, opened 2024-03-27
Resolve issue #4631: Fix FP S2699 (`assertions-in-tests`): Add support for `supertest`

### LINKED ISSUE #4631: Fix FP S2699 (`assertions-in-tests`): Add support for `supertest`
[Supertest](https://www.npmjs.com/package/supertest) is a lib to test API routes, compatible with express. It implements an `expect()` function like the following.

This issue was discovered as part of the March '24 coverage hardening.

Reproducer:
```ts
import request from 'supertest';

describe('foo', function () {
    it('should work', function () {
        return request(app).get(`/foo/bar`).expect('Content-Type', /json/u).expect(200);
    });
});

ericmorand-sonarsource:
Fixes #4631

### DIFF
--- a/its/ruling/src/test/expected/jsts/file-for-rules/javascript-S2699.json
+++ b/its/ruling/src/test/expected/jsts/file-for-rules/javascript-S2699.json
@@ -1,0 +1,5 @@
+{
+"file-for-rules:tests/S2699.js": [
+17
+]
+}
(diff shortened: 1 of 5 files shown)


ilia-kebets-sonarsource (changes requested):
The implementations looks good, could you add a custom ruling file to verify the change in an integration test?

ilia-kebets-sonarsource (approved):
LGTM!

wtospit:
Hi @ericmorand-sonarsource , I came across this because I have found that Supertest assertions can be another form.

For example

`supertest(...).post(...).send(...).expect(...)`
or
`supertest(...).get(...).set(....).expect(...)`


so I think the above examples might break this rule right ?

PR #4637, opened 2024-03-27
Fix FP S6754 (`hook-use-state`): Ignore state variables without a setter

### LINKED ISSUE #4627: Fix FP S6754 (`hook-use-state`): Ignore state variables without a setter
The rule reports an issue when the destructured return value of `useState` only includes the state variable without the corresponding setter. This seems like an antipattern but it's outside the scope of the rule since it should only care about symmetrical naming.

```javascript
const [value /*, missing setValue */] = useState(...);
```

yassin-kammoun-sonarsource:
Fixes #4627


ilia-kebets-sonarsource (approved):
LGTM!

PR #4638, opened 2024-03-27
Create rule S6958: No literal call

### LINKED ISSUE #4509: Create rule S6958: No literal call

ilia-kebets-sonarsource:
Fixes #4509 

Continuing work from https://github.com/SonarSource/SonarJS/pull/4621


packages/jsts/src/rules/S6958/rule.ts
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+import { type Rule } from 'eslint';
+import { type Node } from 'estree';
+
+const NODES = new Set([
+  'Expression',
```
zglicz, line 24: you don't want to reference the AST_NODE_TYPE like here - https://github.com/SonarSource/eslint-plugin-sonarjs/blob/master/src/rules/no-use-of-empty-return-value.ts#L27

I would advise for that if possible, that way these constants will be kept in sync and won't be just magical constants without any tie to any repository.

zglicz (approved):
single suggestion

packages/jsts/src/rules/S6958/rule.ts
zglicz: Sorry, I'm still not very good at not missing github comments. Fair enough.

ilia-kebets-sonarsource: I've decided to go with the variables as you suggested, although, I believe that we should agree on a practice at the squad level.
I still think it's less readable, and makes working with our codebase that mixes types from EStree and TSESTree messy.

Issue #4641, opened 2024-03-28
Remove usage of deprecated logging APIs


saberduck:
Sonar logging API was deprecated in favor of standard slf4j

PR #4646, opened 2024-03-28
Remove usage of deprecated APIs

saberduck:
Fixes https://github.com/SonarSource/SonarJS/issues/4641

### DIFF
--- a/sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/BridgeServer.java
+++ b/sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/BridgeServer.java
@@ -19,7 +19,7 @@
  */
 package org.sonar.plugins.javascript.bridge;
 
-import static org.sonarsource.api.sonarlint.SonarLintSide.MULTIPLE_ANALYSES;
+import static org.sonarsource.api.sonarlint.SonarLintSide.INSTANCE;
 
 import java.io.IOException;
 import java.util.List;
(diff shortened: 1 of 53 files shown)


sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/BridgeServer.java
```diff
@@ -33,7 +33,7 @@
 import org.sonarsource.api.sonarlint.SonarLintSide;
 
 @ScannerSide
-@SonarLintSide(lifespan = MULTIPLE_ANALYSES)
+@SonarLintSide(lifespan = INSTANCE)
```
ilia-kebets-sonarsource, line 36: Is `INSTANCE` the replacement for the deprecated `MULTIPLE_ANALYSES`?

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/BridgeServerImpl.java
```diff
@@ -177,7 +175,8 @@ void deploy(Configuration configuration) throws IOException {
   }
 
   void startServer(SensorContext context, List<Path> deployedBundles) throws IOException {
-    PROFILER.startDebug("Starting server");
+    LOG.debug("Starting server");
+    long start = System.currentTimeMillis();
```
ilia-kebets-sonarsource, line 179: Is there no equivalent for Profiler in slf4j?

ilia-kebets-sonarsource (approved):
Some questions otherwise LGTM!

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/BridgeServer.java
saberduck: yes, you can check the javadoc

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/BridgeServerImpl.java
saberduck: unfortunately no, if really needed we can create something similar on our own

PR #4652, opened 2024-04-05
Change min required Node.JS version to 18.17.0

### LINKED ISSUE #4650: Use 18.17.0 as minimum version for Node.JS
We have some outdated images on different AML services which have 18.17.0 Node.JS and use Alpine so can't use embedded runtime. Since the requirement is formal, we should lower the version to avoid friction.

See original ticket #4440

saberduck:
Fixes #4650

### DIFF
--- a/sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/NodeDeprecationWarning.java
+++ b/sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/NodeDeprecationWarning.java
@@ -54,7 +54,7 @@ public class NodeDeprecationWarning {
    */
   static final Version MIN_SUPPORTED_NODE_VERSION = Version.create(18, 17, 0);
   static final int MIN_RECOMMENDED_NODE_VERSION = 18;
-  static final List<Integer> RECOMMENDED_NODE_VERSIONS = Arrays.asList(18, 20);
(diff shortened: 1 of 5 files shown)


sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/bridge/NodeDeprecationWarning.java
```diff
@@ -52,7 +52,7 @@ public class NodeDeprecationWarning {
    * <p>
    * The minor version is a requirement from the ESLint version that the bridge uses.
    */
-  static final Version MIN_SUPPORTED_NODE_VERSION = Version.create(18, 17, 1);
+  static final Version MIN_SUPPORTED_NODE_VERSION = Version.create(18, 17, 0);
   static final int MIN_RECOMMENDED_NODE_VERSION = 18;
   static final List<Integer> RECOMMENDED_NODE_VERSIONS = Arrays.asList(18, 20);
   static final List<Integer> ALL_RECOMMENDED_NODE_VERSIONS = Arrays.asList(18, 20, 21);
```
ilia-kebets-sonarsource, line 58: as mentioned on Slack, should we set the minor versions here as well?

ilia-kebets-sonarsource (approved):
See minor proposition, otherwise LGTM: 
https://github.com/SonarSource/SonarJS/pull/4652/files#r1553725167

PR #4662, opened 2024-04-16
Resolve JS-3 Stricter rule implementation for detecting a valid header

zglicz:
Trying to expand on the logic in https://github.com/SonarSource/sonar-html/blob/master/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableWithoutHeaderCheck.java

Trying to be stricter about the implementation and the logic applied is as follows:
1. Recursively read the contents of a <table> (skipping nested ```<table>```'s)
2. Expand the colspan's
3. Expand the rowspan's and build the full table "grid"
4. Check that there exists any row or any column that is a header.

For the most part following the spec - https://developer.mozilla.org/en-US/docs/Web/HTML/Element/th 
The only missing thing is not breaking up separately the ```<thead> ```, ``` <tbody> ``` and ``` <tfoot> ``` and running the full grid check.


packages/jsts/src/rules/S5256/rule.ts
```diff
@@ -0,0 +1,220 @@
...
+    }
+  });
+  return row;
+}
+
+function extractRows(context: Rule.RuleContext, tree: TSESTree.JSXElement): TableCell[][] | null {
+  const rows: TableCell[][] = [];
+  let unknownTableStructure = false;
+  tree.children.forEach(child => {
+    if (child.type === 'JSXElement') {
```
vdiez, line 134: reminder: handle `JSXFragment`

vdiez (approved):
LGTM!

PR #4665, opened 2024-04-17
Add rule S5264(`object-alt-content`)

vdiez:
Fixes https://sonarsource.atlassian.net/browse/JS-7

RSPEC: https://github.com/SonarSource/rspec/pull/3889


packages/jsts/src/rules/S5264/rule.ts
```diff
@@ -0,0 +1,77 @@
...
+        const jsxNode = node as unknown as TSESTree.JSXElement;
+        const type = elementType(jsxNode.openingElement);
+        if (type.toLowerCase() !== 'object') {
+          return;
+        }
+
+        if (!hasVisibleChildren(jsxNode.children)) {
+          context.report({
+            node,
+            messageId: 'addContent',
```
zglicz, line 71: whats the benefit of a message in the meta versus using a constant?

zglicz (approved):

packages/jsts/src/rules/S5264/rule.ts
vdiez: just a different way, easier to reference from RuleTester

PR #4666, opened 2024-04-18
[JS-11] Rename rule S6854 to S1090

vdiez:
Fixes https://sonarsource.atlassian.net/browse/JS-11

RSPEC change: https://github.com/SonarSource/rspec/pull/3894


sonar-plugin/javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/Sonar_way_profile.json
```diff
@@ -326,7 +326,7 @@
     "S6851",
     "S6852",
     "S6853",
-    "S6854",
+    "S1090",
```
zglicz, line 329: would it be better to have these sorted? In the future, it'd make it much easier to check if sth is in this list

zglicz (approved):

PR #4669, opened 2024-04-19
Resolve JS-6 Implement S5260 in js/ts

zglicz:
Reusing most of the logic from S5256 to compute the grid structure. Taking that logic to a helper file about tables.


packages/jsts/src/rules/helpers/table.ts
```diff
@@ -0,0 +1,217 @@
...
+  const spanAttr = getProp(tree.openingElement.attributes, spanKey);
+  if (spanAttr) {
+    span = parseInt(String(getLiteralPropValue(spanAttr)));
+  }
+  return span;
+}
+
+function rowSpan(tree: TSESTree.JSXElement): number {
+  let value = computeSpan(tree, 'rowspan');
+  const MAX_ROW_SPAN = 65534;
```
vdiez, line 50: no need to declare const on each call, you can move this to the top of the file.

packages/jsts/src/rules/helpers/table.ts
```diff
@@ -0,0 +1,217 @@
...
+  if (value > MAX_ROW_SPAN) {
+    value = MAX_ROW_SPAN;
+  }
+  return value;
+}
+
+function colSpan(tree: TSESTree.JSXElement): number {
+  let value = computeSpan(tree, 'colspan');
+  const MAX_INVALID_COL_SPAN = 10000;
```
vdiez, line 59: same

packages/jsts/src/rules/helpers/table.ts
```diff
@@ -0,0 +1,217 @@
...
+
+  tree.children.forEach(child => {
+    if (child.type === 'JSXElement') {
+      const childType = getElementType(context)(child.openingElement).toLowerCase();
+      if (childType === 'tr') {
+        rows.push(extractRow(child));
+      } else if (childType === 'table') {
+        // skip
+      } else {
+        const KNOWN_TABLE_STRUCTURE_ELEMENTS = ['thead', 'tbody', 'tfoot'];
```
vdiez, line 137: same, move this declaration to top of the page

packages/jsts/src/rules/helpers/table.ts
```diff
@@ -0,0 +1,217 @@
...
+        if (KNOWN_TABLE_STRUCTURE_ELEMENTS.includes(childType)) {
+          handleInternalStructure(child);
```
vdiez, line 139: why not calling direclty extractRows? don't understand the need for `handleInternalStructure`

packages/jsts/src/rules/helpers/table.ts
```diff
@@ -0,0 +1,217 @@
...
+          unknownTableStructure = true;
+        }
+      }
+    } else if (
+      child.type === 'JSXExpressionContainer' &&
+      child.expression.type !== 'JSXEmptyExpression'
+    ) {
+      unknownTableStructure = true;
+    } else if (child.type === 'JSXFragment') {
+      handleInternalStructure(child);
```
vdiez, line 150: same, why not calling directly extractRows?

vdiez (changes requested):

packages/jsts/src/rules/helpers/table.ts
zglicz: as in to do what?
  rows.concat(extractRows) ?
We need to perform the check if its null to know if its an unknown structure. I do believe we need to run the logic from handleInternalStructure to correctly propagate the unknown structure or add the rows.

packages/jsts/src/rules/helpers/isHtmlElement.ts
```diff
@@ -175,6 +175,6 @@ const HTML_TAG_NAMES = new Set([
 export function isHtmlElement(node: TSESTree.JSXElement) {
   return (
     node.openingElement.name.type === 'JSXIdentifier' &&
-    HTML_TAG_NAMES.has(node.openingElement.name.name.toLowerCase())
+    HTML_TAG_NAMES.has(node.openingElement.name.name)
```
vdiez, line 178: why remove the lowercase comparison?

zglicz: While checking on peach, I found this code snippet - https://peach.aws-prd.sonarsource.com/issues?projects=js%3Agutenberg&resolved=false&rules=javascript%3AS5256&open=AY70ITYshdJEwEPe_w_q

JSX is case sensitive and we can differentiate between, in this case, <Section> vs <section>

`<table
	className={ classes === '' ? undefined : classes }
	style={ { ...colorProps.style, ...borderProps.style } }
>
	<Section type="head" rows={ head } />
	<Section type="body" rows={ body } />
	<Section type="foot" rows={ foot } />
</table> `

packages/jsts/src/rules/helpers/table.ts
vdiez: got it, thx

packages/jsts/src/rules/helpers/table.ts
```diff
@@ -0,0 +1,216 @@
...
+  if (rows.length === 0) {
+    return [];
+  }
+  const nbColumns = rows[0].length;
+  const columns: (TableCellInternal | undefined)[] = Array.from({ length: nbColumns });
+  let row = 0;
+  const result: TableCell[][] = [];
+  while (row < rows.length) {
+    const resultRow: TableCell[] = [];
+    let rowIndex = 0;
```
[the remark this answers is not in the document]
zglicz: Renamed to indexInRow

packages/jsts/src/rules/S5260/rule.ts
```diff
@@ -0,0 +1,130 @@
...
+        const elementType = getElementType(context)(tree.openingElement);
+        if (elementType === 'table') {
+          verifyHeaderReferences(tree);
+        }
+      },
+    };
+  },
+};
+
+function compileBlockInfo(grid: TableCell[][]) {
```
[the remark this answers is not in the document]
zglicz: Added a comment block

vdiez (approved):

PR #4670, opened 2024-04-22
Check for unknown tags in tr's

zglicz:
Talking with @vdiez - we figured out that the tr's aren't checking for custom other tags. Let us be very strict and only perform a check if under a `<tr>` we have `<td>` or `<th>`


vdiez (approved):
LGTM!

PR #4677, opened 2024-04-29
Resolve JS-147 - Adapt SonarJS to the next ESLint Plugin

ericmorand-sonarsource:
Fixes JS-147


package.json
```diff
@@ -79,17 +79,17 @@
...
+    "@typescript-eslint/utils": "^7.7.1",
     "builtin-modules": "3.3.0",
     "bytes": "3.1.2",
     "eslint": "8.57.0",
     "eslint-plugin-import": "2.29.1",
     "eslint-plugin-jsx-a11y": "6.8.0",
     "eslint-plugin-react": "7.34.1",
     "eslint-plugin-react-hooks": "4.6.0",
-    "eslint-plugin-sonarjs": "0.25.0",
+    "eslint-plugin-sonarjs": "^0.26.0-2918",
```
zglicz, line 92: is this an internal only number? As in, it's not possible to get this version from npm

packages/jsts/src/rules/S6647/rule.ts
```diff
@@ -22,6 +22,7 @@ import { Rule } from 'eslint';
 import { eslintRules } from '../core';
 import { decorate } from './decorator';
 import { getVariableFromName } from '../helpers';
+import estree from 'estree';
```
zglicz, line 25: could you make do with just importing the type?

zglicz (approved):
Makes sense. Glad that this works!

packages/jsts/src/rules/S6647/rule.ts
ericmorand-sonarsource: You, it is a mistake. Thanks for noticing, let me remove.

package.json
```diff
@@ -9,7 +9,7 @@
     "build:fast": "npm run bridge:build:fast && npm run _:plugin:prepare-bridge && npm run _:plugin-fetch-node && npm run plugin:build:fast && npm run update-ruling-data",
     "bf": "npm run build:fast",
     "new-rule": "ts-node tools/newRule.ts",
-    "ruling": "node tools/prepare-ruling.js && jest ./packages/ruling/tests/projects/*.ruling.test.ts",
+    "ruling": "node tools/prepare-ruling.js && jest --runInBand ./packages/ruling/tests/projects/*.ruling.test.ts",
```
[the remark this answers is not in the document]
ericmorand-sonarsource: You, it is a mistake. Thanks for noticing, let me remove.

package.json
ericmorand-sonarsource: It is. When we push the 1.0.0 version, I propose that we update this accordingly.

PR #4679, opened 2024-05-02
Resolve JS-81- Options Schemas required for ESLint v9

ericmorand-sonarsource:
Just testing from the CI to prevent my computer from crashing.

Fixes JS-81

### DIFF
--- a/packages/jsts/src/rules/S107/rule.ts
+++ b/packages/jsts/src/rules/S107/rule.ts
@@ -62,7 +62,7 @@ export const rule: RuleModule<Options> = {
     const ruleDecoration: Rule.RuleModule = interceptReport(
       eslintMaxParams,
       function (context: Rule.RuleContext, descriptor: Rule.ReportDescriptor) {
-        const maxParams = (context.options as Options)[0].maximumFunctionParameters;
+        const [{ maximumFunctionParameters }] = context.options as Options;
         if ('node' in descriptor) {
           const functionLike = descriptor.node as TSESTree.FunctionLike;
           if (!isException(functionLike)) {
@@ -76,7 +76,8 @@ export const rule: RuleModule<Options> = {
 
         function isBeyondMaxParams(functionLike: TSESTree.FunctionLike) {
           return (
-            functionLike.params.filter(p => p.type !== 'TSParameterProperty').length <= maxParams
+            functionLike.params.filter(p => p.type !== 'TSParameterProperty').length <=
+            maximumFunctionParameters
           );
         }
 
--- a/packages/jsts/src/rules/S107/rule.ts
+++ b/packages/jsts/src/rules/S107/rule.ts
@@ -30,12 +30,29 @@ import {
   interceptReport,
   mergeRules,
 } from '../helpers';
+import type { RuleModule } from '../../../../shared/src/types/rule';
 
 const eslintMaxParams = eslintRules['max-params'];
 
-export const rule: Rule.RuleModule = {
+export type Options = [
+  {
+    maximumFunctionParameters: number;
+  },
+];
+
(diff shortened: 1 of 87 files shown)


sonar-plugin/javascript-checks/src/test/java/org/sonar/javascript/checks/NoNestedFunctionsCheckTest.java
ericmorand-sonarsource, line 1: Previously, there was no test at all for this one.

sonar-plugin/javascript-checks/src/main/java/org/sonar/javascript/checks/GetterSetterCheck.java
```diff
@@ -43,11 +43,22 @@ public class GetterSetterCheck implements EslintBasedCheck {
 
   @Override
   public List<Object> configurations() {
-    return Collections.singletonList(getWithoutSet);
+    return Collections.singletonList(
+      new Config(getWithoutSet)
```
zglicz, line 47: this.

sonar-plugin/javascript-checks/src/main/java/org/sonar/javascript/checks/HardcodedCredentialsCheck.java
```diff
@@ -43,11 +44,22 @@ public class HardcodedCredentialsCheck implements EslintBasedCheck {
 
   @Override
   public List<Object> configurations() {
-    return Arrays.asList((Object[]) credentialWords.split("\\s*,\\s*"));
+    return Collections.singletonList(
+      new Config(credentialWords.split("\\s*,\\s*"))
```
zglicz, line 48: this or not this?

zglicz (approved):
Talked offline about the typing - very slick.
The Java class serve no purpose, except for storing the default values. Except for that, they could be autogenerated. If we were to move these definitions into js/ts, we could fully auto generate them and remove the need for the additional tests to test passing of simple values....

All in all looks good and a very positive chang
lgtmacro
e!

packages/jsts/src/rules/S107/rule.ts
```diff
@@ -45,7 +62,7 @@ export const rule: Rule.RuleModule = {
     const ruleDecoration: Rule.RuleModule = interceptReport(
       eslintMaxParams,
       function (context: Rule.RuleContext, descriptor: Rule.ReportDescriptor) {
-        const maxParams = context.options[0] as number;
+        const maxParams = (context.options as Options)[0].maximumFunctionParameters;
```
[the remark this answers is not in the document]
ericmorand-sonarsource: Good catch, I'll use destructuring.

packages/jsts/src/rules/S2004/rule.ts
```diff
@@ -25,21 +25,36 @@ import { TSESTree } from '@typescript-eslint/utils';
 import { getMainFunctionTokenLocation } from 'eslint-plugin-sonarjs/lib/utils/locations';
 import { SONAR_RUNTIME } from '../../linter/parameters';
 import { RuleContext, toEncodedMessage } from '../helpers';
+import type { RuleModule } from '../../../../shared/src/types/rule';
 
 const DEFAULT_THRESHOLD = 4;
```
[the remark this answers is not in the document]
ericmorand-sonarsource: I definitely agree.

sonar-plugin/javascript-checks/src/main/java/org/sonar/javascript/checks/HardcodedCredentialsCheck.java
ericmorand-sonarsource: this :smile:

PR #4682, opened 2024-05-06
Change trigger type of perf task to `manual`

saberduck:
Unfortunately current infra doesn't allow to have stable enough results for meaningful perf regression testing. To avoid having alarm on the wallboard I will "disable" the task by setting it to manual execution only.


zglicz (approved):
noidea

PR #4683, opened 2024-05-06
Resolve JS-151 - Add explicit `meta.schema` to the configurable rules that are missing it 

ericmorand-sonarsource:
Fixes JS-151


sonar-plugin/javascript-checks/src/main/java/org/sonar/javascript/checks/ContentLengthCheck.java
```diff
@@ -19,7 +19,6 @@
  */
 package org.sonar.javascript.checks;
 
-import java.util.Arrays;
```
ericmorand-sonarsource, line 22: I took the opportunity to remove unused imports so that the quality gate succeeds.

zglicz (approved):

PR #4696, opened 2024-05-20
Use Cirrus cache for Orchestrator

saberduck:
To avoid traffic to repox when dowloading SQ for ITs we will cache the dir where orchestrator keeps downloads


ilia-kebets-sonarsource (approved):
I see it's similar to what other teams have done like:
- https://github.com/SonarSource/sonar-enterprise/pull/10684/files

Nice touch with the `reupload_on_changes`

PR #4700, opened 2024-05-22
JS-157 Generate protobuf definition of ESTree

ilia-kebets-sonarsource:


tools/protobuf/output/ast.proto
```diff
@@ -0,0 +1,594 @@
+syntax = "proto3";
+// Generated for @types/estree version: 1.0.5
+
+message Program {
+  string sourceType = 1;
+  repeated BaseNodeWithoutComments body = 2;
+}
+message ModuleDeclaration {
+  oneof moduleDeclaration {
+    ImportDeclaration moduleDeclaration_importDeclaration = 3;
```
saberduck, line 10: afaik, field numbers are per message, it would be more logical to start each message with 1

quentin-jaquier-sonarsource: You are right, not sure why I was convinced it was supposed to be unique overall.

ilia-kebets-sonarsource: done

saberduck (approved):

PR #4706, opened 2024-05-24
JS-163 New Java API to process files during the analysis

saberduck:
Fixes JS-163


its/plugin/plugins/consumer-plugin/src/main/java/org/sonar/samples/javascript/consumer/Consumer.java
```diff
@@ -0,0 +1,55 @@
...
+  private boolean done;
+
+  @Override
+  public void accept(JsFile jsFile) {
+    LOG.info("Accepted file: {}", jsFile.inputFile());
+    jsFiles.add(jsFile);
+  }
+
+  @Override
+  public void doneAnalysis() {
```
ilia-kebets-sonarsource, line 43: Maybe rename this `finishAnalysis()` because it might be confusing with `isDone()`?

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/analysis/AbstractAnalysis.java
```diff
@@ -65,17 +72,71 @@ protected static String inputFileLanguage(InputFile file) {
...
+    this.consumers = consumers;
   }
 
   protected boolean isJavaScript(InputFile file) {
     return inputFileLanguage(file).equals(JavaScriptLanguage.KEY);
   }
 
   abstract void analyzeFiles(List<InputFile> inputFiles, List<String> tsConfigs) throws IOException;
+
+  protected void analyzeFile(InputFile file, @Nullable List<String> tsConfigs, @Nullable TsProgram tsProgram) throws IOException {
```
ilia-kebets-sonarsource, line 90: nice job of refactoring duplicate code into this!

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/analysis/AnalysisConsumers.java
```diff
@@ -0,0 +1,57 @@
...
+
+  @Override
+  public void accept(JsFile jsFile) {
+    consumers.forEach(c -> c.accept(jsFile));
+  }
+
+  @Override
+  public void doneAnalysis() {
+    consumers.forEach(JsAnalysisConsumer::doneAnalysis);
+  }
```
ilia-kebets-sonarsource, line 56: `doneAnalysis()` is actually never used for the consumerS, only for each consumer individually?

its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/ConsumerPluginTest.java
ilia-kebets-sonarsource, line 1: So this IT runs both the JS and Consumer plugins and validates that the consumer plugin runs `execute()` if and only if once the JS plugin is done processing all the files? Is that correct?

ilia-kebets-sonarsource (approved):
A few comments and remarks, otherwise, LGTM!

its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/ConsumerPluginTest.java
saberduck: yes, it is asserting that `execute` is invoked only after the analysis. The order of execution of sensors is implemented with `DependsUpon/@DependedUpon` annotation

PR #4714, opened 2024-05-29
JS-174 Generate Java classes for deserialization of ESTree

saberduck:
Fixes #JS-174


tools/protobuf/generate-proto-file.ts
```diff
@@ -350,7 +350,10 @@ function addHandWrittenMessages(messages: Record<string, ProtobufMessage>) {
 
 function writeMessagesToFile(messages: Record<string, ProtobufMessage>) {
   fs.mkdirSync('output', { recursive: true });
-  fs.writeFileSync(path.join('output', 'ast.proto'), addPrefix(translateToProtoFormat(messages)));
+  fs.writeFileSync(
+    path.join('output', 'estree.proto'),
+    addPrefix(translateToProtoFormat(messages)),
+  );
```
ilia-kebets-sonarsource, line 356: Why not write the file to `sonar-plugin/bridge/src/main/protobuf/estree.proto` here as well?

saberduck: hmm, I don't want the tool to touch directories outside of itself; this is OK, IMO. Anyway, it will be used manually.

ilia-kebets-sonarsource (approved):
LGTM

PR #4722, opened 2024-06-03
Implement the concept of Environment Record

ericmorand-sonarsource:
Fixes JS-175


packages/jsts/src/dbd/js-to-dbd/core/expressions/literal.ts
```diff
@@ -1,21 +1,45 @@
...
+  }
+
+  let value = valueByConstantTypeRegistry.get(typeof constant.value);
+
+  if (!value) {
+    value = createValue(typeof constant.value, context.scopeManager.createValueIdentifier());
+
+    valueByConstantTypeRegistry.set(typeof constant.value, value);
+
+    instructions.push(
```
zglicz, line 33: This should be pushed immediately to the current block

zglicz (approved):
Once I actually started having a grasp on the code, you go and re-write it all.

I left 1 comment, as I think you missed a direct push instead of adding to an instructions array.

packages/jsts/src/dbd/js-to-dbd/core/expressions/literal.ts
ericmorand-sonarsource: Nice catch!

ericmorand-sonarsource:
> Once I actually started having a grasp on the code, you go and re-write it all.

That's a fair point, and I'm very sorry about that. When trying to make the first sample pass, I ended up writing so many conditionals that I had the feeling something was wrong. I really hope that this is the last major change that we have to make. Please accept my apologies, we should have made those changes together.

PR #4723, opened 2024-06-04
JS-158 Prepare generator to produce Java classes

saberduck:
Generator is split into reading d.ts and producing proto/java output. I renamed the model to be protobuf agnostic


ilia-kebets-sonarsource (approved):
LGTM

PR #4724, opened 2024-06-04
JS-164 Improve S5122: Detect origin reflection

yassin-kammoun-sonarsource:
The rule was reimplemented from scratch.
It now leverages helper functions implemented in the meantime.
It makes the rule smarter, more reliable, and less noisy.

I have checked locally and confirmed that the original vulnerability reported by the R&D Team is now being detected.

**Please review also the RSPEC update: https://github.com/SonarSource/rspec/pull/3963**


ilia-kebets-sonarsource (approved):
LGTM. Clean job with nice patterns! Early returns FTW!

PR #4725, opened 2024-06-04
JS-158 Geneate Java API for ESTree

saberduck:


sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/estree/ESTree.java
```diff
@@ -0,0 +1,129 @@
...
+
+public class ESTree {
+
+  private ESTree() {
+    // shouldn't be instantiated
+  }
+  
+  sealed interface Node {
+    
+    String type();
```
andrey-tyukin-sonarsource, line 33: Note that `type` is a reserved keyword in Scala. The API would be a bit easier to use if `type` would be renamed to `tpe` or `typ` (both are somewhat standard abbreviations, `tpe` being a bit more standard, I feel). 

If renaming it is problematic for some other reasons, than leaving it as `type` is also not tragic, in Scala one can escape all identifiers with backticks, so `type()` would still be usable, just a bit more awkward.

sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/estree/ESTree.java
```diff
@@ -0,0 +1,129 @@
...
+        
+  public record Program(String type, Location loc, String sourceType, List<Node> body) implements Node {}
+  public record ModuleDeclaration(String type, Location loc, Node moduleDeclaration) implements Node {}
+  public record ExportAllDeclaration(String type, Location loc, Identifier exported, Literal source) implements Node {}
+  public record Literal(String type, Location loc, Node literal) implements Node {}
+  public record BigIntLiteral(String type, Location loc, int value, String bigint, String raw) implements Node {}
+  public record SimpleLiteral(String type, Location loc, Node value, String raw) implements Node {}
+  public record Identifier(String type, Location loc, String name) implements Node {}
+  public record ExportDefaultDeclaration(String type, Location loc, Node declaration) implements Node {}
+  public record Expression(String type, Location loc, Node expression) implements Node {}
```
andrey-tyukin-sonarsource, line 48: The spec says something like

```
interface Expression <: Node { }
```

so it seemed that it might translate into `sealed interface Expression extends Node`, and then be used as 

```
record ArrayExpression(....) implements Expression
```
etc. So, I _personally_ would find it less surprising if it were an interface, not a `record`. But maybe I'm missing something, not sure.

sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/estree/ESTree.java
```diff
@@ -0,0 +1,129 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+package org.sonar.plugins.javascript.api.estree;
+
+import java.util.List;
+
+public class ESTree {
+
+  private ESTree() {
```
andrey-tyukin-sonarsource, line 27: I'm not sure why everything is wrapped in this class. It seems that the `estree` package already does the same, why another nesting level?

sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/estree/ESTree.java
saberduck: I am actually considering dropping this field entirely because it merely duplicates the class. wdyt?

sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/estree/ESTree.java
saberduck: you are right, this is a mistake, it should be an interface

sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/estree/ESTree.java
saberduck: I like that you can have single import for the whole API and then just use `ESTree.Program` . Java doesn't allow to use package name like this (afaik)

sonar-plugin/bridge/src/main/java/org/sonar/plugins/javascript/bridge/ESTreeFactory.java
```diff
@@ -0,0 +1,35 @@
+package org.sonar.plugins.javascript.bridge;
+
+import java.util.List;
+import java.util.stream.Collectors;
+import org.sonar.plugins.javascript.api.estree.ESTree;
+import org.sonar.plugins.javascript.bridge.protobuf.Node;
+import org.sonar.plugins.javascript.bridge.protobuf.SourceLocation;
+
+public class ESTreeFactory {
```
saberduck, line 9: this is preliminary. It will be part of JS-160

sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/estree/ESTree.java
andrey-tyukin-sonarsource: Ah, yes, it's just the marker that's needed only in JS for JSON object, because JSON objects don't have a `prototype` on which one could perform an `instanceof`.

sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/estree/ESTree.java
```diff
@@ -0,0 +1,160 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+package org.sonar.plugins.javascript.api.estree;
+
+import java.util.List;
+
+/**
+  This file is generated. Do not modify it manually. Look at tools/estree instead.
```
quentin-jaquier-sonarsource, line 26: If I'm not wrong, this file was actually manually modified at line 96. At least, I have this difference when I generate it again. Is it expected?

sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/estree/ESTree.java
andrey-tyukin-sonarsource: You might be right: https://stackoverflow.com/a/10274528/2707792 .

quentin-jaquier-sonarsource (approved):
Except for a small manual modification in the generated file, it LGTM.

sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/estree/ESTree.java
```diff
@@ -0,0 +1,160 @@
...
+  
+  public sealed interface Node {    
+    Location loc();
+  }
+  
+  public record Position(int line, int column) {}
+  public record Location(Position start, Position end) {}
+  
+  public sealed interface CallExpression extends Node {
+    Node callee();
```
andrey-tyukin-sonarsource, line 44: I don't understand why it's a `Node`. The [spec says](https://github.com/estree/estree/blob/master/es5.md#callexpression) that callee and arguments are all `Expression`s. 

This would result in a million unnecessary type casts.

sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/estree/ESTree.java
saberduck: for me they are identical, what's the diff?

sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/estree/ESTree.java
```diff
@@ -0,0 +1,160 @@
...
+
+  }
+  public sealed interface ExportDefaultDeclaration extends Node {
+
+  }
+  public sealed interface Expression extends Node {
+
+  }
+  public sealed interface HasBody extends Node {
+    Node body();
```
andrey-tyukin-sonarsource, line 60: Feels surprising to me. Why is this interface necessary? What does it allow you to do except `instanceof`-ing the implementing class, then typecasting it, and then extracting the more precise type of the `body`?

sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/estree/ESTree.java
saberduck: found it, https://github.com/estree/estree/blob/master/es2015.md#expressions

sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/estree/ESTree.java
```diff
@@ -0,0 +1,160 @@
...
+  public sealed interface Pattern extends Node {
+
+  }
+  public sealed interface Statement extends Node {
+
+  }
+        
+  public record ArrayExpression(Location loc, List<Node> elements) implements Expression {}
+  public record ArrayPattern(Location loc, List<Pattern> elements) implements Pattern {}
+  public record ArrowFunctionExpression(Location loc, boolean expression, Node body, List<Pattern> params, boolean generator, boolean async) implements Expression, HasBody {}
```
andrey-tyukin-sonarsource, line 80: This already begins to look like a problem with `HasBody`: the more specific type of `body` is not `Node` here, it's `Either<FunctionBody, Expression>` [according to the spec](https://github.com/estree/estree/blob/master/es2015.md#arrowfunctionexpression)

sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/estree/ESTree.java
```diff
@@ -0,0 +1,160 @@
...
+  public record AssignmentExpression(Location loc, AssignmentOperator operator, Node left, Expression right) implements Expression {}
+  public record AssignmentOperator(Location loc, String assignmentOperator) implements Node {}
```
andrey-tyukin-sonarsource, line 82: I'm wondering whether this could have been an

```
public enum AssignmentOperator {
  Eq("="),
  PlusEq("+=")
  // ...
  ;
  
  public final String raw;
  private AssignmentOperator(String raw) {
    this.raw = raw;
  }
}
```
?

sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/estree/ESTree.java
```diff
@@ -0,0 +1,160 @@
...
+  public record AssignmentPattern(Location loc, Pattern left, Expression right) implements Pattern {}
+  public record AssignmentProperty(Location loc, Pattern value, String kind, boolean method, Node key, boolean shorthand, boolean computed) implements Node {}
```
andrey-tyukin-sonarsource, line 84: The `key`s of `Property`s are `Expression`s, as defined [here under "Property"](https://github.com/estree/estree/blob/master/es2015.md#expressions). Why is this `Node` and not `Expression`?

sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/estree/ESTree.java
```diff
@@ -0,0 +1,160 @@
...
+  public record AwaitExpression(Location loc, Expression argument) implements Expression {}
+  public record BigIntLiteral(Location loc, int value, String bigint, String raw) implements Literal {}
```
andrey-tyukin-sonarsource, line 86: `int value` looks suspicious. The only reliable way to transfer values of `BigInt`s between languages is by using `String bigint` anyway, so why do we need the `value`?

`value` is part of the _evaluation semantics_, it doesn't belong into syntax.

sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/estree/ESTree.java
```diff
@@ -0,0 +1,160 @@
...
+  public record BinaryExpression(Location loc, BinaryOperator operator, Expression left, Expression right) implements Expression {}
+  public record BinaryOperator(Location loc, String binaryOperator) implements Node {}
```
andrey-tyukin-sonarsource, line 88: Same as for `AssignmentOperator`: wouldn't this be better off as an `enum`?

sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/estree/ESTree.java
```diff
@@ -0,0 +1,160 @@
...
+  public record BlockStatement(Location loc, List<Statement> body) implements Statement {}
+  public record BreakStatement(Location loc, Identifier label) implements Statement {}
+  public record CatchClause(Location loc, Pattern param, BlockStatement body) implements Node, HasBody {}
+  public record ChainExpression(Location loc, ChainElement expression) implements Expression {}
+  public record ClassBody(Location loc, List<Node> body) implements Node {}
```
andrey-tyukin-sonarsource, line 93: I guess it was at some point extended with property declarations as well, but shouldn't this be just `MethodDefinition`? Or `Either<PropertyDeclaration, MethodDefinition>` or something similar?

sonar-plugin/bridge/src/main/protobuf/estree.proto
```diff
@@ -591,4 +591,4 @@ message SourceLocation {
 message Position {
```
andrey-tyukin-sonarsource, line 591: Having a comment somewhere whether it's all `0`-based or `1`-based certainly wouldn't hurt.

tools/estree/output/ESTree.java
```diff
@@ -0,0 +1,160 @@
+/*
```
andrey-tyukin-sonarsource, line 1: This entire file seems duplicated. Why do we need two of them?

andrey-tyukin-sonarsource (commented):
Scrolled through it, added a few comments here and there.

Some types of some child nodes seem unnecessarily lax.

sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/estree/ESTree.java
andrey-tyukin-sonarsource: Ah, sorry, yeah, `estree/index.d.ts` is the ultimate source of truth here.

sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/estree/ESTree.java
andrey-tyukin-sonarsource: Ah, found a related comment:

```
// adding some hardcoded interfaces which allow to treat loops in the AST consistently
```

Still not sure what it means, to be honest.

tools/estree/output/ESTree.java
saberduck: we don't, I will remove this

sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/estree/ESTree.java
saberduck: I was thinking that it would allow to treat loops with single type, eg. declare method like `handleLoop(HasBody node)`

saberduck: I removed `HasBody` and `HasTest`, it seems they wouldn't be very useful

sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/estree/ESTree.java
quentin-jaquier-sonarsource: My bad, it was a mismatch on my side; all good.

sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/estree/ESTree.java
saberduck: I converted operators to enums, although it was done manually, not automatically

PR #4733, opened 2024-06-12
JS-166 The AST should be serialized and sent to the plugin only when at least one consumer is listening

ilia-kebets-sonarsource:


packages/jsts/src/analysis/analyzer.ts
```diff
@@ -86,12 +86,18 @@ function analyzeFile(
...
-    return {
+    const result: JsTsAnalysisOutput = {
       issues,
       ucfgPaths,
       ...extendedMetrics,
-      ast: serializeAst(sourceCode, filePath),
     };
+
+    const ast = serializeAst(sourceCode, filePath, input.skipAst);
+    if (ast) {
```
quentin-jaquier-sonarsource, line 96: We can probably avoid the extra argument, and check the value of `input.skipAst` inside the if condition instead.

sonar-plugin/bridge/src/main/java/org/sonar/plugins/javascript/bridge/BridgeServer.java
```diff
@@ -19,6 +19,8 @@
  */
 package org.sonar.plugins.javascript.bridge;
 
+import static org.sonarsource.api.sonarlint.SonarLintSide.INSTANCE;
+
```
quentin-jaquier-sonarsource, line 23: Is this a specific setup for your repo? It does not follow the layout defined in the [common toolbox](https://github.com/SonarSource/sonar-developer-toolset):

1. import all other imports
2. "blank line"
3. import static all other imports

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/analysis/AnalysisConsumers.java
```diff
@@ -54,4 +54,8 @@ public void accept(JsFile jsFile) {
   public void doneAnalysis() {
     consumers.forEach(JsAnalysisConsumer::doneAnalysis);
   }
+
+  public int getSize() {
+    return consumers.size();
+  }
```
quentin-jaquier-sonarsource, line 60: We can replace this with "HasConsumer" and the body with `!isEmpty`.

sonar-plugin/bridge/src/main/java/org/sonar/plugins/javascript/bridge/BridgeServer.java
ilia-kebets-sonarsource: My intellij must be misconfigured.

quentin-jaquier-sonarsource (approved):
LGTM!

PR #4734, opened 2024-06-13
JS-155 SonarJS exposes entry point to the parser for testing

quentin-jaquier-sonarsource:


sonar-plugin/bridge/src/main/java/org/sonar/plugins/javascript/bridge/BridgeServerConfig.java
```diff
@@ -0,0 +1,37 @@
...
+import org.sonar.api.batch.sensor.SensorContext;
+import org.sonar.api.config.Configuration;
+
+/**
+ * {@link BridgeServerImpl} requires information from {@link org.sonar.api.batch.sensor.SensorContext}.
+ * Hovewer, {@link org.sonar.api.batch.sensor.SensorContext} is a big object, containing more than what we need.
+ * This class will contain only information required by {@link BridgeServerImpl}.
+ * This will reduce the dependency on external API, and ease the testing.
+ */
+public record BridgeServerConfig(Configuration config, String workDirAbsolutePath, SonarProduct product) {
```
quentin-jaquier-sonarsource, line 32: It is a first step toward removing the dependency over `sonar.api` in the Bridge. I failed to remove everything, as we rely deeply on the `Configuration` object. I tried to wrap the object, store only the result that we need, but both solution result in a lot of non-trivial changes.

It is a bit annoying, as it means you will need to have the classes from `sonar.api` when using the standalone... But I'm afraid we will have to deal with it for now.

quentin-jaquier-sonarsource:
I still miss a bit of coverage. I'm really surprised to see the lines in `StandaloneTemporaryFolder`, as I have an explicit test for it. It looks like a bug from the code coverage tool.
The rest are exceptional cases, that are cumbersome to test in a meaningful way. I would be in favor of not bothering covering those. Let me know what you think.

sonar-plugin/standalone/src/test/java/org/sonar/plugins/javascript/standalone/StandaloneParserTest.java
```diff
@@ -0,0 +1,77 @@
...
+      });
+    });
+    Program secondSample = parser.parse("let x;");
+    assertThat(secondSample.body()).hasSize(1);
+  }
+
+  @Test
+  void should_throw_exception_when_fail_to_parse_code() {
+    assertThatThrownBy(() -> parser.parse("..."))
+      .isInstanceOf(IllegalStateException.class)
```
ilia-kebets-sonarsource, line 65: maybe throw [ParseException](https://docs.oracle.com/javase/8/docs/api/java/text/ParseException.html) instead?

quentin-jaquier-sonarsource: After thinking about it, this exception takes an offset, I'm not sure it makes sense in our context. It is probably meant to be used when parsing Java Objects.

PR #4738, opened 2024-06-14
Do not serialize the AST if the `skipAst` flag is set

ilia-kebets-sonarsource:
With this change, we skip the AST serialization, not just the sending of the AST in the response.

### DIFF
--- a/packages/jsts/src/analysis/analyzer.ts
+++ b/packages/jsts/src/analysis/analyzer.ts
@@ -92,9 +92,11 @@ function analyzeFile(
       ...extendedMetrics,
     };
 
-    const ast = serializeAst(sourceCode, filePath, input.skipAst);
-    if (ast) {
-      result.ast = ast;
+    if (!input.skipAst) {
+      const ast = serializeAst(sourceCode, filePath);
+      if (ast) {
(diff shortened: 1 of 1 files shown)


packages/jsts/src/analysis/analyzer.ts
```diff
@@ -92,8 +92,8 @@ function analyzeFile(
       ...extendedMetrics,
     };
 
-    const ast = serializeAst(sourceCode, filePath);
-    if (!input.skipAst && ast) {
+    const ast = serializeAst(sourceCode, filePath, input.skipAst);
+    if (ast) {
```
quentin-jaquier-sonarsource, line 96: Oups, that was my initial suggestion that led to this imprecision...
Nice catch anyway.
I still think it feels strange to pass this argument. What about:
```
if (!input.skipAst) {
  const ast = serializeAst(sourceCode, filePath);
  if (ast) {
    result.ast = ast;
  }
}
```

ilia-kebets-sonarsource: Both are not great. I'll add your suggestion.

quentin-jaquier-sonarsource (approved):

PR #4741, opened 2024-06-18
JS-193 Add missing `body` property to static block

ilia-kebets-sonarsource:


ilia-kebets-sonarsource:
Do you think we should address this issue and reuse the visitBlockStatement() function  (probably renaming it to something like visitBlock()) as the issue suggests or accept the duplication?

ilia-kebets-sonarsource:
@quentin-jaquier-sonarsource Do you think we should address [this issue](https://next.sonarqube.com/sonarqube/project/issues?pullRequest=4741&issueStatuses=OPEN%2CCONFIRMED&open=b75b5fa6-75dc-477a-893f-bdd3880d11cf&id=org.sonarsource.javascript%3Ajavascript) and reuse the `visitBlockStatement()` function  (probably renaming it to something like `visitBlock()`) as the issue suggests or accept the duplication?

quentin-jaquier-sonarsource (approved):

quentin-jaquier-sonarsource:
I personally think this file is an exception to the rule, I don't think it makes sense to address the issue. The motivation is that I would prefer to have an exhaustive list of methods, rather than merging two methods. We already apply this logic for other methods.

PR #4742, opened 2024-06-18
JS-188 Directive can be used as a Statement

quentin-jaquier-sonarsource:


sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/estree/ESTree.java
```diff
@@ -104,7 +104,7 @@ public record ClassExpression(Location loc, Optional<Identifier> id, Optional<Ex
   public record ConditionalExpression(Location loc, Expression test, Expression alternate, Expression consequent) implements Expression {}
   public record ContinueStatement(Location loc, Optional<Identifier> label) implements Statement {}
   public record DebuggerStatement(Location loc) implements Statement {}
-  public record Directive(Location loc, Literal expression, String directive) implements DirectiveOrModuleDeclarationOrStatement {}
```
ilia-kebets-sonarsource, line 107: Should we rename the top interface `ModuleDeclarationOrStatement` since directive is a child of Statement?

quentin-jaquier-sonarsource: This interface was introduced as the type of Program is:
```
export interface Program extends BaseNode {
    // ...
    body: Array<Directive | Statement | ModuleDeclaration>;
}
```
On one hand, we might want to match this definition, on the other, you are right that it is redundant. No strong feeling on this one, do you think that it is worth it?

quentin-jaquier-sonarsource: As discussed, let's keep it as it is, and come back at it if there is a need.

ilia-kebets-sonarsource (approved):
LGTM

PR #4753, opened 2024-07-12
JS-230 Java ESTree AST should correctly build array expressions and patterns with empty elements

quentin-jaquier-sonarsource:


ilia-kebets-sonarsource (approved):
LGTM!

PR #4756, opened 2024-07-17
Use new docs URL

vdiez:


.gitignore
```diff
@@ -45,4 +45,5 @@ Desktop.ini
 .scannerwork
 
 # eslint-plugin-sonarjs
-eslint-plugin-sonarjs-*.tgz
\ No newline at end of file
+eslint-plugin-sonarjs-*.tgz
+its/eslint-plugin-sonarjs/package-lock.json
```
yassin-kammoun-sonarsource, line 49: Any reason why you want to gitignore this lock file?

packages/jsts/tests/rules/index.test.ts
```diff
@@ -44,7 +44,7 @@ describe('index', () => {
       }
       expect(configs.recommended.plugins!['sonarjs'].rules).toHaveProperty(eslintId);
       expect(rule.meta.docs!.url).toBe(
-        `https://github.com/SonarSource/rspec/blob/master/rules/${sonarKey}/javascript/rule.adoc`,
+        `https://sonarsource.github.io/rspec/#/rspec/${sonarKey}/javascript`,
```
yassin-kammoun-sonarsource, line 47: Should we also update the `newRule` script to automatically include the URL to the rule rspec?

yassin-kammoun-sonarsource (approved):
Looks good to me, but I left two questions for you before merging :)

.gitignore
vdiez: it was failing to install the tarball of the integration test during ci with the lock file. i don't know the reason, but without the lock file it works

packages/jsts/tests/rules/index.test.ts
vdiez: it's already doing it, right?

.gitignore
yassin-kammoun-sonarsource: 👍

packages/jsts/tests/rules/index.test.ts
yassin-kammoun-sonarsource: Yes, sorry, I missed it 👍

PR #4763, opened 2024-08-12
Register babel presets and plugin from their actual modules

ericmorand-sonarsource:
Testing the pattern that consists at importing the actual Babel modules instead of referencing their packages by their name, as supported by Babel:

https://babeljs.io/docs/options#plugin-and-preset-options
https://babeljs.io/docs/options#entrytarget

It allows dependency resolvers to organically emit the Babel plugins and parsers to the bundler, instead of having to register them explicitely.

Practically, when using esbuild and the package manifest plugin, it allows us to replace the following esbuild configuration:

```js
{
      entryPoints: [
        {
          in: './worker.js',
          out: 'worker'
        },
        {
          in: './main.ts',
          out: 'index'
        }
      ],
      bundle: true,
      outdir: 'target',
      format: "cjs",
      minify: true,
      platform: "node",
      packages: "external",
      plugins: [
        createPackageManifestPlugin({
          dependencies: {
            '@babel/core': '7.24.3',
            '@babel/preset-flow': '7.24.1',
            '@babel/preset-env': '7.24.3',
            '@babel/plugin-proposal-decorators': '7.24.1'
          },
          bundledDependencies: [
            '@babel/core',
            '@babel/preset-flow',
            '@babel/preset-env',
            '@babel/plugin-proposal-decorators'
          ]
        }, '../../../package.json', {
          shouldBundleDependencies: true
        })
      ]
    })
  ]);
}
```

...with this shorter and more maintainable one:

```js
{
      entryPoints: [
        {
          in: './worker.js',
          out: 'worker'
        },
        {
          in: './main.ts',
          out: 'index'
        }
      ],
      bundle: true,
      outdir: 'target',
      format: "cjs",
      minify: true,
      platform: "node",
      packages: "external",
      plugins: [
        createPackageManifestPlugin({}, '../../../package.json', {
          shouldBundleDependencies: true
        })
      ]
    })
  ]);
}
```

### DIFF
--- a/packages/jsts/src/parsers/options.ts
+++ b/packages/jsts/src/parsers/options.ts
@@ -18,6 +18,10 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 import { Linter } from 'eslint';
+const babelReactPreset = require('@babel/preset-react').default;
+const babelFlowPreset = require('@babel/preset-flow').default;
+const babelEnvPreset = require('@babel/preset-env').default;
+const babelDecoratorProposalPlugin = require('@babel/plugin-proposal-decorators').default;
 
 /**
  * Builds ESLint parser options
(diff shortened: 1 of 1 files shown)


packages/jsts/src/parsers/options.ts
```diff
@@ -18,6 +18,10 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 import { Linter } from 'eslint';
+const babelReactPreset = require('@babel/preset-react').default;
```
saberduck, line 21: could we use `import` instead of `require`? I am not sure if it's technically possible

ericmorand-sonarsource: It is, but since there is no available typing for those packages, it would probably involve adding some ts-ignore things or maintaining some declarations ourselves. I can try.

ericmorand-sonarsource: This pass the tests and emit the dependencies to the bundler, so it's technically all good:

```ts
// @ts-ignore
import {default as babelReactPreset} from "@babel/preset-react";
// @ts-ignore
import {default as babelFlowPreset} from "@babel/preset-flow";
// @ts-ignore
import {default as babelEnvPreset} from "@babel/preset-env";
// @ts-ignore
import {default as babelDecoratorProposalPlugin} from "@babel/plugin-proposal-decorators";
```

The `ts-ignore` directives are not looking great (`@babel/preset-env` got typings available, note) but if we can live with them or add some local declarations, we are all good.

ilia-kebets-sonarsource: @ericmorand-sonarsource If I'm not mistaken, we can add partial typings for our dependencies in [this folder](https://github.com/SonarSource/SonarJS/tree/master/typings).

ilia-kebets-sonarsource (changes requested):
Add typings and use `import` as discussed https://github.com/SonarSource/SonarJS/pull/4763/files#r1714107832
I'm open to discuss if you would like to consider another solution.

ilia-kebets-sonarsource (approved):
you can use require instead.

PR #4764, opened 2024-08-14
SONARARMOR-98 SonarArmor JS/TS rules should be in Sonar Way profile

quentin-jaquier-sonarsource:
This PR only makes sense with https://github.com/SonarSource/sonar-armor/pull/176, and should be merged first.

### DIFF
--- a/sonar-plugin/sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/JavaScriptProfilesDefinitionTest.java
+++ b/sonar-plugin/sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/JavaScriptProfilesDefinitionTest.java
@@ -19,12 +19,6 @@
  */
 package org.sonar.plugins.javascript;
 
-import static org.assertj.core.api.Assertions.assertThat;
-import static org.sonar.plugins.javascript.JavaScriptProfilesDefinition.SECURITY_RULES_CLASS_NAME;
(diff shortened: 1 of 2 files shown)


sonar-plugin/sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/JavaScriptProfilesDefinitionTest.java
```diff
@@ -140,26 +140,26 @@ void no_legacy_Key_in_profile_json() {
   @Test
   void should_contains_security_rules_if_available() {
```
renaud-tognelli-sonarsource, line 141: Could it make sense to perform the same test with `SONAR_ARMOR_RULES_CLASS_NAME` and `(ts|js)security-armor` ?

renaud-tognelli-sonarsource (approved):
LGTM
I have one question regarding tests.

sonar-plugin/sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/JavaScriptProfilesDefinitionTest.java
quentin-jaquier-sonarsource: This test unit tests the behavior of the method `getSecurityRuleKeys` by faking the presence of `com.sonar.plugins.security.api.JsRules` in the classpath (the package is redefined in the tests package...). It does not really test what is happening in a real use case, testing the same for armor would be redundant.
That being said, I agree that we are not really testing the real behavior in this PR; the proper test will be done in Armor.

PR #4769, opened 2024-08-20
Resolves JS-278

ericmorand-sonarsource:
Fixes JS-278

Please read the commit messages for more information.

Also, this may seem like a big change according to the number of files changed, but around 440 of these files are Java Check subclasses where the `eslintKey` public getter was removed. You can safely ignore them and focus on the 20 or so relevant changes that this PR brings.


packages/jsts/tests/linter/wrapper.test.ts
```diff
@@ -352,7 +352,9 @@ describe('LinterWrapper', () => {
       }
 
       if (!fixture) {
-        throw new Error(`Failed to find fixture file for rule '${ruleId}' in '${fixtures}'.`);
+        // todo: discuss with the team if we can not throw on missing fixture
```
vdiez, line 355: we didn't discuss this. can we have it thowing again? what happens if we do?

packages/jsts/tests/rules/index.test.ts
```diff
@@ -52,7 +52,8 @@ describe('index', () => {
       }
     }
     expect(missing).toHaveLength(0);
-    expect(Object.keys(rules)).toHaveLength(mappedRules.size);
+    // todo: this test is fragile and does not guarantee anything relevant: that a certain number of symbols are exported does not mean that the expected symbols are exported; we need to implement a relevant test
+    //expect(Object.keys(rules)).toHaveLength(mappedRules.size);
```
vdiez, line 56: agreed, feel free to remove completely then

sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/Check.java
```diff
@@ -0,0 +1,13 @@
+package org.sonar.plugins.javascript.api;
+
+import org.sonar.check.Rule;
+
+public class Check implements EslintBasedCheck {
```
vdiez, line 5: I wonder if we can get rid completely of EslintBasedCheck, instead of hidding it with an intermediate class

vdiez (approved):
LGTM, very nice Eric! thx!

some small comments, from what we discussed you can focus on the quickfixes test and ignore the others.

packages/jsts/tests/linter/wrapper.test.ts
ericmorand-sonarsource: If we do, we must add fixtures named SXXX alongside the ESLint-named ones. Definitely feasible, but time-consuming - we have to copy and paste under another name each available fixture. This added value in quality is debatable in my opinion: if the ESLint-named fixture is successful, then the SXXX alias is covered too.

sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/TestFileCheck.java
```diff
@@ -22,8 +22,10 @@
 import java.util.Collections;
 import java.util.List;
 import org.sonar.api.batch.fs.InputFile;
+import org.sonar.check.Rule;
 
-public abstract class TestFileCheck implements EslintBasedCheck {
+@Rule(key="key")
```
saberduck, line 27: why is this needed? imo it shouldn't be needed on abstract class

sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/Check.java
ericmorand-sonarsource: For posterity, we discussed this by private message and we agree that the ESLintbasedCheck interface can, and should, be removed. There would be some impact on plugins that consume the API of SonarJS, like SonarSecurity, which is why we keep it for now.

sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/TestFileCheck.java
ericmorand-sonarsource: True, I think it is a mistake. Let me remove it and see how it goes.

packages/jsts/tests/rules/index.test.ts
ericmorand-sonarsource: Thanks. At one point, we may want to add a test that checks explicitly what we export as a public API for the ESLint Plugin.

packages/jsts/tests/linter/wrapper.test.ts
vdiez: but you have the mapping between both keys, don't you? I would modify the test to look for the right file no matter the kind of id. if we leave it like this, some valid ids might be ignored. if you want to avoid repetition, just ignore ids `/^S\d+$/`. WDYT?

ericmorand-sonarsource: > but you have the mapping between both keys, don't you? 

Not explicitly - i.e. not in an operable manner. 

>  if you want to avoid repetition, just ignore ids /^S\d+$/.

Yeah, sounds good to me. Will do.

PR #4785, opened 2024-08-28
JS-299 List nodes that need to be supported in TS

ericmorand-sonarsource:
Fixes JS-299

### DIFF
--- a/packages/jsts/src/parsers/ast.ts
+++ b/packages/jsts/src/parsers/ast.ts
@@ -19,8 +19,7 @@
  */
 import * as protobuf from 'protobufjs';
 import * as path from 'node:path';
-import * as estree from 'estree';
-import { AST } from 'eslint';
+import { TSESTree } from '@typescript-eslint/utils';
 import { debug } from '@sonar/shared';
 
 const PATH_TO_PROTOFILE = path.join(__dirname, 'estree.proto');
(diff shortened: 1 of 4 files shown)


packages/jsts/src/parsers/ast.ts
```diff
@@ -636,125 +737,119 @@ export function visitNode(node: estree.BaseNodeWithoutComments | undefined | nul
...
-    } else {
-      return {
-        expression: visitNode(node.expression),
-      };
-    }
+  function visitExpressionStatement(node: TSESTree.ExpressionStatement) {
+    return {
+      expression: visitNode(node.expression),
+      directive: node.directive,
+    };
```
ilia-kebets-sonarsource, line 841: this change will have repercussions

ilia-kebets-sonarsource (changes requested):
Can you adapt the changes to expression statement serializing in the deserialization?

packages/jsts/src/parsers/ast.ts
ericmorand-sonarsource: I'm not sure I understand the issue. If `node.directive` is undefined, `directive` will be undefined and will be serialized as such, just like if it was not present in the seralizable output, isn't it?

ericmorand-sonarsource: I'll revert the change anyway.

ilia-kebets-sonarsource (approved):
LGTM

PR #4786, opened 2024-08-28
JS-280 Support unknown nodes in AST protobuf file

vdiez:
Fixes https://sonarsource.atlassian.net/browse/JS-280


packages/jsts/src/parsers/ast.ts
```diff
@@ -57,7 +57,7 @@ export function visitNode(node: estree.BaseNodeWithoutComments | undefined | nul
   }
 
   return {
-    type: NODE_TYPE_ENUM.values[node.type + 'Type'],
+    type: NODE_TYPE_ENUM.values[node.type + 'Type'] ?? NODE_TYPE_ENUM.values['UnknownType'],
     loc: node.loc,
     [lowerCaseFirstLetter(node.type)]: getProtobufShapeForNode(node),
```
ilia-kebets-sonarsource, line 62: what do you think about storing its type so it could be logged further down the road, for example if we want to debug a behaviour in JaRED, it would provide us more context for debugging.

This would imply to implement some visitUknown() function

packages/jsts/tests/parsers/ast.test.ts
```diff
@@ -62,6 +64,17 @@ describe('ast', () => {
       },
     );
   });
+  test('should encode unknown nodes', async () => {
+    const filePath = path.join(__dirname, 'fixtures', 'ast', 'unknownNode.ts');
+    const sc = await parseSourceCode(filePath, parsers.typescript);
+    const protoMessage = parseInProtobuf(sc.ast);
+    fs.writeFileSync('JS-285-reproducer-ts-proto.json', JSON.stringify(protoMessage, null, 2));
```
ilia-kebets-sonarsource, line 71: You don't need that

packages/jsts/tests/parsers/fixtures/ast/unknownNode.ts
```diff
@@ -0,0 +1,3 @@
+if ("ts" as Object) {
```
ilia-kebets-sonarsource, line 1: we'll have to change this one when we will support this, but I think it's clear what problem we're tackling

packages/jsts/tests/parsers/ast.test.ts
```diff
@@ -62,6 +64,17 @@ describe('ast', () => {
...
+    const serialized = serializeInProtobuf(sc.ast);
+    fs.writeFileSync('JS-285-reproducer-ts.proto', serialized);
```
ilia-kebets-sonarsource, line 73: same here

packages/jsts/src/parsers/estree.proto
```diff
@@ -467,3 +469,7 @@ message FunctionExpression {
   optional bool generator = 4;
   optional bool async = 5;
 }
+
+message UnknownNode {
+  string astNodeType = 1;
```
ilia-kebets-sonarsource, line 474: WDYT?

```suggestion
  string rawType = 1;
```

sonar-plugin/bridge/src/test/java/org/sonar/plugins/javascript/bridge/ESTreeFactoryTest.java
```diff
@@ -57,11 +60,30 @@
...
 import org.sonar.plugins.javascript.bridge.protobuf.WithStatement;
 
-import static org.assertj.core.api.Assertions.assertThat;
-import static org.assertj.core.api.AssertionsForClassTypes.assertThatThrownBy;
-
 class ESTreeFactoryTest {
 
+  @Test
+  void should_not_fail_with_unknown_nodes() throws IOException {
+    // the clear version of serialized.proto is `packages/jsts/tests/parsers/fixtures/ast/JS-285-reproducer.ts`,
```
ilia-kebets-sonarsource, line 67:
```suggestion
    // the clear version of serialized.proto is `packages/jsts/tests/parsers/fixtures/ast/unknownNode.ts`,
```

sonar-plugin/bridge/src/test/java/org/sonar/plugins/javascript/bridge/ESTreeFactoryTest.java
```diff
@@ -57,11 +60,30 @@
...
+    }
+    ESTree.Node root = ESTreeFactory.from(node, ESTree.Node.class);
+    assertThat(root).isInstanceOf(ESTree.Program.class);
+    ESTree.Program program = (ESTree.Program) root;
+    assertThat(program.body()).hasSize(1);
+    assertThat(program.body().get(0)).isInstanceOfSatisfying(ESTree.IfStatement.class, ifStatement -> {
+      assertThat(ifStatement.test()).isInstanceOf(ESTree.UnknownNode.class);
+      ESTree.UnknownNode unknownNode = (ESTree.UnknownNode)ifStatement.test();
+      assertThat(unknownNode.astNodeType()).isEqualTo("TSAsExpression");
+
```
ilia-kebets-sonarsource, line 83: extra newline?

packages/jsts/tests/parsers/ast.test.ts
vdiez: oops, i should not have commited that

packages/jsts/src/parsers/estree.proto
vdiez: btw, why so many repeated proto files? I think one should be source of truth

packages/jsts/tests/parsers/fixtures/ast/unknownNode.ts
vdiez: as long as we don't change the proto file of the test we should be fine

packages/jsts/src/parsers/estree.proto
ilia-kebets-sonarsource: they are repeated because we didn't want to create coupling between the different modules. If we can find a better solution, it would be nice. See https://sonarsource.atlassian.net/browse/JS-197

ilia-kebets-sonarsource: As discussed, we will remove the deprecated scripts in [tools/estree](https://github.com/SonarSource/SonarJS/tree/master/tools/estree) and only keep a script similar to [copy-to-lib.js](https://github.com/SonarSource/SonarJS/blob/master/tools/estree/copy-to-lib.js), named `copy-to-plugin.js` that copies the original [in the bridge](https://github.com/SonarSource/SonarJS/blob/master/packages/jsts/src/parsers/estree.proto) to the copy [in the plugin](https://github.com/SonarSource/SonarJS/blob/master/sonar-plugin/bridge/src/main/protobuf/estree.proto).

We should also adapt the header comments in the protofile to identify:
- which one is the original
- which one the copy
- by what script they are maintained in sync

ilia-kebets-sonarsource (approved):
LGTM

PR #4788, opened 2024-08-29
JS-304 Support TSAsExpression nodes

ericmorand-sonarsource:
Fixes JS-304


ilia-kebets-sonarsource (approved):
We implemented it with Victor and reviewed with Eric.

PR #4795, opened 2024-08-30
JS-310 Do not serialize unsupported known nodes

ilia-kebets-sonarsource:


ilia-kebets-sonarsource:
dropped as it works against the fix introduced in https://github.com/SonarSource/SonarJS/pull/4786

PR #4797, opened 2024-09-03
JS-312 Create rule S1607 (`no-skipped-tests`): Tests should not be skipped without providing a reason

yassin-kammoun-sonarsource:
[RSPEC](https://github.com/SonarSource/rspec/pull/4218) to review


packages/jsts/src/rules/S1607/rule.ts
```diff
@@ -0,0 +1,244 @@
...
+        });
+      }
+    }
+
+    /**
+     * Checks if the node denoting a test has an explanation comment on the line above.
+     */
+    function hasExplanationComment(node: estree.Node): boolean {
+      const comments = context.sourceCode.getCommentsBefore(node);
+      return comments.some(comment => comment.loc!.end.line === node.loc!.start.line - 1);
```
zglicz, line 241: Is a single character enough? Should there be a minimal length of explanation?

packages/jsts/src/rules/S1607/rule.ts
```diff
@@ -0,0 +1,244 @@
...
+  create(context) {
+    const dependencies = getDependencies(context.filename);
+    switch (true) {
+      case dependencies.has('jasmine'):
+        return jasmineListener();
+      case dependencies.has('jest'):
+        return jestListener();
+      case dependencies.has('mocha'):
+        return mochaListener();
+      case getPackageJsonsCount() > 0:
```
zglicz, line 55: Why does this imply the nodejsListener?

packages/jsts/src/rules/S1607/rule.ts
```diff
@@ -0,0 +1,244 @@
...
+      default:
+        return {};
+    }
+
+    /**
+     * Returns a rule listener specific to Jasmine.
+     *
+     * Ignoring tests with Jasmine is done by using `xit`, `xdescribe`, or `xcontext`.
+     */
+    function jasmineListener(): Rule.RuleListener {
```
zglicz, line 66: the jasmine, jest and mocha listeners follow the same convention. They only differ on the isIgnoredTest implementation. Could you merge them?

zglicz (changes requested):
Few small comments for you!

awwyiss2

packages/jsts/src/rules/S1607/rule.ts
yassin-kammoun-sonarsource: What would you suggest as a minimal length?

packages/jsts/src/rules/S1607/rule.ts
yassin-kammoun-sonarsource: I assume that a Node.js project includes at least one package.json file so that the Node-specific listener is worth triggering.

packages/jsts/src/rules/S1607/rule.ts
yassin-kammoun-sonarsource: I have mixed feelings about this because of past experiences. While it's true that the implementation is quite similar, I also feel that merging them may not be beneficial once we receive feedback from users and need to introduce framework-specific exceptions.

packages/jsts/src/rules/S1607/rule.ts
zglicz: I wouldn't consider a single word to be a good enough reason. I'd say 20 characters could be fair, so that it can be a semblance of a valid sentence.

packages/jsts/src/rules/S1607/rule.ts
```diff
@@ -0,0 +1,244 @@
...
+
+export const rule: Rule.RuleModule = {
+  meta: generateMeta(meta as Rule.RuleMetaData, {
+    messages: {
+      removeOrExplainTest: 'Remove this unit test or explain why it is ignored.',
+    },
+  }),
+  create(context) {
+    const dependencies = getDependencies(context.filename);
+    switch (true) {
```
zglicz, line 48: oh, that's also a new one for me as well. Interesting.

packages/jsts/src/rules/S1607/rule.ts
zglicz: I've only ever used jest, so I'm not sure how users normally setup the nodejs-tests. It's a foreign approximation for me.

packages/jsts/src/rules/S1607/rule.ts
zglicz: Hm, well ok. I would say that's a premature separation, but I lack the experience here. Just a small fear, that there might also be improvements that will need to land in all 3 places.

zglicz (approved):
Thanks for answering the comments

packages/jsts/src/rules/S1607/rule.ts
yassin-kammoun-sonarsource: That sounds fair, but I wonder how that would work on comments written That sounds fair, but I wonder how that would work with comments written in Asian languages such as Chinese or Japanese, which can be quite compact.

sonar-plugin/javascript-checks/src/main/java/org/sonar/javascript/checks/NoSkippedTestsCheck.java
```diff
@@ -0,0 +1,33 @@
...
+
+import org.sonar.check.Rule;
+import org.sonar.plugins.javascript.api.Check;
+import org.sonar.plugins.javascript.api.JavaScriptRule;
+import org.sonar.plugins.javascript.api.TypeScriptRule;
+
+@JavaScriptRule
+@TypeScriptRule
+@Rule(key = "S1607")
+public class NoSkippedTestsCheck extends Check {
```
ilia-kebets-sonarsource, line 30: Do we need tests for checks of rules without parameters?

packages/jsts/src/rules/S1607/rule.ts
yassin-kammoun-sonarsource: I have been thinking, and I remember seeing quite a few times comments that just mention a Jira ticket number or a GitHub issue number.

sonar-plugin/javascript-checks/src/main/java/org/sonar/javascript/checks/NoSkippedTestsCheck.java
yassin-kammoun-sonarsource: We never did that until now. Did something change recently?

PR #4798, opened 2024-09-03
JS-59: Create rule S7059 (`no-async-constructor`): Constructors should not contain asynchronous operations

vdiez:
Fixes https://sonarsource.atlassian.net/browse/JS-59

RSPEC PR: https://github.com/SonarSource/rspec/pull/4223


packages/jsts/src/rules/S7059/rule.ts
```diff
@@ -0,0 +1,71 @@
...
+import * as estree from 'estree';
+import { meta } from './meta';
+import { AST_NODE_TYPES } from '@typescript-eslint/utils';
+
+const flaggedConstructors = new Set();
+
+export const rule: Rule.RuleModule = {
+  meta: generateMeta(meta as Rule.RuleMetaData, {
+    messages: {
+      noAsyncConstructor: 'Move this asynchronous operation outside of the constructor.',
```
ilia-kebets-sonarsource, line 33:
```suggestion
      noAsyncConstructor: 'Refactor this asynchronous operation outside of the constructor.',
```

packages/jsts/src/rules/S7059/rule.ts
```diff
@@ -0,0 +1,71 @@
...
+    },
+  }),
+  create(context: Rule.RuleContext) {
+    const services = context.sourceCode.parserServices;
+    if (!isRequiredParserServices(services)) {
+      return {};
+    }
+    function isPromiseLike(expr: estree.Expression) {
+      return isRequiredParserServices(services) && isThenable(expr, services);
```
ilia-kebets-sonarsource, line 42: why call `isRequiredParserServices(services)` again? you do it already at line 38:

```ts
if (!isRequiredParserServices(services)) {
      return {};
    }
```

```suggestion
      return isThenable(expr, services);
```

vdiez: good catch!

packages/jsts/src/rules/S7059/rule.ts
```diff
@@ -0,0 +1,71 @@
...
+        if (classConstructor && isPromiseLike(node) && !flaggedConstructors.has(classConstructor)) {
+          flaggedConstructors.add(classConstructor);
+          context.report({
+            node: node as estree.Node,
+            messageId: 'noAsyncConstructor',
+          });
+        }
+      },
+      'Program:exit'() {
+        flaggedConstructors.clear();
```
ilia-kebets-sonarsource, line 67: The `flaggedConstructors` logic is meant to not raise the issue multiple times in a single constructor, is that right?

```js
class MyClass {
  constructor() {
    Promise.resolve().then(() => this.data = fetchData()); //Noncompliant
    // ... lots of lines
    Promise.resolve().then(() => this.data = fetchData()); //compliant
  }
}
```


Why shouldn't we raise multiple problematic promise calls in a single constructors? I believe we should otherwise, a user will fix 1 issue and be caught on the next one on the next analysis.

If it is needed indeed, could you add a test case for it?

its/ruling/src/test/expected/jsts/desktop/typescript-S7059.json
```diff
@@ -0,0 +1,50 @@
...
+423
+],
+"desktop:app/src/lib/stores/accounts-store.ts": [
+63
+],
+"desktop:app/src/lib/stores/app-store.ts": [
+527
+],
+"desktop:app/src/lib/stores/commit-status-store.ts": [
+167
```
ilia-kebets-sonarsource, line 27: this is :

```js
  public constructor(accountsStore: AccountsStore) {
    accountsStore.getAll().then(this.onAccountsUpdated) // noncompliant
    accountsStore.onDidUpdate(this.onAccountsUpdated)
  }
```

Do we still want to raise when the user calls `Promise.then()`?

packages/jsts/src/rules/S7059/rule.ts
vdiez: the problem is that without flaggedContructors the first non-compliant line in your example will raise 3 times, (fetchData, resolve(), then()). So the problem is repetition in a single line.

its/ruling/src/test/expected/jsts/desktop/typescript-S7059.json
vdiez: yes, I think so, the callback in promise.then will be executed after the constructor has returned

packages/jsts/src/rules/S7059/rule.ts
ilia-kebets-sonarsource: You could use something like `flaggedLines()`, although it would fail when splitting on multiple lines like:
```js
Promise.
  resolve()
  .then(
     () => this.data = fetchData()
)
```

could you use something like `flaggedStatements`? or would that raise for `Promise.resolve().then()` and `fetchData()`?

Would there be some other heuristic that would allow to conveniently group nested promise calls?

ilia-kebets-sonarsource: Could we consider tracking `flaggedCallExpressions` and checking the ancestors of flagged nodes?

its/ruling/src/test/expected/jsts/desktop/typescript-S7059.json
```diff
@@ -0,0 +1,50 @@
...
+97
+],
+"desktop:app/src/lib/databases/github-user-database.ts": [
+69
+],
+"desktop:app/src/lib/databases/issues-database.ts": [
+18
+],
+"desktop:app/src/lib/databases/pull-request-database.ts": [
+89
```
ilia-kebets-sonarsource, line 12: here we should have many more issues

packages/jsts/src/rules/S7059/rule.ts
ilia-kebets-sonarsource: As discussed, we will flag and track ExpressionStatements

ilia-kebets-sonarsource (changes requested):
as discussed, simply change the flag to ExpressionStatement

packages/jsts/src/rules/S7059/rule.ts
ilia-kebets-sonarsource: We dismissed the idea of flagging only the constructor as this would require the user to track the async calls himself, which is worse.

its/ruling/src/test/expected/jsts/desktop/typescript-S7059.json
```diff
@@ -21,16 +42,14 @@
...
 ],
 "desktop:app/src/lib/stores/commit-status-store.ts": [
 167
 ],
 "desktop:app/src/lib/stores/pull-request-coordinator.ts": [
-54
-],
-"desktop:app/src/main-process/ordered-webrequest.ts": [
-146
+62
```
ilia-kebets-sonarsource, line 52: these changes look fishy

packages/jsts/src/rules/S7059/rule.ts
```diff
@@ -38,33 +38,46 @@ export const rule: Rule.RuleModule = {
...
-      return context.sourceCode.getAncestors(node).find(node => {
-        return (
+    function asyncStatementInsideConstructor(node: estree.Expression) {
+      let classConstructor: estree.MethodDefinition | undefined;
+      let statement: estree.Statement | undefined;
+      context.sourceCode.getAncestors(node).forEach(node => {
+        if (
           node.type === AST_NODE_TYPES.MethodDefinition &&
           node.key.type === AST_NODE_TYPES.Identifier &&
           node.key.name === 'constructor'
```
ilia-kebets-sonarsource, line 49: what about checking this instead:
```js
node.kind === 'constructor'
```

packages/jsts/src/rules/S7059/rule.ts
```diff
@@ -38,33 +38,46 @@ export const rule: Rule.RuleModule = {
...
+          classConstructor = node;
+        }
+        if (classConstructor && node.type.endsWith('Statement')) {
+          statement = node as estree.Statement;
+        }
+        // If we find a function declaration it should not be considered as part of the constructor
+        if (classConstructor && statement && isFunctionNode(node)) {
+          statement = undefined;
+          classConstructor = undefined;
+        }
```
ilia-kebets-sonarsource, line 60: I don't understand this case

packages/jsts/src/rules/S7059/rule.ts
```diff
@@ -38,33 +38,46 @@ export const rule: Rule.RuleModule = {
...
-        if (classConstructor && isPromiseLike(node) && !flaggedConstructors.has(classConstructor)) {
-          flaggedConstructors.add(classConstructor);
+        if (!isThenable(node, services)) {
+          return;
+        }
+        const statement = asyncStatementInsideConstructor(node);
+        if (statement && !flaggedStatements.has(statement)) {
+          flaggedStatements.add(statement);
           context.report({
             node: node as estree.Node,
```
ilia-kebets-sonarsource, line 74: why not raise it on the whole statement?

its/ruling/src/test/expected/jsts/desktop/typescript-S7059.json
ilia-kebets-sonarsource: thanks for the explanation that is now commented in the rule implementation

packages/jsts/src/rules/S7059/rule.ts
ilia-kebets-sonarsource: all good with the comment now

ilia-kebets-sonarsource (approved):
LGTM, nice job!

packages/jsts/src/rules/S7059/rule.ts
ilia-kebets-sonarsource: all good, we do now

PR #4799, opened 2024-09-04
JS-313 Create rule S2301 (`no-selector-parameter`): Methods should not contain selector parameters

ericmorand-sonarsource:
Fixes JS-313


packages/jsts/src/rules/S2301/index.ts
```diff
@@ -0,0 +1,126 @@
...
+              isRequiredParserServices(context.parserServices)
+            ) {
+              const type = getTypeFromTreeNode(definition.name, context.parserServices);
+
+              if (isBooleanType(type)) {
+                booleanParameters.push(definition.name);
+              }
+            }
+
+            for (const booleanParameter of booleanParameters) {
```
zglicz, line 94: I can't seem to understand how can this have more than 1 element? Shouldn't we just report above in line 90?

packages/jsts/src/rules/S2301/index.ts
```diff
@@ -0,0 +1,126 @@
...
+          }
+
+          return isAChildOf(identifier.parent, node);
+        };
+
+        const testNode = testNodes.find(testNode => {
+          return testNode === node || isAChildOf(node, testNode);
+        });
+
+        if (testNode !== undefined) {
```
zglicz, line 75: should we prefer an early-return (and below in 78)? that wouldn't lead to this nested code? I re-call we wanted to add this explicit rule in this sprint

packages/jsts/src/rules/plugin.ts
```diff
@@ -678,6 +679,7 @@ export const rules: Record<string, Rule.RuleModule> = {
   'x-powered-by': S5689,
   'xml-parser-xxe': S2755,
   xpath: S4817,
+  s2301: S2301,
```
zglicz, line 682: why not provide a name to the rule? I _think_ this drives what we show here - https://github.com/SonarSource/SonarJS/blob/master/packages/jsts/src/rules/README.md

packages/jsts/src/rules/S2301/index.ts
```diff
@@ -0,0 +1,126 @@
...
+          if (identifier.parent === node) {
+            return true;
+          }
+
+          if (identifier.parent === null) {
+            return false;
+          }
+
+          return isAChildOf(identifier.parent, node);
+        };
```
zglicz, line 69: This could probably live in `helpers/ancestor.ts`

packages/jsts/src/rules/S2301/index.ts
```diff
@@ -0,0 +1,126 @@
...
+
+        const testNode = testNodes.find(testNode => {
+          return testNode === node || isAChildOf(node, testNode);
```
zglicz, line 72: does the second part of this logical check, makes this rule trigger on any usage of a boolean argument in the test condition? Is that not very strict? Doesn't that make all boolean parameters unusable? Are they only to be passed to other functions, I guess?

zglicz (commented):
Asking a few questions to better understand :)

packages/jsts/src/rules/S2301/index.ts
ericmorand-sonarsource: Good catch. Will fix.

packages/jsts/src/rules/plugin.ts
ericmorand-sonarsource: True, I forgot to find a proper key for this rule. Let me make a proposal.

packages/jsts/src/rules/S2301/index.ts
ericmorand-sonarsource: I don't know, not sure what the policy is about helpers. I assumed it was only for tools that were used by multiple rules but this logic is very specific to this use case. Let's ask the squad during the daily.

packages/jsts/src/rules/S2301/index.ts
ericmorand-sonarsource: I'm not sure yet if it is too broad to not. The idea is that every boolean parameter that creates a branch inside the body is suspect. That may be too noisy, which I'd like to check with the squad and the web guild during the verify step on Peach.

packages/jsts/src/rules/S2301/index.ts
ericmorand-sonarsource: There may be more than one boolean parameter by function - like `getCat(onlyBrown: boolean, emitAsJSON: boolean)`. In this case, both should trigger the issue.

ericmorand-sonarsource: OK, I understand, good catch. I'll add a test to cover my example and fix the code following your feedback.

yassin-kammoun-sonarsource (changes requested):
Another reviewer may have a different opinion, but from what I could see regarding the ruling outcome, we should adjust the approach and the scope because we are raising too many false positives. Currently, the rule flags any boolean parameter used in the condition of an `if`-statement. The way I understand the RSPEC, we are mainly interested in code patterns where the body of a function/method definition is an if-statement that dispatches the execution path of the definition based on a parameter:

```
function f(p1, p2, ... pn) {
  if (pX) {
    doSomething(...);
  } else {
    doSomethingElse(...);
  }
}
```

I would therefore advise to apply the following algorithm:
1. Visit function-like nodes whose body is an if-else construct.
2. Check that the condition is a test on the parameter
3. Check that both branches are call expressions
4. Raise an issue if 2) and 3) are true

yassin-kammoun-sonarsource:
@ericmorand-sonarsource We talked with the rest of the team, and we all agreed that the current approach is too broad. As mentioned earlier and discussed with the two of us yesterday, let's revisit the scope of the rule and stick to the noncompliant code pattern of the RSPEC. The ruling results show that the rule is already quite noisy. It would be a waste of time to merge the rule first and then look at the results on Peach, as the ruling results already provide evidence of the noise.

yassin-kammoun-sonarsource (changes requested):
Let's implement a different approach based on the rule pattern of the RSPEC noncompliant code.

ericmorand-sonarsource:
The rule perimeter is now much narrower; it only triggers for:

* tests that are the only children of a function body _and_ have an alternate path (i.e., either if with else or conditional expressions)
* return statements that are the only children of a function body and have a conditional expression as expression

The ruling tests demonstrate that the rule now identifies only relevant issues.

packages/jsts/src/rules/S2301/rule.ts
```diff
@@ -0,0 +1,161 @@
...
+      FunctionExpression: node => {
+        handleFunctionBody(node.body);
+      },
+      ArrowFunctionExpression: node => {
+        if (node.body.type === 'BlockStatement') {
+          handleFunctionBody(node.body);
+        }
+      },
+      Identifier: node => {
+        const isAChildOf = (identifier: Node, node: Node): boolean => {
```
vdiez, line 72: move this function definition out, no need to redefine it on each identifier

packages/jsts/src/rules/S2301/rule.ts
```diff
@@ -0,0 +1,161 @@
...
+        }
+
+        const variable = getVariableFromIdentifier(node, context.getScope());
+
+        if (variable) {
+          const definition = variable.defs[variable.defs.length - 1];
+
+          if (
+            definition?.type === 'Parameter' &&
+            isRequiredParserServices(context.parserServices)
```
vdiez, line 102: this check should be moved at the top of the create method, if we don't have type information we can return an empty object without any selectors, as we will never raise any issue

vdiez (changes requested):
I like a lot your implementation using the stack. Just some minor changes requested, let me know what you think

packages/jsts/src/rules/S2301/rule.ts
ericmorand-sonarsource: Good catch!

packages/jsts/src/rules/S2301/rule.ts
```diff
@@ -0,0 +1,161 @@
...
+type Node = ESTreeNode & NodeParentExtension;
+
+const message =
+  'Provide multiple methods instead of using "{{parameterName}}" to determine which action to take.';
+
+/**
+ * A suspect test node is a test node that is the only child of a function body
+ */
+export const S2301: Rule.RuleModule = {
+  meta: generateMeta(rspecMeta as Rule.RuleMetaData, {}, true),
```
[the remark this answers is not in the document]
ericmorand-sonarsource: Out of curiosity, is it mandatory for ESLint to work correctly even if we never reference the message ID?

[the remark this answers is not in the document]
vdiez: no, it's not mandatory. both message or messageId can be used and the existence of a messageId in the meta does not require it to be used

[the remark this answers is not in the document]
ericmorand-sonarsource: Thanks for the clarification. :pray:

sonar-plugin/javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S2301.json
```diff
@@ -0,0 +1,27 @@
...
+    },
+    "attribute": "FOCUSED"
+  },
+  "status": "ready",
+  "remediation": {
+    "func": "Constant\/Issue",
+    "constantCost": "15min"
+  },
+  "tags": [
+    "design"
```
vdiez, line 16: one minor request: I just saw that we don't use the tag 
    "type-dependent"

Can you please update rspec to add it? It's used to generate the table in the README to add if the rule needs Typescript type-checking

vdiez (approved):
LGTM! Just one final thing to be added in rspec

sonar-plugin/javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S2301.json
ericmorand-sonarsource: Yes, nice catch, let me fix RSpec and update this PR.

ericmorand-sonarsource: @vdiez We should be good now.

packages/jsts/src/rules/S2301/meta.ts
```diff
@@ -0,0 +1,33 @@
...
+
+// DO NOT EDIT! This file is autogenerated by "npm run generate-meta"
+
+export const meta = {
+  type: 'suggestion',
+  docs: {
+    description: 'Methods should not contain selector arguments',
+    recommended: false,
+    url: 'https://sonarsource.github.io/rspec/#/rspec/S2301/javascript',
+    requiresTypeChecking: false,
```
yassin-kammoun-sonarsource, line 29: This should be true since we use type information, right?

```suggestion
    requiresTypeChecking: true,
```

ericmorand-sonarsource: Good catch Yassin, thanks.

yassin-kammoun-sonarsource (approved):

PR #4802, opened 2024-09-04
JS-323 Create rule S6627 (`no-internal-api-use`): Users should not use internal APIs

ilia-kebets-sonarsource:
RSPEC PR https://github.com/SonarSource/rspec/pull/4234


ilia-kebets-sonarsource:
I just need to update the ruling

packages/jsts/src/rules/S6627/rule.ts
```diff
@@ -0,0 +1,56 @@
...
+const messages = {
+  default: 'Do not use internal APIs of your dependencies',
+};
+
+export const rule: Rule.RuleModule = {
+  meta: generateMeta(meta as Rule.RuleMetaData, { messages }),
+  create(context: Rule.RuleContext) {
+    return {
+      CallExpression(node: estree.CallExpression) {
+        if (isIdentifier(node.callee, 'require') && node.arguments.length === 1) {
```
yassin-kammoun-sonarsource, line 36: We can maybe export and refine this helper:

https://github.com/SonarSource/SonarJS/blob/522c9c029cb46b66dacc761cab19e7446c39b26f/packages/jsts/src/rules/helpers/module.ts#L55-L62

as

```js
function isRequire(node: Node): node is estree.CallExpression {
  return (
    node.type === 'CallExpression' &&
    node.callee.type === 'Identifier' &&
    node.callee.name === 'require' &&
    node.arguments.length === 1
  );
}
```

packages/jsts/src/rules/S6627/meta.ts
```diff
@@ -0,0 +1,33 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+// DO NOT EDIT! This file is autogenerated by "npm run generate-meta"
+
+export const meta = {
+  type: 'suggestion',
+  docs: {
+    description: 'Users should not use internal APIs',
```
yassin-kammoun-sonarsource, line 26: I wonder whether we should specialize the rule title for JavaScript and emphasize "nodes modules" instead of "internal APIs" because an internal API is quite broad. WDYT?

packages/jsts/src/rules/S6627/rule.ts
```diff
@@ -0,0 +1,56 @@
...
+          if (isStringLiteral(arg) && arg.value.includes('node_modules')) {
+            context.report({
+              node,
+              messageId: 'default',
+            });
+          }
+        }
+      },
+      ImportDeclaration(node: estree.ImportDeclaration) {
+        if (isStringLiteral(node.source) && node.source.value.includes('node_modules')) {
```
yassin-kammoun-sonarsource, line 47: ESLint typing is imprecise. We can remove the `isStringLiteral` check because the ECMAScript specification says that the source is always a string literal (see [ModuleSpecifier](https://tc39.es/ecma262/multipage/ecmascript-language-scripts-and-modules.html#prod-ModuleSpecifier)).

```suggestion
         const moduleName = node.source.value as string;
        if (moduleName.includes('node_modules')) {
```

packages/jsts/src/rules/S6627/cb.fixture.ts
```diff
@@ -0,0 +1,7 @@
+
+import foo from 'bar';
+
+import foo from '../node_modules/bar'; // Noncompliant {{Do not use internal APIs of your dependencies}}
```
yassin-kammoun-sonarsource, line 4: We can maybe add a line assertion as well.

packages/jsts/src/rules/S6627/cb.fixture.ts
```diff
@@ -0,0 +1,7 @@
...
+import { foo } from '../node_modules/bar'; // Noncompliant
+
+require('../node_modules/bar'); // Noncompliant
```
yassin-kammoun-sonarsource, line 7: Same here.

yassin-kammoun-sonarsource (approved):

PR #4804, opened 2024-09-06
JS-312 Improve S1607 (`no-skipped-tests`): Allow explanation comments adjacent to the disabled test

yassin-kammoun-sonarsource:
While validating S1607 on Peach, I encountered some false positives where a disabled test has an explanation comment that follows the test, such as this [issue](https://peach.aws-prd.sonarsource.com/project/issues?issues=2e27df59-cecf-46c6-b452-f8653568b6bd&open=2e27df59-cecf-46c6-b452-f8653568b6bd&id=armor-js%3Ajuice-shop).

[RSPEC's change](https://github.com/SonarSource/rspec/pull/4243) to review

### DIFF
--- a/packages/jsts/src/rules/S1607/rule.ts
+++ b/packages/jsts/src/rules/S1607/rule.ts
@@ -234,14 +234,21 @@ export const rule: Rule.RuleModule = {
     }
 
     /**
-     * Checks if the node denoting a test has an explanation comment on the line above.
+     * Checks if the node denoting a test has an adjacent explanation comment.
      */
-    function hasExplanationComment(node: estree.Node): boolean {
-      const comments = context.sourceCode.getCommentsBefore(node);
-      return comments.some(
-        comment =>
-          comment.loc!.end.line === node.loc!.start.line - 1 && comment.value.trim().length > 0,
-      );
+    function hasExplanationComment(node: estree.CallExpression) {
+      function isAdjacent(comment: estree.Comment, node: estree.Node) {
+        const commentLine = comment.loc!.end.line;
+        const nodeLine = node.loc!.start.line;
+        return commentLine === nodeLine - 1 || commentLine === nodeLine + 1;
+      }
+
+      function hasContent(comment: estree.Comment) {
+        return /\p{L}/u.test(comment.value.trim());
+      }
+
+      const comments = context.sourceCode.getAllComments();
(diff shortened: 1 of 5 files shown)


ilia-kebets-sonarsource:
@yassin-kammoun-sonarsource Great idea to tackle the line after, but why not handle when the explanation comment is on the same line as well?

packages/jsts/src/rules/S1607/rule.ts
```diff
@@ -234,14 +234,21 @@ export const rule: Rule.RuleModule = {
...
-      const comments = context.sourceCode.getCommentsBefore(node);
-      return comments.some(
-        comment =>
-          comment.loc!.end.line === node.loc!.start.line - 1 && comment.value.trim().length > 0,
-      );
+    function hasExplanationComment(node: estree.CallExpression) {
+      function isAdjacent(comment: estree.Comment, node: estree.Node) {
+        const commentLine = comment.loc!.end.line;
+        const nodeLine = node.loc!.start.line;
+        return commentLine === nodeLine - 1 || commentLine === nodeLine + 1;
```
ilia-kebets-sonarsource, line 243: If we consider [my comment here](https://github.com/SonarSource/SonarJS/pull/4804#issuecomment-2334112099),

This could be:

```suggestion
        return Math.abs(commentLine - nodeLine) <= 1;
```

ilia-kebets-sonarsource: I'm not sure how it could be tested in the comment-based framework though.

yassin-kammoun-sonarsource: Yeah, I was going to, but I just don't know how to test that.
Any suggestions?

ilia-kebets-sonarsource: Another option is to use relative line increments (`@+line_increment`) or decrements (`@-line_decrement`):

```javascript
// Noncompliant@+1
some.faulty.code();

another.faulty.code();
// another comment
// Noncompliant@-2
```

from DEV.md

ilia-kebets-sonarsource (approved):
LGTM, good job!

packages/jsts/src/rules/S1607/rule.ts
yassin-kammoun-sonarsource: Thank you so much for remembering it!

PR #4807, opened 2024-09-09
JS-65 Create rule S7063 (`no-top-level-side-effects`): Module loading should not have side effects

zglicz:


packages/jsts/src/rules/README.md
```diff
@@ -255,6 +257,7 @@ SonarJS rules for ESLint to help developers produce [Clean Code](https://www.son
 | [object-shorthand](https://sonarsource.github.io/rspec/#/rspec/S3498/javascript)                        | Object literal shorthand syntax should be used                                                                                 |     | ✅  | 🔧  |     |     |     |
 | [operation-returning-nan](https://sonarsource.github.io/rspec/#/rspec/S3757/javascript)                 | Arithmetic operations should not result in "NaN"                                                                               |     | ✅  |     |     | 💭  |     |
 | [os-command](https://sonarsource.github.io/rspec/#/rspec/S4721/javascript)                              | Using shell interpreter when executing OS commands is security-sensitive                                                       | ✅  |     |     |     |     |     |
+| [pluginRules-of-hooks](https://sonarsource.github.io/rspec/#/rspec/S6440/javascript)                    | React Hooks should be properly called                                                                                          | ✅  |     |     |     |     |     |
```
zglicz, line 260: @vdiez - can you tell me how this is generated? Seems like `no-async-constructor` and `no-skipped-test` didn't manage to update this file. Also, this rule name is a bit messed up.

I also don't have my new rule in here as well, hence, I'd like to understand...

vdiez: `npm run eslint-docs` from packages/jsts/src/rules/

I can show you tomorrow

zglicz: No worries, I've updated it. This rule name is I guess on purpose, it just wasn't re-generated...

zglicz: Addressing this in - https://github.com/SonarSource/SonarJS/pull/4808

packages/jsts/src/rules/S7063/rule.ts
```diff
@@ -0,0 +1,102 @@
...
+  topLevelStatement: Directive | Statement | ModuleDeclaration,
+) {
+  if (LOOP_STATEMENTS.includes(topLevelStatement.type)) {
+    context.report({
+      message: 'Do not include loop statements on module top level',
+      node: topLevelStatement,
+    });
+  } else if (CONDITIONAL_STATEMENTS.includes(topLevelStatement.type)) {
+    context.report({
+      message: 'Do not include conditional statements on module top level',
```
yassin-kammoun-sonarsource, line 72:
```suggestion
      message: 'Do not include conditional statements on module top level.',
```

packages/jsts/src/rules/S7063/unit.test.ts
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+import { RuleTester } from 'eslint';
+import { rule } from './';
+
+const ruleTester = new RuleTester({ parserOptions: { ecmaVersion: 2018, sourceType: 'module' } });
+ruleTester.run('"Array.reduce()" calls should include an initial value', rule, {
```
yassin-kammoun-sonarsource, line 24:
```suggestion
ruleTester.run('Module with exports should not include side effects, rule, {
```

yassin-kammoun-sonarsource (changes requested):
Let's rename the PR's title based on the [commit guidelines](https://xtranet-sonarsource.atlassian.net/wiki/spaces/LANG/pages/2567241836/JS+Commit+Guidelines#Examples).

packages/ruling/tests/tools/testProject.ts
```diff
@@ -85,7 +85,7 @@ const DEFAULT_EXCLUSIONS = [
   '**/dist/**',
   '**/vendor/**',
   '**/external/**',
-].map(pattern => new Minimatch(pattern, { nocase: true }));
+].map(pattern => new Minimatch(pattern, { nocase: true, dot: true }));
```
zglicz, line 88: @yassin-kammoun-sonarsource - this fixes the issue described in Slack, with multiple .dot folders not matching this - https://sonarsource.slack.com/archives/C03DLUN6GR3/p1726045054704339?thread_ts=1725888470.811169&cid=C03DLUN6GR3

yassin-kammoun-sonarsource: Thanks for taking care of that!

Maybe we should extract this fix into a separate PR. Alternatively, we could have two distinct commits in this PR that we don't squash on merging:

1. One commit for the rule implementation.
2. One commit for the ruling fix.

As you prefer.

packages/jsts/src/rules/S7063/unit.test.ts
```diff
@@ -0,0 +1,97 @@
...
+  ],
+  invalid: [
+    {
+      code: `
+        for (let i = 0; i < 10; i++) {
+          console.log(i);
+        }
+        export const x = 10;
+      `,
+      errors: 1,
```
yassin-kammoun-sonarsource, line 47: Let's assert at least once the issue message and location instead of just asserting the number of expected issues.

yassin-kammoun-sonarsource (commented):
@zglicz Looks good overall, but I have one minor concern about the locations of the issues we raise.

We are currently raising issues on whole statements and expressions, which could affect the user experience. To illustrate my point, consider the following Peach issues raised by two different rules:

- [S4139](https://peach.aws-prd.sonarsource.com/project/issues?issues=c712ee26-2cc1-4634-b4cb-b05a1a885c1f&open=c712ee26-2cc1-4634-b4cb-b05a1a885c1f&id=armor-js%3Abtlejuice) raises on the `for`-keyword
- [S1535](https://peach.aws-prd.sonarsource.com/project/issues?issues=332e5c01-6816-44fd-b803-6cf8808710a7&open=332e5c01-6816-44fd-b803-6cf8808710a7&id=armor-js%3Apdfjs) raises at the whole `for`-loop.

I think we can agree that the first issue is less aggressive on the user's eye. Now things get worse when different rules raise issues on the same node (see [example](https://peach.aws-prd.sonarsource.com/code?id=armor-js%3Abtlejuice&selected=armor-js%3Abtlejuice%3Acore.js&line=278)). It's obviously unavoidable to have issues on overlapping code portions, but we can mitigate this to some extent.

This is just a FYI, as I don't have a strong opinion on the matter. I let you decide what you think is best, but if you want to "improve" this, we could adjust the locations where we raise issues:

- `for`-statement -> `if`-keyword
- `if`-statement -> `if`-keyword
- variable declaration -> `var`-keyword, `let`-keyword, `const`-keyword
- assignment -> `=`-operator
- call -> callee

For instance, [here](https://github.com/SonarSource/SonarJS/blob/4b817fda7cad74aa412f23733a5c1d450ed51c1f/packages/jsts/src/rules/S4139/rule.ts#L53) is how S4139 reports issues at the token level. It should be straightforward for conditionals, loops, and declarations as it's always the first token we are interested in.

Please let me know what are your thoughts so that I can approve the PR if you decide to keep the current implementation (which is perfectly fine).

PR #4808, opened 2024-09-10
Remove a column in the readme of plugin rules

zglicz:
First thing that was confusing to me was the table in the plugin rules. Let us remove the column with the disabled rules - it only adds confusion.


packages/jsts/src/rules/plugin.ts
```diff
@@ -620,7 +620,7 @@ export const rules: Record<string, Rule.RuleModule> = {
   'redundant-type-aliases': S6564,
   'regex-complexity': S5843,
   'regular-expr': S4784,
-  'pluginRules-of-hooks': S6440,
+  'rules-of-hooks': S6440,
```
ericmorand-sonarsource, line 623: I don't know if this rule is used or not but if it is, it would create a breaking change to change its key.

Maybe it's OK because the plugin V2 is still young enough.

vdiez (approved):
LGTM!

PR #4810, opened 2024-09-10
JS-54 Add rule S7060 Modules should not import themselves

saberduck:
Fix JS-54


packages/jsts/src/rules/S7060/decorator.ts
```diff
@@ -0,0 +1,75 @@
...
+    reportWithQuickFix,
+  );
+}
+
+function reportWithQuickFix(context: Rule.RuleContext, reportDescriptor: Rule.ReportDescriptor) {
+  if (!('node' in reportDescriptor)) {
+    return;
+  }
+  const { node } = reportDescriptor;
+  let suggestion: Rule.SuggestionReportDescriptor | undefined;
```
yassin-kammoun-sonarsource, line 45: Do you think we could turn the `suggestion` variable into an array that we would initialize to an empty array? By doing so, we could
- make it constant,
- push to it a quick fix only if possible, and
- avoid the ternary expression that conditionally injects the quickfix.

I am not entirely sure, but this should work since the `suggest` property of the report expects an array. We would report an empty array of suggestions if no quickfix applies.

packages/jsts/src/rules/S7060/meta.ts
```diff
@@ -0,0 +1,33 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+// DO NOT EDIT! This file is autogenerated by "npm run generate-meta"
+
+export const meta = {
```
yassin-kammoun-sonarsource, line 23: cc @victor-diez-sonarsource because I am not sure.

- Since the rule provides a quickfix, should the rule metadata include the key-value `fixable: 'code'`?
- Do we need to update the eslint-plugin [README](https://github.com/SonarSource/SonarJS/blob/master/packages/jsts/src/rules/README.md) to include an entry for this new rule `no-self-import`? If yes, how?

vdiez: * Yes, it should. `npm run generate-meta` should put that field in the `meta.ts` file if the rspec contains the `quickfix === 'covered'` attribute. [Link](https://github.com/SonarSource/SonarJS/blob/4b817fda7cad74aa412f23733a5c1d450ed51c1f/tools/generate-meta.ts#L71)
* Yes, that is part of also of the `generate-meta` [script](https://github.com/SonarSource/SonarJS/blob/4b817fda7cad74aa412f23733a5c1d450ed51c1f/package.json#L12). You will see that it calls `cd packages/jsts/src/rules && npm run eslint-docs`, which updates the README file.

yassin-kammoun-sonarsource: Thanks 👍

packages/jsts/src/rules/README.md
```diff
@@ -215,6 +215,7 @@ SonarJS rules for ESLint to help developers produce [Clean Code](https://www.son
 | [no-same-argument-assert](https://sonarsource.github.io/rspec/#/rspec/S5863/javascript)                 | Assertions should not be given twice the same argument                                                                         | ✅  |     |     |     |     |     |
 | [no-same-line-conditional](https://sonarsource.github.io/rspec/#/rspec/S3972/javascript)                | Conditionals should start on new lines                                                                                         | ✅  |     | 🔧  | 💡  |     |     |
 | [no-self-compare](https://sonarsource.github.io/rspec/#/rspec/S6679/javascript)                         | "Number.isNaN()" should be used to check for "NaN" value                                                                       | ✅  |     | 🔧  | 💡  |     |     |
+| [no-self-import](https://sonarsource.github.io/rspec/#/rspec/S7060/javascript)                          | Module should not import itself                                                                                                | ✅  |     |     | 💡  |     |     |
```
yassin-kammoun-sonarsource, line 218:
```suggestion
| [no-self-import](https://sonarsource.github.io/rspec/#/rspec/S7060/javascript)                          | Module should not import itself                                                                                                | ✅  |     |  🔧  | 💡  |     |     |
```

yassin-kammoun-sonarsource (approved):
LGTM!

packages/jsts/src/rules/README.md
saberduck: actually not - this rule doesn't provide the `fix` , it's a bug in metadata generation

PR #4814, opened 2024-09-12
JS-327 Support TSExportAssignment nodes

ilia-kebets-sonarsource:
For the AST, you can use this in [ASTexplorer](https://astexplorer.net/):
```js
export = TAINT_SOURCE();
```

which yields:
```json
{
  "type": "Program",
  "body": [
    {
      "type": "TSExportAssignment",
      "expression": {
        "type": "CallExpression",
        "callee": {
          "type": "Identifier",
          "name": "TAINT_SOURCE",
          "range": [
            9,
            21
          ]
        },
        "arguments": [],
        "optional": false,
        "range": [
          9,
          23
        ]
      },
      "range": [
        0,
        24
      ]
    }
  ],
  "sourceType": "module",
  "range": [
    0,
    24
  ]
}
```

### DIFF
--- a/packages/jsts/src/parsers/ast.ts
+++ b/packages/jsts/src/parsers/ast.ts
@@ -274,6 +274,9 @@ function getProtobufShapeForNode(node: TSESTree.Node) {
     case 'EmptyStatement':
       shape = visitEmptyStatement(node);
       break;
+    case 'TSExportAssignment':
+      shape = visitExportAssignment(node);
+      break;
     case 'ExpressionStatement':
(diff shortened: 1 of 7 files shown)


ilia-kebets-sonarsource:
I could write a simple test in [ESTreeTest.java](https://github.com/SonarSource/SonarJS/blob/master/sonar-plugin/api/src/test/java/org/sonar/plugins/javascript/api/estree/ESTreeTest.java) to improve the coverage, but I believe it's not worth it. WDYT?

saberduck (approved):

PR #4815, opened 2024-09-12
JS-296 Fix rule S128 to support eslint v9

zglicz:
Reported in community - https://sonarsource.atlassian.net/jira/software/c/projects/JS/issues/JS-296?jql=project%20%3D%20%22JS%22%20AND%20text%20~%20%22SQ%2010.7%22%20ORDER%20BY%20created%20DESC

Following the migration guidelines outlined here - https://eslint.org/docs/latest/use/migrate-to-9.0.0#removed-codepath-currentsegments

Test:
- fix rule S128
- npm run --prefix packages/jsts/src/rules/ build 
- cd to its/eslint9-plugin-sonarjs
- npm run build 
- npm run test

Perform the same check with its/eslint8-plugin-sonarjs

### DIFF
--- a/its/eslint9-plugin-sonarjs/file.js
+++ b/its/eslint9-plugin-sonarjs/file.js
@@ -28,6 +28,18 @@ function S2376() {
   }
 }
 
+function doSomething() {
+  doSmth();
+}
+switch (x) {
+  case 0:
+    doSomething();
+  case 1:
+    doSomething();
+  default:
+    doSomethingElse();
+}
+
 /*
 
 function S125() {
--- a/its/eslint9-plugin-sonarjs/index.test.js
+++ b/its/eslint9-plugin-sonarjs/index.test.js
@@ -2,15 +2,18 @@ const { test } = require('node:test');
 const assert = require('node:assert');
 const spawn = require('cross-spawn');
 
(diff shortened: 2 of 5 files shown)


.cirrus.yml
```diff
@@ -313,7 +313,7 @@ eslint_plugin_test_task:
     - npm run test
     - cd ../eslint9-plugin-sonarjs
     - npm run build
-    - npx tsc --noEmit # check typings for tseslint.config.ts
+    #    - npx tsc --noEmit # check typings for tseslint.config.ts
```
zglicz, line 316: Currently, the generated plugin.d.ts types from the plugin, are not valid for Eslint v9 types. We cannot assert that this runs without an error. We should work on improving this. Potentially, check how typescript-eslint is doing this - https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/index.d.ts#L9

ThomasTrepanier:
Do you have a timeframe of when this fix will be released? I think I'm encountering this issue and I would like to see if this fixes it. Thanks!

vdiez:
Hi @ThomasTrepanier, end of next week at the latest.

PR #4816, opened 2024-09-12
JS-329 Upgrade TypeScript to 5.6.2

saberduck:
Behavior wrt. project references has changed. Previously, if a project overlapped with the referenced project, it would include the files of the referenced project regardless. Now, the files from the referenced project are not included. This is better for us since it avoids analyzing the files duplicitly.

### DIFF
--- a/its/ruling/src/test/expected/jsts/ace/typescript-S1874.json
+++ b/its/ruling/src/test/expected/jsts/ace/typescript-S1874.json
@@ -1,0 +1,6 @@
+{
+"ace:demo/kitchen-sink/docs/typescript.ts": [
+35,
+46
+]
+}
--- a/its/ruling/src/test/expected/jsts/ant-design/typescript-S4325.json
+++ b/its/ruling/src/test/expected/jsts/ant-design/typescript-S4325.json
@@ -8,8 +8,7 @@
 144
 ],
 "ant-design:components/anchor/Anchor.tsx": [
-27,
-258
+27
 ],
 "ant-design:components/anchor/__tests__/Anchor.test.tsx": [
 61,
@@ -129,9 +128,6 @@
 "ant-design:components/mentions/index.tsx": [
 142
 ],
-"ant-design:components/message/index.tsx": [
-107
-],
 "ant-design:components/popconfirm/index.tsx": [
 152
 ],
@@ -167,8 +163,9 @@
 "ant-design:components/tag/index.tsx": [
 55
 ],
-"ant-design:components/timeline/Timeline.tsx": [
-48
+"ant-design:components/time-picker/index.tsx": [
+72,
+74
 ],
 "ant-design:components/transfer/__tests__/dropdown.test.tsx": [
 54,
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -285,6 +285,8 @@ js_ts_ruling_task:
     - npm run build:fast
     - npm run update-ruling-data
     - npm run ruling -- --maxWorkers=8
+  on_failure:
+    debug_script: diff -rq its/ruling/src/test/expected/jsts packages/ruling/tests/actual/jsts
   cleanup_before_cache_script: cleanup_maven_repository
 
 eslint_plugin_test_task:
--- a/its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/EslintCustomRulesTest.java
+++ b/its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/EslintCustomRulesTest.java
@@ -127,7 +127,7 @@ class EslintCustomRulesTest {
       )
     )
       .isEmpty();
-    assertThat(buildResult.getLogsLines(l -> l.contains("TS API in custom rule: TS version 5.4.3")))
+    assertThat(buildResult.getLogsLines(l -> l.contains("TS API in custom rule: TS version 5.6.2")))
       .hasSize(2);
     List<Issue> issues = findIssues("eslint-custom-rules:sqKey", orchestrator);
     assertThat(issues).hasSize(2);
--- a/its/ruling/src/test/expected/jsts/TypeScript/typescript-S4325.json
+++ b/its/ruling/src/test/expected/jsts/TypeScript/typescript-S4325.json
@@ -316,19 +316,7 @@
 ],
 "TypeScript:src/harness/harness.ts": [
 36,
-743,
 1102
-],
-"TypeScript:src/harness/harnessLanguageService.ts": [
-151,
-190
-],
-"TypeScript:src/harness/virtualFileSystem.ts": [
-45,
-50,
-69,
-152,
-170
 ],
 "TypeScript:src/server/editorServices.ts": [
 1123
--- a/its/ruling/src/test/expected/jsts/ant-design/typescript-S6606.json
+++ b/its/ruling/src/test/expected/jsts/ant-design/typescript-S6606.json
@@ -16,7 +16,6 @@
 217
 ],
 "ant-design:components/descriptions/index.tsx": [
-38,
 76
 ],
 "ant-design:components/form/FormItem.tsx": [
--- a/its/ruling/src/test/expected/jsts/courselit/typescript-S4325.json
+++ b/its/ruling/src/test/expected/jsts/courselit/typescript-S4325.json
@@ -2,14 +2,8 @@
 "courselit:apps/web/components/admin/media/media-preview.tsx": [
 98
 ],
-"courselit:apps/web/components/admin/settings.tsx": [
-310
-],
 "courselit:apps/web/pages/api/payment/initiate.ts": [
 55,
 65
-],
-"courselit:apps/web/payments/stripe-payment.ts": [
-72
 ]
 }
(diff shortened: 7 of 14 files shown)


pmaieref:
Hi, I was about to open a thread in the forum regarding the TypeScript update, but as I see this here now, I thought I might just ask here. If it is too off-topic, please tell me and I'll move it to the community forum.

I am wondering if there are any guidelines, processes, documentation on how TypeScript versions are handled within SonarQube / SonarJS. The documentation does not provide any insights into:
- Can my local project use a higher / lower TS version than SQ/SonarJS
  - Considering this PR makes it into SQ 10.7, do I need my project on TS 5.6 as well to work in SQ 10.7?)
- What is the release cadence of TS updates in SonarJS? I see that 5.5 was skipped and you update directly to TS 5.6. Considering TS 5.6 is now 4 days old, this update is "early" (thanks for that!), but "support" for TS 5.5 is quite late.
- How do certain TS compiler flags play into this? Does SQ need to explicitly support / drop support for them?

The project I am working on, strives to use current / up-to-date versions of Angular, but in the past SQ's TS version was a blocking issue that prevented updating Angular as they raise the minimum versions of TS every 6 months with every major update. We ran into (compilation / execution) issues due to different TS versions and had to downgrade again. Now we don't even try if it would maybe work and just wait it out.

Thanks in advance.

gabriel-vivas-sonarsource:
Hi @pmaieref,

Indeed, I think this would be better to discuss in [the Community](https://community.sonarsource.com/).
I'll still provide an answer here while the PR is open :-)

We do regular upgrades as they come. However, sometimes conflicting dependencies prevent us from updating faster.

You should be fine using older versions of TypeScript. We don't officially support newer versions. You might experience issues with new compiler options, which can be solved with a dedicated `tsconfig.sonar.json`. In other cases, we won't yet support new syntax in files that use it, but the rest of your analysis might work well. YMMV.

Keep in mind that if you are using SonarQube, you'll only get the updates when SonarQube itself is released and you update.
The SonarQube releases happen every ~2 months (latest version) or every ~18 months (long-term version).

Hope that helps!

pmaieref:
@gabriel-vivas-sonarsource Highly appreciated the reply. If needed I'll create a post in the forum for any further discussions. Yes, we keep our SQ always on latest, thanks. :) 

Maybe you could discuss internally if a small update to https://docs.sonarsource.com/sonarqube/latest/analyzing-source-code/languages/javascript-typescript-css/ would make sense, i.e. adding some information regarding TS version support, tsconfig.sonar.json, etc.. 

Thank you very much for the quick reply here and good luck with the update! :)

ilia-kebets-sonarsource:
@saberduck What are all the ruling diffs? Is there a general pattern?

its/ruling/src/test/expected/jsts/ace/typescript-S1874.json
```diff
@@ -0,0 +1,6 @@
+{
+"ace:demo/kitchen-sink/docs/typescript.ts": [
+35,
+46
```
ilia-kebets-sonarsource, line 4: these are weird:

```ts
module Sayings { // line 35
    export class Greeter {
        greeting: string;
        constructor (message: string) {
            this.greeting = message;
        }
        greet() {
            return "Hello, " + this.greeting;
        }
    }
}
module Mankala { // line 46
   export class Features {
       public turnContinues = false;
       public seedStoredCount = 0;
       public capturedCount = 0;
       public spaceCaptured = NoSpace;

       public clear() {
           this.turnContinues = false;
           this.seedStoredCount = 0;
           this.capturedCount = 0;
           this.spaceCaptured = NoSpace;
       }

       public toString() {
           var stringBuilder = "";
           if (this.turnContinues) {
               stringBuilder += " turn continues,";
           }
           stringBuilder += " stores " + this.seedStoredCount;
           if (this.capturedCount > 0) {
               stringBuilder += " captures " + this.capturedCount + " from space " + this.spaceCaptured;
           }
           return stringBuilder;
       }
   }
}
```

its/ruling/src/test/expected/jsts/ant-design/typescript-S4325.json
```diff
@@ -8,8 +8,7 @@
 144
 ],
 "ant-design:components/anchor/Anchor.tsx": [
-27,
-258
```
ilia-kebets-sonarsource, line 12: this one is not great as well

its/ruling/src/test/expected/jsts/vuetify/typescript-S139.json
```diff
@@ -41,9 +41,6 @@
 "vuetify:packages/vuetify/src/components/VDatePicker/VDatePicker.ts": [
 131
 ],
-"vuetify:packages/vuetify/src/components/VDatePicker/VDatePickerDateTable.ts": [
-48
-],
```
ilia-kebets-sonarsource, line 46: This is really weird:

```ts
      return this.weekdayFormatter
        ? createRange(7).map(i => this.weekdayFormatter!(`2017-01-${first + i + 15}`)) // 2017-01-15 is Sunday <- this is line 48
        : createRange(7).map(i => ['S', 'M', 'T', 'W', 'T', 'F', 'S'][(i + first) % 7])
```

its/ruling/src/test/expected/jsts/ant-design/typescript-S4325.json
saberduck: doesn't look incorrect https://github.com/microsoft/TypeScript/blob/main/src/lib/dom.generated.d.ts#L15864

its/ruling/src/test/expected/jsts/vuetify/typescript-S1438.json
```diff
@@ -6949,54 +6949,6 @@
 535,
 537
 ],
-"vuetify:packages/vuetify/src/components/VDatePicker/VDatePickerDateTable.ts": [
-5,
```
ilia-kebets-sonarsource, line 6953: that's weird as well, is it the result of the TSconfig handling? all the missing semicolon issues disappeared on this file.

its/ruling/src/test/expected/jsts/vuetify/typescript-S2260.json
```diff
@@ -0,0 +1,5 @@
+{
+"vuetify:packages/vuetify/src/components/VDatePicker/VDatePickerDateTable.ts": [
+null
+]
+}
```
ilia-kebets-sonarsource, line 5: What is null here?

ilia-kebets-sonarsource:
It seems that all ruling issues disappeared from file `packages/vuetify/src/components/VDatePicker/VDatePickerDateTable.ts`

its/ruling/src/test/expected/jsts/vuetify/typescript-S1438.json
ilia-kebets-sonarsource: see https://github.com/SonarSource/SonarJS/pull/4816#issuecomment-2348993221

its/ruling/src/test/expected/jsts/vuetify/typescript-S139.json
ilia-kebets-sonarsource: see https://github.com/SonarSource/SonarJS/pull/4816#issuecomment-2348993221

its/ruling/src/test/expected/jsts/vuetify/typescript-S2260.json
ilia-kebets-sonarsource: see https://github.com/SonarSource/SonarJS/pull/4816/#issuecomment-2348993221

ilia-kebets-sonarsource (approved):
There are changes in the ruling for these rules which depend on the TS type checker:
- [S4325: Redundant casts and non-null assertions should be avoided](https://sonarsource.github.io/rspec/#/rspec/S4325)
- [S1874: Code annotated as deprecated should not be used](https://sonarsource.github.io/rspec/#/rspec/S1874)

and weirdly one issue gone for rule [S6606: Nullish coalescing should be preferred](https://sonarsource.github.io/rspec/#/rspec/S6606/javascript) in [this location](https://github.com/SonarSource/SonarJS/pull/4816/files#diff-b169e621debe554d982c82f0ab53670018f4fc3fa9091d651e34cecb2bcd2926L19)

and one weirder thing: All issues are gone from the `vuetify:packages/vuetify/src/components/VDatePicker/VDatePickerDateTable.ts` file.

packages/ruling/tests/projects/vuetify.ruling.test.ts
ilia-kebets-sonarsource, line 1: If you like the changes here, you should apply them to all the `packages/ruling/tests/projects/*.ruling.test.ts` files

saberduck:
> It seems that all ruling issues disappeared from file `packages/vuetify/src/components/VDatePicker/VDatePickerDateTable.ts`

It seems that on Mac, we reached `Maximum call stack size exceeded` in file `vuetify:packages/vuetify/src/components/VDatePicker/VDatePickerDateTable.ts.` I had to pass `--stack-size=1000` for a ruling to pass on my Mac machine. On CI, the limit is probably higher (Linux). I will fix the CI in another PR so the failure is easier to debug.

PR #4818, opened 2024-09-16
ESLINTJS-53 Fix incompatibilities with ESLint 9

vdiez:
Fixes https://sonarsource.atlassian.net/browse/ESLINTJS-53
Fixes https://sonarsource.atlassian.net/browse/JS-297


packages/jsts/src/rules/helpers/decorators/interceptor.ts
```diff
@@ -93,6 +93,11 @@ export function interceptReport(
             return originalContext.sourceCode;
           },
 
+          // @ts-ignore
+          getSource(...args) {
+            return originalContext.sourceCode.getText(...args);
+          },
+
```
zglicz, line 100: above you are using `originalContext.getAncestors();`, `originalContext.getScope(); ` and `originalContext.markVariableAsUsed(name); `
Apparently, these now requires a node as an argument https://eslint.org/docs/latest/use/migrate-to-9.0.0#-removed-multiple-context-methods

zglicz (commented):
Could you see if these are real issues that would fail in eslint v9

vdiez:
Sorry there are many changes related completly removing ESLint ids, work started in [JS-278](https://github.com/SonarSource/SonarJS/pull/4769/files)

packages/jsts/src/rules/helpers/decorators/interceptor.ts
vdiez: the ones in sourceCode require a node. This is overriding the context and forwarding same call. Tomorrow I'll check how many rules of the decorated rules use those methods.

packages/jsts/tests/linter/fixtures/wrapper/eslint.js
ericmorand-sonarsource, line 1: Out of curiosity, were these fixtures not used?

PR #4819, opened 2024-09-17
ESLINTJS-49 Rule `no-implicit-dependencies` doesn't work

ericmorand-sonarsource:
Fixes ESLINTJS-49

### DIFF
--- a/packages/jsts/src/rules/S4328/rule.ts
+++ b/packages/jsts/src/rules/S4328/rule.ts
@@ -23,7 +23,7 @@ import { Rule } from 'eslint';
 import * as estree from 'estree';
 import builtins from 'builtin-modules';
 import * as ts from 'typescript';
-import { generateMeta, getDependenciesFromPackageJson } from '../helpers';
+import { generateMeta, getDependenciesFromPackageJson, toUnixPath } from '../helpers';
 import { FromSchema } from 'json-schema-to-ts';
 import { meta, schema } from './meta';
 import { Minimatch } from 'minimatch';
@@ -38,7 +38,12 @@ export const rule: Rule.RuleModule = {
   meta: generateMeta(meta as Rule.RuleMetaData, { messages, schema }),
   create(context: Rule.RuleContext) {
     // we need to find all the npm manifests from the directory of the analyzed file to the context working directory
-    const manifests = getManifests(context.filename, context.cwd, fs);
+    const cwd = toUnixPath(context.cwd);
+    const fileName = toUnixPath(context.filename);
+    const manifests = getManifests(fileName, cwd, fs);
+
+    console.log('S4328, manifests', manifests);
+
     const dependencies: Set<string> = new Set();
 
     manifests.forEach(manifest => {
--- a/packages/jsts/src/rules/S4328/rule.ts
+++ b/packages/jsts/src/rules/S4328/rule.ts
@@ -23,10 +23,12 @@ import { Rule } from 'eslint';
 import * as estree from 'estree';
 import builtins from 'builtin-modules';
 import * as ts from 'typescript';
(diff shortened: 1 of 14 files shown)


packages/jsts/src/rules/S4328/unit.test.ts
```diff
@@ -20,11 +20,6 @@
 import { RuleTester } from 'eslint';
 import { rule } from './';
 import path from 'path';
```
vdiez, line 22: I think we should do the same for all rules calling loadPackageJsons in the unit tests

packages/jsts/src/rules/helpers/package-json.ts
```diff
@@ -183,3 +183,53 @@ function addDependency(result: Set<string | Minimatch>, dependency: string, isGl
...
+export const getManifests = (
+  fileName: string,
+  workingDirectory: string,
+  fileSystem: {
+    readdirSync: (path: string) => Array<string>;
+    readFileSync: (path: string) => Buffer;
+  },
+): Array<PackageJson> => {
+  // if the fileName is not a child of the working directory, we bail
+  const relativePath = Path.relative(workingDirectory, fileName);
```
vdiez, line 200: seems qa in failing in windows, it may be due to backward slashes. Check other helpers where we always transform paths with `toUnixPath` helper

packages/jsts/src/rules/S4328/rule.ts
```diff
@@ -35,8 +37,19 @@ const messages = {
 export const rule: Rule.RuleModule = {
   meta: generateMeta(meta as Rule.RuleMetaData, { messages, schema }),
   create(context: Rule.RuleContext) {
+    // we need to find all the npm manifests from the directory of the analyzed file to the context working directory
+    const manifests = getManifests(context.filename, context.cwd, fs);
```
vdiez, line 41: I would like this logic to sit in the `package-json` helpers, I think other rules are affected by this

vdiez (approved):

packages/jsts/src/rules/helpers/package-json.ts
ericmorand-sonarsource: Fair enough. The `path` module is OS-aware, though. If this fails, it means that at some point we are using explicit path separator.

packages/jsts/src/rules/S4328/rule.ts
ericmorand-sonarsource: You mean the rest of the code, the one that gather the dependencies?

packages/jsts/src/rules/S4328/unit.test.ts
ericmorand-sonarsource: I agree, but it would require another broader ticket. What do you think?

packages/jsts/src/rules/helpers/package-json.ts
vdiez: but some paths are sent from java. to be sure we are working with the proper paths, always use `path.unix` methods. Windows supports in any case forward slashes also

packages/jsts/src/rules/S4328/rule.ts
vdiez: ok fair enough. can you please also check what other rules depend on having the package-json cache? just create tickets for those as well pls and we'll see how we prepare the helper

packages/jsts/src/rules/S4328/unit.test.ts
vdiez: yes, check my comment below, we can create tickets for all rules having same issue

vdiez:
would be nice to add an integration test so that we are sure the rule works now in eslint context

vdiez (approved):
LGTM!

PR #4821, opened 2024-09-17
JS-85 Update usages of old apis

zglicz:


its/ruling/src/test/expected/jsts/ace/javascript-S2189.json
```diff
@@ -1,21 +1,9 @@
...
-117,
-117,
-123,
-123,
-134,
-134,
-194,
-194,
-242,
-270
```
zglicz, line 18: I have mixed feelings about this. We stopped raising more issues as we probably aren't as thorough as the previous rule was, when it was getting the context.getScope() -> called on Program:exit. However, I've checked the issues we stopped triggering on and it was only on the same variables, so maybe it's alright.

vdiez: `context.sourceCode.getScope(node)` now returns a different scope as it depends on the node we give as argument. 

In the decorator, we only have the node that is raising the issue. Before, calling `context.getScope()` returned the scope of node in selector, but from the decorator we don't know what node that is. I don't think we can

vdiez (approved):

PR #4822, opened 2024-09-17
JS-238 Fix response.ast serialization with node 20.13

zglicz:
I narrowed this down to a regression between 20.12.2 -> 20.13.0, but I was not able to find it in the changelog.
With this update, we are now receiving a "File" object consistently.


saberduck (dismissed):

saberduck:
@zglicz, could we add the test that asserts the type of response?

saberduck (approved):

PR #4823, opened 2024-09-17
JS-231 Deprecate the analyzer property sonar.nodejs.forceHost

yassin-kammoun-sonarsource:
This branch depends on #4820, which should be reviewed first.


vdiez (approved):
LGTM

PR #4829, opened 2024-09-18
JS-67 Fix FP S4782 (`no-redundant-optional`): Ignore when "exactOptionalPropertyTypes" is enabled

yassin-kammoun-sonarsource:
Please have a look also at this [RSPEC change](https://github.com/SonarSource/rspec/pull/4275), where I mention the behavior of the rule if `exactOptionalPropertyTypes` is enabled.

Note that I leave the update of the HTML rule description for when we run the rule-api tool before the next release.


vdiez (approved):
LGTM!

PR #4832, opened 2024-09-18
Remove mapping between Sonar and ESLint keys

vdiez:
Fixes #


yassin-kammoun-sonarsource (approved):
Thanks for getting rid of that!

PR #4833, opened 2024-09-19
JS-255 Restore cognitive complexity calculation at file level

yassin-kammoun-sonarsource:
The complexity calculation at file level is restored to its previous state the SonarJS ESLint plugin before this [change](https://github.com/SonarSource/eslint-plugin-sonarjs/pull/444). However, when the rule is activated for cognitive complexity at the function level, the new behavior for a nesting function that does not account for its nested function's complexity is preserved. The [RSPEC](https://github.com/SonarSource/rspec/pull/4303) of the rule was also slightly improved.

It might not be obvious to review this PR. I think that the changes on the unit test assertions and the absence of changes in the ruling is proof enough that everything is working as it should. Therefore, reviewing those might be enough.

Last but not least, I think the rule should be implemented from scratch as it's becoming quite difficult to understand what it actually does. However, I suggest we postpone this to later and create a ticket for this.


zglicz (approved):
In general makes sense. The unit.test also seems to get an increased count, which gives more confidence as well.

PR #4834, opened 2024-09-19
JS-70 More descriptive log when extracting node fails

zglicz:
Hopefully, this allows the user to optimally set the location of the node deployment in their project. This isn't documented in the docs, so maybe the logs will help.


yassin-kammoun-sonarsource (approved):
LGTM

PR #4837, opened 2024-09-19
!chore: use react and vue dependencies as optional

rtritto:
Change to option peer dependencies:
- @babel/preset-react
- eslint-plugin-react
- eslint-plugin-react-hooks
- vue-eslint-parser

### Note
Please run `npm i --lockfile-version 3` to update `packages\jsts\src\rules\package-lock.json`

### DIFF
--- a/packages/jsts/src/linter/config/linter-config.ts
+++ b/packages/jsts/src/linter/config/linter-config.ts
@@ -53,9 +53,12 @@ export function createLinterConfig(
     globals,
     parserOptions,
     rules: {},
-    /* using "max" version to prevent `eslint-plugin-react` from printing a warning */
-    settings: { react: { version: '999.999.999' } },
   };
+  try {
+    require.resolve('eslint-plugin-react');
+    /* using "max" version to prevent `eslint-plugin-react` from printing a warning */
+    config.settings = { react: { version: '999.999.999' } };
+  } catch {}
   enableRules(config, inputRules, linterRules);
   enableInternalCustomRules(config);
   return config;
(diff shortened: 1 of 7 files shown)


vdiez:
Hello @rtritto,

Thank you very much for contributing! 

Correct me if I'm wrong, but with this PR, only users having those dependencies will have S125 enabled right? 

Is the goal to reduce our number of dependencies?

rtritto:
> Correct me if I'm wrong, but with this PR, only users having those dependencies will have [S125](https://sonarsource.github.io/rspec/#/rspec/S125/javascript) enabled right?

This is a draft that I started, it needs to be continued by anyone.

> Is the goal to reduce our number of dependencies?

Yes, in my use case I use SolidJS instead of React (next step is to introduce optional `eslint-plugin-solid` dependency)

ericmorand-sonarsource:
I think that we should not use `try...catch` statements to drive business logic. But, maybe more importantly, I believe that the plugin, in its current state, is an atomic entity that exposes all of its rules, and that if we want to export a discrete set of rules, the package should be designed to handle this.

Typically, if we want the React rules available separately, we should provide them either through a separate plugin or through a factory that takes the dependency as arguments.

Something like that:

> eslint.config.mjs

```js
const {createPluginWithReactRules} from "eslint-plugin-sonarjs";
const reactPlugin from "eslint-plugin-react";

// createPluginWithReactRules expects an instance of an ESLint plugin as first parameter
const sonarJSPlugin = createPluginWithReactRules(reactPlugin);

// export ESLint configuration
```

This way, the user of the plugin remains in control.

That said, this is not how the plugin is designed today: it exposes all the rules, unconditionally, and as such requires that all the rule dependencies are installed. I don't think that we should depend on `try...catch` hacks to change that behaviour.

PR #4839, opened 2024-09-20
Js-261 Update stylint to v16

zglicz:


zglicz:
After digging a bit deeper with @vdiez, turns out we are not able to perform this update. The reason is: whilst Stylelint v16 would be fine to update, as it supports CommonJS. A rule: 'no-extra-semicolon' has been deprecated in Stylelint v15 and was removed in v16. This rule has been ported to @stylistic/styleline-plugin, which in turn, doesn't support CommonJS and only has ESModule support. Therefore, we wouldn't be able to maintain the same rules. This might be yet another reason to migrate from CommonJS to ESModules....

PR #4840, opened 2024-09-20
ESLINTJS-56 Improve the performances of package manifest search

ericmorand-sonarsource:
Fixes ESLINTJS-56

### DIFF
--- a/packages/jsts/src/rules/helpers/find-up.ts
+++ b/packages/jsts/src/rules/helpers/find-up.ts
@@ -1,0 +1,128 @@
+/*
+ * SonarQube JavaScript Plugin
+ * Copyright (C) 2011-2024 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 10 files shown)


vdiez (changes requested):
Really nice! Just please try to avoid adding a new dependency (`sinon`) as `jest` already provides spy functionality

packages/jsts/src/rules/helpers/find-up.ts
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    let cacheContent = cache.get(from);
+
+    if (cacheContent === undefined) {
+      cacheContent = [];
+
+      cache.set(from, cacheContent);
+
+      for (const entry of filesystem.readdirSync(from)) {
```
vdiez, line 83: if you are listing and stating, better use `filesystem.readdirSync(dir, { withFileTypes: true });` which makes both in the same call

packages/jsts/src/rules/helpers/find-up.ts
```diff
@@ -0,0 +1,128 @@
...
+          // todo: this is testable and should be tested
+          stats = {
+            isFile: () => false,
+          };
+        }
+
+        if (stats.isFile()) {
+          const basename = Path.basename(fullEntryPath);
+
+          if (new Minimatch(pattern).match(basename)) {
```
vdiez, line 101: `new Minimatch(pattern)` will always be the same. I think we could extract from here and pass it in the constructor `createFindUp`. WDYT?

vdiez (changes requested):

vdiez (approved):
LGTM!

PR #4843, opened 2024-09-23
JS-97 Fix FP S1128 (`unused-import`): Ignore imported symbols with v-prefixed used as Vue.js directives

yassin-kammoun-sonarsource:


packages/jsts/src/rules/S1128/rule.ts
```diff
@@ -219,8 +219,12 @@ export const rule: Rule.RuleModule = {
             const {
               name: { name },
             } = node;
-            vueIdentifiers.add(toCamelCase(name));
-            vueIdentifiers.add(toPascalCase(name));
+            const camelCased = toCamelCase(name);
```
zglicz, line 222: in the lines for VElement, there is some similar code... Might this also need to be v-prefixed?

zglicz (approved):

packages/jsts/src/rules/S1128/rule.ts
yassin-kammoun-sonarsource: From what I read from Vue.js documentation, this v-prefix thingy is only used for directives.

PR #4844, opened 2024-09-23
JS-74 Update S6747 (`no-unknown-property`): Enable the rule only on React projects

yassin-kammoun-sonarsource:


zglicz (approved):
successkid

PR #4849, opened 2024-09-24
JS-145 Fix FP S6847 (`no-noninteractive-element-interactions`): Reduce the set of handlers to consider

yassin-kammoun-sonarsource:


vdiez (approved):
LGTM

PR #4851, opened 2024-09-25
Resolve false postive in S5693 for multer storage engine init

pmckee11:
The express multer middleware module has several member exports in addition to the main export. This change fixes the logic in S5693 so that it doesn't check for the content length limit parameter being passed to those members, since it's not applicable.

Googled a bit and have seen several folks in the community stumped by this false positive, e.g. https://stackoverflow.com/questions/76305839/sonar-security-warning-for-multer-express-js


vdiez:
external contribution don't have permissions for our CI. Created a new PR [here](https://github.com/SonarSource/SonarJS/pull/4886)

PR #4858, opened 2024-09-26
Always look for package.json files up to root directory

vdiez:


vdiez:
we leave as is, we depend on `cwd` on many of our tests to avoid going up to SonarJS `package.json`

PR #4863, opened 2024-10-08
JS-345 Migrate from CommonJS to ESModules

vdiez:
Fixes https://sonarsource.atlassian.net/browse/JS-345


kebetsi:
It looks weird how all the imports in TS code use the `.js` extension. How does this work? Do we already execute tests against compiled JS files?

kebetsi:
why aren't we using the packageJsons in project analysis?

kebetsi:
Is this the pattern to load some CJS code?
```js
import Module from 'node:module';
const require = Module.createRequire(import.meta.url);
require.resolve('@typescript-eslint/parser'),
```

yassin-kammoun-sonarsource (changes requested):

package.json
```diff
@@ -62,6 +62,8 @@
     "@types/tmp": "0.2.6",
     "cpy-cli": "5.0.0",
     "dir-compare": "5.0.0",
+    "esprima": "4.0.1",
```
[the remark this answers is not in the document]
vdiez: i would say that's another ticket, here we are just adding a dependency that was missing in an already existing test

.cirrus/nodejs.Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM ${CIRRUS_AWS_ACCOUNT}.dkr.ecr.eu-central-1.amazonaws.com/base:j17-latest
 
 USER root
 
-ARG NODE_VERSION=18
```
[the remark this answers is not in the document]
vdiez: correct

packages/ruling/tests/filter/JavaScriptExclusionsFilter.ts
```diff
@@ -17,9 +17,9 @@
  * along with this program; if not, write to the Free Software Foundation,
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
-import { bundleAssessor } from './bundleAssessor';
-import { minificationAssessor } from './minificationAssessor';
-import { sizeAssessor } from './sizeAssessor';
+import { bundleAssessor } from './bundleAssessor.js';
```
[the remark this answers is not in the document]
vdiez: that's how it is at the moment, further reads https://stackoverflow.com/a/77150985

vdiez:
> It looks weird how all the imports in TS code use the `.js` extension. How does this work? Do we already execute tests against compiled JS files?
> 
> It looks very verbose when replacing things like `../parameters` by `../parameters/index.js`

same answer as I did to Yassin, it's the way it is right now. https://stackoverflow.com/a/77150985

vdiez:
> why aren't we using the packageJsons in project analysis?

that was from another PR about better packagejsons lookup from eric, appears here due to rebase

vdiez:
> Is this the pattern to load some CJS code?
> 
> ```js
> import Module from 'node:module';
> const require = Module.createRequire(import.meta.url);
> require.resolve('@typescript-eslint/parser'),
> ```

correct, otherwise the alternative is dynamic import which is not always possible in top level

packages/jsts/src/rules/external.ts
```diff
@@ -17,10 +17,12 @@
  * along with this program; if not, write to the Free Software Foundation,
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
-import { eslintRules } from './core';
-import { rules as reactPluginRules } from 'eslint-plugin-react';
-import { tsEslintRules } from './typescript-eslint';
-import { rules as a11yPluginRules } from 'eslint-plugin-jsx-a11y';
+import { eslintRules } from './core/index.js';
+import pkg from 'eslint-plugin-react';
+const reactPluginRules = pkg.rules;
```
kebetsi, line 22: Can't you do `import { rules: reactPluginRules } from 'eslint-plugin-react';` ?

kebetsi: or `import { rules as reactPluginRules } from 'eslint-plugin-react'; `

kebetsi: ah it's because the package is CJS, right?

vdiez: yes, it does not recognize the exports properly. michal found this workaround

pom.xml
```diff
@@ -11,7 +11,7 @@
 
   <groupId>org.sonarsource.javascript</groupId>
   <artifactId>javascript</artifactId>
-  <version>10.16.0-SNAPSHOT</version>
+  <version>10.17.0-SNAPSHOT</version>
```
kebetsi: you could have done this one in its own commit, but whatever.

vdiez: this is already on master, part of the merge with master commit, you can ignore

sonar-plugin/bridge/src/test/java/org/sonar/plugins/javascript/bridge/RulesBundlesTest.java
```diff
@@ -39,16 +38,6 @@ class RulesBundlesTest {
...
-  @Test
-  void test() throws Exception {
-    TestRulesBundle rulesBundle = new TestRulesBundle("/test-bundle.tgz");
-    RulesBundles rulesBundles = new RulesBundles(new TestRulesBundle[] { rulesBundle });
-    List<Path> paths = rulesBundles.deploy(tempDir);
-    assertThat(paths).hasSize(1);
-    assertThat(paths.get(0)).exists();
-    assertThat(paths.get(0).resolve("bin/server")).hasContent("#!/usr/bin/env node\n\n");
-  }
-
```
kebetsi, line 51: why remove this test?

vdiez: RulesBundles should be used for rules bundles, not for bridge bundle, don't know why it was there at all this test. in any case, it does not work now, because rulesBundles has been changed to provide a path to a rules.js file, so no bin/server can be expected there

tools/fetch-node/node-distros.mjs
```diff
@@ -1,4 +1,4 @@
-export const NODE_VERSION = 'v20.9.0';
```
kebetsi, line 1: we should mention this in the changelog

vdiez: already in master, also part of the merge with master, you can ignore

packages/bridge/tests/memory.test.ts
```diff
@@ -17,40 +17,43 @@
...
   it('should log out-of-memory troubleshooting guide', () => {
-    console.error = jest.fn();
+    console.error = mock.fn();
 
     logMemoryError({ code: 'ERR_WORKER_OUT_OF_MEMORY' });
 
-    expect(console.error).toHaveBeenCalledWith(
-      expect.stringContaining(
-        `The analysis will stop due to the Node\.js process running out of memory`,
-      ),
```
kebetsi, line 31: can't we assert this message anymore?

kebetsi: I think you forgot to move it as it's asserted below

packages/bridge/tests/router.test.ts
```diff
@@ -274,10 +276,6 @@ describe('router', () => {
   });
 
   it('should route /new-tsconfig requests', async () => {
-    /**
-     * There is no easy way to test that a module was unloaded, because jest is modifying require calls for tests
-     * @see https://github.com/facebook/jest/issues/6725
-     */
```
kebetsi, line 280: Is it overkill to test this part now that we aren't using Jest anymore?

packages/bridge/tests/router.test.ts
```diff
@@ -311,13 +309,13 @@ describe('router', () => {
...
   it('should forward /tsconfig-files failures', async () => {
-    console.error = jest.fn();
+    console.error = mock.fn();
     const tsConfig = path.join(fixtures, 'malformed.json');
     const data = { tsConfig };
     const response = (await request(server, '/tsconfig-files', 'POST', data)) as string;
     const { error } = JSON.parse(response);
     expect(error).toEqual('Debug Failure.');
-    expect(console.error).toHaveBeenCalled();
+    assert((console.error as Mock<typeof console.error>).mock.calls.length > 0);
```
kebetsi, line 318: Is there no nicer API, like doesn't `mock.fn()` return something that we can use instead of `(console.error as Mock<typeof console.error>).mock`?

packages/bridge/tests/server.test.ts
```diff
@@ -37,21 +40,20 @@ describe('server', () => {
   });
 
   it('should start', async () => {
-    expect.assertions(5);
```
kebetsi, line 40: why remove this one?

packages/jsts/tests/linter/wrapper.test.ts
```diff
@@ -97,14 +103,14 @@ describe('LinterWrapper', () => {
       expect.objectContaining({
         ruleId: customRuleId,
         message:
-          `Visited 'Today here, tomorrow the world!' literal from a custom rule ` +
+          `Visited 'sonar-context' literal from a custom rule ` +
```
kebetsi, line 106: where does this change come from?

packages/jsts/tests/rules/index.test.ts
```diff
@@ -18,26 +18,33 @@
...
+      const { rule } = await import(
+        pathToFileURL(path.join(ruleFolder, sonarKey, 'index.js')).toString()
+      );
       expect(rule.meta.docs!.url).toBe(
         `https://sonarsource.github.io/rspec/#/rspec/${sonarKey}/javascript`,
       );
-      const { meta } = require(path.join(ruleFolder, sonarKey, 'meta.ts'));
+      const { meta } = await import(
+        pathToFileURL(path.join(ruleFolder, sonarKey, 'meta.js')).toString()
+      );
```
kebetsi, line 47: there is no importSync?

packages/bridge/tests/router.test.ts
vdiez: it actually returns the mock, but the types do not seem to behave as expected for TypeScript right now it seems. This is more about the node types, so probably there is something better once we update them.

packages/bridge/tests/server.test.ts
vdiez: because a number of assertions is not working anymore. I could use any number and it would always pass

packages/jsts/tests/linter/wrapper.test.ts
vdiez: before there were a `custom-rule.js` and a `custom-rule.ts`, the js version only contained a string. it was a very strange test. not we only use the typescript version, which is a custom rule, to parse the file containing the rule.

packages/jsts/tests/rules/index.test.ts
vdiez: No, AFAIK

kebetsi: Sad

packages/bridge/tests/server.test.ts
kebetsi: There are still a few usages of `expect.assertions()` (I haven't found how to share a search on a certain branch)

Screenshot 2024-10-09 at 13 58 03

vdiez: ok, gonna remove those

ericmorand-sonarsource:
@vdiez , @yassin-kammoun-sonarsource , @kebetsi , about the `.js` extension, the explanation is in the TypeScript documentation:

https://www.typescriptlang.org/docs/handbook/modules/theory.html#module-specifiers-are-not-transformed

> While the module compiler option can transform imports and exports in input files to different module formats in output files, the module specifier (the string from which you import, or pass to require) is always emitted as-written.

With CommonJS, importing without an extension is allowed because of how the algorithm is specified (roughly: first, lookup for a file, then lookup for a directory that contains an index). With ESM, importing a directory is not supported and the extension is mandatory. Since TS doesn't transform the module specifier, the extension must be one of the final artifacts.

packages/bridge/tests/router.test.ts
vdiez: not sure what that comment actually meant, tbh. I don't see the relation with the test

packages/bridge/tests/server.test.ts
vdiez: done

packages/bridge/tests/memory.test.ts
vdiez: added in latest commit, good catch

kebetsi (approved):
Amazing job @vdiez and @zglicz !

yassin-kammoun-sonarsource (approved):

pumano:
would love to get fresh release with it!

vdiez:
Hi @pumano,

for the moment we will still only publish CommonJS target in the eslint-plugin-sonarjs package. We plan to provide both targets soon.

Cheers

PR #4870, opened 2024-10-14
Reintroduce support to Node 18

vdiez:


saberduck (approved):

tychenjiajun:
Now the eslint plugin depends on minimatch^10, and minimatch^10 requires node>=20, make it unable to be installed on node 18

vdiez:
Hello @tychenjiajun,

yes, you are right. We spotted the minimatch requirement too late after release. In any case, it's still compatible with node 18 as it seems minimatch just [upgraded ](https://github.com/isaacs/minimatch/commit/632e0da294b644ce2b537df4d64d4316fd84e0f1)engines requirement without actual need. I may be wrong though, but our CI runs with node 18 without any issues. 

Next release we'll move back to minimatch 9 to avoid raising the warning.

Cheers

PR #4874, opened 2024-10-15
JS-372 Introduce caches to improve SonarLint performance

zglicz:
As part of this improvement, we include:
1) TsConfigCache <- special cache for SonarLint runtime. It supports 2 levels of caching, of the file system traversal to detect tsconfigs and consequently a cache for a filename to tsconfig mapping. It is generous with the clearing of the cache whenever, a tsconfig is modified or a new project file (javascript or typescript) is created.
2) Optimization of the filename to tsconfig search, by reordering pending tsconfigs to process the most likely tsconfig first.
3) Removing superfluous calls to Path.realPath() when receiving filenames from the Typescript compiler and later when matching files to tsconfig. For SonarJS, this proved important to reduce querying all tsconfigs (195 configs) found in SonarJS from 20s to 3s.


.cirrus.yml
```diff
@@ -129,8 +129,6 @@ build_task:
   build_and_deploy_script:
     - source cirrus-env BUILD
     - node --version
-    - npm ci
-    - npm run check-format
```
vdiez, line 133: why did we remove this?

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/JavaScriptPlugin.java
```diff
@@ -348,7 +350,7 @@ static class SonarLintPluginAPIVersion {
 
     boolean isDependencyAvailable() {
       try {
-        Class.forName("org.sonarsource.sonarlint.plugin.api.module.file.ModuleFileSystem");
```
vdiez, line 351: We should also check for ModuleFileSystemClass, keep both checks

vdiez (approved):
LGTM! just a couple of minor things I'm having questions about

PR #4875, opened 2024-10-16
Fix typedoc action

vdiez:


kebetsi (approved):
LGTM! Thanks for fixing this!

PR #4877, opened 2024-10-17
Update sonar-api and orchestrator

saberduck:


its/ruling/src/test/java/org/sonar/javascript/it/CssRulingTest.java
```diff
@@ -112,7 +112,6 @@ void test() throws Exception {
       .setProjectKey(PROJECT_KEY)
       .setProjectName(PROJECT_KEY)
       .setProjectVersion("1")
-      .setLanguage("css")
```
ericmorand-sonarsource, line 115: Just one question, not blocking for the merge: was it never required to begin with?

ericmorand-sonarsource (approved):
Looks good

PR #4881, opened 2024-10-18
JS-355 Fix SQ coverage

vdiez:


pom.xml
```diff
@@ -111,7 +107,7 @@
     <sonar.exclusions>packages/**/*.test.ts,packages/**/*.fixture.*,packages/**/fixtures/**/*</sonar.exclusions>
     <sonar.tests>packages/</sonar.tests>
     <sonar.test.inclusions>packages/**/*.test.ts</sonar.test.inclusions>
-    <sonar.javascript.lcov.reportPaths>coverage/lcov.info</sonar.javascript.lcov.reportPaths>
+    <sonar.javascript.lcov.reportPaths>lcov.info</sonar.javascript.lcov.reportPaths>
```
zglicz, line 110: I would have sworn I tried changing this and it was still picked up from the inner folder.

Why does this change the coverage results?

zglicz (approved):

PR #4886, opened 2024-10-28
Fix S5693: Resolve false positive for multer storage engine init

vdiez:
Copied from https://github.com/SonarSource/SonarJS/pull/4851


zglicz (approved):
lgtmacro

PR #4888, opened 2024-10-31
JS-379 Merge file traversals to count project size and tsconfigs together

zglicz:
[JS-379](https://sonarsource.atlassian.net/browse/JS-379)

[JS-379]: https://sonarsource.atlassian.net/browse/JS-379?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/JavaScriptPlugin.java
```diff
@@ -121,10 +120,15 @@ public class JavaScriptPlugin implements Plugin {
   public static final String TS_EXCLUSIONS_KEY = "sonar.typescript.exclusions";
   public static final String[] EXCLUSIONS_DEFAULT_VALUE = new String[] {
     "**/node_modules/**",
+    "**/node_modules",
```
vdiez, line 123: I think we need to rethink this. Users adding exclusions will not be aware that they need to be repeated with both globs.

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/analysis/JsTsSensor.java
```diff
@@ -57,30 +56,11 @@ public JsTsSensor(
     AnalysisWithProgram analysisWithProgram,
     AnalysisWithWatchProgram analysisWithWatchProgram,
     AnalysisConsumers consumers
-  ) {
```
vdiez, line 60: we can remove comment at the top of constructor

vdiez (changes requested):
Looking good, but I have doubts about the need to double the exclusions, specially because this impacts how the user need to pass properties. We can have a look on Monday on how to make it work with the already existing glob

vdiez (approved):
LGTM!

PR #4893, opened 2024-11-05
JS-400 Single entrypoint for worker and server

vdiez:
[JS-400](https://sonarsource.atlassian.net/browse/JS-400)

[JS-400]: https://sonarsource.atlassian.net/browse/JS-400?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/bridge/src/delegate.ts
```diff
@@ -0,0 +1,97 @@
...
+ */
+export function createDelegator(worker: Worker | undefined) {
+  return function (type: RequestType) {
+    return worker ? createWorkerHandler(worker, type) : createHandler(type);
+  };
+}
+
+/**
+ * Handler to analyze in the same thread as HTTP server. Used for testing purposes
+ * @param type
```
zglicz, line 39: Is this comment `@param type` helping?

packages/bridge/src/delegate.ts
```diff
@@ -0,0 +1,97 @@
...
+  next: express.NextFunction,
+) {
+  switch (message.type) {
+    case 'success':
+      if (typeof message.result === 'object' && outputContainsAst(message.result)) {
+        sendFormData(message.result, response);
+      } else {
+        response.send(message.result);
+      }
+      break;
```
zglicz, line 76: no curly braces around multiple statements? or is an if simple enough?

ericmorand-sonarsource:
I find the pattern strange, and more generally, I don't see why we need a different script for the server and the worker.

I would have gone with this pattern:

```js
#!/usr/bin/env node
import { Worker, isMainThread, parentPort }  from "node:worker_threads";

if (isMainThread) {
  // this is the parent, aka the CLI script
  const worker = new Worker(import.meta.filename);

  worker.once("message", (message) => {
    console.log('The worker did its thing', message);
  });

  worker.postMessage('Do something');
}
else {
  // this is the worker
  console.log('ParentPort is guaranteed to exist since we are not in the main thread');

  parentPort.once("message", (message) => {
    if (message === 'Do something') {
      parentPort.postMessage("I did what I am supposed to do");
    }
  });
}
```

There seems to be some partial implementation of this pattern - the `isMainThread` in `server.mjs` and the `if (parentPort)` in `worker.ts` - which adds to the confusion.

zglicz (approved):
lgtm

PR #4897, opened 2024-11-07
JS-380 Use Rollup to bundle SonarJS

zglicz:


zglicz:
The bundling will be done with esbuild on PR - https://github.com/SonarSource/SonarJS/pull/4901

PR #4901, opened 2024-11-12
JS-380 Bundle bridge using esbuild

zglicz:
[JS-380](https://sonarsource.atlassian.net/browse/JS-380)

[JS-380]: https://sonarsource.atlassian.net/browse/JS-380?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


typings/postcss-syntax/index.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'postcss-syntax';
```
zglicz, line 1: sad-panda-sitting-vector-illustation_952832-137127

esbuild.mjs
```diff
@@ -0,0 +1,103 @@
...
+  plugins: [
+    textReplace({
+      include: /server\.mjs$/,
+      pattern: [['new URL(import.meta.url)', '__filename']],
+    }),
+    textReplace({
+      include: /lib[\/\\]jsts[\/\\]src[\/\\]parsers[\/\\]ast\.js$/,
+      pattern: [['path.dirname(fileURLToPath(import.meta.url))', '__dirname']],
+    }),
+    // Simplify the loader function in babel. At the end it's just importing Babel parser
```
zglicz, line 25: Do you think it would be worth to point a specific line in a github repo? Perhaps this way, if something changes, it would be a good starting point to investigate.

Also, it would be great, if we could assert that these replacements actually happened. If we could make the bundling fail if any of the replacements didn't occur....

its/plugin/plugins/eslint-custom-rules-plugin/bundle/dist/rules.js
```diff
@@ -20,9 +20,6 @@ exports.rules = [
         ],
       },
       create(context) {
```
zglicz, line 22: For posteriority, the lines below were removed because typescript isn't installed and available

packages/css/tests/linter/config.test.ts
```diff
@@ -29,13 +29,10 @@ describe('createStylelintConfig', () => {
       { key: 'bar', configurations: [42] },
     ];
     const config = createStylelintConfig(rules);
-    expect(config).toEqual({
-      customSyntax: 'postcss-syntax',
```
zglicz, line 33: we now pass the imported module and we can't really compare it

packages/jsts/src/builders/build.ts
```diff
@@ -70,7 +70,7 @@ export function buildSourceCode(input: JsTsAnalysisInput, language: JsTsLanguage
     return parseForESLint(
       input.fileContent,
       parser.parse,
-      buildParserOptions({ parser: vueFile ? parsers.javascript.parser : undefined }, true),
+      buildParserOptions({ parser: vueFile ? parsers.javascript : undefined }, true),
```
zglicz, line 73: Again, we want to pass in a resolved module. parsers.javascript.parser was the library name `@babel/eslint-parser` and it would get invoked dynamically and esbuild wouldn't be able to resolve it.

server.mjs
```diff
@@ -35,8 +35,5 @@ if (isMainThread) {
   }
 
   context.setContext({ workDir, shouldUseTypeScriptParserForJS, sonarlint, debugMemory, bundles });
-
-  server
-    .start(Number.parseInt(port), host, createWorker(new URL(import.meta.url), getContext()))
-    .catch(() => {});
+  server.start(Number.parseInt(port), host, createWorker(new URL(import.meta.url), getContext()));
```
zglicz, line 38: Consciously we remove the catch. There doesn't seem to be a benefit of silently swallowing the error.

sonar-plugin/bridge/src/main/java/org/sonar/plugins/javascript/bridge/BundleImpl.java
```diff
@@ -37,7 +37,7 @@ public class BundleImpl implements Bundle {
 
   // this archive is created in the bridge module
   private static final String BUNDLE_LOCATION = "/sonarjs-1.0.0.tgz";
-  private static final String DEFAULT_STARTUP_SCRIPT = "package/bin/server.mjs";
+  private static final String DEFAULT_STARTUP_SCRIPT = "package/bin/server.cjs";
```
zglicz, line 40: One step forward, two steps back ...

sad-panda-sitting-vector-illustation_952832-137127

sonar-plugin/bridge/src/main/java/org/sonar/plugins/javascript/nodejs/NodeCommandBuilderImpl.java
```diff
@@ -46,7 +46,7 @@ public class NodeCommandBuilderImpl implements NodeCommandBuilder {
 
   public static final String NODE_EXECUTABLE_DEFAULT = "node";
   private static final String NODE_EXECUTABLE_DEFAULT_MACOS =
-    "package/node_modules/run-node/run-node";
+    "package/bin/run-node";
```
zglicz, line 49: `run-node` is now a part of our project, we don't need to expect it to be in the node_modules

zglicz (commented):
Commenting with a bit more details, as I'm the author it seems :)

vdiez (approved):

PR #4903, opened 2024-11-14
JS-409 CssRulingTest change mvn verify -> mvn test

zglicz:
[JS-409](https://sonarsource.atlassian.net/browse/JS-409)

I noticed that at the end of the CssRulingTest we download some jars and try to publish a package. As @saberduck suggested, we could try to use `mvn test` instead of `mvn verify`.

[JS-409]: https://sonarsource.atlassian.net/browse/JS-409?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -102,7 +102,7 @@ plugin_qa_body: &PLUGIN_QA_BODY
   qa_script:
     - source cirrus-env QA
     - source set_maven_build_version $BUILD_NUMBER
-    - mvn -f its/plugin/pom.xml -Dsonar.runtimeVersion=${SQ_VERSION} ${MVN_TEST} -B -e -V verify surefire-report:report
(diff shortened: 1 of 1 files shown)


saberduck:
@zglicz did you check if this brings any timing benefits? I think it's quite risky change, I would keep it only if there are significant improvements and maybe let's start with Css ruling only at first

zglicz:
Comparing: https://github.com/SonarSource/SonarJS/runs/32957225989 (last master) and this one
CSS ruling, now: 1:23, before: 1:36
plugin qa win1, now: 12:25, before: 11:40
plugin qa win2, now: 12:40, before: 13:25
win_sonarlint: now:  7:48, before: 8:25


So I wouldn't say conclusively that it is faster, but just the fact, that we aren't trying to `package` and run `tests` - which only includes maven IntegrationTest, which as I understand, we don't use...

saberduck (approved):

PR #4907, opened 2024-11-15
JS-381 Update pom parent version

zglicz:
[JS-381](https://sonarsource.atlassian.net/browse/JS-381)

Based on the releasability, we should update this parent version in the pom

[JS-381]: https://sonarsource.atlassian.net/browse/JS-381?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/its/plugin/plugins/consumer-plugin/src/main/java/org/sonar/samples/javascript/consumer/package-info.java
+++ b/its/plugin/plugins/consumer-plugin/src/main/java/org/sonar/samples/javascript/consumer/package-info.java
@@ -1,6 +1,6 @@
 /*
  * SonarQube JavaScript Plugin
- * Copyright (C) 2011-2024 SonarSource SA
+ * Copyright (C) 2012-2024 SonarSource SA
  * mailto:info AT sonarsource DOT com
  *
  * This program is free software; you can redistribute it and/or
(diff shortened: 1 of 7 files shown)


its/plugin/plugins/consumer-plugin/src/main/java/org/sonar/samples/javascript/consumer/package-info.java
```diff
@@ -1,6 +1,6 @@
 /*
  * SonarQube JavaScript Plugin
- * Copyright (C) 2011-2024 SonarSource SA
+ * Copyright (C) 2012-2024 SonarSource SA
```
vdiez, line 3: why 2012?

vdiez (changes requested):
there are many files with 2011. why this change? was this done by maven or manually?

its/plugin/plugins/consumer-plugin/src/main/java/org/sonar/samples/javascript/consumer/package-info.java
zglicz: its consistent across the directory. I could update all the 2012 files to 2011?

vdiez (approved):

PR #4913, opened 2024-11-19
JS-371 - Deprecate Node.js 18

ericmorand-sonarsource:
[JS-371](https://sonarsource.atlassian.net/browse/JS-371)

[JS-371]: https://sonarsource.atlassian.net/browse/JS-371?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
   },
   "homepage": "https://github.com/SonarSource/SonarJS#readme",
   "engines": {
-    "node": "^18.17.0 || ^20.9.0 || >=21.1.0"
+    "node": ">=22"
   },
   "type": "module",
   "devDependencies": {
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ If you want to report a bug, request a feature, or provide other kind of feedbac
 
 ## Prerequisites
 
-To work on this Project, it is required to have the following tools installed:
+To work on this project, it is required to have the following tools installed:
 
 - [JDK 17](https://docs.aws.amazon.com/corretto/latest/corretto-17-ug/what-is-corretto-17.html)
 - [Node.js](https://nodejs.org/en) >= 22
--- a/README.md
+++ b/README.md
@@ -30,15 +30,26 @@ If you want to report a bug, request a feature, or provide other kind of feedbac
 
 # Contributing
 
-#### 1. Request a new feature
+## Prerequisites
+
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -56,7 +56,7 @@
   },
   "homepage": "https://github.com/SonarSource/SonarJS#readme",
   "engines": {
-    "node": "^18.17.0 || ^20.9.0 || >=21.1.0"
+    "node": ">=22"
```
yassin-kammoun-sonarsource, line 59: Since we are only deprecating, not dropping support, maybe we should keep this as it was before.

.cirrus/nodejs.Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM ${CIRRUS_AWS_ACCOUNT}.dkr.ecr.eu-central-1.amazonaws.com/base:j17-latest
 
 USER root
 
-ARG NODE_VERSION=18
+ARG NODE_VERSION=22
```
yassin-kammoun-sonarsource, line 6: Let's double-check with the rest of the squad whether we still want to test SonarJS against Node.js 18 on CI.

yassin-kammoun-sonarsource (commented):
Hey @ericmorand-sonarsource, as mentioned during the standup, deprecating the support of a Node.js version requires also a bit of work on the Java side. In particular, we need to update the following:

https://github.com/SonarSource/SonarJS/blob/ac71cb1a9e18dd9f78dc4c60ddb9f86eb4cdb6f1/sonar-plugin/bridge/src/main/java/org/sonar/plugins/javascript/bridge/NodeDeprecationWarning.java#L40-L44

Maybe we should document the process of deprecating a Node.js version to avoid tribal knowledge. Not sure where to write this down...

package.json
ericmorand-sonarsource: This change establishes that the project needs Node 22 to be operated: you can't run the unit tests without at least Node 22.

ericmorand-sonarsource:
@yassin-kammoun-sonarsource , issues should be fixed:

* NodeDeprecationWarning was updated
* The formatting issues on documentation were fixed

PR #4917, opened 2024-11-20
JS-421 Fix new-rule script

kebetsi:
[JS-421](https://sonarsource.atlassian.net/browse/JS-421)

[JS-421]: https://sonarsource.atlassian.net/browse/JS-421?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

* Refactored the scripts inside `tools` folder to make use of the same helpers.
  * generate-meta.ts
  * new-rule.ts
  * generate-external-rule-docs.ts
  * generate-rule-indexes.ts
* The new rule will prompt the user for all needed input (using inquirer.js)
* Java class will now implement the correct parent class depending if it's a rule for MAIN or TEST
* All Java check classes have been renamed to use the Sonar rule id. This makes it much easier to synchronize both sides
* The index with all java checks is now autogenerated on `npm ci` (as we do in the node.js side) in a git ignored file (AllChecks.java)


kebetsi:
@vdiez Can you include rule parameters in the java check template, as I did create the corresponding test file?
Also, could you handle the issue of `generate-meta` updating all metas with double quotes instead of single?

vdiez:
yes, I was planning to update the java template. single-quotes is already done in the base of this PR

vdiez:
seems the code coverage is not working very well, I don't think that can be improved if it counts the class declaration

ericmorand-sonarsource:
Do we want to track generated files? I assumed that the point was to not track them.

yassin-kammoun-sonarsource (commented):
@vdiez Could you consider rewriting the git history of this pull request? It would really help to have a single commit for each aspect, as reviewing the entire pr is challenging (my web browser is struggling to keep up).

On another note, we now have the following list of tools:

- count-rules.js
- external-rules-metadata.js
- generate-external-rules-docs.ts
- generate-meta.ts
- generate-rule-indexes.ts
- generate-rules-list.js
- new-rule.mts
- prepare-ruling.js

While the name of a tool provides an idea of its purpose, most of them are missing a brief introductory comment at the top of each script. More importantly, it's not easy to tell what is the expected outcome, output, or side effects. Ideally, creating a README.md file in this folder could help provide more context about the general workflow of how rule creation works.

Overall, I feel like we are lacking comments on both older and new scripts, and there is too much tribal knowledge within these scripts.

vdiez:
@yassin-kammoun-sonarsource , I agree, but this PR was originally about the new-rule and it expanded well above its scope, refactoring and improving other scripts as well. I can take care of all the rest as you propose and add documentation.

About git history, I'll do that as well.

@ericmorand-sonarsource good point, I will ignore all generated-meta.ts files as well and add it to scripts/prepare

vdiez:
Cleaned up git history and added generated-meta.ts to git ignore

I will remove unused scripts in another PR:
* external-rules-metadata.js
* generate-rules-list.js

tools/generate-meta.ts
```diff
@@ -14,121 +14,16 @@
...
-    }
-  }
+/**
+ * Generate packages/jsts/src/rules/SXXXX/generated-meta.ts on each rule
+ * with data coming from the RSPEC json files. This data fills in the Rule ESLint metadata
+ * as well as the JSON schema files available in
+ * "sonar-plugin/javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/schemas"
+ */
+for (const file of await listRulesDir()) {
+  await generateMetaForRule(file);
```
zglicz, line 26: not sure what the proper art is, but usually you shouldn't have await in a for loop. Better to hand it off to the language to manage these promises. As in:
```suggestion
Promise.all((await listRulesDir()).map( async (file) => await generateMetaForRule(file)));
```

Come to think of it, we didn't have many rules/lints in PHP/Hack, but not having await's in loops was one of them. Usually as you were looping though friends lists or photos to retrieve some data, as its a slow operation (DB reads), having it in an abstract map was one of the rules.

tools/generate-rule-indexes.ts
```diff
@@ -14,82 +14,76 @@
...
+await updateIndexes();
+
+/**
+ * Updates the following rules indexes, which are autogenerated and
+ * should NOT be edited manually
  *
- * This program is free software; you can redistribute it and/or
- * modify it under the terms of the Sonar Source-Available License Version 1, as published by SonarSource SA.
+ * Java index:
+ * sonar-plugin/javascript-checks/src/main/java/org/sonar/javascript/checks/AllChecks.java
```
zglicz, line 37: awesome! forgetting updating this was then always failing the test AllChecksTest. Can we then remove it?

tools/helpers.ts
```diff
@@ -0,0 +1,239 @@
...
+  'src',
+  'main',
+  'resources',
+  'org',
+  'sonar',
+  'l10n',
+  'javascript',
+  'rules',
+  'javascript',
+);
```
zglicz, line 43: 🙄

tools/header.ts
```diff
@@ -0,0 +1,16 @@
+/*
```
zglicz, line 1: why does this have a .ts extension? Not sure what other extension would be better, but .ts isn't good at all. If someone starts writing Typescript code in there, wouldn't be good ...

tools/helpers.ts
```diff
@@ -0,0 +1,239 @@
...
+  tags: string[];
+};
+
+const sonarWayProfile = JSON.parse(
+  await readFile(join(METADATA_FOLDER, `Sonar_way_profile.json`), 'utf-8'),
+);
+
+// Array sorter for Sonar rule IDs
+const getInt = (sonarKey: string) => parseInt(/^S(\d+)/.exec(sonarKey)[1]);
+const sonarKeySorter = (a: string, b: string) => getInt(a) - getInt(b);
```
zglicz, line 67: best sorter ever

tools/helpers.ts
```diff
@@ -0,0 +1,239 @@
...
+  return str.replace(/([.*+?^=!:${}()|\[\]\/\\])/g, '\\$1');
+}
+
+/**
+ * Inflate string template with given dictionary
+ * @param text template string
+ * @param dictionary object with the keys to replace
+ */
+export function inflateTemplate(text: string, dictionary: { [x: string]: string }): string {
+  for (const tok in dictionary) {
```
zglicz, line 93: what is `tok`? Do you mean `key`? Theory of knowledge? https://www.google.com/search?q=tok+meaning

tools/helpers.ts
```diff
@@ -0,0 +1,239 @@
...
+}
+
+export function escapeRegExp(str: string) {
+  return str.replace(/([.*+?^=!:${}()|\[\]\/\\])/g, '\\$1');
+}
+
+/**
+ * Inflate string template with given dictionary
+ * @param text template string
+ * @param dictionary object with the keys to replace
```
zglicz, line 90: keys are regexp expressions?

tools/helpers.ts
```diff
@@ -0,0 +1,239 @@
...
+ */
+export function inflateTemplate(text: string, dictionary: { [x: string]: string }): string {
+  for (const tok in dictionary) {
+    text = text.replace(new RegExp(escapeRegExp(tok), 'g'), dictionary[tok]);
```
zglicz, line 94: Does it have to conform to a regexp? Would this work without the wrapping in regex?

yassin-kammoun-sonarsource (approved):
All good on my side. Thank you for the comments.
This will really make easy getting rid of the Java classes.

tools/helpers.ts
```diff
@@ -0,0 +1,239 @@
...
+  dict: { [x: string]: string },
+) {
+  const template = await readFile(templatePath, 'utf8');
+  await writePrettyFile(dest, inflateTemplate(template, dict));
+}
+
+/**
+ * From the RSPEC json file, creates a generated-meta.ts file with ESLint formatted metadata
+ *
+ * @param sonarKey rule ID for which we need to create the generated-meta.ts file
```
zglicz, line 119: sonarKey is the SXXX format? Might be nice to add an example in the code
```suggestion
 * @param sonarKey rule ID ("SXXX") for which we need to create the generated-meta.ts file
```

tools/helpers.ts
```diff
@@ -0,0 +1,239 @@
...
+    console.log(`Type not found for rule ${sonarKey}`);
+  }
+
+  const schemaFile = join(METADATA_FOLDER, 'schemas', `${sonarKey}-schema.json`);
+  let schema = '';
+  if (await exists(schemaFile)) {
+    try {
+      schema = `\nimport type { JSONSchema4 } from '@typescript-eslint/utils/json-schema';\nexport const schema = ( ${await readFile(schemaFile, 'utf-8')} ) as const satisfies JSONSchema4;`;
+    } catch {}
+  }
```
zglicz, line 151: wow, impressive one can do an await in `` (backticks) throwing an exception that you swallow..

tools/helpers.ts
```diff
@@ -0,0 +1,239 @@
...
+}
+
+/**
+ * List all Java checks classes
+ */
+export async function getAllJavaChecks() {
+  const files = await readdir(javaChecksPath('main'), { withFileTypes: true });
+  return files
+    .filter(file => ruleRegex.test(file.name) && !file.isDirectory())
+    .map(file => file.name.slice(0, -5)) // remove .java extension
```
zglicz, line 198: wow, -5, nice that that works

zglicz (commented):
Some random comments from me

package.json
```diff
@@ -29,7 +29,7 @@
     "plugin:build:fast": "mvn install -DskipTests && npm run update-ruling-data",
     "pbf": "npm run plugin:build:fast",
     "td": "npm --prefix typedoc/searchable-parameters-plugin run setup && npx typedoc --options typedoc/typedoc.js",
-    "prepare": "husky install && npm run create-rule-indexes",
+    "prepare": "husky install && npm run generate-meta",
```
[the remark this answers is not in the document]
vdiez: fixed in https://github.com/SonarSource/SonarJS/pull/4969

tools/generate-meta.ts
vdiez: I don't agree here... i prefer to do sequential to not launch 1k+ read IO at the same time.

tools/generate-rule-indexes.ts
vdiez: it's on gitignore now

tools/header.ts
vdiez: I just want that the maven license update script updates the header in this file, so that the autogenerated scripts get the new header out of the box

PR #4920, opened 2024-11-21
JS-359 Create rule S6418 (`hardcoded-secrets`): Hard-coded secrets are security-sensitive (broken)

kebetsi:
Based on [the sonar-python S6418 rule](https://github.com/SonarSource/sonar-python/pull/1941).

[RSPEC PR](https://github.com/SonarSource/rspec/pull/4518)

[JS-359](https://sonarsource.atlassian.net/browse/JS-359)

- [x] Create base Rspec
- [x] Create base files
- [x] Understand algorithm
- [x] Write test cases for hardcoded secrets
- [x] write hardcoded secrets rule
- [x] Rename S2068 to HardcodedPassword
- [x] Update ruling
- [x] Refine coding example in Rspec
- [x] handle the diff in parameter numbers

### Other ticket tackling S2068
- [ ] write test cases for hardcoded passwords
- [ ] refactor hardcoded passwords rule

[JS-359]: https://sonarsource.atlassian.net/browse/JS-359?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/jsts/src/rules/S6418/unit.test.ts
kebetsi, line 1: I need to delete this file, as I was using the comment-based test file.

packages/jsts/src/rules/S6418/rule.ts
```diff
@@ -0,0 +1,206 @@
...
+  isLogicalExpression,
+  isStringLiteral,
+} from '../helpers/index.js';
+import { meta } from './meta.js';
+import { JSONSchema4 } from '@typescript-eslint/utils/json-schema';
+import { FromSchema } from 'json-schema-to-ts';
+import { TSESTree } from '@typescript-eslint/utils';
+
+const DEFAULT_SECRET_WORDS = 'api[_.-]?key,auth,credential,secret,token';
+const DEFAULT_RANDOMNESS_SENSIBILITY = 3.0;
```
kebetsi, line 35: I'm not sure if we need default values here as well, or if they are somehow generated from the Java check classes to the meta.

its/sources/jsts/custom/S6418.ts
kebetsi, line 1: I don't need that one anymore

sonar-plugin/javascript-checks/src/main/java/org/sonar/javascript/checks/HardcodedSecretsCheck.java
```diff
@@ -0,0 +1,68 @@
...
+    key = "randomness-sensibility",
+    description = "Minimum shannon entropy threshold of the secret",
+    defaultValue = DEFAULT_RANDOMNESS_SENSIBILITY
+  )
+  public String randomnessSensibility = DEFAULT_RANDOMNESS_SENSIBILITY;
+
+  @Override
+  public List<Object> configurations() {
+    return Collections.singletonList(
+      new HardcodedSecretsCheck.Config(secretWords, randomnessSensibility)
```
yassin-kammoun-sonarsource, line 54: You don't really need to prefix the inner class with the outer class because the inner class is static (Java FTW).

```suggestion
      new Config(secretWords, randomnessSensibility)
```

its/ruling/src/test/expected/jsts/Ghost/javascript-S6418.json
```diff
@@ -0,0 +1,6 @@
+{
+"Ghost:core/client/app/mirage/config.js": [
+59,
+61
+]
```
kebetsi, line 5: those are TPs:

```js
    this.post('/authentication/token', function () {
        return {
            access_token: '5JhTdKI7PpoZv4ROsFoERc6wCHALKFH5jxozwOOAErmUzWrFNARuH1q01TYTKeZkPW7FmV5MJ2fU00pg9sm4jtH3Z1LjCf8D6nNqLYCfFb2YEKyuvG7zHj4jZqSYVodN2YTCkcHv6k8oJ54QXzNTLIDMlCevkOebm5OjxGiJpafMxncm043q9u1QhdU9eee3zouGRMVVp8zkKVoo5zlGMi3zvS2XDpx7xsfk8hKHpUgd7EDDQxmMueifWv7hv6n',
            expires_in: 3600,
            refresh_token: 'XP13eDjwV5mxOcrq1jkIY9idhdvN3R1Br5vxYpYIub2P5Hdc8pdWMOGmwFyoUshiEB62JWHTl8H1kACJR18Z8aMXbnk5orG28br2kmVgtVZKqOSoiiWrQoeKTqrRV0t7ua8uY5HdDUaKpnYKyOdpagsSPn3WEj8op4vHctGL3svOWOjZhq6F2XeVPMR7YsbiwBE8fjT3VhTB3KRlBtWZd1rE0Qo2EtSplWyjGKv1liAEiL0ndQoLeeSOCH4rTP7',
            token_type: 'Bearer'
        };
    });
```

sonar-plugin/sonar-javascript-plugin/pom.xml
```diff
@@ -398,7 +398,7 @@
             <configuration>
               <rules>
                 <requireFilesSize>
-                  <minsize>70000000</minsize>
+                  <minsize>65000000</minsize>
```
kebetsi, line 401: I had to lower it when I compiled this last week.

packages/jsts/src/rules/S6418/rule.ts
yassin-kammoun-sonarsource: Until recently, we wouldn't need them, but now that we expose most of rules in the ESLint plugin, I am not really sure. In any case, it doesn't hurt to have them here.

packages/jsts/src/rules/S6418/rule.ts
```diff
@@ -0,0 +1,206 @@
...
+    };
+  },
+};
+
+function handleAssignmentExpression(
+  context: Rule.RuleContext,
+  node: TSESTree.AssignmentExpression,
+) {
+  const keySuspect = findKeySuspect(node.left);
+  const ValueSuspect = findValueSuspect(extractDefaultOperatorIfNeeded(node));
```
yassin-kammoun-sonarsource, line 106: We can use camelCase instead of PascalCase here and below.

packages/jsts/src/rules/S6418/rule.ts
```diff
@@ -0,0 +1,206 @@
...
+}
+
+function findKeySuspect(node: TSESTree.Node): string | undefined {
+  if (isIdentifier(node) && secretWordRegexps.some(pattern => pattern.test(node.name))) {
+    return node.name;
+  } else {
+    return undefined;
+  }
+}
+function findValueSuspect(node: TSESTree.Node | undefined | null): TSESTree.Node | undefined {
```
yassin-kammoun-sonarsource, line 163:
```suggestion

function findValueSuspect(node: TSESTree.Node | undefined | null): TSESTree.Node | undefined {
```

packages/jsts/src/rules/S6418/rule.ts
```diff
@@ -0,0 +1,206 @@
...
+    return undefined;
+  }
+}
+
+function valuePassesPostValidation(value: string): boolean {
+  return POSTVALIDATION_PATTERN.test(value);
+}
+
+function buildSecretWordRegexps(secretWords: string) {
+  return secretWords.split(',').map(word => new RegExp(`(${word})`, 'i'));
```
yassin-kammoun-sonarsource, line 181: We should try-catch this statement because RegExp throws a syntax error if the pattern is malformed.

```
% node
Welcome to Node.js v22.11.0.
Type ".help" for more information.
> new RegExp("[a-z")
Uncaught SyntaxError: Invalid regular expression: /[a-z/: Unterminated character class
```

Let's add test case with a malformed regular expression.

yassin-kammoun-sonarsource: Are all the other implementations also using regular expressions? It's not really safe :/

sonar-plugin/javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S6418.html
```diff
@@ -0,0 +1,92 @@
...
+</ul>
+<p>There would be a risk if you answered yes to any of those questions.</p>
+<h2>Recommended Secure Coding Practices</h2>
+<ul>
+  <li> Store the secret in a configuration file that is not pushed to the code repository. </li>
+  <li> Use your cloud provider’s service for managing secrets. </li>
+  <li> If a secret has been disclosed through the source code: revoke it and create a new one. </li>
+</ul>
+<h2>Sensitive Code Example</h2>
+<pre>
```
yassin-kammoun-sonarsource, line 26: You will need to update the RSPEC with JavaScript code snippets.

sonar-plugin/javascript-checks/src/test/java/org/sonar/javascript/checks/HardcodedPasswordCheckTest.java
```diff
@@ -24,11 +24,11 @@
 import com.google.gson.Gson;
 import org.junit.jupiter.api.Test;
 
-class HardcodedCredentialsCheckTest {
+class HardcodedPasswordCheckTest {
 
   @Test
   void configurations() {
-    HardcodedCredentialsCheck check = new HardcodedCredentialsCheck();
+    HardcodedPasswordCheck check = new HardcodedPasswordCheck();
```
yassin-kammoun-sonarsource, line 31: We can add another Java test class for the new rule asserting the default configuration and a custom one.

yassin-kammoun-sonarsource (changes requested):
Looks great overall, but there are a few things to address.

packages/jsts/src/rules/S6418/rule.ts
kebetsi: https://sonarsource.atlassian.net/browse/JS-426

yassin-kammoun-sonarsource (approved):
Great job!

packages/jsts/src/rules/S6418/rule.ts
```diff
@@ -0,0 +1,212 @@
...
+    const secretWords =
+      (context.options as FromSchema<typeof schema>)[0]?.['secretWords'] ?? DEFAULT_SECRET_WORDS;
+    secretWordRegexps = buildSecretWordRegexps(secretWords);
+    randomnessSensibility =
+      (context.options as FromSchema<typeof schema>)[0]?.['randomnessSensibility'] ??
+      DEFAULT_RANDOMNESS_SENSIBILITY;
+
+    return {
+      AssignmentExpression(node) {
+        handleAssignmentExpression(context, node as TSESTree.AssignmentExpression);
```
vdiez, line 81: why do you cast all nodes to TSESTree nodes? I see you had to change the helpers, but I don't understand the reasoning behind. Shouldn't it be enough to handle estree nodes directly?

kebetsi: maybe it's not needed anymore, I was struggling with that because some helpers required TSESTree nodes.

vdiez: I think it's better to cast only at the very end, meaning when you use a helper that requires TSEStree types. So you keep using estree types and no need to propagate your types to the helpers which do not use TSEStree

kebetsi: I adapted the code to only use estree: https://github.com/SonarSource/SonarJS/pull/4920/commits/aa0fbe5dff44abbe9267e223639d3e492b8726ad

vdiez: Thx! Can you also revert the changes on the helpers now?

kebetsi: I removed all the changes in the helpers now

PR #4921, opened 2024-11-21
JS-286 Fix license check and add missing license headers

yassin-kammoun-sonarsource:


yassin-kammoun-sonarsource:
Closing this in favor of #4922

PR #4927, opened 2024-11-26
Typedoc as a prerequisite

ericmorand-sonarsource:
Part of

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "@types/express": "4.17.21",
     "@types/functional-red-black-tree": "1.0.6",
     "@types/lodash.clone": "4.5.9",
-    "@types/node": "20.17.6",
+    "@types/node": "20.17.8",
     "@types/semver": "7.5.8",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -27,7 +27,7 @@
     "plugin:build": "mvn install && npm run update-ruling-data",
     "plugin:build:fast": "mvn install -DskipTests && npm run update-ruling-data",
     "pbf": "npm run plugin:build:fast",
-    "td": "npm --prefix typedoc/searchable-parameters-plugin run setup && npx typedoc --options typedoc/typedoc.js",
+    "td": "NODE_PATH=$(npm root -g) typedoc --plugin ./typedoc/searchable-parameters-plugin --options typedoc/typedoc.js",
```
ericmorand-sonarsource, line 30: This NODE_PATH assignment declares the global packages location to Node.js. We have a local Typedoc plugin that requires typedoc.

Note that, ideally, this plugin should be packed and installed globally too. But I wanted to keep things simple here.

PR #4928, opened 2024-11-26
JS-367 Survive runtime failures on saving malformed issues

yassin-kammoun-sonarsource:
[JS-367](https://sonarsource.atlassian.net/browse/JS-367)

[JS-367]: https://sonarsource.atlassian.net/browse/JS-367?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/analysis/AnalysisProcessor.java
+++ b/sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/analysis/AnalysisProcessor.java
@@ -189,7 +189,7 @@ public class AnalysisProcessor {
         saveIssue(issue);
       } catch (RuntimeException e) {
         LOG.warn("Failed to save issue in {} at line {}", file.uri(), issue.line());
-        LOG.warn("Reason: {}", e.getMessage());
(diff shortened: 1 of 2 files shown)


sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/analysis/AnalysisProcessor.java
```diff
@@ -185,7 +185,12 @@ private void saveIssues(List<Issue> issues) {
         file,
         issue.line()
       );
-      saveIssue(issue);
+      try {
+        saveIssue(issue);
+      } catch (RuntimeException e) {
+        LOG.warn("Failed to save issue in {} at line {}", file.uri(), issue.line());
+        LOG.warn("Reason: {}", e.getMessage());
```
saberduck, line 192: I would suggest to use the signature with `Throwable`, it will also include the stacktrace in the log
```suggestion
        LOG.warn("Exception cause", e);
```

saberduck (approved):

PR #4934, opened 2024-11-27
JS-435 Clear dependencies cache when a package.json is modified.

zglicz:
[JS-435](https://sonarsource.atlassian.net/browse/JS-435)

As it might happen when the package.json is changed, we should dirty the cache.

I've tested, building the jar, copying it into the plugins and verifying that the 2 logs I've added are being triggered. I will add tests once the integration tests for SonarIDE are available.

I've verified, that opening a file from node_modules directory, does not trigger this event:
```
Trigger: EDITOR_OPEN
[EDITOR_OPEN] 1 file(s) submitted
File 'package.json' excluded: file is excluded or ignored in project structure
```

[JS-435]: https://sonarsource.atlassian.net/browse/JS-435?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/jsts/src/analysis/analysis.ts
+++ b/packages/jsts/src/analysis/analysis.ts
@@ -49,6 +49,7 @@ export interface JsTsAnalysisInput extends AnalysisInput {
   tsConfigs?: string[];
   programId?: string;
   skipAst?: boolean;
+  shouldClearDependenciesCache?: boolean;
 }
 
 export interface ParsingError {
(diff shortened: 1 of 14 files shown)


vdiez (approved):
LGTM! just a small remark on AnalysisWithWatchProgram.java

PR #4980, opened 2024-12-02
SonarLint can configure persistent bundle path

zglicz:
Part of

### DIFF
--- a/sonar-plugin/bridge/src/test/java/org/sonar/plugins/javascript/bridge/BridgeServerImplTest.java
+++ b/sonar-plugin/bridge/src/test/java/org/sonar/plugins/javascript/bridge/BridgeServerImplTest.java
@@ -854,6 +854,11 @@ class BridgeServerImplTest {
     }
 
     @Override
+    public void setDeployLocation(Path deployLocation) {
+      // no-op for unit test
+    }
+
+    @Override
     public void deploy(Path deployLocation) {
       // no-op for unit test
     }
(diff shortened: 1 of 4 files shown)


sonar-plugin/bridge/src/test/java/org/sonar/plugins/javascript/bridge/BridgeServerImplTest.java
```diff
@@ -827,18 +827,47 @@ void should_not_deploy_runtime_if_node_force_host_is_set() throws Exception {
...
+    var config = BridgeServerConfig.fromSensorContext(context);
+    assertThatThrownBy(() -> bridgeServer.startServerLazily(config))
+      .isInstanceOf(NodeCommandException.class);
+  }
+
+  private BridgeServerImpl createBridgeServer(Bundle bundle) {
     return new BridgeServerImpl(
       builder(),
       TEST_TIMEOUT_SECONDS,
-      new TestBundle(startServerScript),
```
zglicz, line 834: this TestBundle doesn't seem to be a good idea. It just shouldn't be used, as with that, we are skipping the actual BundleImpl where there actually is some logic...

sonar-plugin/bridge/src/test/java/org/sonar/plugins/javascript/bridge/BridgeServerImplTest.java
```diff
@@ -858,6 +887,11 @@ static class TestBundle implements Bundle {
       this.startServerScript = startServerScript;
     }
 
+    @Override
+    public void setDeployLocation(Path deployLocation) {
+      // no-op for unit test
+    }
```
zglicz, line 893: this is unused in the tests, only for the interface implementation

vdiez (approved):

PR #4986, opened 2024-12-03
Fix prettier for java files

zglicz:
`npm run format` hasn't been formatting java files for quite some time. Let us explicitly invoke the `prettier-plugin-java` to make it run.


package.json
```diff
@@ -3,8 +3,8 @@
   "version": "1.0.0",
   "description": "SonarJS code analyzer",
   "scripts": {
-    "format": "prettier --write .",
-    "check-format": "prettier --list-different .",
+    "format": "prettier --write . --plugin=./node_modules/prettier-plugin-java/dist/index.js",
```
saberduck, line 6: perhaps we can create config file as described here https://github.com/jhipster/prettier-java?tab=readme-ov-file#usage instead of using CLI parameters?

package.json
```diff
@@ -146,7 +146,10 @@
     "trailingComma": "all",
     "singleQuote": true,
     "arrowParens": "avoid",
-    "endOfLine": "lf"
+    "endOfLine": "lf",
+    "plugins": [
+      "prettier-plugin-java"
+    ]
```
zglicz, line 152: with this setup, it also works.

saberduck (approved):

PR #5010, opened 2024-12-10
JS-478 Upgrade to ESLint 9

vdiez:
https://eslint.org/docs/latest/use/migrate-to-9.0.0

## topics to present
1. new ruleTester objects
	- clarify which one to use when
2. mandatory fields for test cases
	- mandatory: messages, suggestions (if provided)
	- the rest is optional
	- this has a big impact on the changes of this PR
5. how to run tests in intelliJ / VScode
	- ruletester
	- cb framework
	- other
6. schemas are strict:
	- for configurable rules, we need to provide the default values from the java check classes, otherwise, they might fail, as is done for S4275
7. init-linter: there is a comment in LinterWrapper, should we also mention it to the squad?
8. disallowAutomaticSingleRunInference parser option
	- see the comment in the code
9. formatting rules were migrated to [stylistic/eslint-plugin](https://www.npmjs.com/package/@stylistic/eslint-plugin)

### ruling
- go rule by rule
  - check if options changed and need to be added to the ruling config, as they seem to be mostly from decorated rules
- fix or open tickets for rules that have worse behaviours like the exhaustive switch case one


packages/jsts/src/linter/wrapper.ts
```diff
@@ -32,33 +35,47 @@ import { createLinterConfig } from './config/linter-config.js';
  * @param inputRules the quality profile rules, enabled rules
  * @param environments the JavaScript environments
  * @param globals the global variables
- * @param ruleBundles the bundles of rules to load in the linter
+ * @param rulesPlugins the bundles of rules to load in the linter
  * @param customRules array of rules to load in the linter
```
zglicz, line 39: these params are no longer present

tools/foo.ts
```diff
@@ -0,0 +1,36 @@
...
+    `import { describe } from 'node:test';`,
+    '',
+    `describe('${rule}', () => {`,
+  );
+  lines.push('});');
+  await writePrettyFile(join(ruleFolder, UNIT_TEST), lines.join('\n'));
+  //console.log(ruleFolder);
+}
+
+console.log(count, rules.length);
```
zglicz, line 36: Remember to remove this file

packages/jsts/tests/tools/testers/rule-tester.ts
```diff
@@ -14,10 +14,84 @@
...
+/**
+ * Rule tester for JS/TS, using @typescript-eslint parser with type-checking.
+ */
+class RuleTester extends NoTypeCheckingRuleTester {
+  constructor(options?: Linter.LanguageOptions) {
+    super(merge({}, typeCheckingLanguageOptions, options));
+  }
+}
+
+(RuleTester as any).it = it;
```
zglicz, line 95: think if we want to keep this

zglicz (commented):
leaving a few comments. It's not helping much though.

its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/PRAnalysisTest.java
```diff
@@ -325,6 +325,8 @@ private static SonarScanner getScanner(Path projectDir, String projectKey) {
 
   private static BuildResult scanWith(SonarScanner scanner) {
     var result = orchestrator.executeBuild(scanner);
+
+    System.out.println(result.getLogs());
```
saberduck, line 329: Remove sout

its/eslint8-plugin-sonarjs/package.json
```diff
@@ -5,12 +5,12 @@
...
     "cross-spawn": "7.0.6",
-    "eslint": "8.57.1",
-    "eslint-plugin-import": "2.31.0",
-    "typescript": "5.6.3",
-    "typescript-eslint": "7.18.0"
+    "eslint": "^8",
+    "eslint-plugin-sonarjs": "file:../../lib/eslint-plugin-sonarjs-0.0.0-SNAPSHOT.tgz",
+    "typescript": "^5",
+    "typescript-eslint": "^7",
+    "@typescript-eslint/parser": "^7"
```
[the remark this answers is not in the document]
kebetsi: the goal is to simulate a user's setup which would be set to the latest of a particular version, so we do not want to lock a particular one, but get the latest.
cross-spawn is used to execute the test, so it's fixed.

kebetsi (approved):

PR #5011, opened 2024-12-10
JS-493 Add 'get-telemetry' endpoint on bridge-server

zglicz:
[JS-493](https://sonarsource.atlassian.net/browse/JS-493)

Added a new endpoint 'get-telemetry'. This is assumed to be called after an analysis has occurred and the dependency cache is filled.

Tested in 2 ways:
1) with postman and verified that after a) initialize-linter and b) analyze-file, a request to 'get-telemetry' returned results
2) Added a simple unit-test, with a custom rule implementation, that verified that a simple package.json was picked up.

[JS-493]: https://sonarsource.atlassian.net/browse/JS-493?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/jsts/src/rules/helpers/package-json.ts
+++ b/packages/jsts/src/rules/helpers/package-json.ts
@@ -20,7 +20,6 @@ import { toUnixPath, stripBOM } from './files.js';
 import { Minimatch } from 'minimatch';
 import { type Filesystem, createFindUp } from './find-up.js';
 import fs from 'fs';
-import { Telemetry } from '../../../../bridge/src/request.js';
 
 export const PACKAGE_JSON = 'package.json';
 
--- a/packages/jsts/src/rules/helpers/package-json.ts
+++ b/packages/jsts/src/rules/helpers/package-json.ts
@@ -20,6 +20,7 @@ import { toUnixPath, stripBOM } from './files.js';
 import { Minimatch } from 'minimatch';
 import { type Filesystem, createFindUp } from './find-up.js';
 import fs from 'fs';
+import { Telemetry } from '../../../../bridge/src/request.js';
 
 export const PACKAGE_JSON = 'package.json';
 
--- a/packages/jsts/src/rules/helpers/package-json.ts
+++ b/packages/jsts/src/rules/helpers/package-json.ts
@@ -47,6 +47,11 @@ type Dependency = MinimatchDependency | NamedDependency;
  */
 const cache: Map<string, Set<Dependency>> = new Map();
 
+/**
+ * Returns the dependencies of the root package.json file collected in the cache.
+ * As the cache is populated lazily, it could be null in case no rule execution has touched it.
+ * This removes duplicate dependencies and keeps the last occurrence.
+ */
 export function getAllDependencies(): NamedDependency[] {
   const dependencies = [...cache.values()]
     .flatMap(dependencies => [...dependencies])
--- a/packages/jsts/src/rules/helpers/package-json.ts
+++ b/packages/jsts/src/rules/helpers/package-json.ts
@@ -30,10 +30,37 @@ const findPackageJsons = createFindUp(PACKAGE_JSON);
 
 const DefinitelyTyped = '@types/';
 
+type MinimatchDependency = {
+  name: Minimatch;
+  version?: string;
+};
+
+export type NamedDependency = {
+  name: string;
+  version?: string;
+};
+
+type Dependency = MinimatchDependency | NamedDependency;
+
 /**
  * Cache for the available dependencies by dirname.
  */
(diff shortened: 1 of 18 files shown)


saberduck (approved):

packages/jsts/tests/analysis/analyzer.test.ts
```diff
@@ -899,6 +899,50 @@ describe('analyzeJSTS', () => {
...
   });
 
+  it('should populate dependencies after analysis', async () => {
+    const baseDir = path.join(currentPath, 'fixtures', 'dependencies');
+    const linter = new Linter();
+    linter.defineRule('custom-rule-file', {
+      create(context) {
+        return {
+          CallExpression(node) {
+            // Necessarily call 'getDependencies' to populate the cache of dependencies
```
kebetsi, line 909: Does this mean that we would not receive any dependencies in the telemetry in case the analysis never populated the cache?

packages/jsts/tests/analysis/analyzer.test.ts
```diff
@@ -899,6 +899,50 @@ describe('analyzeJSTS', () => {
...
+    } as Rule.RuleModule);
+    const filePath = path.join(currentPath, 'fixtures', 'dependencies', 'index.js');
+    const sourceCode = await parseJavaScriptSourceFile(filePath);
+    linter.verify(
+      sourceCode,
+      { rules: { 'custom-rule-file': 'error' } },
+      { filename: filePath, allowInlineConfig: false },
+    );
+    const telemetry = getTelemetry();
+    const dependencies = telemetry.dependencies;
```
kebetsi, line 929:
```suggestion
    const { dependencies } = getTelemetry();
```

packages/jsts/tests/analysis/analyzer.test.ts
kebetsi: Is there a good chance that this happens?

zglicz: Yes, we will only have dependencies once the analysis populates it. Chance is low. I guess one would have to remove all of the rules that call the package-jsons.ts cache. In that case, we wouldn't populate anything for them. That's a tradeoff. This way, this call is very cheap and doesn't incur an extra time-cost.

kebetsi: and could you document this behavior somewhere, preferably in the code base like in the JS doc of [the getDependencies() function](https://github.com/SonarSource/SonarJS/pull/5011/files#diff-60ba2e4de214ee6ebf5fc93da8c6f337ff31319f0b5bceaee4a45c334380bd93R50):

```
Returns the dependencies of the root package.json file collected in the cache.

As the cache is populated lazily, it could be null in case no rule execution has touched it.
```

packages/jsts/src/rules/helpers/package-json.ts
```diff
@@ -30,10 +30,37 @@ const findPackageJsons = createFindUp(PACKAGE_JSON);
...
+
 /**
  * Cache for the available dependencies by dirname.
  */
-const cache: Map<string, Set<string | Minimatch>> = new Map();
+const cache: Map<string, Set<Dependency>> = new Map();
+
+export function getAllDependencies(): NamedDependency[] {
+  const dependencies = [...cache.values()]
+    .flatMap(dependencies => [...dependencies])
```
kebetsi, line 52: if we have duplicates, would it simply overwrite and keep the last one?

packages/jsts/src/rules/helpers/package-json.ts
```diff
@@ -30,10 +30,37 @@ const findPackageJsons = createFindUp(PACKAGE_JSON);
...
+    .filter((dependency): dependency is NamedDependency => typeof dependency.name === 'string');
+  return Object.values(
+    dependencies.reduce(
+      (result, dependency) => ({
+        ...result,
+        [dependency.name]: dependency,
+      }),
+      {},
+    ),
+  );
```
kebetsi, line 62: what is this step doing?

packages/jsts/src/rules/helpers/package-json.ts
```diff
@@ -46,9 +73,9 @@ export function getDependencies(filename: string, cwd: string) {
   const dirname = Path.dirname(toUnixPath(filename));
   const cached = cache.get(dirname);
   if (cached) {
-    return cached;
+    return new Set([...cached].map(item => item.name));
```
kebetsi, line 76: what is this doing? Just a clone or something more?

kebetsi (approved):
See my few nitpicking comments about documenting in JS doc some things, otherwise, good job!

PR #5012, opened 2024-12-10
JS-494 Sensor telemetry Java part

zglicz:
[JS-494](https://sonarsource.atlassian.net/browse/JS-494)

[JS-494]: https://sonarsource.atlassian.net/browse/JS-494?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/analysis/PluginTelemetry.java
```diff
@@ -0,0 +1,63 @@
...
+      // addTelemetryProperty is added in 10.9:
+      // https://github.com/SonarSource/sonar-plugin-api/releases/tag/10.9.0.2362
+      return;
+    }
+    try {
+      var keyMapToSave = telemetry
+        .dependencies()
+        .stream()
+        .collect(
+          Collectors.toMap(dependency -> KEY_PREFIX + dependency.name(), Dependency::version)
```
saberduck, line 54: I think it would be good if we have prefix like `javascript.dependency` , so if in the future we decide to add other kinds of keys they will be not mixed between dependencies

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/JavaScriptPlugin.java
```diff
@@ -281,7 +282,8 @@ public void define(Context context) {
         EslintRulesDefinition.class,
         TslintReportSensor.class,
         TslintRulesDefinition.class,
-        AnalysisWithProgram.class
+        AnalysisWithProgram.class,
+        PluginTelemetry.class
```
saberduck, line 286: why is it an extension? it doesn't have any DI annotations?

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/analysis/AnalysisWithProgram.java
```diff
@@ -111,6 +111,8 @@ void analyzeFiles(List<InputFile> inputFiles) throws IOException {
             )
           );
       }
+      var telemetry = bridgeServer.getTelemetry();
+      (new PluginTelemetry(context)).reportTelemetry(telemetry);
```
saberduck, line 115: parenthesis is redundant, does it format like this?

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/analysis/PluginTelemetry.java
```diff
@@ -0,0 +1,63 @@
...
+        );
+      keyMapToSave.forEach(ctx::addTelemetryProperty);
+      LOG.debug("Telemetry saved: {}", keyMapToSave);
+    } catch (UnsupportedOperationException e) {
+      // Ignore if api is not supported yet. In production, it is currently noop, but in tests it throws UnsupportedOperationException:
```
saberduck, line 59: I think this is not needed in the latest version, we could remove this https://github.com/SonarSource/sonarqube/blob/24.12.0.100206/sonar-plugin-api-impl/src/main/java/org/sonar/api/batch/sensor/internal/SensorContextTester.java#L447

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/JavaScriptPlugin.java
zglicz: No, it isn't used with DI. Is it only necessary there? I though here, we need to delcare all of the classes that are used. Is it only those classes which use DI?

PR #5017, opened 2024-12-16
JS-498 Add runtime telemetry

zglicz:
[JS-498](https://sonarsource.atlassian.net/browse/JS-498)

Adding runtime telemetry as per the *bonus* part of the MMF.
Refactoring the code a bit, so only querying the node server, if telemetry is actually enabled.

[JS-498]: https://sonarsource.atlassian.net/browse/JS-498?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sonar-plugin/bridge/src/main/java/org/sonar/plugins/javascript/bridge/BridgeServer.java
```diff
@@ -278,7 +281,7 @@ public String toString() {
 
   record TsProgramRequest(String tsConfig) {}
 
-  record TelemetryResponse(List<Dependency> dependencies) {}
+  record TelemetryNodeResponse(List<Dependency> dependencies) {}
```
saberduck, line 284: why `Node` in the name? Seems a bit redudant/not important?  (imagine we don't use Node anymore, would you want to rename this?)

sonar-plugin/bridge/src/main/java/org/sonar/plugins/javascript/bridge/BridgeServer.java
```diff
@@ -68,7 +69,9 @@ void initLinter(
 
   TsConfigFile createTsConfigFile(String content) throws IOException;
 
-  TelemetryResponse getTelemetry();
+  TelemetryNodeResponse getTelemetry();
+
+  NodeCommand command();
```
saberduck, line 74: I don't think it's very nice to expose this in the interface directly. I would prefer this remain internal to the bridge module. Perhaps we could just collect the required data in `Telemetry` object and only expose that.

saberduck (changes requested):
I think it would be better not to expose `NodeCommand` outside of the `bridge` module. We could instead just add more fields to the `Telemetry` record

sonar-plugin/bridge/src/main/java/org/sonar/plugins/javascript/bridge/BridgeServer.java
```diff
@@ -281,7 +279,9 @@ public String toString() {
 
   record TsProgramRequest(String tsConfig) {}
 
-  record TelemetryNodeResponse(List<Dependency> dependencies) {}
+  record TelemetryEslintBridgeResponse(List<Dependency> dependencies) {}
+
+  record TelemetryData(List<Dependency> dependencies, Map<String, String> runtime) {}
```
saberduck, line 284: I think it Java it would be nicer to have proper `record` for runtime info instead of generic `Map<String, String>`

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/analysis/AnalysisWithProgram.java
```diff
@@ -111,8 +111,7 @@ void analyzeFiles(List<InputFile> inputFiles) throws IOException {
             )
           );
       }
-      var telemetry = bridgeServer.getTelemetry();
-      new PluginTelemetry(context).reportTelemetry(telemetry);
+      new PluginTelemetry(context, bridgeServer).reportTelemetry();
```
saberduck, line 114: optional - I would only pass value from `bridgeServer.getTelemtry()` instead of passing the whole `bridgeServer`

saberduck (approved):

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/analysis/AnalysisWithProgram.java
zglicz: This way, if telemetry is not available or not necessary to compute for some reason, we wouldn't be computing it.

PR #5051, opened 2025-01-06
JS-20 - Fix FP S6754 (`hook-use-state`): Add exception when using `us…

ericmorand-sonarsource:
…eState` for components


yassin-kammoun-sonarsource (approved):
LGTM!

PR #5075, opened 2025-01-23
JS-526 Support ESLint configuration comment syntax

vdiez:


patches/eslint+9.18.0.patch
```diff
@@ -0,0 +1,34 @@
...
++                            const rule = (providedOptions.getRule ?? getRuleFromConfig)(ruleId, config);
+                             const ruleValue = inlineConfig.rules[ruleId];
+ 
+                             if (!rule) {
+@@ -1896,11 +1896,19 @@ class Linter {
+         const commentDirectives = options.allowInlineConfig && !options.warnInlineConfig
+             ? getDirectiveCommentsForFlatConfig(
+                 sourceCode,
+-                ruleId => getRuleFromConfig(ruleId, config),
++                ruleId => (providedOptions.getRule ?? getRuleFromConfig)(ruleId, config),
```
ericmorand-sonarsource, line 19: Just a comment, because I love this pattern: it would throws a TypeError is `providedOptions` happens to be a class instance.

Classes are evil. :D

ericmorand-sonarsource:
```
import assert from "node:assert";

class Foo {
  #name = 'foo';
  
  name() {
    return this.#name; 
  }
}

const foo2 = {
  _name: 'foo',
  name() {
    return this._name;
  }
}

const foo = new Foo();

assert(foo.name() === 'foo'); // works
assert(foo2.name() === 'foo'); // works

try {
  assert((foo.name || (() => 'unknown'))() === 'foo');
}
catch (exception) {
  assert(exception.message === 'Cannot read properties of undefined (reading \'#name\')');
}

try {
  assert((foo2.name || (() => 'unknown'))() === 'foo');
}
catch (exception) {
  assert(exception.message === 'Cannot read properties of undefined (reading \'_name\')');
}
```

tools/generate-rule-indexes.ts
```diff
@@ -65,6 +65,17 @@ export async function updateIndexes() {
     },
   );
 
+  await inflateTemplateToFile(
```
ericmorand-sonarsource, line 68: For my own understanding, is it needed, or is it to widen our API by exposing the metas?

vdiez: this is to easily expose the metas (and access the eslint ids) outside of the rules folder. This will still not be included in the eslint-plugin-sonarjs

patches/eslint+9.18.0.patch
vdiez: good point, do you think it's a possibility we need to consider? I would say the user providing the options will need to wrap the class in a helper function to keep `this` value

ericmorand-sonarsource: I don't think we should change anything to your code. I like it this way - I would have written it exactly the same. My comment was merely a way to re-affirm my dislike for the `this` pattern. :laughing:

ericmorand-sonarsource (approved):
Looks good to me. I did not check each ITS expectation change, but the few ones that I checked are legit.

PR #5078, opened 2025-01-24
BUILD-7042: fix maven cache fingerprint script

hedinasr:
[BUILD-7042](https://sonarsource.atlassian.net/browse/BUILD-7042)

Fix maven cache fingerprint script.

[BUILD-7042]: https://sonarsource.atlassian.net/browse/BUILD-7042?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -29,7 +29,13 @@ container_definition: &CONTAINER_DEFINITION
 maven_cache_definition: &MAVEN_CACHE
   maven_cache:
     folder: ${CIRRUS_WORKING_DIR}/.m2/repository
-    fingerprint_script: cat **/pom.xml
+    fingerprint_script: |
+      if [ "$CIRRUS_OS" = "windows" ]; then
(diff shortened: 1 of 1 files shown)


.cirrus.yml
```diff
@@ -29,7 +29,13 @@ container_definition: &CONTAINER_DEFINITION
 maven_cache_definition: &MAVEN_CACHE
   maven_cache:
     folder: ${CIRRUS_WORKING_DIR}/.m2/repository
-    fingerprint_script: cat **/pom.xml
+    fingerprint_script: |
+      if [ "$CIRRUS_OS" = "windows" ]; then
```
saberduck, line 33: is this really required when shell is configured to be `bash` above with `CIRRUS_SHELL`?

hedinasr: The thing is that the `find` command does not work in Windows. That's why I had to use the Get-ChildItem instead.

PR #5138, opened 2025-02-26
JS-571 Create config.ts files to store the initial configuration of rules

zglicz:
[JS-571](https://sonarsource.atlassian.net/browse/JS-571)

Different approach than before, now, we will have the following:
1) If a rule doesn't override any options/properties/configurations - no change
2) If a rule provides options, we will have a `config.ts` file
   --- If a property can be overwritten by SQ, i.e. it is a @RuleProperty, it will include a "description"
   --- all properties that we want to set, need to provide a 'default'

We will have access to these defaults in the Linter.initialize() method, so that we can enrich the data coming from Java/SQ. Also, in the Java classes, we will only have properties that possible to configure in SQ, no more defaults in the Java Check classes :)

Based on these, I'm quite confident we can also construct the CheckClasses

I will remove the generator from the PR, but that was just to show how the config.ts files were generated. Now, they've needed a bit of manual changes to add the descriptions and defaults and can now be removed.

[JS-571]: https://sonarsource.atlassian.net/browse/JS-571?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vdiez:
for next time, I think it's better to rebase

vdiez (approved):

PR #5142, opened 2025-02-26
JS-571 Add extending the RuleConfig.options with the default configurations

zglicz:
[JS-571](https://sonarsource.atlassian.net/browse/JS-571)

Keeping this as a small PR, so that it's digestible. Here, we augment the method `extendRuleConfig` to accept the `ESLintConfiguration` object and update the options. To note, the `RuleConfig.configuration` takes priority and overrides what is in the default `ESLintConfiguration`.

[JS-571]: https://sonarsource.atlassian.net/browse/JS-571?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/jsts/src/linter/config/rule-config.ts
+++ b/packages/jsts/src/linter/config/rule-config.ts
@@ -20,6 +20,8 @@ import { FileType } from '../../../../shared/src/helpers/files.js';
 import { JsTsLanguage } from '../../../../shared/src/helpers/language.js';
 import type { JSONSchema4 } from 'json-schema';
 import { AnalysisMode } from '../../analysis/analysis.js';
+import { defaultOptions, type ESLintConfiguration } from '../../rules/helpers/configs.js';
+import merge from 'lodash.merge';
 
 /**
  * An input rule configuration for linting
@@ -46,22 +48,27 @@ export interface RuleConfig {
  * Extends an input rule configuration
  *
  * A rule configuration might be extended depending on the rule definition.
- * The purpose of the extension is to activate additional features during
- * linting, e.g., secondary locations.
+ * Primarily, this includes adding the default options stored in `schema.json` files in the rule directory
+ * Also, it allows the extension is to activate additional features during linting, e.g., secondary locations.
  *
  * _A rule extension only applies to rules whose implementation is available._
  *
  * @param schema the rule schema
  * @param inputRule the rule configuration
  * @param workDir the working directory used by rules using the 'sonar-context' flag
+ * @param defaultConfiguration possibly undefined ESLint default options configuration added to the
+ *  provided inputRule.configuration
  * @returns the extended rule configuration
  */
 export function extendRuleConfig(
   schema: JSONSchema4 | JSONSchema4[] | undefined,
   inputRule: RuleConfig,
   workDir?: string,
+  defaultConfiguration?: ESLintConfiguration,
 ) {
-  const options = [...inputRule.configurations];
+  const options = Object.values(
+    merge(defaultOptions(defaultConfiguration), [...inputRule.configurations]),
+  );
   if (hasSonarRuntimeOption(schema)) {
     options.push(SONAR_RUNTIME);
   }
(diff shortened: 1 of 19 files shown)


packages/jsts/tests/linter/config/rule-config.test.ts
```diff
@@ -63,4 +64,53 @@ describe('extendRuleConfig', () => {
...
+      { default: 100, type: 'integer' },
+      [
+        {
+          field: 'format',
+          type: 'string',
+          default: 'allow',
+        },
+      ],
+    ];
+    const config = extendRuleConfig(
```
vdiez, line 86: this tests shows that schema is not detached from the passed options, so there is a possibility that defaults and ruleConfig are not in sync with the schema. do we want to fail? it may happen that the external rule has a different schema now

packages/jsts/src/linter/config/rule-config.ts
```diff
@@ -46,22 +48,27 @@ export interface RuleConfig {
...
  */
 export function extendRuleConfig(
   schema: JSONSchema4 | JSONSchema4[] | undefined,
   inputRule: RuleConfig,
   workDir?: string,
+  defaultConfiguration?: ESLintConfiguration,
 ) {
-  const options = [...inputRule.configurations];
+  const options = Object.values(
+    merge(defaultOptions(defaultConfiguration), [...inputRule.configurations]),
```
vdiez, line 70: do we want to allow rule configurations if the defaults don't exist? right now, we are allowing it if i'm not wrong

vdiez (approved):
looking good, just some existential doubts

zglicz:
Mea culpa, I wanted to make this to be another PR, but I forgot to branch out. I'll re-request review as I would consider this to be complete now on the JS side.

Now, we loading the RuleMetas in the Linter.intialize and connecting the plumbing. It has test coverage, as the tests touch S100 which has a `config.ts` file.

packages/jsts/src/linter/config/rule-config.ts
zglicz: I don't think it would do anything useful. As in, the configurations will come from the Java class, which will be generated based on the EsLintConfiguration object.

Are you thinking of adding a log or actually breaking?

packages/jsts/src/linter/linter.ts
```diff
@@ -269,6 +271,11 @@ export class Linter {
   ): ESLintLinter.RulesRecord {
     return {
       ...rules.reduce((rules, rule) => {
+        const ruleMeta = ruleMetas[rule.key as keyof typeof ruleMetas];
+        let eslintConfiguration;
+        if (ruleMeta && 'fields' in ruleMeta) {
+          eslintConfiguration = ruleMeta.fields as ESLintConfiguration;
+        }
```
zglicz, line 278: I didn't find a cleaner way to do this. `ruleMeta.esLintId` or `ruleMeta.implementation` were being inferred by Typescript, but the `fields` aren't. Perhaps its because the former, after present on all of the `meta` objects and the `fields` are only present on a subset of them.

packages/jsts/src/rules/S1105/config.ts
```diff
@@ -20,7 +20,7 @@ import { ESLintConfiguration } from '../helpers/configs.js';
 
 export const fields = [
   {
-    default: '1tbps',
+    default: '1tbs',
```
zglicz, line 23: A test actually caught this. The validator inside of eslint was throwing an error. That's nice :)

packages/jsts/tests/linter/config/rule-config.test.ts
zglicz: `schema` is only used to derive if it is in the Sonar Runtime. We could change the `schema` argument to the 2 flags: `isSonarRuntime` and `isSonarContext`

 I do believe, that afterwards, as we pass the configuration, ESLint will verify it. As it showed me in one of the tests, where an invalid option was passed. So I do think that occurs.

PR #5144, opened 2025-02-27
JS-572 Generate all of the Java check classes

zglicz:
[JS-572](https://sonarsource.atlassian.net/browse/JS-572)

As in title.

[JS-572]: https://sonarsource.atlassian.net/browse/JS-572?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


zglicz:
To further show there are no descrepancies. In the Js/TsRuling, I save all of the rule configurations for all of the linters. Here is the comparison (link will expire in 24h) - https://www.diffchecker.com/BUHyfiy7/

I add a write to file here - https://github.com/SonarSource/SonarJS/blob/cc4540287e0f9911206bdd811ab0bef9ffa252ee/packages/jsts/src/analysis/projectAnalysis/projectAnalyzer.ts#L71

The only meaningful difference is on rule S139, but I do not understand that, as the generated S139 class in java has the exact same regex.

sonar-plugin/sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/analysis/JavaScriptEslintBasedSensorTest.java
```diff
@@ -535,9 +536,10 @@ void should_have_configured_rules() {
     assertThat(rules).hasSize(3);
 
     assertThat(rules.get(0).getKey()).isEqualTo("S1192");
-    assertThat(new Gson().toJson(rules.get(0).getConfigurations())).isEqualTo(
-      "[{\"threshold\":3,\"ignoreStrings\":\"application/json\"}]"
-    );
+    @SuppressWarnings("unchecked")
+    var configuration = (Map<String, Object>) rules.get(0).getConfigurations().get(0);
+    assertThat(configuration).containsEntry("threshold", 3);
+    assertThat(configuration).containsEntry("ignoreStrings", "application/json");
```
zglicz, line 542: this came sometimes out of order, so just checking the contents of the map

sonar-plugin/sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/utils/RulesMetadataForSonarLintTest.java
```diff
@@ -82,11 +82,7 @@ void test() throws Exception {
       "      }\n" +
       "    ],\n" +
       "    \"defaultParams\": [\n" +
-      "      \"single\",\n" +
-      "      {\n" +
-      "        \"avoidEscape\": true,\n" +
-      "        \"allowTemplateLiterals\": true\n" +
-      "      }\n" +
+      "      \"single\"\n" +
```
zglicz, line 85: this only populates based on the Java Check classes and currently, we stopped writing the default configurations into the Java Check class, so this is expected

vdiez (approved):

vdiez:
* Generate diff always from JsTsRuling. compare master against this pr
* For the future: would be useful to check with SQ/SC guys about merging the property name for 1541 and remove the special case
* Another task for the future: clean up helpers. Probably organize in their own folder `helpers/`

zglicz:
Checked diff with master on JSTSRuling of the linter configs. Found discrepencies in:
- S6426 and fixing based on rspec - https://github.com/SonarSource/rspec/pull/4716
- S2486 - was incorrectly marked in SonarJS as only to run on tests, will be updates with this change

PR #5147, opened 2025-02-28
BUILD-7627: remove cleanup script

jayadeep-km-sonarsource:
Part of


pom.xml
```diff
@@ -342,4 +342,5 @@
     </profile>
   </profiles>
 
+
```
jayadeep-km-sonarsource, line 345: This is just to invalidate the existing cache. Will be removed before merge

kebetsi (approved):

jayadeep-km-sonarsource:
Thanks for the review @kebetsi ! Is this good for merge?

PR #5150, opened 2025-02-28
SONARARMOR-735 Support parsing of ImportEqualsDeclaration nodes

roberto-orlandi-sonarsource:
I have two issues about duplicated code, but according to the structure of the way the class is implemented, I would be inclined to mark them as 'won't fix'. Please let me know if I should fix them instead :)

### DIFF
--- a/sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/estree/ESTree.java
+++ b/sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/estree/ESTree.java
@@ -37,22 +37,6 @@ public class ESTree {
     Location loc();
   }
 
-  public sealed interface IdentifierOrTSQualifiedNameOrTSExternalModuleReference extends Node {}
-
-  public record TSQualifiedName(Location loc, TSQualifiedName left, Identifier right)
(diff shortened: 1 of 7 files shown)


sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/estree/ESTree.java
```diff
@@ -24,8 +24,8 @@
 
 /**
   This file is generated. Do not modify it manually. Look at tools/estree instead.
-  
-  This is !EXPERIMENTAL UNSUPPORTED INTERNAL API! It can be modified or removed without prior notice.   
+
+  This is !EXPERIMENTAL UNSUPPORTED INTERNAL API! It can be modified or removed without prior notice.
 */
```
kebetsi, line 29: can you remove this whole comment?

roberto-orlandi-sonarsource: Sure, I forgot about that!

roberto-orlandi-sonarsource: And done

kebetsi (approved):
LGTM

PR #5151, opened 2025-03-02
JS-594 Remove eslint scope

zglicz:
[JS-594](https://sonarsource.atlassian.net/browse/JS-594)

Digging into why we have an outdated version of @types/eslint@9.6, it is because the types of @types/eslint-scope requires @types/eslint which is a deprecated package - https://github.com/DefinitelyTyped/DefinitelyTyped/pull/70735 .

eslint package now publishes their own types and by removing the dependency on `eslint-scope`, we get perfectly typed `eslint` package.

[JS-594]: https://sonarsource.atlassian.net/browse/JS-594?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vdiez (approved):
Nice catch!

PR #5152, opened 2025-03-02
JS-579 Use defaultOptions on Rule.meta to store default options

zglicz:
[JS-579](https://sonarsource.atlassian.net/browse/JS-579)

This is a new feature of eslint as of 9.15 . With this, we don't need to do any magic with ruleMetas and loading. All of our rules will always ship with our defaults. This also gives us the benefit of the configurations being validated against the schema.

[JS-579]: https://sonarsource.atlassian.net/browse/JS-579?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


zglicz:
Closing in favor of https://github.com/SonarSource/SonarJS/pull/5168/checks

PR #5160, opened 2025-03-03
JS-597 Remove dependency on sonar-runtime to determine if rule has secondaries

zglicz:
[JS-597](https://sonarsource.atlassian.net/browse/JS-597)

First step to remove using 'sonar-runtime' from adding it to rules. Here, we stop depending on it to decode the rule.

We introduce a new property in the rule meta, i.e. SXXXX/meta.ts called `hasSecondaries`. This will indicate if the rule has the capability to produce secondaryLocations.

[JS-597]: https://sonarsource.atlassian.net/browse/JS-597?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


its/plugin/plugins/eslint-custom-rules-plugin/bundle/dist/rules.js
```diff
@@ -2,42 +2,15 @@ exports.rules = [
   {
     ruleId: 'customrule',
     ruleModule: {
-      meta: {
-        schema: [
-          {
-            // internal parameter for rules having secondary locations
-            enum: ['sonar-runtime'],
```
zglicz, line 9: we remove the capability for external custom rules to have secondaries. As far as I see, ucfg's don't use it, architecture doesn't use it.

its/plugin/plugins/eslint-custom-rules-plugin/src/main/java/org/sonar/samples/javascript/EslintCustomRulesPlugin.java
```diff
@@ -26,7 +27,8 @@ public void define(Context context) {
       EslintRulesBundle.class,
       CustomRulesDefinition.class,
       RuleRepository.class,
-      TsRepository.class
+      TsRepository.class,
+      AnalysisMode.class
```
zglicz, line 31: without this, locally for me it was crashing

packages/jsts/src/rules/helpers/generate-meta.ts
```diff
@@ -42,11 +48,10 @@ export function generateMeta(
   }
 
   metadata.messages.sonarRuntime = '{{sonarRuntimeData}}';
-
-  if (hasSecondaries) {
+  if (sonarMeta.hasSecondaries) {
     const sonarOptions = {
       type: 'string',
-      enum: [SONAR_RUNTIME, 'metric'], // 'metric' only used by S3776
+      enum: ['sonar-runtime', 'metric'], // 'metric' only used by S3776
```
zglicz, line 54: not sure why I changed that but in next PR, it will be fully removed.

ericmorand-sonarsource:
@zglicz,

> We introduce a new property in the rule meta, i.e. SXXXX/meta.ts called hasSecondaries. This will indicate if the rule has the capability to produce secondaryLocations.

What is the purpose of this property? If every rule has a property named `flow` or `secondaryLocations`, which is an array, doesn't it cover every possible case?

vdiez (approved):

zglicz:
@ericmorand-sonarsource - we can't easily change all of the rules to be encoded into a common format. We would need to update all of the external rules that call eslint.report directly. At least that's how I see it right now.

ericmorand-sonarsource:
@zglicz good point, thanks.

PR #5175, opened 2025-03-05
JS-586 Ruling rewrite

zglicz:
[JS-586](https://sonarsource.atlassian.net/browse/JS-586)

With the usage of default options, and not storing them in the Java classes, we would break the ruling. Therefore, we have to update where does the ruling get the default options from. It used to be a file like: `sonarlint-metadata.json`. However, thats far from perfect, or even good. We should use the metas!

[JS-586]: https://sonarsource.atlassian.net/browse/JS-586?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/ruling/tests/tools/testProject.ts
```diff
@@ -203,10 +215,37 @@ async function analyzeFiles(
...
  */
-async function loadRules() {
-  const rulesPath = join(toUnixPath(import.meta.dirname), '..', 'data', 'rules.json');
-  const rulesContent = await fs.readFile(rulesPath, 'utf8');
-  return JSON.parse(rulesContent);
+function applyRulingConfig(rule: RuleConfig) {
+  switch (rule.key) {
+    case 'S2486': {
+      // for some reason the scope is different
+      rule.fileTypeTargets = ['TEST'];
```
vdiez, line 224: wasn't that already fixed in RSPEC?

vdiez (approved):

packages/ruling/tests/tools/testProject.ts
```diff
@@ -14,8 +14,9 @@
  * You should have received a copy of the Sonar Source-Available License
  * along with this program; if not, see https://sonarsource.com/license/ssal/
  */
-import { join, basename, dirname } from 'node:path/posix';
+import path, { join, basename, dirname } from 'node:path/posix';
```
zglicz, line 17:
```suggestion
import { join, basename, dirname } from 'node:path/posix';
```

packages/ruling/tests/tools/testProject.ts
```diff
@@ -14,8 +14,9 @@
...
 import fs from 'node:fs/promises';
+import fsSync from 'fs';
```
zglicz, line 19:
```suggestion
```

tools/prepare-ruling.js
```diff
@@ -16,22 +16,6 @@
...
-const missing = findMissingData();
-if (missing.length > 0) {
-  console.error(`Missing ${missing} files`);
-  console.error(
-    'Please run "npm run update-ruling-data" before running "npm run ruling". You can also run "npm run build" or "npm run build:fast"',
-  );
-  process.exit(1);
-}
 
 const TARGET = path.join(import.meta.dirname, '..', 'its', 'sources');
```
vdiez, line 20: should this just be merged into testProject.ts and remove this file?

packages/ruling/tests/tools/testProject.ts
zglicz: I think so, but I don't want to update the ruling in this PR, small pr's ....

tools/prepare-ruling.js
zglicz: I think it should, but it's not obvious to me how to do it. Cause if it runs in parallel the sym-linking, I'm not sure.

zglicz:
Merging as the JS/TS ruling is passing

packages/ruling/tests/tools/testProject.ts
vdiez: pls remember to remove then :)

PR #5176, opened 2025-03-06
COMMUNITY-129686 - Update program.ts

ericmorand-sonarsource:
The sixth argument of `ts.parseJsonConfigFileContent` is wrongly set to the passed path. The method expect a _resolved path_. Passing a non-resolved path makes the program unable to resolve the included files.

Note that this is not the fix to COMMUNITY-129686. It is a fix required to debug the issue.

### DIFF
--- a/packages/jsts/src/program/program.ts
+++ b/packages/jsts/src/program/program.ts
@@ -100,7 +100,7 @@ export function createProgramOptions(
     {
       noEmit: true,
     },
-    tsConfig,
+    path.resolve(tsConfig),
     undefined,
     [
       {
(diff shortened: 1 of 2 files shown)


vdiez (approved):
I thought tsconfig was already the resolved path

vdiez:
I mean that in prod `tsconfig` will always be a resolved path, or that's what I thought. I saw the test, and I see that indeed it was not properly handled if the path was not resolved.
The only way to reach that function in prod is from the the endpoints `tsconfig-files` and `create-program`, and in both cases the path comes from Java as an absolute path

PR #5178, opened 2025-03-06
JS-580 Update new rule script

zglicz:
[JS-580](https://sonarsource.atlassian.net/browse/JS-580)

Cleaned up the script. Now the generated comment based test can be ran :)

[JS-580]: https://sonarsource.atlassian.net/browse/JS-580?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tools/templates/ts/meta.template
+++ b/tools/templates/ts/meta.template
@@ -2,4 +2,6 @@ ___HEADER___// https://sonarsource.github.io/rspec/#/rspec/___RULE_KEY___/javasc
 export * from './generated-meta.js';
 export const implementation = '___IMPLEMENTATION___';
 export const eslintId = '___ESLINT_ID___';
+export const scope = '___SCOPE___';
+export const languages: ('JAVASCRIPT' | 'TYPESCRIPT')[] = ___LANGUAGES___;
 ___EXTRA___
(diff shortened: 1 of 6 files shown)


tools/templates/ts/meta.template
```diff
@@ -2,4 +2,6 @@ ___HEADER___// https://sonarsource.github.io/rspec/#/rspec/___RULE_KEY___/javasc
 export * from './generated-meta.js';
 export const implementation = '___IMPLEMENTATION___';
 export const eslintId = '___ESLINT_ID___';
+export const scope = '___SCOPE___';
```
vdiez, line 5: aren't these 2 in `generated-meta.ts` already?

tools/templates/ts/rule.template
```diff
@@ -9,10 +9,9 @@ import {
 } from '../helpers/index.js';
 import estree from 'estree';
 import { FromSchema } from 'json-schema-to-ts';
-// if rule has schema, provide it in
-// sonar-plugin/javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/schemas/
-// and re-run npm run generate-meta, otherwise remove the schema import
-import { meta, schema } from './meta.js';
+// if rule has schema, provide it in a 'schema.json' file
```
vdiez, line 12: the schema was extracted into a separated file because initially the plan was to read it from Java. If I'm not mistaken, it is only used from js after all your changes, so we can probably get rid of it and we can initialize it to `undefined` in meta.ts

vdiez (commented):
Looks good, but don't understand why we change `meta.ts`, for me scope and languages should be in `generated-meta.ts`

vdiez (approved):

tools/templates/ts/meta.template
zglicz: good catch. This way, they would end up in both.

PR #5179, opened 2025-03-06
JS-587 Remove type from SQ property

zglicz:
[JS-587](https://sonarsource.atlassian.net/browse/JS-587)

This can be derived from the `property.customDefault ?? property.default`, so lets remove it

[JS-587]: https://sonarsource.atlassian.net/browse/JS-587?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tools/helpers.ts
+++ b/tools/helpers.ts
@@ -222,27 +222,28 @@ function generateBody(config: ESLintConfiguration, imports: Set<string>) {
     }
 
     const getJavaType = () => {
-      switch (property.type) {
-        case 'integer':
+      const defaultValue = property.customDefault ?? property.default;
+      switch (typeof defaultValue) {
+        case 'number':
           return 'int';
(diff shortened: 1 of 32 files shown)


tools/helpers.ts
```diff
@@ -284,7 +289,7 @@ function generateBody(config: ESLintConfiguration, imports: Set<string>) {
             return undefined;
           }
           let value: string;
-          if (namedProperty.type === 'array') {
+          if (typeof namedProperty.default === 'object') {
```
vdiez, line 292: both objects and arrays will match here. are we sure there are no defaults which are objects instead of arrays?

vdiez (approved):

PR #5180, opened 2025-03-06
JS-610 Move schema.json -> meta.ts

zglicz:
[JS-610](https://sonarsource.atlassian.net/browse/JS-610)

As we have all of the master logic of ESLint configuration now in js world, no need to keep these schema.json files around, when we can include them in our meta object to get validation and simpler generation scripts.

[JS-610]: https://sonarsource.atlassian.net/browse/JS-610?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vdiez (approved):
GREAT!

PR #5195, opened 2025-03-11
Undeprecate CustomRuleRepository

vdiez:
Part of


sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/Language.java
```diff
@@ -0,0 +1,46 @@
...
+  JAVASCRIPT("js"),
+  TYPESCRIPT("ts");
+
+  private final String lang;
+
+  Language(String language) {
+    this.lang = language;
+  }
+
+  private static final Map<String, Language> stringMap = Arrays.stream(values()).collect(
```
vdiez, line 34: I think it was you who proposed to use the literal map with the 2 entries instead of the "magic". I would say this is the right moment

zglicz (approved):

sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/Language.java
saberduck: “Any sufficiently advanced technology is indistinguishable from magic” Arthur C. Clarke :D

PR #5198, opened 2025-03-13
JS-635 - Include the RSPEC Maven Plugin to SonarJS codebase and Maven lifecycle

ericmorand-sonarsource:
This looks like an intimidating PR, but among the 900+ files that changed, nearly 900 are from the Rule Data held in `src/main/resources/org/sonar/l10n/javascript/rules/javascript`, that are now generated from RSPEC and that we unfortunately have to track to honor the requirements of the rules.sonarsource.com project.

The rest of the changes are due to the rework of the Maven lifecycle, which is now complete (i.e. we can build the plugin with one single `mvn install` command), and implemented in the correct modules (e.g. the rule data generation phases are located in the `javascript-checks` module, the bridge compilation in the `bridge` module, etc.).

There are also 2 Maven plugin dedicated to the project:

* rspec-maven-plugin: this one is temporary; it will be hosted in a dedicated project in the near future, and will eventually be published as an artifact of the RSPEC project
* sonarjs-maven-plugin: this one is here to stay; it implements the custom Maven goals that are required to build the project; it will be augmented in the future.

All in all, this is not such a big PR.


sonar-plugin/javascript-checks/pom.xml
```diff
@@ -49,4 +49,67 @@
...
+            <id>Generate the rules data</id>
+            <phase>generate-resources</phase>
+            <goals>
+              <goal>generate-rule-data</goal>
+            </goals>
+            <configuration>
+              <ruleSubdirectory>javascript</ruleSubdirectory>
+              <targetDirectory>${project.basedir}/../../resources/rule-data</targetDirectory>
+              <!--          <vcsRepositoryUrl>https://github.com/ericmorand-sonarsource/rspec.git</vcsRepositoryUrl>-->
+              <vcsBranchName>RULEAPI-836</vcsBranchName>
```
ericmorand-sonarsource, line 69: We can remove this as soon as RULEAPI-386 is merged into master.

maven-plugins/sonarjs-maven-plugin/README.md
kebetsi, line 1: what about renaming this one to illustrate what it does with something like `js-runtime-provider-maven-plugin`?

maven-plugins/sonarjs-maven-plugin/src/main/java/DownloadRuntimesMojo.java
```diff
@@ -0,0 +1,119 @@
...
+    getLog().info("Downloading " + url + " to " + destinationFile);
+
+    var inputStream = url.openStream();
+
+    ReadableByteChannel readableByteChannel = Channels.newChannel(inputStream);
+
+    var fileOutputStream = new FileOutputStream(destinationFile.toString());
+    var fileChannel = fileOutputStream.getChannel();
+
+    fileChannel.transferFrom(readableByteChannel, 0, Long.MAX_VALUE);
```
kebetsi, line 71: where is the logic to download it from artifactory first, then fallback on nodejs.org?

kebetsi:
Could you also mention that you refactored the fetching of the runtime into Java in the description?

kebetsi:
Would you consider splitting the PR into 3 commits? It's tough to follow otherwise.
1. the rspec maven plugin
2. the changes in the JSON files
3. the port of the fetch node logic

maven-plugins/sonarjs-maven-plugin/src/test/java/CompressMojoTest.java
```diff
@@ -0,0 +1,105 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+import org.tukaani.xz.XZInputStream;
+
+class CompressMojoTest {
```
kebetsi, line 31: How much effort would it be to put back these tests and get some coverage?

sonar-plugin/bridge/src/test/java/org/sonar/plugins/javascript/bridge/EmbeddedNodeTest.java
kebetsi, line 1: How is removing this test file necessary?

sonar-plugin/bridge/src/test/java/org/sonar/plugins/javascript/nodejs/NodeCommandTest.java
```diff
@@ -423,9 +423,9 @@ void test_embedded_runtime() throws Exception {
       .embeddedNode(en)
       .build();
     // For some reason, using mockProcessWrapper to test for the used command does not yield the expected result
-    var expectedCommand = Paths.get(en.binary().toString()) + " " + PATH_TO_SCRIPT;
-    assertThat(nodeCommand.toString()).isEqualTo(expectedCommand);
-    assertThat(nodeCommand.getNodeExecutableOrigin()).isEqualTo("embedded");
+    var expectedCommand = "node " + PATH_TO_SCRIPT;
+    //assertThat(nodeCommand.toString()).isEqualTo(expectedCommand);
+    //assertThat(nodeCommand.getNodeExecutableOrigin()).isEqualTo("none");
```
kebetsi, line 428: Why are you removing this logic and assertions?

sonar-plugin/javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S108.json
```diff
@@ -9,19 +9,26 @@
...
   "tags": [
     "suspicious"
   ],
+  "extra": {
+    "replacementRules": [],
+    "legacyKeys": [
+      "EmptyBlock",
+      "EmptyCompoundStatement"
+    ]
+  },
```
kebetsi, line 24: is it worth it to filter out this property?

sonar-plugin/javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S1131.html
```diff
@@ -1,4 +1,3 @@
-<p>This rule is deprecated, and will eventually be removed.</p>
```
kebetsi, line 1: can you have the plugin put these lines back?

sonar-plugin/sonar-javascript-plugin/pom.xml
```diff
@@ -104,15 +109,118 @@
...
+        <version>1</version>
+        <executions>
+          <execution>
+            <id>Download the Node.js runtimes</id>
+            <phase>generate-resources</phase>
+            <goals>
+              <goal>download-runtimes</goal>
+            </goals>
+            <configuration>
+              <!-- to preserve backward compatibility, we have to pass the redundant version argument -->
```
kebetsi, line 135: where is it also used?

maven-plugins/sonarjs-maven-plugin/src/test/java/CompressMojoTest.java
ericmorand-sonarsource: They are tools I would gladly add tests and coverage as soon as we add tests and coverage for every other tools that we maintain in the `tools` directory.

sonar-plugin/javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S1131.html
ericmorand-sonarsource: No. If they are not in RSPEC, they must not be there. You can open a PR in RSPEC to add these lines there.

sonar-plugin/sonar-javascript-plugin/pom.xml
ericmorand-sonarsource: We use this version to know if Node.js must ne re-deployed at runtime. We should have used the checksum instead, but now it is too late.

maven-plugins/sonarjs-maven-plugin/README.md
ericmorand-sonarsource: No. It is a plugin dedicated to SonarJS project, as explained in its README. It is not dedicated to providing the runtime. It is dedicated to providing whatever we need that is specific to the SonarJS project.

sonar-plugin/javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S1131.html
vdiez: this comes from rule-api. This need to be generated by our plugin:

https://github.com/SonarSource/sonar-rule-api/blob/14fe6f151016d195d51207416601ef6119f8cd7f/src/main/java/com/sonarsource/ruleapi/github/GitHubRuleMaker.java#L187

maven-plugins/sonarjs-maven-plugin/src/test/java/CompressMojoTest.java
zglicz: If the tests in the current state don't succeed due to logical issues with the setup of the project, lets remove for now and add a ticket to track this?

tools/deploy-rule-data.ts
```diff
@@ -0,0 +1,78 @@
...
+});
+
+const sonarWayRuleNames: Array<string> = [];
+
+for (const ruleName of ruleNames) {
+  for (const extension of ['json', 'html']) {
+    const fileName = `${ruleName}.${extension}`;
+
+    copyFileSync(join(sourceFolder, fileName), join(RULE_DATA_FOLDER, fileName));
+  }
```
zglicz, line 57: interesting  that you don't copy the contents of the source folder with filtering based on the `listRulesDir` to the destination folder and instead you rely only on the list rules read from the implementations.

What happens if there is a folder, you get it in ruleNames, but its lacking any of the json and html files? Does the copyFileSync file if the source is not present?

Could a safer way be:
1) load all rules
2) load all files in the generated src folder

Perform the copy, based on files from 2, filtering by rules from 1). Then, we can produce info, or fail the generation if there is any mismatch. wdyt?

tools/helpers.ts
```diff
@@ -173,12 +171,7 @@ async function exists(file: string) {
 }
 
 export async function writePrettyFile(filepath: string, contents: string) {
-  await writeFile(
-    filepath,
-    await prettier.format(contents, {
```
zglicz, line 178: why is the prettier removed? Don't we rely on it that the newly generated code looks nice from the start?

zglicz (approved):
huge, huge, huge PR . Many good things and ideas that will make development better. Lets make this green and build on top of it.

I've included a few questions/suggestions, but please don't add anything more in this one.

maven-plugins/pom.xml
```diff
@@ -0,0 +1,22 @@
...
+  </parent>
+
+  <modules>
+    <module>sonarjs-maven-plugin</module>
+    <module>rspec-maven-plugin</module>
+  </modules>
+
+  <properties>
+    <!-- sonar scanner properties -->
+    <sonar.skip>true</sonar.skip>
```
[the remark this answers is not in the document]
ericmorand-sonarsource: We don't want these sources to be analyzed by Sonar, they would be marked as non-covered, and they are tools.

We already do that for the `tools` directory, although using an opposite approach in the root `pom.xml`:

> <sonar.sources>packages/</sonar.sources>

For reference:

https://docs.sonarsource.com/sonarqube-server/9.6/analyzing-source-code/scanners/sonarscanner-for-maven/#excluding-module

.cirrus.yml
```diff
@@ -163,21 +160,19 @@ analyze_task:
...
   <<: *WINDOWS_VM_DEFINITION
   <<: *ONLY_SONARSOURCE_QA
   <<: *MAVEN_CACHE
   <<: *RUNTIME_CACHE
   <<: *NPM_CACHE
   <<: *NPMRC_SCRIPT_DEFINITION
   build_script:
-    - source cirrus-env CI
-    - npm run build
+    - source cirrus-env BUILD
```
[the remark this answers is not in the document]
ericmorand-sonarsource: I don't know. :laughing: 

@vdiez , @kebetsi , do you know?

.cirrus.yml
```diff
@@ -163,21 +160,19 @@ analyze_task:
...
-    - npm run build:cov
+    - npm run bridge:test:cov
     - PULL_REQUEST_SHA=$GIT_SHA1 regular_mvn_build_deploy_analyze -P-deploy-sonarsource,-release,-sign -Dcommercial -Dmaven.shade.skip=true -Dmaven.install.skip=true -Dmaven.deploy.skip=true -Dsonar.coverage.jacoco.xmlReportPaths=$CIRRUS_WORKING_DIR/report-aggregate/target/site/jacoco-aggregate/jacoco.xml
 
 build_win_task:
-  depends_on:
-    - build
   <<: *WINDOWS_VM_DEFINITION
   <<: *ONLY_SONARSOURCE_QA
   <<: *MAVEN_CACHE
```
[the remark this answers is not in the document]
ericmorand-sonarsource: I think we don't need them. I didn't want to remove them before perfectly understanding the build lifecycle that we were using, but now I assume the cache is not needed.

kebetsi (dismissed):
I would like to request porting the removed tests about the embeddednode class. I cannot approve simply removing them because it slows down the delivery.

Put this on pause for the hardening sprint if necessary.

sonar-plugin/sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/JavaScriptProfilesDefinitionTest.java
```diff
@@ -87,7 +87,7 @@ void sonar_way_js() {
     assertThat(profile.rules()).extracting("repoKey").containsOnly(CheckList.JS_REPOSITORY_KEY);
     assertThat(profile.rules().size()).isGreaterThan(100);
 
-    assertThat(deprecatedRulesInProfile(profile, deprecatedJsRules)).isEmpty();
+    assertThat(deprecatedRulesInProfile(profile, deprecatedJsRules)).hasSize(1);
```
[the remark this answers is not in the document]
ericmorand-sonarsource: This is because rule S2310 is in the Sonar Way profile and is now correctly marked as deprecated, as expected. For some reason, it was never updated in the repository even though it was deprecated in RSPEC.

ericmorand-sonarsource:
@kebetsi please create a ticket, so that we can take care of this during the hardening if it seems important.

tools/helpers.ts
ericmorand-sonarsource: I put it back.

sonar-plugin/sonar-javascript-plugin/pom.xml
ericmorand-sonarsource: Good catch. Done.

sonar-plugin/sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/utils/ProgressReportTest.java
```diff
@@ -111,43 +111,44 @@ void testCancel() throws InterruptedException {
...
-    });
-    t.start();
-    t.interrupt();
-    t.join(1000);
-    assertThat(interruptFlagPreserved.get()).isTrue();
-
-    assertThat(messages).isNotEmpty().contains("1/1" + " source file has been analyzed");
-  }
+  // todo: fails randomly, needs to be fixed ot removed
+  //  @Test
```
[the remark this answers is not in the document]
ericmorand-sonarsource: Done.

https://sonarsource.atlassian.net/browse/JS-649

sonar-plugin/bridge/src/test/java/org/sonar/plugins/javascript/nodejs/NodeCommandTest.java
ericmorand-sonarsource: It was restored to its initial state.

maven-plugins/sonarjs-maven-plugin/src/test/java/CompressMojoTest.java
ericmorand-sonarsource: I removed the test entirely.

As I said, there is no reason why we would add some tests to these tools and not to the scripts that are hosted in `tools`. Also, if the functional tests pass, then the tools that operate the projects are de-facto doing their job perfectly.

maven-plugins/sonarjs-maven-plugin/src/main/java/DownloadRuntimesMojo.java
kebetsi: This is not adressed

.cirrus.yml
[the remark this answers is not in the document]
kebetsi: why change it if you have no clue?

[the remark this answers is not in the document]
ericmorand-sonarsource: For consistency with the build task. There is no point at testing the windows build if it is not in the exact same context has the non-windows build, platform excluded.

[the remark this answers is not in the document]
ericmorand-sonarsource: @kebetsi actually you are the one that knows the meaning of these values since you changed it in the build task:

https://github.com/SonarSource/SonarJS/commit/eabc8948e2fd0a2577c7a0090bfef21d747b2601#diff-62587956f943bb2503db7bc6dd27d0d888074a1c0ecaab3f570ad611aff0f7bb

Please, can you explain what they mean?

tools/deploy-rule-data.ts
ericmorand-sonarsource: My motive was to have the build fail if a rule is implemented (i.e. is returned by `listRulesDir()`) and does not have an RSPEC entry, so that we detect that kind of discrepancy early on.

From my understanding, your proposal would do the opposite.

maven-plugins/sonarjs-maven-plugin/src/main/java/DownloadRuntimesMojo.java
ericmorand-sonarsource: This has been addressed two days ago.

sonar-plugin/sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/JavaScriptProfilesDefinitionTest.java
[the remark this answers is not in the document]
vdiez: keeping track of this: https://sonarsource.atlassian.net/browse/JS-651

vdiez (approved):
LGTM!

tools/deploy-rule-data.ts
ericmorand-sonarsource: Oh, great. Can we discuss this tomorrow? And we create a ticket to improve the situation using your proposal. I'm all in for more safety checks at built-time.

PR #5207, opened 2025-03-17
JS-642 Register sonar-architecture rules in default Sonar-Way quality profile

kebetsi:
[JS-642](https://sonarsource.atlassian.net/browse/JS-642)


kebetsi:
fixed in https://github.com/SonarSource/SonarJS/pull/5228

PR #5211, opened 2025-03-18
JS-648 Load external architecture rules in the Sonar way using reflection

kebetsi:
[JS-648](https://sonarsource.atlassian.net/browse/JS-648)


kebetsi:
fixed in https://github.com/SonarSource/SonarJS/pull/5228

PR #5214, opened 2025-03-19
BUILD-7754: Use monthly cache for orchestrator

jayadeep-km-sonarsource:
[BUILD-7754](https://sonarsource.atlassian.net/browse/BUILD-7754)

Part of BUILD-7754: Use monthly cache for orchestrator

This PR changes the cache logic to invalidate only on a monthly basis for tasks which rely on latest releases. For the task which uses `dev` version, it will continue to use the existing logic to invalidate daily

[BUILD-7754]: https://sonarsource.atlassian.net/browse/BUILD-7754?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -50,19 +50,6 @@ orchestrator_cache_preparation_definition_monthly: &ORCHESTRATOR_CACHE_MONTHLY
     fingerprint_script: echo ${THIS_MONTH}
     reupload_on_changes: 'true'
 
-orchestrator_cache_preparation_definition: &ORCHESTRATOR_CACHE
-  set_orchestrator_home_script: |
-    export TODAY=$(date '+%Y-%m-%d')
-    echo "TODAY=${TODAY}" >> $CIRRUS_ENV
-    echo "ORCHESTRATOR_HOME=${CIRRUS_WORKING_DIR}/orchestrator/${TODAY}" >> $CIRRUS_ENV
-  mkdir_orchestrator_home_script: |
-    echo "Create dir ${ORCHESTRATOR_HOME} if needed"
-    mkdir -p ${ORCHESTRATOR_HOME}
-  orchestrator_cache:
-    folder: ${ORCHESTRATOR_HOME}
-    fingerprint_script: echo ${TODAY}
-    reupload_on_changes: 'true'
-
 nodejs_runtimes_cache_definition: &RUNTIME_CACHE
   runtime_cache:
     folder: tools/fetch-node/downloads/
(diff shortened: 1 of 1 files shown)


julien-carsique-sonarsource (approved):

ericmorand-sonarsource (changes requested):
Before we go further, isn't the orchestrator package already cached in the Maven cache?

jayadeep-km-sonarsource:
It is not. orchestrator is downloading the binary separately and saving it to the path defined in the variable `ORCHESTRATOR_HOME`. this folder is what we are caching

ericmorand-sonarsource (approved):

kebetsi:
Hey @jayadeep-km-sonarsource, is there a reason why you left `orchestrator_cache_preparation_definition`?

Wouldn't it have been easier to simply use the monthly cache everywhere, including the CSS ruling check which you left as-is?

kebetsi (changes requested):
I left a comment about why we have 2 cache definitions (monthly and daily) that are very redundant as they only differ by the invalidation frequency and the old one is only used in the CSS ruling check.

jayadeep-km-sonarsource:
Hi @kebetsi . I fixed the CSS ruling check.

The task `plugin_qa_sq_dev_task` uses `SQ_VERSION: 'DEV'` to run the ITs. Since dev releases are happening more frequently, the cache needs to be invalidated daily to avoid accumulating the cache. If it is okay to change this to `SQ_VERSION: LATEST_RELEASE`, we can use monthly caching everywhere and remove the redundant code.

kebetsi:
Oh, I understand. The IT check using the DEV version of SQS is meant to test that we won't break future releases, so I believe it's necessary.
Would you know if other analyzers test against SQS DEV in their CI pipeline?

kebetsi:
@jayadeep-km-sonarsource If we're still using the daily cache for one check, I don't see how there is less cache invalidation by introducing a monthly one for the other checks. Am I missing something?

jayadeep-km-sonarsource:
Currently all the QA task caches are invalidated daily. So every morning it has to pull the artifacts from repox to fill the cache. Since they all run in parallel, each of tasks will pull from repox and upload to cache. This is around ~1GB per task per day x 7 tasks.

By introducing monthly cache for 6 tasks, they will be able to reuse the artifacts from cache instead of hitting repox. Thus reducing the bandwidth consumption. for the 7th one it will continue to hit repox once every day, but this is already a 80%+ improvement in terms of bandwidth.

kebetsi (approved):

kebetsi:
As discussed with @jayadeep-km-sonarsource, this is a quick win, and the next step is to improve the pipeline to populate the cache once prior to running the checks in parallel, which will hit the cache.

PR #5216, opened 2025-03-19
JS-33 Use typescript-eslint rule 'await-thenable' to replace S4123 implementation

zglicz:
[JS-33](https://sonarsource.atlassian.net/browse/JS-33)

As it seems like @typescript-eslint has an implementation of this rule. Let us check how it performs.

I'm showing which unit-tests would have to be removed. They were quite contrived though.

[JS-33]: https://sonarsource.atlassian.net/browse/JS-33?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


zglicz:
Only difference with the ruling is not raising on this piece of code. Which I think is fair, as A could be an awaitable.
If you agree, I would stay with this PR instead of fixing that old rule?
```
async function withDefaultOnErrorAsync<A, B>(cb: () => A, defaultValue: B): Promise<Awaited<A> | B> {
  try {
    return await cb();
  }
  catch (_) {
    return defaultValue;
  }
}
```

ericmorand-sonarsource (approved):

kebetsi (approved):
The samples that are gone are indeed very contrived. As the rule had just an awful performance in JS: ~20% FP rate, I believe any change is worth a shot.

> As it seems like https://github.com/typescript-eslint is a copy of this rule.

What do you mean by this?

zglicz:
> The samples that are gone are indeed very contrived. As the rule had just an awful performance in JS: ~20% FP rate, I believe any change is worth a shot.
> 
> > As it seems like https://github.com/typescript-eslint is a copy of this rule.
> 
> What do you mean by this?

I've updated the summary. @typescript-eslint has an implementation of this rule

PR #5224, opened 2025-03-20
JS-654 Don't use maven profile for ITS

vdiez:
[JS-654](https://sonarsource.atlassian.net/browse/JS-654)

[JS-654]: https://sonarsource.atlassian.net/browse/JS-654?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -126,7 +126,7 @@ plugin_qa_body: &PLUGIN_QA_BODY
   qa_script:
     - source cirrus-env QA
     - source set_maven_build_version $BUILD_NUMBER
-    - mvn -f its/plugin/pom.xml -Dsonar.runtimeVersion=${SQ_VERSION} ${MVN_TEST} -B -e -V verify surefire-report:report
+    - mvn -f its/plugin/pom.xml -DskipTests=false -Dsonar.runtimeVersion=${SQ_VERSION} ${MVN_TEST} -B -e -V verify surefire-report:report
 
 plugin_qa_body_monthly: &PLUGIN_QA_BODY_MONTHLY
   depends_on:
@@ -142,7 +142,7 @@ plugin_qa_body_monthly: &PLUGIN_QA_BODY_MONTHLY
   qa_script:
     - source cirrus-env QA
     - source set_maven_build_version $BUILD_NUMBER
-    - mvn -f its/plugin/pom.xml -Dsonar.runtimeVersion=${SQ_VERSION} ${MVN_TEST} -B -e -V verify surefire-report:report
+    - mvn -f its/plugin/pom.xml -DskipTests=false -Dsonar.runtimeVersion=${SQ_VERSION} ${MVN_TEST} -B -e -V verify surefire-report:report
 
 build_task:
   eks_container:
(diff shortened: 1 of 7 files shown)


.cirrus.yml
```diff
@@ -126,7 +126,7 @@ plugin_qa_body: &PLUGIN_QA_BODY
   qa_script:
     - source cirrus-env QA
     - source set_maven_build_version $BUILD_NUMBER
-    - mvn -f its/plugin/pom.xml -Dsonar.runtimeVersion=${SQ_VERSION} ${MVN_TEST} -B -e -V verify surefire-report:report
+    - mvn -f its/plugin/pom.xml -DskipTests=false -Dsonar.runtimeVersion=${SQ_VERSION} ${MVN_TEST} -B -e -V verify surefire-report:report
```
zglicz, line 129: hm... all of these tasks will depend on the `build` task and the build task will run the tests. I do not think we should run them again in each of the qa-body

its/pom.xml
```diff
@@ -56,5 +56,14 @@
       </plugin>
     </plugins>
   </build>
+  <properties>
+    <maven.deploy.skip>true</maven.deploy.skip>
+    <skipTests>true</skipTests>
```
zglicz, line 61: hm... so you do skipTests, why say 'skipTests=false' above? I guess, in the CI, we know that we run after the build so the tests are always ran. But here, perhaps it's worthyto keep the unit-tests before the its run?

its/pom.xml
```diff
@@ -56,5 +56,14 @@
...
+    <!-- sonar scanner properties -->
+    <sonar.sources>src/main</sonar.sources>
+    <sonar.tests>src/test</sonar.tests>
+    <sonar.javascript.lcov.reportPaths/>
```
zglicz, line 65: why is this here? lcov is only produced based on the unit-test runs and these are completely unrelated

pom.xml
```diff
@@ -59,9 +59,7 @@
   <modules>
     <module>maven-plugins</module>
     <module>sonar-plugin</module>
-    <!--
-      its module is added via profile below
-    -->
+    <module>its</module>
```
zglicz, line 62: now learning of profiles, why move it in by default? Won't this cause all of our builds to have to go through these always? Otherwise, they were only invoked with `mvn build -Pits`

zglicz (commented):
Few questions for you, now that I understand maven profiles :)

its/pom.xml
vdiez: we don't want to run the tests as they are actually ITS, we just want to compile the sources in the its folder

.cirrus.yml
vdiez: here we override the skipTests=true of the pom.xml, as in qa tasks we do want to run the tests

its/pom.xml
vdiez: otherwise it will inherit from parent pom

pom.xml
vdiez: by adding this here, it's taken when running the build task to compile sources, without depending on a profile

zglicz (approved):

PR #5225, opened 2025-03-20
SONARARMOR-820 Parser should analyze files containing unsupported nodes

roberto-orlandi-sonarsource:
[SONARARMOR-820](https://sonarsource.atlassian.net/browse/SONARARMOR-820)

Part of 

[SONARARMOR-820]: https://sonarsource.atlassian.net/browse/SONARARMOR-820?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/jsts/src/analysis/analyzer.ts
+++ b/packages/jsts/src/analysis/analyzer.ts
@@ -26,7 +26,7 @@ import type { TSESTree } from '@typescript-eslint/utils';
 import { Linter } from '../linter/linter.js';
 import { build } from '../builders/build.js';
 import { APIError } from '../../../shared/src/errors/error.js';
-import { serializeInProtobuf, UNSUPPORTED_NODE_TYPES } from '../parsers/ast.js';
+import { serializeInProtobuf } from '../parsers/ast.js';
 import { SymbolHighlight } from '../linter/visitors/symbol-highlighting.js';
 import { computeMetrics, findNoSonarLines } from '../linter/visitors/metrics/index.js';
 import { getSyntaxHighlighting } from '../linter/visitors/syntax-highlighting.js';
@@ -101,17 +101,7 @@ export async function analyzeJSTS(
 
 function serializeAst(sourceCode: SourceCode, filePath: string) {
   try {
-    UNSUPPORTED_NODE_TYPES.clear();
-    const serialized = serializeInProtobuf(sourceCode.ast as TSESTree.Program);
-    if (UNSUPPORTED_NODE_TYPES.size > 0) {
-      debug(
-        `Not supported syntax nods in file "${filePath}": ` +
-          Array.from(UNSUPPORTED_NODE_TYPES)
-            .map(([key, value]) => `${key}(${value})`)
-            .join(', '),
-      );
-    }
-    return serialized;
+    return serializeInProtobuf(sourceCode.ast as TSESTree.Program, filePath);
   } catch (e) {
     info(`Failed to serialize AST for file "${filePath}"`);
     return null;
--- a/packages/jsts/src/analysis/analyzer.ts
+++ b/packages/jsts/src/analysis/analyzer.ts
@@ -26,7 +26,7 @@ import type { TSESTree } from '@typescript-eslint/utils';
 import { Linter } from '../linter/linter.js';
 import { build } from '../builders/build.js';
 import { APIError } from '../../../shared/src/errors/error.js';
-import { serializeInProtobuf } from '../parsers/ast.js';
+import { serializeInProtobuf, UNSUPPORTED_NODE_TYPES } from '../parsers/ast.js';
 import { SymbolHighlight } from '../linter/visitors/symbol-highlighting.js';
 import { computeMetrics, findNoSonarLines } from '../linter/visitors/metrics/index.js';
 import { getSyntaxHighlighting } from '../linter/visitors/syntax-highlighting.js';
@@ -101,7 +101,17 @@ export async function analyzeJSTS(
 
 function serializeAst(sourceCode: SourceCode, filePath: string) {
   try {
-    return serializeInProtobuf(sourceCode.ast as TSESTree.Program);
+    UNSUPPORTED_NODE_TYPES.clear();
+    const serialized = serializeInProtobuf(sourceCode.ast as TSESTree.Program);
+    if (UNSUPPORTED_NODE_TYPES.size > 0) {
+      debug(
+        `Not supported syntax nods in file "${filePath}": ` +
+          Array.from(UNSUPPORTED_NODE_TYPES)
+            .map(([key, value]) => `${key}(${value})`)
+            .join(', '),
+      );
+    }
+    return serialized;
   } catch (e) {
     info(`Failed to serialize AST for file "${filePath}"`);
     return null;
--- a/packages/jsts/src/parsers/ast.ts
+++ b/packages/jsts/src/parsers/ast.ts
@@ -16,7 +16,6 @@
  */
 import protobuf from 'protobufjs';
 import type { TSESTree } from '@typescript-eslint/utils';
-import { debug } from '../../../shared/src/helpers/logging.js';
 
 import path from 'path';
 import { fileURLToPath } from 'node:url';
@@ -25,6 +24,7 @@ const PATH_TO_PROTOFILE = path.join(path.dirname(fileURLToPath(import.meta.url))
 const PROTO_ROOT = protobuf.loadSync(PATH_TO_PROTOFILE);
 const NODE_TYPE = PROTO_ROOT.lookupType('Node');
 export const NODE_TYPE_ENUM = PROTO_ROOT.lookupEnum('NodeType');
+export const UNSUPPORTED_NODE_TYPES = new Map<string, number>();
 
 export function serializeInProtobuf(ast: TSESTree.Program): Uint8Array {
   const protobufAST = parseInProtobuf(ast);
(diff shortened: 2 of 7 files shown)


packages/jsts/tests/parsers/ast.test.ts
```diff
@@ -76,14 +76,6 @@ describe('ast', () => {
       }),
     );
   });
-  test('should encode unknown nodes', async () => {
```
roberto-orlandi-sonarsource, line 79: I tried to find a way to preserve this test, but I could not find an unsupported syntax to use which would not be right under the `Program` node, which means that is just ignored and not converted to an `UnknownNode`.

I also tried to change a bit the nature of the test by creating a dummy `UnknownNode` but I could not get it done (you can find it in the first commit).

Let me know if you have some better ideas.

packages/jsts/src/analysis/analyzer.ts
```diff
@@ -101,7 +101,17 @@ export async function analyzeJSTS(
 
 function serializeAst(sourceCode: SourceCode, filePath: string) {
   try {
-    return serializeInProtobuf(sourceCode.ast as TSESTree.Program);
+    UNSUPPORTED_NODE_TYPES.clear();
+    const serialized = serializeInProtobuf(sourceCode.ast as TSESTree.Program);
+    if (UNSUPPORTED_NODE_TYPES.size > 0) {
+      debug(
+        `Not supported syntax nods in file "${filePath}": ` +
```
zglicz, line 108: there is a typo here, nods -> nodes

packages/jsts/src/analysis/analyzer.ts
```diff
@@ -101,7 +101,17 @@ export async function analyzeJSTS(
 
 function serializeAst(sourceCode: SourceCode, filePath: string) {
   try {
-    return serializeInProtobuf(sourceCode.ast as TSESTree.Program);
+    UNSUPPORTED_NODE_TYPES.clear();
```
zglicz, line 104: Oh, this is global state from `ast.ts` and you are managing it's state here in `analyzer.ts`. Let's not do that. Could you push this down int the `serializeInProtobuf` and to the `ast.ts` file. And then either, return a struct with the serialized and unsupported node types, or if it's needed just for the logs, print the logs in `ast.ts`

packages/jsts/src/parsers/ast.ts
```diff
@@ -25,6 +24,7 @@ const PATH_TO_PROTOFILE = path.join(path.dirname(fileURLToPath(import.meta.url))
 const PROTO_ROOT = protobuf.loadSync(PATH_TO_PROTOFILE);
 const NODE_TYPE = PROTO_ROOT.lookupType('Node');
 export const NODE_TYPE_ENUM = PROTO_ROOT.lookupEnum('NodeType');
+export const UNSUPPORTED_NODE_TYPES = new Map<string, number>();
```
zglicz, line 27: Why is the name in all uppercase? As a convention, we use that for constants. Here, this is anything but. Its a normal variable with global state.

packages/jsts/src/parsers/ast.ts
```diff
@@ -498,8 +416,8 @@ function visitThisExpression(_node: TSESTree.ThisExpression) {
 
 function visitTemplateLiteral(node: TSESTree.TemplateLiteral) {
   return {
-    quasis: node.quasis.map(visitNode),
-    expressions: node.expressions.map(visitNode),
+    quasis: filterUnknownTypeNodes(node.quasis.map(visitNode)),
+    expressions: filterUnknownTypeNodes(node.expressions.map(visitNode)),
```
zglicz, line 420: Oh, we can't do it like this. This is adding this filtering call to each and every `visitXXX` method. Anytime someone adds a new node type, they will need to magically call this random filtering method. This is very error-prone. Any time a new node type would be added, if this is omitted, we wouldn't perform this filtering.

Why not just do that in line: 325 on the shape? This way, all of these below can be removed?

packages/jsts/src/parsers/ast.ts
```diff
@@ -310,108 +316,20 @@ function getProtobufShapeForNode(node: TSESTree.Node) {
...
-      // do not log unsupported known nodes
-      break;
     default:
-      debug(`Unknown node type: ${node.type}`);
+      UNSUPPORTED_NODE_TYPES.set(node.type, (UNSUPPORTED_NODE_TYPES.get(node.type) ?? 0) + 1);
   }
   return {
     type: NODE_TYPE_ENUM.values[node.type + 'Type'] ?? NODE_TYPE_ENUM.values['UnknownNodeType'],
     loc: node.loc,
     [lowerCaseFirstLetter(node.type)]: shape || visitUnknownNode(node),
```
zglicz, line 325: can you add the filtering here?

zglicz (changes requested):
Back to you to fix. 2 main issues:
1) Shared variable from other module. This is stateful and it's better to avoid as it's life-cycle can be managed by the callee instead.
2) Adding filtering in very-many places instead of just 1

packages/jsts/src/parsers/ast.ts
roberto-orlandi-sonarsource: It did not feel right, but I got trick by the `const` and the examples right above :D Thanks for pointing it out.

packages/jsts/src/analysis/analyzer.ts
roberto-orlandi-sonarsource: That's a fair point. I think the best solution is to add it to `ast.ts` in the method `serializeInProtobuf`. 
However, I would like to add a string parameter to specify the path so that we can print it in the log. Please let me know if you have any concerns about that.

packages/jsts/src/parsers/ast.ts
roberto-orlandi-sonarsource: I totally agree with your concern but how do we apply it at line 325? As we discussed yesterday, an unsupported node could happen at any point, and the only unknown nodes that we can aim to filter out at the moment are the ones appearing inside arrays.
In the visits, we know that we are handling arrays, while at line 325 there could be any kind of object. Or did I misunderstand your suggestion?

packages/jsts/src/parsers/ast.ts
zglicz: 2 things here:
1) Please rename the `visitUnknownNode` -> you are not visiting this node, you are just getting a representation for this unknown node. *visit* implies you are actually visiting it. Thinking of this more, why even return this object representing an unknown node if you are removing it just a second later? Just return the undefined and filter for undefineds.
2) You can santize all of the properties in the shape and only return that.

packages/jsts/src/analysis/analyzer.ts
zglicz: If this is important to serializeInProtobuf to know the filePath, then you shall pass it, no? Even if for a log. It is not related to the `analyzer.ts` whatsoever.

packages/jsts/src/parsers/ast.ts
roberto-orlandi-sonarsource: This is related to my comment above: we don't always remove it. As discussed during the call, we just filter it out of the arrays. But this method is also used to get the shape of elements that are not in an array, and doing as you suggest would risk having the effect of creating some shape with fields set to `undefined`.

Maybe we can discuss this on a call when you will be available.

zglicz: Hm.. You have removed the list of known unsupported types. Hence, all of the nodes that you don't handle in the switch, will be treated exactly the same, no?
Therefore, there is nothing special about any of the keys. You can:
1) loop through all of the keys in the `shape`
2) get the value under that key
3) if this value is an array
4) sanitize it

Would that not work?


Re: undefined <- I still understand this would work. You wouldn't put in an undefined inside of the shape. It would be the return of the `visitUnknownNode`. And then you would filter out the values that have an '...'.type === undefined. But this isn't as necessary and depends on the business logic of what you expect to have as the shape for those nodes.

roberto-orlandi-sonarsource: Oh, I see now what you mean, I totally missed the idea of iterating over the keys of shape. I will try to follow your suggestions, and see if there are any blockers. Thanks a lot!

roberto-orlandi-sonarsource: I think I implemented more or less what you were proposing, except for the fact that I don't assign any `undefined `and that I do the _sanitization_ based on the `UnknownNodeType` type.
  
Also, while before the `shape` of un `unknownNode` would be an object with a `rawType` field, now its `shape` is only an empty object.  In this way, I also got rid of the `visitUnknownNode` method.

Let me know what you think. In any case I will (hopefully) soon assign it for review. once the PR is green.

This is how the new implementation looks like: 
```
 ....
    default:
      unsupportedNodeTypes.set(node.type, (unsupportedNodeTypes.get(node.type) ?? 0) + 1);
      shape = {};
  }
  // Visiting UnknownNodeTypes can cause the failure of the ESTreeFactory. For this reason where possible (for example in the arrays) we try to remove them.
  for (const key in shape) {
    if (Array.isArray(shape[key])) {
      shape[key] = shape[key].filter(
        node => node?.type !== NODE_TYPE_ENUM.values['UnknownNodeType'],
      );
    }
  }
  return {
    type: NODE_TYPE_ENUM.values[node.type + 'Type'] ?? NODE_TYPE_ENUM.values['UnknownNodeType'],
    loc: node.loc,
    [lowerCaseFirstLetter(node.type)]: shape,
  };
```

zglicz (approved):
Thanks for applying the review comments!

roberto-orlandi-sonarsource:
> Thanks for applying the review comments!

Thank you for the super helpful review!

PR #5226, opened 2025-03-20
SONARARMOR-817 Parser should serialize TSModuleDeclaration nodes

roberto-orlandi-sonarsource:
[SONARARMOR-817](https://sonarsource.atlassian.net/browse/SONARARMOR-817)

Part of 

[SONARARMOR-817]: https://sonarsource.atlassian.net/browse/SONARARMOR-817?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/jsts/src/parsers/ast.ts
+++ b/packages/jsts/src/parsers/ast.ts
@@ -320,6 +320,12 @@ function getProtobufShapeForNode(node: TSESTree.Node) {
     case 'TSExternalModuleReference':
       shape = visitTSExternalModuleReference(node);
       break;
+    case 'TSModuleBlock':
+      shape = visitTSModuleBlock(node);
+      break;
+    case 'TSModuleDeclaration':
+      shape = visitTSModuleDeclaration(node);
+      break;
     default:
       unsupportedNodeTypes.set(node.type, (unsupportedNodeTypes.get(node.type) ?? 0) + 1);
       shape = {};
@@ -908,6 +914,20 @@ function visitTSQualifiedName(node: TSESTree.TSQualifiedName) {
 function visitTSExternalModuleReference(node: TSESTree.TSExternalModuleReference) {
   return {
     expression: visitNode(node.expression),
+  };
+}
+
+function visitTSModuleBlock(node: TSESTree.TSModuleBlock) {
+  return {
+    body: node.body.map(visitNode),
+  };
+}
+
+function visitTSModuleDeclaration(node: TSESTree.TSModuleDeclaration) {
+  return {
+    id: visitNode(node.id),
+    body: visitNode(node.body),
+    kind: node.kind,
   };
 }
 
--- a/packages/jsts/src/parsers/ast.ts
+++ b/packages/jsts/src/parsers/ast.ts
@@ -58,7 +58,8 @@ export function deserializeProtobuf(serialized: Uint8Array): any {
   return decoded;
 }
 
-type VisitNodeReturnType = {
+// Exported for testing purpose
+export type VisitNodeReturnType = {
   type: number;
   loc: TSESTree.SourceLocation;
   [key: string]: any;
--- a/packages/jsts/tests/parsers/ast.test.ts
+++ b/packages/jsts/tests/parsers/ast.test.ts
@@ -29,6 +29,7 @@ import {
   parseInProtobuf,
   serializeInProtobuf,
   visitNode,
+  VisitNodeReturnType,
 } from '../../src/parsers/ast.js';
 import { parse } from '../../src/parsers/parse.js';
 
@@ -70,9 +71,7 @@ describe('ast', () => {
         const filePath = path.join(import.meta.dirname, 'fixtures', 'ast', 'base.js');
         const sc = await parseSourceFile(filePath, parser, usingBabel);
         const protoMessage = parseInProtobuf(sc.sourceCode.ast as TSESTree.Program);
-        const serialized = serializeInProtobuf(sc.sourceCode.ast as TSESTree.Program, filePath);
-        const deserializedProtoMessage = deserializeProtobuf(serialized);
-        compareASTs(protoMessage, deserializedProtoMessage);
+        checkAstIsProperlySerializedAndDeserialized(sc.sourceCode.ast, protoMessage, filePath);
       }),
     );
   });
@@ -173,9 +172,7 @@ describe('ast', () => {
     expect(moduleReference.type).toEqual(NODE_TYPE_ENUM.values['IdentifierType']);
     expect(moduleReference.identifier.name).toEqual('foo');
 
-    const serialized = serializeInProtobuf(ast as TSESTree.Program, 'foo.ts');
-    const deserializedProtoMessage = deserializeProtobuf(serialized);
-    compareASTs(protoMessage, deserializedProtoMessage);
+    checkAstIsProperlySerializedAndDeserialized(ast as TSESTree.Program, protoMessage, 'foo.ts');
   });
 
   test('should support TSImportEquals nodes with TSQualifiedName module reference', async () => {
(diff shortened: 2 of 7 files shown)


roberto-orlandi-sonarsource:
There is one issue on Next which, according to the way the file is implemented, I am not sure I should fix. Please let me know if I can accept it.

packages/jsts/tests/parsers/ast.test.ts
```diff
@@ -275,6 +275,66 @@ describe('ast', () => {
...
+      NODE_TYPE_ENUM.values['TSModuleDeclarationType'],
+    );
+    const tSModuleDeclaration = protoMessage.program.body[0].tSModuleDeclaration;
+    expect(tSModuleDeclaration.kind).toEqual('namespace');
+    expect(tSModuleDeclaration.id.type).toEqual(NODE_TYPE_ENUM.values['IdentifierType']);
+    expect(tSModuleDeclaration.id.identifier.name).toEqual('Foo');
+    expect(tSModuleDeclaration.body.type).toEqual(NODE_TYPE_ENUM.values['TSModuleBlockType']);
+    expect(tSModuleDeclaration.body.tSModuleBlock.body[0].type).toEqual(
+      NODE_TYPE_ENUM.values['VariableDeclarationType'],
+    );
```
zglicz, line 294: what is this `tSModuleDeclaration` ? is it a class or a plain object? If its an object and you are asserting many properties, would it be more pleasant to compare it like so? With this setup, it's more visible to the reader of the code of the exact nesting and/or structure. However, as I was typing this out, it did feel a bit longish..
```suggestion
    expect(protoMessage.program.body[0].tSModuleDeclaration).toStrictEqual({ // or toEqual or toMatchObject
      kind: 'namespace',
      id: {
        type: NODE_TYPE_ENUM.values['IdentifierType'],
        identifier: {
          name: 'Foo',
        }
      },
      body: {
        type: NODE_TYPE_ENUM.values['TSModuleBlockType'],
        tSModuleBlock: {
          body: [
            {
              type: NODE_TYPE_ENUM.values['VariableDeclarationType']
            }
          ]
        }
      }
    });
```

packages/jsts/tests/parsers/ast.test.ts
```diff
@@ -275,6 +275,66 @@ describe('ast', () => {
...
+      NODE_TYPE_ENUM.values['TSModuleDeclarationType'],
+    );
+    const tSModuleDeclaration = protoMessage.program.body[0].tSModuleDeclaration;
+    expect(tSModuleDeclaration.kind).toEqual('module');
+    expect(tSModuleDeclaration.id.type).toEqual(NODE_TYPE_ENUM.values['LiteralType']);
+    expect(tSModuleDeclaration.id.literal.valueString).toEqual('Foo');
+    expect(tSModuleDeclaration.body).toEqual(undefined);
+    const serialized = serializeInProtobuf(ast as TSESTree.Program, 'foo.ts');
+    const deserializedProtoMessage = deserializeProtobuf(serialized);
+    compareASTs(protoMessage, deserializedProtoMessage);
```
zglicz, line 336: is this the same in the entire test suite? Can this be extracted someplace or at least a single call across all of these tests

sonar-plugin/api/src/test/java/org/sonar/plugins/javascript/api/estree/ESTreeTest.java
```diff
@@ -26,21 +26,21 @@ class ESTreeTest {
...
 
     //filter all classes that are interface
     var ifaceCount = Arrays.stream(classes).filter(Class::isInterface).count();
-    assertThat(ifaceCount).isEqualTo(27);
+    assertThat(ifaceCount).isEqualTo(28);
 
     var recordCount = Arrays.stream(classes).filter(Class::isRecord).count();
-    assertThat(recordCount).isEqualTo(81);
+    assertThat(recordCount).isEqualTo(83);
   }
```
zglicz, line 37: What is this testing? You add unit tests to verify the business logic and that should be enough.

Should this be removed? It's not adding value. It didn't catch that you are doing anything wrong, just added noise by making you update this.

sonar-plugin/api/src/test/java/org/sonar/plugins/javascript/api/estree/ESTreeTest.java
```diff
@@ -26,21 +26,21 @@ class ESTreeTest {
...
 
   @Test
   void test_node_subclasses() {
     Class<?> sealedClass = ESTree.Node.class;
     Class<?>[] permittedSubclasses = sealedClass.getPermittedSubclasses();
-    assertThat(permittedSubclasses).hasSize(24);
+    assertThat(permittedSubclasses).hasSize(25);
   }
```
zglicz, line 44: same question. Perhaps, I'm not understanding something about the value of these.

sonar-plugin/bridge/src/test/java/org/sonar/plugins/javascript/bridge/ESTreeFactoryTest.java
```diff
@@ -833,6 +835,67 @@ void should_create_ts_import_equals_declaration() {
...
+      .setTSModuleDeclaration(tsModuleDeclaration)
+      .build();
+
+    ESTree.Node estree = ESTreeFactory.from(protobufNode, ESTree.Node.class);
+    assertThat(estree).isInstanceOfSatisfying(ESTree.TSModuleDeclaration.class, module -> {
+      assertThat(module.id()).isInstanceOf(ESTree.Identifier.class);
+      assertThat(module.body()).isEmpty();
+      assertThat(module.kind()).isEqualTo("module");
+    });
+  }
```
zglicz, line 897: wow, this is so verbose. Almost make me think that a simpler, more maintainable way would be to have a generic node, with a field: type (for the TS/JS node name) and props (for the map of properties, which can again be primitives or further nodes) - almost like a json.

zglicz (approved):
Not much to add about the newly added functionality. It follows the same pattern as before.

It would be interesting to revisit how all of this code is generated and maintained. There are many 1000s of repetitive lines that do not contain difficult logic. These comments are outside of the scope of the PR though.

sonar-plugin/api/src/test/java/org/sonar/plugins/javascript/api/estree/ESTreeTest.java
roberto-orlandi-sonarsource: Same here. This test was just adapted.

sonar-plugin/api/src/test/java/org/sonar/plugins/javascript/api/estree/ESTreeTest.java
zglicz: Maybe lets remove?

roberto-orlandi-sonarsource: While I agree with not seeing value in those UTs, I would not make removing them part of this PR. I don't feel comfortable removing tests that I did not write, and that I don't know why they were added in the first place.

packages/jsts/src/parsers/ast.ts
```diff
@@ -911,6 +917,20 @@ function visitTSExternalModuleReference(node: TSESTree.TSExternalModuleReference
...
+}
+
+function visitTSModuleDeclaration(node: TSESTree.TSModuleDeclaration) {
+  return {
+    id: visitNode(node.id),
+    body: visitNode(node.body),
+    kind: node.kind,
+  };
+}
+
```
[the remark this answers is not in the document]
roberto-orlandi-sonarsource: I agree that this could be improved. Do you think it should be us to rethink this implementation? As far as I know, the Web squad owns and is responsible for this part of the code. I may be wrong because I was not part of the effort when this was created. In such a case, we could consider it in the future (but probably not very soon since we have a lot of other priorities at the moment)

packages/jsts/tests/parsers/ast.test.ts
roberto-orlandi-sonarsource: Good idea, I extracted those three lines to a function

roberto-orlandi-sonarsource:
> Not much to add about the newly added functionality. It follows the same pattern as before.
> 
> It would be interesting to revisit how all of this code is generated and maintained. There are many 1000s of repetitive lines that do not contain difficult logic. These comments are outside of the scope of the PR though.

Thanks for the review. As you mentioned, I think your comments are valid but probably out of the scope of this PR (at least the ones regarding bigger refactoring or changing tests not strictly related to this PR). I will merge as soon as the PR is green again, but happy to continue the open discussions.

PR #103, opened 2015-11-13
SONARJS-529 Add SQALE to S1451 and S1067 rules

vilchik-elena:

PR #5228, opened 2025-03-21
SONARCH-621 Add JS architecture rules to default quality profile

kaufco:
[SONARCH-621](https://sonarsource.atlassian.net/browse/SONARCH-621)

[SONARCH-621]: https://sonarsource.atlassian.net/browse/SONARCH-621?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ericmorand:
Sonar Architecture Ticket to support a SonarJS PR. It only took you 24 hours.

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/JavaScriptProfilesDefinition.java
```diff
@@ -51,53 +55,80 @@ public class JavaScriptProfilesDefinition implements BuiltInQualityProfilesDefin
...
   static {
-    REPO_BY_LANGUAGE.put(JavaScriptLanguage.KEY, CheckList.JS_REPOSITORY_KEY);
-    REPO_BY_LANGUAGE.put(TypeScriptLanguage.KEY, CheckList.TS_REPOSITORY_KEY);
+    REPO_BY_LANGUAGE.put(Language.JAVASCRIPT, CheckList.JS_REPOSITORY_KEY);
+    REPO_BY_LANGUAGE.put(Language.TYPESCRIPT, CheckList.TS_REPOSITORY_KEY);
+  }
+
+  private final Map<Language, ArrayList<RuleKey>> additionalRulesByLanguage;
+
+  JavaScriptProfilesDefinition(ProfileRegistrar[] profileRegistrars) {
```
kebetsi, line 67: Don't we need to provide an empty constructor anymore?

sonar-plugin/sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/JavaScriptProfilesDefinitionTest.java
```diff
@@ -84,7 +98,11 @@ void sonar_way_js() {
 
     assertThat(profile.language()).isEqualTo(JavaScriptLanguage.KEY);
     assertThat(profile.name()).isEqualTo(JavaScriptProfilesDefinition.SONAR_WAY);
-    assertThat(profile.rules()).extracting("repoKey").containsOnly(CheckList.JS_REPOSITORY_KEY);
+    assertThat(profile.rules())
+      .extracting("repoKey")
+      .containsOnly(CheckList.JS_REPOSITORY_KEY, "additionalRepository")
+      .contains(CheckList.JS_REPOSITORY_KEY)
+      .contains("additionalRepository");
```
kebetsi, line 105: what's the difference between the assertion at #103 and the ones at 104-5?

kebetsi (approved):
LGTM, I left a comment for some code I don't get, but it's for my own understanding

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/JavaScriptProfilesDefinition.java
ericmorand: So, this is your concern? Seriously?

That you **had** to provide an empty contructor to begin with did not raise any question from you? That was perfectely fine?

kaufco: Hey Ilia, that's a valid question, since in principle it could be that if DI doesn't find an injectable instance, it would expect a default constructor to call. However, I checked, and at least in SQS it manages to call the parameterized constructor with an empty array in that case, while with a default constructor it doesn't work: it always calls the default constructor when present, even if `ProfileRegistar` s are available for injection.

What we still should do (but outside the context of this PR) is to check why SonarJava requires a default constructor in that place, and why the parameterized one still works.

sonar-plugin/sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/JavaScriptProfilesDefinitionTest.java
kaufco: Good catch! 104, 105 are not needed. Fixed.

PR #5245, opened 2025-03-25
SONARARMOR-852 Fix ESTree.java hierarchy

roberto-orlandi-sonarsource:
https://sonarsource.atlassian.net/browse/SONARARMOR-852

### DIFF
--- a/sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/estree/ESTree.java
+++ b/sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/estree/ESTree.java
@@ -114,7 +114,7 @@ public class ESTree {
     Location loc,
     boolean expression,
     BlockStatementOrExpression body,
-    List<Pattern> params,
+    List<PatternOrTSParameterProperty> params,
     boolean generator,
     boolean async
   )
@@ -260,7 +260,7 @@ public class ESTree {
     Location loc,
     Optional<Identifier> id,
     BlockStatement body,
-    List<Pattern> params,
+    List<PatternOrTSParameterProperty> params,
     boolean generator,
     boolean async
   )
--- a/sonar-plugin/bridge/src/main/java/org/sonar/plugins/javascript/bridge/ESTreeFactory.java
+++ b/sonar-plugin/bridge/src/main/java/org/sonar/plugins/javascript/bridge/ESTreeFactory.java
@@ -567,7 +567,7 @@ public class ESTreeFactory {
       fromLocation(node.getLoc()),
       arrowFunctionExpression.getExpression(),
       from(arrowFunctionExpression.getBody(), ESTree.BlockStatementOrExpression.class),
-      from(arrowFunctionExpression.getParamsList(), ESTree.Pattern.class),
+      from(arrowFunctionExpression.getParamsList(), ESTree.PatternOrTSParameterProperty.class),
       arrowFunctionExpression.getGenerator(),
       arrowFunctionExpression.getAsync()
     );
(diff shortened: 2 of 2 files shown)


zglicz (commented):
Why does this need to be fixed? As in, how did you notice it? Did it compile? Is there a test missing that would show that this change is required?

roberto-orlandi-sonarsource:
All good questions. Let me go in order:

> Why does this need to be fixed?

With the previous PR we added support for the `TSParameterType`. This node differs from a "normal" parameter because comes with either the explicit `readonly` or the `accessibility`. For example:
- given `function foo(readonly bar)` then `readonly bar` is a `TSParameterType` 
- given `function foo(bar)` then `bar` is not a `TSParameterType` 

As for the "normal" parameter of the second example, the `TSParameterType` can be used by:
- `FunctionExpression`
- `FunctionDeclaration`
- `ArrowFunctionExpression`

This information can be extracted (not in a straightforward way unfortunately, it needs some "digging")  from: https://github.com/typescript-eslint/typescript-eslint/blob/6428670b94de0027189803e5a291858d2e45362d/packages/types/src/ts-estree.ts#L230

In the way I implemented the previous PR, the `TSParameterType` would be exepcted only in `FunctionExpression`. It needs to be fixed so that we can expect it also in `FunctionDeclaration` and `ArrowFunctionExpression`

> As in, how did you notice it? Did it compile?

Being the hierarchy in the `ESTree.java` completely manually defined, there is no obvious way to notice that. We define it, so it compiles. I realized this incidentally when starting the implementation of the semantics of this newly added node on the SonarJasmin  side. 

> Is there a test missing that would show that this change is required?

Being the hierarchy manually defined, I am afraid there is no test missing, we just need to be careful when we add nodes in the `ESTree.java` that is defined correctly and we link it to ALL the correct existing nodes. Said so, not having this fixed would not break anything, it would just not allow to give proper support of those nodes on the SonarJasmin engine side

zglicz:
Does that mean that the translation from the TS ast to Java AST representation was incomplete? Do we not have tests in the like: "for this piece of JS code, we expect the resulting Java AST to have this specific representation"? And there you would be able to assert that the expected Java nodes have the extended interfaces that in this case cover the union of `FunctionDeclaration` and `ArrowFunctionExpression`

roberto-orlandi-sonarsource:
> Does that mean that the translation from the TS ast to Java AST representation was incomplete? Do we not have tests in the like: "for this piece of JS code, we expect the resulting Java AST to have this specific representation"? And there you would be able to assert that the expected Java nodes have the extended interfaces that in this case cover the union of `FunctionDeclaration` and `ArrowFunctionExpression`

Exactly, the representation was incomplete. Actually the Java representation will be as complete as we define it. For example in the  TS ast, the `TSParameterProperty` has some childeren nodes that are not interesting for us so they were not translated (the [static](https://github.com/typescript-eslint/typescript-eslint/blob/1a9ab8fbc9a45b5ee738dca61f91ec815197ceef/packages/ast-spec/src/parameter/TSParameterProperty/spec.ts#L16C3-L16C9) field for example). If in the future we will need it for some reason, we will need to add it in the representation.

As far as I know we don't have that kind of tests in place, although it could be interesting to have in the future (I see something similar to what you propose is done [here](https://github.com/SonarSource/SonarJS/blob/22493d6be397bbec2a1cf068f95ed60f1b9efdfc/sonar-plugin/standalone/src/test/java/org/sonar/plugins/javascript/standalone/StandaloneParserTest.java#L43), so we may re-use this tecnique)

zglicz (approved):
Thanks for explaining!

PR #5248, opened 2025-03-26
Add default constructor for Pico compatibility

kaufco:


sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/JavaScriptProfilesDefinition.java
```diff
@@ -120,7 +127,7 @@ void activateAdditionalRules(NewBuiltInQualityProfile profile) {
       return;
     }
     rules.forEach(it -> profile.activateRule(it.repository(), it.rule()));
-    LOG.debug("Adding extra {} ruleKeys {}", language, rules);
+    LOG.info("Adding extra {} ruleKeys {}", language, rules);
```
zglicz, line 130: is this info worthy?

zglicz (commented):
As with these things? Could you summarise why this change is necessary? Could it be exposed via one of the its or in a unit test?

kaufco:
@zglicz I added details about this to the [JIRA ticket](https://sonarsource.atlassian.net/browse/JS-664) (somehow it did not link this PR automatically).

We didn't observe it failing, but
1) they had a similar issue in SonarJava about 1.5 years ago,
2) I have validated with @steve-marion-sonarsource in a local test that PicoContainer indeed needs the default constructor.

Since you only can observe it fail in SQC, there is no real way to provide a unit or plugin test for this

zglicz (approved):
I think I encountered it at some point. I'm a bit disappointed that we can't cover this with tests.

kebetsi:
@zglicz I referenced your request in https://sonarsource.atlassian.net/browse/JS-665

PR #5252, opened 2025-03-27
Update all of the rule testers to only use the required parser

zglicz:
As part of optimizing the CI, the bottleneck are the JS unit-tests. For some background. We have 3 different rule testers:
1) DefaultParserRuleTester - this is the default parser eslint ships with, it has no typescript syntax or type information
2) NoTypeCheckingRuleTest - this includes the parser from @typescript-elint, so it has typescript syntax, but no type information
3) RuleTester - the full shebang. It has typescript syntax along with type information.

Looking at the behavior of those in tests, 1 and 2 are quite quick, and 3) is much slower. On a single test in terminal the difference is: 0.8s vs 1.4s, hence 0.6s per test. With this excercise, I've dropped everything to use the DefaultParserRuleTester and checked which tests break, then verify in each of them do they need option 2) or 3).

The results, based on running 2 times each on my machine `npm run bridge:test`:
1) Before: ~200s 
2) After  : ~160s

This is another 20% gain. I hope that these will stack and we will get to around 45% improvements. I also think that with the lower load on the CPU, as the tests are lighter, we might be able to increase the concurrency to improve even more.

### DIFF
--- a/packages/jsts/src/rules/S100/unit.test.ts
+++ b/packages/jsts/src/rules/S100/unit.test.ts
@@ -15,10 +15,10 @@
  * along with this program; if not, see https://sonarsource.com/license/ssal/
  */
 import { rule } from './index.js';
-import { RuleTester } from '../../../tests/tools/testers/rule-tester.js';
+import { NoTypeCheckingRuleTester } from '../../../tests/tools/testers/rule-tester.js';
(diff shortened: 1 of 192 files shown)


kebetsi (approved):
Nice improvement

PR #5256, opened 2025-03-28
SONARCH-631 Replace delegator stub per rule with single delegator in JS frontend

kaufco:
[SONARCH-631](https://sonarsource.atlassian.net/browse/SONARCH-631)

[SONARCH-631]: https://sonarsource.atlassian.net/browse/SONARCH-631?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/EslintHook.java
+++ b/sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/EslintHook.java
@@ -21,7 +21,14 @@ import java.util.List;
 import org.sonar.api.batch.fs.InputFile;
 
 /**
- * Descriptor for a hook into the ES Linter.
+ * Descriptor for a hook into ESLint.
+ *
+ * ESLint hooks are a mechanism for other plugins to execute custom logic during
+ * the SonarJS analysis, with a parser context available.
+ * They are not associated with rule keys and hence, don't raise any Sonar issues
+ * and are executed independently of rule activation.
+ *
+ * A current use-case is the collection of data for cross-file analyzers.
  */
 public interface EslintHook {
   /**
--- a/sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/EslintHook.java
+++ b/sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/EslintHook.java
@@ -1,0 +1,38 @@
+package org.sonar.plugins.javascript.api;
+
+import java.util.Collections;
+import java.util.List;
+import org.sonar.api.batch.fs.InputFile;
+
+/**
+ * Descriptor for a hook into the ES Linter.
+ */
+public interface EslintHook {
+  /**
+   * Key for the hook to be executed on JS side.
+   */
+  String eslintKey();
+
+  default List<Object> configurations() {
+    return Collections.emptyList();
+  }
+
+  default List<InputFile.Type> targets() {
+    return List.of(InputFile.Type.MAIN);
+  }
+
+  default List<AnalysisMode> analysisModes() {
+    return List.of(AnalysisMode.DEFAULT);
+  }
+
+  default List<String> blacklistedExtensions() {
+    return Collections.emptyList();
+  }
+
+  /**
+   * Whether the hook should be executed on JS side.
+   */
+  default boolean isEnabled() {
+    return true;
+  }
+}
--- a/sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/EslintHook.java
+++ b/sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/EslintHook.java
@@ -1,3 +1,19 @@
+/*
+ * SonarQube JavaScript Plugin
+ * Copyright (C) 2011-2025 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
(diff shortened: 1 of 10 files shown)


sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/EslintHook.java
```diff
@@ -0,0 +1,54 @@
...
+ * along with this program; if not, see https://sonarsource.com/license/ssal/
+ */
+package org.sonar.plugins.javascript.api;
+
+import java.util.Collections;
+import java.util.List;
+import org.sonar.api.batch.fs.InputFile;
+
+/**
+ * Descriptor for a hook into the ES Linter.
```
kebetsi, line 24:
```suggestion
 * Descriptor for a hook into ESLint.
```

zglicz (commented):
Super-interesting @kaufco. I'm still a bit hazy how the api's we expose on the Java side can be used. Seems to me this is similar to the approach with have with the CustomRuleRepository, just a little bit slimmed down.

I added a few comments, just to make sure I understand all that is happening. Makes sense to me. Would love to have someone more experienced from the team to review if it makes holistically. Ideally, we should be able to include the `java-security` ucfg rules here as well. I'll take a look later at how you changed the `architecture/frontend/js` code to enable these. As well, I'll look into how we are currently registering the `ucfg` rule so that it works there as well. Luckily, this is very complicated so it should be "easy" to reason about :)

sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/EslintHookRegistrar.java
```diff
@@ -0,0 +1,49 @@
...
+ *       registrarContext.registerEslintHook(jsRuleKeys, Language.JAVASCRIPT);
+ *       registrarContext.registerDefaultQualityProfileRules(tsRuleKeys, Language.TYPESCRIPT);
+ *     }
+ *   }
+ * </pre>
+ */
+@ScannerSide
+public interface EslintHookRegistrar {
+  /**
+   * This method is called on the server side and during an analysis to register ES linter hooks.
```
kebetsi, line 36:
```suggestion
   * This method is called on the server side and during an analysis to register ESLint hooks.
```

sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/EslintHook.java
```diff
@@ -0,0 +1,54 @@
...
+ */
```
kebetsi, line 25:
```suggestion
  * 
  * ESLint hooks are executed during the SonarJS analysis, but they do not raise and Sonar issues.
  * This is meant to be used by other analyzers that rely on the SonarJS parser's context to produce an Intermediate Representation (IR) that they will analyze and raise issues on.
  */
```

sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/EslintHookRegistrar.java
```diff
@@ -0,0 +1,49 @@
...
+   */
+  void register(RegistrarContext registrarContext);
+
+  interface RegistrarContext {
+    /**
+     * Registers an ES linter hook to be executed on the JS side if enabled.
+     *
+     * @param language lamguage for which to execute the hook
+     * @param hook ES linter hook
+     */
```
kebetsi, line 46:
```suggestion
    /**
     * Registers an ESLint hook to be executed on the JS side if enabled.
     *
     * @param language language for which to execute the hook
     * @param hook ES linter hook
     */
```

zglicz:
On Security they make use of the CustomRuleRepository and declare these dummy EslintBasedCheck classes, that are just indicators that a `ucfg` rule should be registered, or using this nomenclature 'hooked'. As a by-product of how they use it, if there are no security rules enabled, then the key 'ucfg' won't be passed into the bridge and the IR will not be generated. I commented on your PR in sonar-architecture about this. Otherwise, this looks good to me.

sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/EslintHook.java
kaufco: Reformulated your suggestion a bit more generic. While what you write is the current use-case, the interface is for _any_ custom logic that analyzers want to inject into the SonarJS analysis phase.

sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/EslintHookRegistrar.java
```diff
@@ -0,0 +1,49 @@
...
+
+import org.sonar.api.scanner.ScannerSide;
+
+/**
+ * This class can be extended to provide ES linter hooks to be executed on the JS side during analysis.
+ *
+ * <pre>
+ *   {@code
+ *     public void register(RegistrarContext registrarContext) {
+ *       registrarContext.registerEslintHook(jsRuleKeys, Language.JAVASCRIPT);
```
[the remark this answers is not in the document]
kaufco: Nice catch! And whoops, there was even more wrong in the example: a copy and paste relic from `ProfileRegistrar`

kebetsi (approved):
LGTM

PR #5257, opened 2025-03-28
SONARARMOR-860 Parser should serialize popular syntax

roberto-orlandi-sonarsource:
[SONARARMOR-860](https://sonarsource.atlassian.net/browse/SONARARMOR-860)

Part of 

[SONARARMOR-860]: https://sonarsource.atlassian.net/browse/SONARARMOR-860?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/jsts/tests/parsers/ast.test.ts
+++ b/packages/jsts/tests/parsers/ast.test.ts
@@ -331,6 +331,7 @@ describe('ast', () => {
       checkAstIsProperlySerializedAndDeserialized(ast as TSESTree.Program, protoMessage, 'foo.ts');
     }),
   );
+
   test('should serialize TSEmptyBodyFunctionExpression node to empty object', async () => {
     const code = `class Foo { bar() }`;
     const ast = await parseSourceCode(code, parsersMap.typescript);
@@ -347,6 +348,7 @@ describe('ast', () => {
     expect(methodDefinition.tSEmptyBodyFunctionExpression).toEqual({});
     checkAstIsProperlySerializedAndDeserialized(ast as TSESTree.Program, protoMessage, 'foo.ts');
   });
+
   test('should serialize TSAbstractMethodDefinition nodes to empty objects', async () => {
     const code = `class Foo { abstract bar() }`;
     const ast = await parseSourceCode(code, parsersMap.typescript);
--- a/packages/jsts/tests/parsers/ast.test.ts
+++ b/packages/jsts/tests/parsers/ast.test.ts
@@ -316,7 +316,7 @@ describe('ast', () => {
     expect(tSModuleDeclaration.body).toEqual(undefined);
     checkAstIsProperlySerializedAndDeserialized(ast as TSESTree.Program, protoMessage, 'foo.ts');
   });
-  test('should serialize some nodes to empty objects', async () => {
+  test('should serialize some nodes to empty object', async () => {
     const codeMap = new Map<string, string>();
     codeMap.set('TSTypeAliasDeclaration', `type A = { a: string };`);
     codeMap.set('TSInterfaceDeclaration', `interface A { a: string; }`);
@@ -333,7 +333,7 @@ describe('ast', () => {
       checkAstIsProperlySerializedAndDeserialized(ast as TSESTree.Program, protoMessage, 'foo.ts');
     }
   });
-  test('should serialize TSEmptyBodyFunctionExpression nodes to empty objects', async () => {
+  test('should serialize TSEmptyBodyFunctionExpression node to empty object', async () => {
     const code = `class Foo { bar() }`;
     const ast = await parseSourceCode(code, parsersMap.typescript);
     const protoMessage = visitNode(ast as TSESTree.Program);
--- a/packages/jsts/tests/parsers/ast.test.ts
+++ b/packages/jsts/tests/parsers/ast.test.ts
@@ -316,23 +316,21 @@ describe('ast', () => {
     expect(tSModuleDeclaration.body).toEqual(undefined);
     checkAstIsProperlySerializedAndDeserialized(ast as TSESTree.Program, protoMessage, 'foo.ts');
   });
-  test('should serialize some nodes to empty object', async () => {
-    const codeMap = new Map<string, string>();
-    codeMap.set('TSTypeAliasDeclaration', `type A = { a: string };`);
-    codeMap.set('TSInterfaceDeclaration', `interface A { a: string; }`);
-    codeMap.set('TSEnumDeclaration', `enum Direction {}`);
-    codeMap.set('TSDeclareFunction', `declare function foo()`);
-
-    for (const [nodeType, code] of codeMap) {
+  [
+    { nodeType: 'TSTypeAliasDeclaration', code: `type A = { a: string };` },
+    { nodeType: 'TSInterfaceDeclaration', code: `interface A { a: string; }` },
+    { nodeType: 'TSEnumDeclaration', code: `enum Direction {}` },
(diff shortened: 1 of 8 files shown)


packages/jsts/tests/parsers/ast.test.ts
```diff
@@ -262,15 +263,6 @@ describe('ast', () => {
     checkAstIsProperlySerializedAndDeserialized(ast as TSESTree.Program, protoMessage, 'foo.ts');
   });
 
-  test('Unknown node types in program body are not serialized', async () => {
```
roberto-orlandi-sonarsource, line 265: This test was previously implemented to check that from a list of nodes we are able to filter out the _unknown_ ones. With this PR we implement _basic support_ for some nodes previously unsupported. After this, I could not find another test code to use to reproduce this scenario, so I removed this test, while I kept the business logic, as I am sure that it is still a possible scenario in production.

sonar-plugin/api/src/main/java/org/sonar/plugins/javascript/api/estree/ESTree.java
```diff
@@ -673,7 +673,7 @@ public record TSModuleDeclaration(
     Optional<TSModuleBlock> body,
     String kind
   )
-    implements ModuleDeclaration {}
+    implements Declaration {}
```
roberto-orlandi-sonarsource, line 676: This fixes an error in the hierarchy introduced during the previous PR to support `TSModuleDeclaration` module.

sonar-plugin/standalone/src/test/java/org/sonar/plugins/javascript/standalone/StandaloneParserTest.java
roberto-orlandi-sonarsource, line 1: Inspired by [this comment](https://github.com/SonarSource/SonarJS/pull/5245#issuecomment-2753813938) in the previous PR, I added some UTs in this class, to check that given a piece of code we create the expected java nodes.

roberto-orlandi-sonarsource:
This PR contains the following:
- Implement basic support for some of the most _popular_ TS unsupported nodes, where basic support means that we serialize and deserialize them as empty objects, with the correct type. This will prevent us from stopping the analysis on SonarJasmin, when those nodes are encountered
- Fixes an error introduced in the `ESTree.java` hierarchy during a previous PR: The `TSModuleDeclaration` should be a subclass of `Declaration` and not of `ModuleDeclaration`. This was spotted because we would fail handling code such as `export namespace A {}`, because the `export` syntax node expects a `Declaration` as a child.
- A few more UTs added in the `StandaloneParserTest`, inspired by a previous [PR's comment](https://github.com/SonarSource/SonarJS/pull/5245#issuecomment-2753813938)

packages/jsts/tests/parsers/ast.test.ts
```diff
@@ -324,6 +316,52 @@ describe('ast', () => {
...
+    for (const [nodeType, code] of codeMap) {
+      const ast = await parseSourceCode(code, parsersMap.typescript);
+      const protoMessage = visitNode(ast as TSESTree.Program);
+
+      expect(protoMessage.program.body[0].type).toEqual(NODE_TYPE_ENUM.values[`${nodeType}Type`]);
+      const tSModuleDeclaration = protoMessage.program.body[0][lowerCaseFirstLetter(nodeType)];
+      expect(tSModuleDeclaration).toEqual({});
+      checkAstIsProperlySerializedAndDeserialized(ast as TSESTree.Program, protoMessage, 'foo.ts');
+    }
+  });
```
zglicz, line 335: I think a better practice to implement would be to run the test in a loop. Sth like this:

```suggestion
  [
    { nodeType: 'TSTypeAliasDeclaration', code: `type A = { a: string };` },
    { nodeType: 'TSInterfaceDeclaration', code: `interface A { a: string; }`},
    { nodeType: 'TSEnumDeclaration', code: `enum Direction {}`},
    { nodeType: 'TSDeclareFunction', code: `declare function foo()`},
  ].forEach(({ nodeType, code}) => test(`should serialize ${nodeType} to empty object`, async () => {
      const ast = await parseSourceCode(code, parsersMap.typescript);
      const protoMessage = visitNode(ast as TSESTree.Program);

      expect(protoMessage.program.body[0].type).toEqual(NODE_TYPE_ENUM.values[`${nodeType}Type`]);
      const tSModuleDeclaration = protoMessage.program.body[0][lowerCaseFirstLetter(nodeType)];
      expect(tSModuleDeclaration).toEqual({});
      checkAstIsProperlySerializedAndDeserialized(ast as TSESTree.Program, protoMessage, 'foo.ts');
    })
  });
```

Pardon if this is invalid js code, I was typing in this textbox freehand.
This way, you get the name of the node if it fails and its an actual unit-test.

sonar-plugin/api/src/test/java/org/sonar/plugins/javascript/api/estree/ESTreeTest.java
```diff
@@ -26,14 +26,14 @@ class ESTreeTest {
   @Test
   void test() {
     Class<?>[] classes = ESTree.class.getDeclaredClasses();
-    assertThat(classes).hasSize(118);
+    assertThat(classes).hasSize(125);
```
zglicz, line 29: 😢

sonar-plugin/bridge/src/test/java/org/sonar/plugins/javascript/bridge/ESTreeFactoryTest.java
```diff
@@ -961,4 +961,57 @@ void throw_exception_for_incorrect_cast() {
...
+    );
+  }
+
+  @Test
+  void should_create_ts_empty_body_function_expression() {
+    assertNodeTypeIsParsedToExpectedClass(
+      NodeType.TSEmptyBodyFunctionExpressionType,
+      ESTree.TSEmptyBodyFunctionExpression.class
+    );
+  }
```
zglicz, line 1011: I'm not fluent in Java, but couldn't this be a @parametrizedTest with these test cases provided with a @methodSource? No need naming these tests and they all share a common implementation from line 1013?

sonar-plugin/standalone/src/test/java/org/sonar/plugins/javascript/standalone/StandaloneParserTest.java
```diff
@@ -75,4 +76,87 @@ void test_empty_configuration() {
...
+    );
+  }
+
+  @Test
+  void should_parse_exported_ts_declare_function() {
+    assertExportedNodeIsParsedToCorrectObjectType(
+      "export declare function foo()",
+      ESTree.TSDeclareFunction.class
+    );
+  }
```
zglicz, line 118: Same comment as before, could be parametrized to reduce code-bloat.

zglicz (approved):

packages/jsts/tests/parsers/ast.test.ts
roberto-orlandi-sonarsource: Nice, thanks for the suggestion!

sonar-plugin/bridge/src/test/java/org/sonar/plugins/javascript/bridge/ESTreeFactoryTest.java
roberto-orlandi-sonarsource: Good idea!

PR #5260, opened 2025-03-31
JS-657 Bring back JS code coverage

kebetsi:
[JS-657](https://sonarsource.atlassian.net/browse/JS-657)

I organized this into 4 separate commits as these changes are independent.

`nyc` does not support ESM out-of-the-box. As suggested [in this comment](https://github.com/istanbuljs/nyc/issues/1287#issuecomment-832957610), I am using [@istanbuljs/esm-loader-hook](https://github.com/istanbuljs/esm-loader-hook/?tab=readme-ov-file) to make it compatible.

We're getting 96.3% coverage on JS code now: [lcov.info](https://gist.github.com/kebetsi/82281f06f5b09ff477d5efdb3133d79d)

*FIXED* This yielded many times (once per file?) the following warning message:

```
(node:26325) ExperimentalWarning: `--experimental-loader` may be removed in the future; instead use `register()`:
--import 'data:text/javascript,import { register } from "node:module"; import { pathToFileURL } from "node:url"; register("%40istanbuljs/esm-loader-hook", pathToFileURL("./"));'
(Use `node --trace-warnings ...` to show where the warning was created)
```

### The following remarks were relevant when using the Node.js' native coverage tool

This path was dropped because the native tool has very poor precision such as:
- not ignoring multi-line comments `/**/`, penalizing coverage in many places, including all license headers

`nyc` [does not support ESM](https://github.com/istanbuljs/nyc/issues/1353), so we're rolling back to Node.js' native experimental coverage tool [which was used right after migrating to ESM](https://github.com/SonarSource/SonarJS/blob/dd54c90a6620bc2a193c4adc8a7cd4511a9f7e31/package.json#L21).

The alternatives would be:
- c8
- try ts-node instead of tsx
- fix nyc

I tested the JS coverage with a local SQS instance, picking up the [sonar parameters from the root pom.xml](https://github.com/SonarSource/SonarJS/blob/acdb990da4bca9f8cd31b785fd40a22d8c242469/pom.xml#L106-L116), and here is [the report](https://gist.github.com/kebetsi/5c7bc65d19e4cb880af7740d8b355eb4):

```
sonar-scanner
  -Dsonar.projectKey=<my-project-id> \
  -Dsonar.sources=packages/ \
  -Dsonar.exclusions="packages/**/*.test.ts,packages/**/*.fixture.*,packages/**/fixtures/**/*" \
  -Dsonar.tests=packages/ \
  -Dsonar.test.inclusions="packages/**/*.test.ts" \
  -Dsonar.javascript.lcov.reportPaths=bridge-coverage/lcov.info \
  -Dsonar.typescript.tsconfigPath="./packages/tsconfig.app.json,./packages/tsconfig.test.json" \
  -Dsonar.cpd.exclusions="sonar-plugin/javascript-checks/src/main/resources/**/*.html" \
  -Dsonar.coverage.exclusions="packages/ruling/**,packages/jsts/tests/tools/**" \
  -Dsonar.host.url=http://localhost:9000 \
  -Dsonar.token=<my-token>
```

[JS-657]: https://sonarsource.atlassian.net/browse/JS-657?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.nycrc
```diff
@@ -0,0 +1,10 @@
+{
+  "extension": [".ts"],
+  "include": ["packages/**/src/**/*.ts"],
+  "exclude": ["**/*.test.ts", "**/*.d.ts", "node_modules/**"],
+  "reporter": ["lcov", "text-summary"],
+  "report-dir": "bridge-coverage",
+  "all": true,
+  "sourceMap": true,
+  "instrument": true
```
zglicz, line 9: do you use the sourceMap or instrumentation? Why do you set these flags? Could you explain?

.nycrc
```diff
@@ -0,0 +1,10 @@
+{
+  "extension": [".ts"],
+  "include": ["packages/**/src/**/*.ts"],
+  "exclude": ["**/*.test.ts", "**/*.d.ts", "node_modules/**"],
+  "reporter": ["lcov", "text-summary"],
+  "report-dir": "bridge-coverage",
```
zglicz, line 6: before we were saving the coverage in the root at `lcov.info`, why change this?

zglicz (changes requested):
Can you fix it so that it doesn't have the log spew. As you wrote on Slack, these are not deprecation warnings, they are helpful suggestions that you should just apply.

Please also see inline comments with questions.

.nycrc
kebetsi: removed

.nycrc
kebetsi: As discussed, it would be nice to have a dedicated folder for the JS coverage at some point to avoid ambiguity and conflict with the Java coverage.
Since you already plan to do this later, I'll leave the honor to you.

.cirrus.yml
```diff
@@ -168,7 +168,7 @@ build_task:
     - source cirrus-env BUILD
     - node --version
     - npm ci
-    - regular_mvn_build_deploy_analyze -T1C -Dsonar.skip=true -DartifactoryToken=$ARTIFACTORY_ACCESS_TOKEN
+    - regular_mvn_build_deploy_analyze -T1C -Dsonar.skip=true -DartifactoryToken=$ARTIFACTORY_ACCESS_TOKEN -DskipTests
```
zglicz, line 171: why skip the tests here?

zglicz (commented):
What is the resulting coverage? See also inline comments.

nyc-esm-hook-loader.js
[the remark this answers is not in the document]
kebetsi: moved to `tools/`

.cirrus.yml
kebetsi: The JS tests are not done in the `build` task anymore. There is no point of doing the Java ones here anymore

zglicz (approved):

PR #5283, opened 2025-04-07
JS-677 deploy and analyze in single step

zglicz:
[JS-677](https://sonarsource.atlassian.net/browse/JS-677)

Part of


zglicz:
The build step has to deploy as the its depend on it. Will not pursue this futher.

PR #5289, opened 2025-04-10
JS-681 Generate rspec for CSS using plugin

zglicz:
[JS-681](https://sonarsource.atlassian.net/browse/JS-681)

Continuing on the effort to remove manual steps from the release process. We can remove the usage of the `rule-api.jar` during the release.

[JS-681]: https://sonarsource.atlassian.net/browse/JS-681?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sonar-plugin/css/src/main/resources/org/sonar/l10n/css/rules/rules.json
zglicz, line 1: I've added this file to indicate, which rules we have implemented. In Javascript, we rely on the folder names in the `packages/src/jsts/rules`. For CSS, afaik, we don't have a similar mechanism, hence adding this list. In a perhaps ideal world, we would have parity between what rspec has and what we keep track of here.

The difference for us is: S3502 <- deprecated rule and S5240 <- a 'closed' rule. Perhaps these could be directly filtered by the maven plugin, we could specify, which statuses we want to only accept. However, this change doesn't require a change in the plugin and it makes the Javascript and CSS folders sync in a similar way.

vdiez (approved):
LGTM

PR #5303, opened 2025-04-15
JS-690 Update message on latest supported Typescript

vdiez:
[JS-690](https://sonarsource.atlassian.net/browse/JS-690)

This will happen again every time renovate updates TS version. We should get rid of this TS hardcoded version in Java side

[JS-690]: https://sonarsource.atlassian.net/browse/JS-690?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/JavaScriptPlugin.java
+++ b/sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/JavaScriptPlugin.java
@@ -57,7 +57,7 @@ import org.sonar.plugins.javascript.sonarlint.TsConfigCacheImpl;
 
 public class JavaScriptPlugin implements Plugin {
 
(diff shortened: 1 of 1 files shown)


zglicz (commented):
So can we do it better? If we keep this on the Java side it should be either verified or filled during the build process. Or do you want to wait until we migrate to analyzerProject.ts?

vdiez:
I think we can do the same log catching the exception in the Node.js process and logging this same message. Not sure it's worthy overengineering the whole thing to have the semver string on the java side. This is just a fix to have this matching in the next release, but I would like to remove it completely from Java

zglicz (approved):
Create a ticket to move it to analyzeProject.ts under your master mmf? Or is this verified by an integration test?

vdiez:
no, it's not. just from a unit test. Ticket created: https://sonarsource.atlassian.net/browse/JS-692

PR #5312, opened 2025-04-17
JS-633 Improve S3735: if no types, don't raise

zglicz:
[JS-633](https://sonarsource.atlassian.net/browse/JS-633)

If we don't have typescript types, lets be permissive and if we have a function call that it can be a promise.

[JS-633]: https://sonarsource.atlassian.net/browse/JS-633?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/jsts/src/rules/S3735/rule.ts
+++ b/packages/jsts/src/rules/S3735/rule.ts
@@ -18,7 +18,12 @@
 
 import type { Rule } from 'eslint';
 import estree from 'estree';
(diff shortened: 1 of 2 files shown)


vdiez (approved):
just what we discussed

PR #5313, opened 2025-04-17
JS-545 Configure renovate to use repox private npm registry

zglicz:
[JS-545](https://sonarsource.atlassian.net/browse/JS-545)

As we noticed, renovate uses the public npm registry. Let us update this to first look in repox as the registry!
@jayadeep-km-sonarsource @hedinasr (I found your names as you've either previously touched this config or added a new source - https://sonarsource.atlassian.net/browse/BUILD-5885 ) - 2 questions:
1) how can I test this?
2) should this be added in the parent configs of renovate? I see that for maven we already have this in a common setup - https://github.com/SonarSource/renovate-config/blob/master/default.json#L22 
Could this come preconfigured for npm to use repox? As example see this pull request and the package-lock.json file - 
https://github.com/SonarSource/SonarJS/pull/5233/files
We change all of the urls from repox to the public repository.

[JS-545]: https://sonarsource.atlassian.net/browse/JS-545?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5319, opened 2025-04-23
Update transitive packages

zglicz:
We have a an issue on mend - https://saas-eu.whitesourcesoftware.com/Wss/WSS.html#!securityVulnerability;id=CVE-2025-27789;orgToken=b55a9158-7eca-400e-809d-5d5f0a5ea7df when using @babel/core < 7.26.10 . The causing package `json-schema-to-ts` only requires ^7.18. Therefore, I removed `node_modules` and re-installed with npm while updating the package-lock.json file.

This also updated many packages to rely on repox instead of the public repository. For renovate to start working off of repox, we have to wait for: https://github.com/SonarSource/SonarJS/pull/5313


jayadeep-km-sonarsource (approved):
I reviewed the first few lines. Looks good

PR #5320, opened 2025-04-23
JS-72 JS-342 Calls to node:assert library should count as valid assertions

zglicz:
[JS-41](https://sonarsource.atlassian.net/browse/JS-41)

This way, if a user is using the node built-in library 'assert', we will consider it a valid assertion.

[JS-41]: https://sonarsource.atlassian.net/browse/JS-41?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/jsts/src/rules/S2699/assert-esm.fixture.js
```diff
@@ -0,0 +1,16 @@
+import assert, { deepStrictEqual } from "assert";
+import vitest from 'vitest';
```
zglicz, line 2: This can be any import from the 4 frameworks that identify a file to be a test file.

packages/jsts/src/rules/S2699/assert-esm.fixture.js
```diff
@@ -0,0 +1,16 @@
...
+describe('global assert', () => {
+  it('should recognize assert', () => { // Compliant
+    assert(4);
+  });
+
+  it('should recognize assert.XXX methods', () => {  // Compliant
+    deepStrictEqual({ a: 1 }, { a: '1' });
+  });
+
+  it('should recognize issue', () => { // Noncompliant {{Add at least one assertion to this test case.}}
```
zglicz, line 13: Necessary to have an issue that raises in each test

packages/jsts/src/rules/S2699/rule.ts
```diff
@@ -125,3 +137,67 @@ class TestCaseAssertionVisitor {
...
+        if (
+          (identifier.escapedText as string) === 'require' &&
+          initializer.arguments.length === 1 &&
+          initializer.arguments[0].kind === SyntaxKind.StringLiteral
+        ) {
+          moduleName = (initializer.arguments[0] as StringLiteral).text;
+        }
+      }
+    }
+  }
```
zglicz, line 201: this I died a slow death whilst typing it. I'm guessing there is a cleaner way of doing this. However, I checked in the `helpers/module.ts` and there we usually go through import/require statements, but not necessarily try to resolve the methods we use...

packages/jsts/src/rules/S2699/rule.ts
```diff
@@ -125,3 +126,15 @@ class TestCaseAssertionVisitor {
...
+function isGlobalAssertion(context: Rule.RuleContext, node: estree.Node): boolean {
+  if (isFunctionCall(node) && node.callee.name === 'expect') {
+    return true;
+  }
+  return isFunctionCallFromNodeAssert(context, node);
+}
+
+function isFunctionCallFromNodeAssert(context: Rule.RuleContext, node: estree.Node) {
+  const fullyQualifiedName = getFullyQualifiedName(context, node);
+  return !!fullyQualifiedName && fullyQualifiedName.split('.')[0] === 'assert';
```
vdiez, line 139:
```suggestion
  return fullyQualifiedName?.split('.')[0] === 'assert';
```

vdiez (approved):

PR #5344, opened 2025-05-01
S2068 should be case-insensitive and support "passphrase"

GabinL21:


vdiez (changes requested):
Thank you @GabinL21,

looks good! However I see a possible issue as the strings can be customize by the user, and he can pass uppercased strings. We should also lowercase the variableNames array I would say.

GabinL21:
Very fair point @vdiez, thanks! It should look better now 😄

vdiez (approved):
Excellent, thank you!

PR #5361, opened 2025-05-08
JS-712 Remove multi-part response for BridgeResponse

zglicz:
[JS-712](https://sonarsource.atlassian.net/browse/JS-712)

Background: In order to support transporting binary data in the response from the Node service a multi-part response was added.

Issues: Using the multi part response to transport the AST binary data added complexity to the codebase and blocked any extension we might want to add. There was convoluted logic into how to extract it with string manipulation +-2,4 in random places. There is no clear way to extend it to support the work on `analyzeProject` where the response will be for multiple files and include multiple asts.

Solution: Base64 encode the binary data and send that together with the standard `application-json` response. This allows for simplicity and a homogenous treatment of the response. The only downside I see is that there is a bit of overhead with this approach (75% efficiency in the encoding).

Adding Security folks so that you are aware @roberto-orlandi-sonarsource, @karim-ouerghemmi-sonarsource 

[JS-712]: https://sonarsource.atlassian.net/browse/JS-712?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sonar-plugin/bridge/src/main/java/org/sonar/plugins/javascript/bridge/AstProtoUtils.java
```diff
@@ -0,0 +1,59 @@
...
+import com.google.protobuf.CodedInputStream;
+import com.google.protobuf.InvalidProtocolBufferException;
+import java.io.IOException;
+import java.util.Base64;
+import javax.annotation.CheckForNull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.sonar.plugins.javascript.bridge.protobuf.Node;
+
+public class AstProtoUtils {
```
zglicz, line 28: This is the previous file FormDataUtils renamed and only kep the `parseProtobuf` method

sonar-plugin/bridge/src/main/java/org/sonar/plugins/javascript/bridge/BridgeServer.java
```diff
@@ -100,11 +100,18 @@ record CssAnalysisRequest(
...
   ) {}
 
-  record BridgeResponse(String json, @Nullable Node ast) {
-    public BridgeResponse(String json) {
-      this(json, null);
-    }
-  }
+  record BridgeResponse(String json) {}
+
+  record AnalysisResponseDTO(
```
zglicz, line 105: this is the assumed parsed json response from the Bridge

karim-ouerghemmi-sonarsource:
Thanks for making us aware @zglicz ! I'll validate with the testbench we have wiht `~200` projects if anything goes wrong related to SonarJasmin https://cirrus-ci.com/build/6283518793220096

vdiez (approved):

PR #5365, opened 2025-05-09
JS-713 S2699 new implementation could cause endless loop

zglicz:
[JS-713](https://sonarsource.atlassian.net/browse/JS-713)

Fuond based on peachee job - and failing on the last test-case here - https://github.com/SonarSource/expected-issues/blob/master/javascript/test/RSPEC_6092/uncertainAssertion.test.js

If we jump to 'ArrowExpression', we already know its a local variable and we don't need to extract the fqn.

[JS-713]: https://sonarsource.atlassian.net/browse/JS-713?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/jsts/src/rules/S2699/chai.fixture.js
+++ b/packages/jsts/src/rules/S2699/chai.fixture.js
@@ -1,5 +1,6 @@
 const chai = require('chai');
 const chaiExpect = require('chai').expect;
+const shouldFunc = require('chai').should();
 const { assert, expect, should, expect: chaiExpectRenamed } = chai;
 
 should();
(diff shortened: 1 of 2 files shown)


packages/jsts/src/rules/S2699/chai.fixture.js
```diff
@@ -69,6 +70,15 @@ describe('chai test cases', () => {
     check();
   });
 
+  const throwsTypeError = () => { throw new TypeError() }
+
+  it("uses chai 'should'", function() { // Noncompliant {{Add at least one assertion to this test case.}}
+    // The same is true for "should" assertions.
+    throwsTypeError.should.to.not.throw(ReferenceError)
```
zglicz, line 77: this testcase would do an endless loop. Extracted from - https://github.com/SonarSource/expected-issues/blob/master/javascript/test/RSPEC_6092/uncertainAssertion.test.js#L43C5-L48C8

vdiez (approved):

PR #5383, opened 2025-05-15
JS-636 Implement progressReport in Node.js

vdiez:
[JS-636](https://sonarsource.atlassian.net/browse/JS-636)

[JS-636]: https://sonarsource.atlassian.net/browse/JS-636?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/shared/src/helpers/progress-report.ts
```diff
@@ -0,0 +1,67 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the Sonar Source-Available License for more details.
+ *
+ * You should have received a copy of the Sonar Source-Available License
+ * along with this program; if not, see https://sonarsource.com/license/ssal/
+ */
+import { info } from './logging.js';
+
+/**
+ * Logs the current file being analyzed every N seconds
```
zglicz, line 20: every X milis

packages/shared/src/helpers/progress-report.ts
```diff
@@ -0,0 +1,67 @@
...
+ */
+
+export class ProgressReport {
+  private currentFileNumber = 0;
+  private currentFilename = '';
+  private intervalId: NodeJS.Timeout | undefined = undefined;
+
+  constructor(
+    private readonly count: number,
+    private readonly interval: number = 10_000,
```
zglicz, line 32: Is this ever used by anything else? Does it need to be parametriczed?

packages/shared/src/helpers/progress-report.ts
```diff
@@ -0,0 +1,67 @@
...
+  ) {
+    info(`${this.count} source ${this.pluralizeFile(this.count)} to be analyzed.`);
```
zglicz, line 34: I don't believe the original java implementation adds period at the end?

packages/shared/src/helpers/progress-report.ts
```diff
@@ -0,0 +1,67 @@
...
+import { info } from './logging.js';
+
+/**
+ * Logs the current file being analyzed every N seconds
+ *
+ * @param interval the interval between two logs, in milliseconds
+ */
+
+export class ProgressReport {
+  private currentFileNumber = 0;
```
zglicz, line 26: shouldn't this be Count instead of Number? Number sounds more like ID.

zglicz (approved):

packages/shared/src/helpers/progress-report.ts
vdiez: makes sense. count right now is the total. i will change that

packages/shared/src/helpers/progress-report.ts
vdiez: I actually prefer to have it parametrized, but ok i will change it

packages/shared/tests/helpers/progress-report.test.ts
```diff
@@ -0,0 +1,53 @@
...
+    const consoleLogMock = (console.log as Mock<typeof console.log>).mock;
+    expect(consoleLogMock.calls[0].arguments[0]).toEqual(`5 source files to be analyzed`);
+    expect(consoleLogMock.callCount()).toEqual(1);
+    // we pass next file
+    progressReport.nextFile('file1.ts');
+    // passing next file should not trigger a console.log
+    expect(consoleLogMock.callCount()).toEqual(1);
+
+    // after 10 seconds have passed, there is a new log
+    mock.timers.tick(10000);
```
zglicz, line 35: to be more precise and actually check sth cool, could you check after 9999 that there is nothing and then at 10000 that its there. And again, it'd be good to tie it with the default in the ProgressReport so that if we change it to 15s in the future, one doesn't have to fix the unit tests

zglicz (approved):

packages/shared/tests/helpers/progress-report.test.ts
vdiez: done!

PR #5394, opened 2025-05-19
JS-727 Remove old analysis path (/analyze-js)

vdiez:
[JS-727](https://sonarsource.atlassian.net/browse/JS-727)

[JS-727]: https://sonarsource.atlassian.net/browse/JS-727?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/jsts/src/analysis/projectAnalysis/projectAnalyzer.ts
```diff
@@ -43,6 +43,9 @@ export async function analyzeProject(
   incrementalResultsChannel?: (result: WsIncrementalResult) => void,
 ): Promise<ProjectAnalysisOutput> {
   const { rules, baseDir, files, configuration = {}, bundles = [], rulesWorkdir } = input;
+  if (!baseDir) {
+    throw new Error('baseDir is required');
+  }
```
zglicz: if it's required and will throw, shouldn't it be annotated as such on the ProjectAnalysisInput?

packages/tsconfig.test.json
```diff
@@ -2,6 +2,7 @@
   "extends": "./tsconfig.json",
   "compilerOptions": {
     "strictNullChecks": false,
+    "lib": ["ES2024"],
```
zglicz, line 5: why is that necessary? is it for the Promise.withResolvers?

sonar-plugin/sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/analysis/JsTsSensorTest.java
```diff
@@ -150,51 +136,40 @@ class JsTsSensorTest {
   @TempDir
   Path workDir;
 
-  private AnalysisProcessor processAnalysis;
+  private AnalysisProcessor analysisProcessor;
 
   @BeforeEach
   void setUp() throws Exception {
-    MockitoAnnotations.initMocks(this);
+    MockitoAnnotations.openMocks(this).close();
```
zglicz, line 143: do you need this if you insta close it? Asking my AI friend it says:

**What does this mean in practice?**

Mocks are initialized for the current test instance.
Mocks are immediately cleaned up (closed) after initialization, which is usually not useful unless you only need the mocks for a very short scope.
Typically, you would assign the result of openMocks(this) to a variable and close it in a @After or @AfterEach method, to ensure mocks are available for the duration of the test.

sonar-plugin/sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/analysis/JsTsSensorTest.java
zglicz, line 1: Much nicer, but still, I think it'd be best to break this test up into multiple themed ones. Some are to do with the communication with the BridgeServer, some are about metrics/highlights/telemetry. Some is basic analyzer functionality. And what I'm guessing is that some of these helpers are similar in other files, so those could go to TestUtils. At 1.6k lines, it's still a beast of a test.

packages/tsconfig.test.json
vdiez: correct

sonar-plugin/sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/analysis/JsTsSensorTest.java
vdiez: so what do you propose then? both lines do the same but one was deprecated so I left the non-deprecated. 
```
    MockitoAnnotations.openMocks(this).close(); // used in JavaScriptEslintBasedSensorTest.java
    MockitoAnnotations.initMocks(this); //used in JsTsSenssorTest.java, deprecated
```

packages/jsts/src/analysis/projectAnalysis/projectAnalyzer.ts
vdiez: it is already. I just wanted to force the exception at runtime. but i'll remove that check, not adding much

PR #5399, opened 2025-05-20
JS-731 AST consumers can be disabled

renaud-tognelli-sonarsource:
[JS-731](https://sonarsource.atlassian.net/browse/JS-731)

[JS-731]: https://sonarsource.atlassian.net/browse/JS-731?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/analysis/AnalysisConsumers.java
+++ b/sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/analysis/AnalysisConsumers.java
@@ -39,10 +39,8 @@ public class AnalysisConsumers implements JsAnalysisConsumer {
 
   public AnalysisConsumers(List<JsAnalysisConsumer> consumers) {
     this.consumers = consumers.stream().filter(JsAnalysisConsumer::isEnabled).toList();
-    if (LOG.isDebugEnabled()) {
-      LOG.debug("Enabled JsAnalysisConsumers {}/{}", this.consumers.size(), consumers.size());
-      LOG.debug("Registered JsAnalysisConsumers {}", this.consumers);
-    }
+    LOG.debug("Enabled JsAnalysisConsumers {}/{}", this.consumers.size(), consumers.size());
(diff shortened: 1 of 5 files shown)


sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/analysis/AnalysisConsumers.java
```diff
@@ -38,8 +38,11 @@ public AnalysisConsumers() {
   }
 
   public AnalysisConsumers(List<JsAnalysisConsumer> consumers) {
-    this.consumers = List.copyOf(consumers);
-    LOG.debug("Registered JsAnalysisConsumers {}", this.consumers);
+    this.consumers = consumers.stream().filter(JsAnalysisConsumer::isEnabled).toList();
+    if (LOG.isDebugEnabled()) {
```
vdiez, line 42: why the need for this check? if log level is higher it will just not be logged, right?

vdiez (approved):
LGTM!!

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/analysis/AnalysisConsumers.java
renaud-tognelli-sonarsource: indeed, it's a not needed optimisation, as computing the value is not expensive here, I will remove it for simplicity.

vdiez: that's ok, no need to remove. I just wanted to make sure

renaud-tognelli-sonarsource: To late 😄 
I think it is more readable without, 
the code is called once per analsysis, 
less is more 👍 

I also added some comments on the deprecated properties.

PR #5405, opened 2025-05-22
JS-739 Clean up analyze-project files finder

vdiez:
[JS-739](https://sonarsource.atlassian.net/browse/JS-739)

[JS-739]: https://sonarsource.atlassian.net/browse/JS-739?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/jsts/src/analysis/projectAnalysis/file-stores/files.ts
```diff
@@ -0,0 +1,142 @@
...
+    this.filenames = [];
+    this.paths = new Set<string>();
+    this.testPaths = getTestPaths()?.map(test => toUnixPath(join(baseDir, test)));
+  }
+
+  async process(file: Dirent, filePath: string) {
+    if (isAnalyzableFile(file.name)) {
+      const fileType = this.getFiletype(filePath, this.testPaths);
+      if (isJsTsFile(file.name)) {
+        const fileContent = await readFile(filePath, 'utf8');
```
zglicz, line 109: Isn't it possible that we actually have the contents of a file already present coming from the request in the JsTsFiles object? Even reading the PR from SonarLint, they promise to provide the content as well.

packages/jsts/src/analysis/projectAnalysis/file-stores/package-jsons.ts
```diff
@@ -0,0 +1,111 @@
...
+  stripBOM,
+} from '../../../rules/index.js';
+import { Dirent } from 'node:fs';
+import { readFile } from 'node:fs/promises';
+import { warn } from '../../../../../shared/src/helpers/logging.js';
+import { FileStore } from './store-type.js';
+import { SourceFileStore } from './files.js';
+
+export const UNINITIALIZED_ERROR =
+  'package.json cache has not been initialized. Call loadFiles() first.';
```
zglicz, line 34: do you call it cache or store? better to be consistent

zglicz: Also, I do not like that the loadFiles() is just a global, floating function. It would be nice to somehow connect it to the stores.

packages/jsts/src/analysis/projectAnalysis/file-stores/store-type.ts
```diff
@@ -0,0 +1,28 @@
...
+ * See the Sonar Source-Available License for more details.
+ *
+ * You should have received a copy of the Sonar Source-Available License
+ * along with this program; if not, see https://sonarsource.com/license/ssal/
+ */
+
+import { Dirent } from 'node:fs';
+
+export abstract class FileStore {
+  abstract isInitialized(baseDir: string): boolean;
```
zglicz, line 21: Could the property baseDir be added here? As in, should the FileStore always have a baseDir string present?

packages/jsts/src/analysis/projectAnalysis/file-stores/tsconfigs.ts
```diff
@@ -0,0 +1,269 @@
...
+} from '../../../../../shared/src/helpers/configuration.js';
+import { basename } from 'node:path';
+import { toUnixPath } from '../../../../../shared/src/helpers/files.js';
+import { join } from 'node:path/posix';
+import { Minimatch } from 'minimatch';
+import { Dirent } from 'node:fs';
+import { FileStore } from './store-type.js';
+import { SourceFileStore } from './files.js';
+
+tmp.setGracefulCleanup();
```
zglicz, line 44: why does this need it? You can put it in a comment. Or is that the comment in lines 18-19, then move it closer.

packages/jsts/src/analysis/projectAnalysis/file-stores/tsconfigs.ts
```diff
@@ -0,0 +1,269 @@
...
+  pattern: Minimatch;
+};
+type TsConfigOrigin = 'property' | 'lookup' | 'fallback';
+
+export class TsConfigStore implements FileStore {
+  // tsconfig.json files in the project tree
+  private foundLookupTsConfigs: string[] = [];
+  // tsconfig.json files specified in sonar.typescript.tsconfigPaths
+  private foundPropertyTsConfigs: string[] = [];
+  private providedTsConfigs: ProvidedTsConfig[] | undefined = undefined;
```
zglicz, line 60: are providedTsConfigs the ones related to `foundPropertyTsConfigs`? If so, something should connect them, at least the name.

packages/jsts/src/analysis/projectAnalysis/file-stores/tsconfigs.ts
```diff
@@ -0,0 +1,269 @@
...
+  private origin: TsConfigOrigin | undefined = undefined;
+  private cacheMap: { [type in TsConfigOrigin]: Cache } = {
+    property: new Cache(),
+    lookup: new Cache(),
+    fallback: new Cache(),
+  };
+
+  constructor(private readonly filesStore: SourceFileStore) {}
+
+  getCacheOrigin() {
```
zglicz, line 70: what if the getCacheOrigin is undefined? Do you not want to throw?

packages/jsts/src/analysis/projectAnalysis/file-stores/tsconfigs.ts
```diff
@@ -0,0 +1,269 @@
...
+
+  isInitialized(baseDir: string) {
+    this.dirtyCachesIfNeeded(baseDir);
+    return this.origin !== undefined && this.cacheMap[this.origin].initialized;
+  }
+
+  getTsConfigs() {
+    if (!this.origin) {
+      throw new Error(UNINITIALIZED_ERROR);
+    }
```
zglicz, line 89: can there be a guard, or it could return? I wonder that cpd doesn't kick in?

packages/jsts/src/analysis/projectAnalysis/file-stores/tsconfigs.ts
```diff
@@ -0,0 +1,269 @@
...
+    if (this.foundPropertyTsConfigs.length) {
+      this.origin = 'property';
+    } else if (this.foundLookupTsConfigs.length) {
+      this.origin = 'lookup';
+    } else {
+      this.cacheMap.fallback.initializeOriginalTsConfigs(await this.getFallbackTsConfig(baseDir));
+      this.origin = 'fallback';
+    }
+    info(
+      `Found ${this.getTsConfigs().length} tsconfig.json file(s): [${this.getTsConfigs().join(', ')}]`,
```
zglicz, line 116: do you also want to add the origin here?

packages/jsts/src/analysis/projectAnalysis/file-stores/tsconfigs.ts
```diff
@@ -0,0 +1,269 @@
...
+  }
+
+  /**
+   * Create and return a TSConfig object.
+   *
+   * @param files array of files included in the TS program
+   * @param include inclusion paths of the TS Program
+   * @returns the TSConfig object
+   */
+  createTSConfigFile(files?: string[], include?: string[]): TsConfigJson {
```
zglicz, line 222: this doesn't actually create a file, no? Should it be `createTSConfigObject` ?

packages/jsts/src/analysis/projectAnalysis/file-stores/tsconfigs.ts
```diff
@@ -0,0 +1,269 @@
...
+      },
+      include,
+      files,
+    } as TsConfigJson;
+  }
+
+  setup(baseDir: string) {
+    this.foundPropertyTsConfigs = [];
+    this.foundLookupTsConfigs = [];
+    this.providedTsConfigs = getTsConfigPaths().map(tsConfigPath => {
```
zglicz, line 236: would a better name be: `providedPropertyTsConfigs`?

zglicz (approved):
Good job! Mostly cosmetic comments.

packages/jsts/src/analysis/projectAnalysis/file-stores/files.ts
vdiez: ah, no... we need the contents to check if the file is minified and thus is ignored... still, I will not read if the contents are present in the request

packages/jsts/src/analysis/projectAnalysis/file-stores/tsconfigs.ts
vdiez: it's only used for tests

packages/jsts/src/analysis/projectAnalysis/file-stores/package-jsons.ts
vdiez: i think it's more meaningful to the user mentioning cache tbh. about loadFiles... would you prefer to pass the stores as parameter?

packages/jsts/src/analysis/projectAnalysis/file-stores/store-type.ts
vdiez: tsconfigStore does not have an attribute baseDir, it's inside the cacheMap elements

packages/jsts/src/analysis/projectAnalysis/file-stores/tsconfigs.ts
vdiez: not enough lines I think. I prefer return to avoid TS inferring possible undefined return value

PR #5444, opened 2025-06-05
NOJIRA Try out new caching

zglicz:


packages/jsts/src/rules/helpers/package-json.ts
```diff
@@ -67,6 +71,19 @@ export function getAllDependencies(): NamedDependency[] {
...
 }
 
+export function getClosestPackageJSONDir(filename: string, cwd: string): string {
+  const dirname = Path.dirname(toUnixPath(filename));
+  if (!dirNameToClosestPackageJSONCache.has(dirname)) {
+    const files = findPackageJsons(dirname, cwd, fs);
+    // take the longest filepath as that will be the closest package.json to the provided file
+    dirNameToClosestPackageJSONCache.set(
+      dirname,
+      files.sort((a, b) => b.path.length - a.path.length).at(0)?.path ?? cwd,
```
zglicz: I'll update this with a reduce

vdiez: we need to check the cache instead of traversing FS again. in another PR

PR #5445, opened 2025-06-06
JS-758 Handle Cancelled SensorContext

vdiez:
[JS-758](https://sonarsource.atlassian.net/browse/JS-758)

[JS-758]: https://sonarsource.atlassian.net/browse/JS-758?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/jsts/src/analysis/projectAnalysis/analyzeWithProgram.ts
+++ b/packages/jsts/src/analysis/projectAnalysis/analyzeWithProgram.ts
@@ -24,6 +24,7 @@ import ts from 'typescript';
 import { ProgressReport } from '../../../../shared/src/helpers/progress-report.js';
 import { handleFileResult } from './handleFileResult.js';
 import type { WsIncrementalResult } from '../../../../bridge/src/request.js';
+import { isAnalysisCancelled } from './projectAnalyzer.js';
 
 /**
  * Analyzes JavaScript / TypeScript files using TypeScript programs. Files not
@@ -45,6 +46,9 @@ export async function analyzeWithProgram(
 ) {
   const processedTSConfigs: Set<string> = new Set();
   for (const tsConfig of tsConfigStore.getTsConfigs()) {
+    if (isAnalysisCancelled()) {
+      return;
+    }
     await analyzeProgram(
       files,
       tsConfig,
@@ -69,7 +73,7 @@ async function analyzeProgram(
   progressReport: ProgressReport,
   incrementalResultsChannel?: (result: WsIncrementalResult) => void,
 ) {
-  if (processedTSConfigs.has(tsConfig)) {
+  if (isAnalysisCancelled() || processedTSConfigs.has(tsConfig)) {
     return;
   }
   processedTSConfigs.add(tsConfig);
@@ -98,6 +102,9 @@ async function analyzeProgram(
   }
   results.meta?.programsCreated.push(tsConfig);
   for (const filename of filenames) {
+    if (isAnalysisCancelled()) {
+      return;
+    }
     // only analyze files which are requested
     if (files[filename] && pendingFiles.has(filename)) {
       progressReport.nextFile(filename);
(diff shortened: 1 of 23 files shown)


packages/jsts/src/analysis/projectAnalysis/analyzeWithProgram.ts
```diff
@@ -69,7 +73,7 @@ async function analyzeProgram(
   progressReport: ProgressReport,
   incrementalResultsChannel?: (result: WsIncrementalResult) => void,
 ) {
-  if (processedTSConfigs.has(tsConfig)) {
+  if (isAnalysisCancelled() || processedTSConfigs.has(tsConfig)) {
```
zglicz: Is this or the one above excessive? I think this one is a bit unnecessary, no?

packages/jsts/src/analysis/projectAnalysis/projectAnalyzer.ts
```diff
@@ -27,10 +27,22 @@ import {
...
+};
+
+export function cancelAnalysis() {
+  analysisStatus.cancelled = true;
+}
+
+export function isAnalysisCancelled() {
+  return analysisStatus.cancelled;
+}
+
```
zglicz, line 45: how about storing this within the Result object? This way, this is only a concept if you have this Result object present. Otherwise, we could be dropping this ifs anywhere, just because.

sonar-plugin/bridge/src/main/java/org/sonar/plugins/javascript/bridge/BridgeServerImpl.java
```diff
@@ -410,7 +410,7 @@ public void analyzeProject(WebSocketMessageHandler<ProjectAnalysisRequest> handl
     var request = handler.getRequest();
     request.setBundles(deployedBundles.stream().map(Path::toString).toList());
     request.setRulesWorkdir(workdir);
-    this.client.send(GSON.toJson(request));
+    this.client.send(new Gson().toJson(Map.of("type", "on-analyze-project", "data", request)));
```
zglicz:
```suggestion
    this.client.send(GSON.toJson(Map.of("type", "on-analyze-project", "data", request)));
```

sonar-plugin/bridge/src/main/java/org/sonar/plugins/javascript/bridge/JSWebSocketClient.java
```diff
@@ -64,6 +66,9 @@ public void onMessage(String message) {
     JsonObject jsonObject = JsonParser.parseString(message).getAsJsonObject();
     for (WebSocketMessageHandler<?> handler : messageHandlers) {
       handler.handleMessage(jsonObject);
+      if (handler.getContext().isCancelled()) {
+        this.send(new Gson().toJson(Map.of("type", "on-cancel-analysis")));
+      }
```
zglicz: so this only occurs in the `onMessage` when there is a message coming back from Node to the Java server. Shouldn't this be proactive coming from Java side? Without the need to have an onMessage trigger to check it?

zglicz (approved):
Few questions from my side. Overall looks slick

sonar-plugin/bridge/src/main/java/org/sonar/plugins/javascript/bridge/JSWebSocketClient.java
vdiez: yes, we are keeping the logic as it was in the old path, same with css, html and yaml sensor. I agree it would be nice to have some kind on cancellation event listener to be proactive, but not sure if it's possible. We would need to do a little bit of research

packages/jsts/src/analysis/projectAnalysis/projectAnalyzer.ts
vdiez: needs to be global to make it accessible from other other modules.

PR #5453, opened 2025-06-11
Use filters on Node.js

vdiez:
We had the filters implemented in both Node.js and Java. We can use the ones on Node.js side all the time and remove the Java ones, which will also avoid side-effects on other Sensors


zglicz:
Perhaps you could also remove the tests that count the number of classes added. I've never like those ... Just have them in the ballpark and not exact numbers?

packages/css/src/analysis/analyzer.ts
```diff
@@ -32,7 +34,13 @@ import { error } from '../../../shared/src/helpers/logging.js';
  * @returns a promise of the CSS analysis output
  */
 export async function analyzeCSS(input: CssAnalysisInput): Promise<CssAnalysisOutput> {
-  const { filePath, fileContent, rules } = await fillFileContent(input);
+  const { filePath, fileContent, rules, configuration } = await fillFileContent(input);
+  if (configuration) {
+    setGlobalConfiguration(configuration);
```
zglicz: would a nicer pattern be to add the if inside the `setGlobalConfiguration`?

packages/css/src/analysis/analyzer.ts
```diff
@@ -32,7 +34,13 @@ import { error } from '../../../shared/src/helpers/logging.js';
...
+    if (!accept(filePath, fileContent)) {
+      return { issues: [] };
+    }
```
zglicz: I'm not sure why this should also always be guarded? I would hope that with an empty configuration, the accept() also works?

packages/jsts/src/analysis/projectAnalysis/file-stores/index.ts
```diff
@@ -39,15 +45,11 @@ export async function initFileStores(baseDir: string, inputFiles?: JsTsFiles) {
...
-  await findFiles(
-    baseDir,
-    async (file, filePath) => {
-      for (const store of pendingStores) {
-        await store.process(file, filePath);
-      }
-    },
-    getExclusions(),
-  );
+  await findFiles(baseDir, async (file: Dirent, filePath: string, fileType: FileType) => {
```
zglicz: curious why you opted to put the types in here?

packages/jsts/src/analysis/projectAnalysis/file-stores/source-files.ts
```diff
@@ -127,57 +120,52 @@ export class SourceFileStore implements FileStore {
...
-    this.setFiles('found', this.newFiles);
+    await this.setFiles('found', this.newFiles, false);
   }
 
-  setFiles(store: keyof typeof SourceFileStore.prototype.store, files: JsTsAnalysisInput[]) {
+  async setFiles(
+    store: keyof typeof SourceFileStore.prototype.store,
+    files: JsTsAnalysisInput[],
+    filter = true,
+  ) {
```
zglicz: Is this a private method? If so, could you mark it as so. If not, isn't this a classic case of using a boolean to drive the method. Instead it is preferable in the interface to have 2 method: setFilesWithFilter and setFiles. That way, as the reader of the code, it's also obvious what is the dangling boolean parameter.

packages/jsts/src/analysis/projectAnalysis/file-stores/source-files.ts
```diff
@@ -127,57 +120,52 @@ export class SourceFileStore implements FileStore {
...
-      const filename = toUnixPath(file.filePath);
-      file.filePath = filename;
+      if (filter) {
+        // We need to apply filters if the files come from the request
+        const filename = toUnixPath(file.filePath);
+        file.filePath = filename;
+        file.fileContent = file.fileContent ?? (await readFile(filename));
+        if (isJsTsExcluded(filename) || !accept(filename, file.fileContent)) {
+          continue;
+        }
```
zglicz: this could be a helper method

packages/jsts/src/analysis/projectAnalysis/file-stores/source-files.ts
```diff
@@ -127,57 +120,52 @@ export class SourceFileStore implements FileStore {
...
+      }
       if (store === 'found') {
-        this.paths!.add(dirname(filename));
+        this.paths!.add(dirname(file.filePath));
       }
-      this.store[store].filenames.push(filename);
-      this.store[store].files[filename] = file;
+      this.store[store].filenames.push(file.filePath);
+      this.store[store].files[file.filePath] = file;
```
zglicz: this as well could be a helper method. So that you will only have to call: saveFileContents(store, path)

packages/jsts/src/embedded/analysis/analyzer.ts
```diff
@@ -48,6 +50,13 @@ export function analyzeEmbedded(
   input: CompleteEmbeddedAnalysisInput,
   languageParser: LanguageParser,
 ): EmbeddedAnalysisOutput {
+  const { filePath, fileContent, configuration } = input;
+  if (configuration) {
+    setGlobalConfiguration(configuration);
+    if (!accept(filePath, fileContent)) {
+      return { issues: [], metrics: { ncloc: [] } };
```
zglicz: Why are the `metrics` included here and not in the other `analyzeJsTs`, `analyzeCSS`, ...

Otherwise, it'd be nice to have a constant representing an empty file analysis output

packages/shared/src/helpers/configuration.ts
```diff
@@ -175,6 +283,10 @@ export function shouldClearTsConfigCache() {
   return configuration.clearTsConfigCache;
 }
 
+export function shouldDetectBundles() {
+  return configuration.detectBundles !== false;
+}
```
zglicz, line 291: could this be: `shouldIgnoreBundlesFilter`? That way you don't need another negative when writing the condition.

packages/shared/src/helpers/filter/filter-bundle.ts
```diff
@@ -14,12 +14,27 @@
  * You should have received a copy of the Sonar Source-Available License
  * along with this program; if not, see https://sonarsource.com/license/ssal/
  */
+import { debug, info } from '../logging.js';
+
 const READ_CHARACTERS_LIMIT = 2048;
 const COMMENT_OPERATOR_FUNCTION = buildBundleRegex();
+let isInfoLogged = false;
```
zglicz: I'd use a past tense.
```suggestion
let hasInfoBeenLogged = false;
```

packages/shared/src/helpers/find-files.ts
```diff
@@ -14,32 +14,28 @@
...
-  const files = await fs.readdir(dir, { recursive: true, withFileTypes: true });
+  const directories = [dir];
 
-  for (const file of files) {
-    const filePath = toUnixPath(join(file.parentPath, file.name));
-    const relativePath = filePath.substring(prefixLength);
-    if (file.isFile() && !exclusions.some(exclusion => exclusion.match(relativePath))) {
-      await onFile(file, filePath, relativePath);
+  while (directories.length > 0) {
+    const directory = directories.shift()!;
```
zglicz: array.shift() is linear time complexity. Use pop() instead. The order won't be perfect layer by layer, given if you had say 1000 items in a queue, it could take 1000*1000 operations to go through them. Not worth it...

sonar-plugin/sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/JavaScriptPluginTest.java
```diff
@@ -36,7 +36,7 @@
 
 class JavaScriptPluginTest {
 
-  private static final int BASE_EXTENSIONS = 35;
+  private static final int BASE_EXTENSIONS = 34;
```
zglicz: grrr, remove this test or make it less exact. No one cares about the exact count.

packages/shared/src/helpers/configuration.ts
vdiez: the thing is that we detect bundles by default, and the sonar property needs to be set explicitly to `false` in order to disable it. I will keep it this way, to be coherent with all the other `should*()` helpers

packages/jsts/src/embedded/analysis/analyzer.ts
vdiez: the types are defined differently, did not investigate deeper why is that. some do not have optional metrics like this. we can tackle this later on, I agree would be better to use same return types everywhere

packages/css/src/analysis/analyzer.ts
vdiez: `setGlobalConfiguration` has a default empty object as default param value, so yeah no need for this if

vdiez: oh but yeah I need to handle it in `setGlobalConfiguration` not to override the already existing global config if we pass undefined

packages/jsts/src/analysis/projectAnalysis/file-stores/index.ts
vdiez: they were all `any` before, I prefer to write the types

zglicz: in this case, isn't the findFiles method argument typed? So here this is extraneous?

vdiez: ok, now I see your point. you're totally right, I removed them

PR #5458, opened 2025-06-12
Handle exceptions during AnalyzeProject

zglicz:
Part of


vdiez (approved):

vdiez (approved):
seems we are missing some coverage on the js side

PR #5470, opened 2025-06-19
JS-771 Support extra Sonar property for timeout of node bridge

zglicz:
[JS-771](https://sonarsource.atlassian.net/browse/JS-771)

[JS-771]: https://sonarsource.atlassian.net/browse/JS-771?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server.mjs
+++ b/server.mjs
@@ -19,7 +19,7 @@ if (isMainThread) {
   const port = process.argv[2];
   const host = process.argv[3];
   const debugMemory = process.argv[4] === 'true';
-  const timeoutSeconds = Number(process.argv[5]) || undefined;
+  const timeoutSeconds = Number(process.argv[5]) || 0;
 
   Promise.resolve().then(async () => {
     return start(
--- a/server.mjs
+++ b/server.mjs
@@ -18,6 +18,9 @@ if (isMainThread) {
   const port = process.argv[2];
   const host = process.argv[3];
   const debugMemory = process.argv[4] === 'true';
+  const timeoutArg = process.argv[5];
+  const timeoutSeconds =
+    timeoutArg !== undefined && !isNaN(Number(timeoutArg)) ? Number(timeoutArg) : undefined;
 
   Promise.resolve().then(async () => {
     return start(
@@ -25,6 +28,7 @@ if (isMainThread) {
       host,
       await createWorker(new URL(import.meta.url), { debugMemory }),
       debugMemory,
+      timeoutSeconds,
     );
   });
 }
--- a/server.mjs
+++ b/server.mjs
@@ -13,14 +13,13 @@ if (isMainThread) {
    * port - port number on which server.mjs should listen
    * host - host address on which server.mjs should listen
    * debugMemory - print memory usage
+   * timeoutSeconds - timeout for the node server to wait before shutting down. If not provided or 0,
    */
 
   const port = process.argv[2];
   const host = process.argv[3];
   const debugMemory = process.argv[4] === 'true';
-  const timeoutArg = process.argv[5];
-  const timeoutSeconds =
-    timeoutArg !== undefined && !isNaN(Number(timeoutArg)) ? Number(timeoutArg) : undefined;
+  const timeoutSeconds = Number(process.argv[5]) || undefined;
 
   Promise.resolve().then(async () => {
     return start(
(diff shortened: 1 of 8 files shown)


server.mjs
```diff
@@ -18,13 +18,17 @@ if (isMainThread) {
...
+  const timeoutSeconds =
+    timeoutArg !== undefined && !isNaN(Number(timeoutArg)) ? Number(timeoutArg) : undefined;
 
   Promise.resolve().then(async () => {
     return start(
       Number.parseInt(port, 10),
       host,
       await createWorker(new URL(import.meta.url), { debugMemory }),
       debugMemory,
+      timeoutSeconds,
```
zglicz, line 30: I didn't know, but if an undefined is passed to a method with a default value, it will still use the default value. That's nice :)

vdiez: yes, passing no argument or passing undefined is the same

server.mjs
```diff
@@ -18,13 +18,17 @@ if (isMainThread) {
   const port = process.argv[2];
   const host = process.argv[3];
   const debugMemory = process.argv[4] === 'true';
+  const timeoutArg = process.argv[5];
+  const timeoutSeconds =
+    timeoutArg !== undefined && !isNaN(Number(timeoutArg)) ? Number(timeoutArg) : undefined;
```
vdiez:
```suggestion
  const timeoutSeconds = Number.parseInt(timeoutArg) || undefined;
```

vdiez (approved):
just a small simplification

can you test in SonarLint pls before merging? kill IDE and node should stay forever in background

server.mjs
vdiez: forget about this, 0 needs to be supported

sonar-plugin/bridge/src/main/java/org/sonar/plugins/javascript/bridge/BridgeServerImpl.java
```diff
@@ -296,6 +276,10 @@ private NodeCommand initNodeCommand(BridgeServerConfig serverConfig, File script
     return nodeCommandBuilder.build();
   }
 
+  public static int getNodeTimeoutSeconds(Configuration config) {
+    return config.getInt(NODE_TIMEOUT_PROPERTY).orElse(DEFAULT_TIMEOUT_SECONDS);
+  }
+
```
vdiez: I would rather have all these helper function to get properties to `JsTsContext.java`

sonar-plugin/standalone/src/main/java/org/sonar/plugins/javascript/standalone/StandaloneParser.java
```diff
@@ -41,27 +40,19 @@
 
 public class StandaloneParser implements AutoCloseable {
 
-  private static final int DEFAULT_TIMEOUT_SECONDS = 5 * 60;
-
   private final BridgeServerImpl bridge;
 
   public StandaloneParser() {
-    this(Http.getJdkHttpClient());
```
vdiez: I think we need to check with security about this change. Why was it needed in the first place.

sonar-plugin/bridge/src/main/java/org/sonar/plugins/javascript/bridge/BridgeServerImpl.java
```diff
@@ -64,19 +64,21 @@ private enum Status {
...
+  // internal property to set the timeout in seconds for Node Bridge, 0 - indicates no timeout
+  public static final String NODE_TIMEOUT_PROPERTY = "sonar.javascript.node.timeout";
   public static final String SONARJS_EXISTING_NODE_PROCESS_PORT =
     "SONARJS_EXISTING_NODE_PROCESS_PORT";
   private static final Gson GSON = new Gson();
   private static final String BRIDGE_DEPLOY_LOCATION = "bridge-bundle";
 
   private final NodeCommandBuilder nodeCommandBuilder;
-  private final int timeoutSeconds;
+  private int timeoutSeconds = DEFAULT_TIMEOUT_SECONDS;
```
vdiez: revert back and use different variables for request timeouts and node timeout

server.mjs
```diff
@@ -13,18 +13,21 @@ if (isMainThread) {
    * port - port number on which server.mjs should listen
    * host - host address on which server.mjs should listen
    * debugMemory - print memory usage
+   * timeoutSeconds - timeout for the node server to wait before shutting down. If not provided or 0,
    */
 
   const port = process.argv[2];
   const host = process.argv[3];
   const debugMemory = process.argv[4] === 'true';
+  const timeoutSeconds = Number(process.argv[5]) || undefined;
```
vdiez:
```suggestion
  const timeoutSeconds = Number(process.argv[5]) || 0;
```

vdiez (approved):

PR #5482, opened 2025-06-25
JS-777 Clean up cirrus.yml

vdiez:
[JS-777](https://sonarsource.atlassian.net/browse/JS-777)

Removing SonarJsIntegrationTest.java. It was testing the availability of our private REST API, as SonarLint for Visual Studio was using it directly, without the Java plugin. Now all IDEs use the plugin directly, we don't have any public contract

[JS-777]: https://sonarsource.atlassian.net/browse/JS-777?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -103,7 +103,7 @@ npm_cache_definition: &NPM_CACHE
 eslint_build_cache_definition: &ESLINT_BUILD_CACHE
   eslint_build_cache:
     folder: ${CIRRUS_WORKING_DIR}/lib
-    fingerprint_script: echo "JAVA$(git rev-parse HEAD)"
+    fingerprint_script: echo "ESLINT$(git rev-parse HEAD)"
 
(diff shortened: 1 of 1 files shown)


.cirrus.yml
```diff
@@ -71,7 +103,7 @@ npm_cache_definition: &NPM_CACHE
 eslint_build_cache_definition: &ESLINT_BUILD_CACHE
   eslint_build_cache:
     folder: ${CIRRUS_WORKING_DIR}/lib
-    fingerprint_script: cat package-lock.json
+    fingerprint_script: echo "JAVA$(git rev-parse HEAD)"
```
zglicz: why is that? Why do you add the JAVA? is that a copy paste of the java_coverage_cache_definition?

PR #5485, opened 2025-06-26
JS-780 Do not apply analysis scope exclusions/inclusions to tsconfig.json and package.json files

vdiez:
[JS-780](https://sonarsource.atlassian.net/browse/JS-780)

[JS-780]: https://sonarsource.atlassian.net/browse/JS-780?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/jsts/src/analysis/projectAnalysis/file-stores/source-files.ts
+++ b/packages/jsts/src/analysis/projectAnalysis/file-stores/source-files.ts
@@ -23,6 +23,7 @@ import { JsTsAnalysisInput } from '../../analysis.js';
 import { accept, shouldIgnoreFile } from '../../../../../shared/src/helpers/filter/filter.js';
 import { readFile, toUnixPath } from '../../../../../shared/src/helpers/files.js';
 import { filterPathAndGetFileType } from '../../../../../shared/src/helpers/filter/filter-path.js';
+import { dirname } from 'node:path/posix';
 
 export const UNINITIALIZED_ERROR = 'Files cache has not been initialized. Call loadFiles() first.';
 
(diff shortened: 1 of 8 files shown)


packages/jsts/src/analysis/projectAnalysis/file-stores/source-files.ts
```diff
@@ -112,25 +106,32 @@ export class SourceFileStore implements FileStore {
...
 
   setup(baseDir: string) {
     this.baseDir = baseDir;
     this.newFiles = [];
   }
 
-  async process(file: Dirent, filePath: string, fileType: FileType) {
-    if (isAnalyzableFile(file.name)) {
+  async processFile(file: Dirent, filePath: string) {
+    if (isAnalyzableFile(file.name) && !this.anyParentIsIgnored(filePath)) {
```
zglicz, line 118: ignored check should be first. As in, if something is ignored, no need to think if its even analyzable

packages/jsts/src/analysis/projectAnalysis/file-stores/source-files.ts
```diff
@@ -112,25 +106,32 @@ export class SourceFileStore implements FileStore {
...
       const fileContent = await this.getFileContent(filePath);
+      const fileType = filterPathAndGetFileType(filePath);
       // we don't call shouldIgnoreFile because the isJsTsExcluded method has already been
       // called while walking the project tree
```
zglicz, line 122: should the isJsTsExcluded method be a type-guard? so that we don't get in sticky situations?

zglicz (approved):

packages/jsts/src/analysis/projectAnalysis/file-stores/source-files.ts
vdiez: we can tackle this once the analyze-[jsts|css|yaml|html] endpoints do not exist

PR #5521, opened 2025-07-04
JS-793 Add SensorContext as parameter to JsAnalysisConsumer::doneAnalysis

vdiez:
org.sonar.api.batch.sensor.Sensor is a module sensor, invoked as many times as modules are in a project. This seems to have caused the problem in Jasmine/.Net scanner, as modules were set up by the dotnet scanner and this caused the JsTsSensor to be invoked multiple times (with different workDir paths). While we explore the option to use ProjectSensor, we can add the context (which contains `workdir`) so that consumers can orchestrate what files belong together my module

[JS-793](https://sonarsource.atlassian.net/browse/JS-793)

[JS-793]: https://sonarsource.atlassian.net/browse/JS-793?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


edward-gonzales-sonarsource:
@karim-ouerghemmi-sonarsource -- this is still pending review.

PR #5522, opened 2025-07-04
JS-794 Implement ProjectSensor instead of Sensor

vdiez:


vdiez:
Closing as it seems projectSensors are not working propertly with lits

PR #5523, opened 2025-07-04
JS-794 Make all Sensors global

vdiez:


karim-ouerghemmi-sonarsource:
(empty push to get a fresh CI build that I can use for testing the impact on sec analysis)

its/plugin/plugins/consumer-plugin/src/main/java/org/sonar/samples/javascript/consumer/ConsumerSensor.java
```diff
@@ -45,7 +45,7 @@ public ConsumerSensor(Consumer consumer) {
 
   @Override
   public void describe(SensorDescriptor descriptor) {
-    descriptor.name("Consumer Sensor");
+    descriptor.global().name("Consumer Sensor");
```
karim-ouerghemmi-sonarsource: I know this is just an IT but out of curiosity and because the situation is similar to SonarJasmin, was this needed? I would expect the fact that this extends `ProjectSensor` to be enough.

karim-ouerghemmi-sonarsource (approved):
LGTM!

- Tested with SonarJasmin locally and it seems to work fine
- Tested on all ITs and Benchmarks of SonarSecurity and it seems to work fine: https://github.com/SonarSource/sonar-security/pull/6219

For alignment, was [AbstractExternalIssuesSensor](https://github.com/SonarSource/SonarJS/blob/e69a1270a2a2343e4fede94b01a2fdc24724ab9a/sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/external/AbstractExternalIssuesSensor.java#L43) not adapted for a reason or just forgotten/intentionally left out?

its/plugin/plugins/consumer-plugin/src/main/java/org/sonar/samples/javascript/consumer/ConsumerSensor.java
vdiez: I would expect so, let me try

vdiez:
thanks! totally overlooked [AbstractExternalIssuesSensor](https://github.com/SonarSource/SonarJS/blob/e69a1270a2a2343e4fede94b01a2fdc24724ab9a/sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/external/AbstractExternalIssuesSensor.java#L43)

PR #5530, opened 2025-07-09
JS-799 Generate licenses for Java dependencies

zglicz:
[JS-799](https://sonarsource.atlassian.net/browse/JS-799)

For the most part, managed to use the existing maven plugin to generate this.
Few caveats:
1) The files in the THIRD_PARTY_FOLDERS have format: <library_name>-<license_name>.txt
2) We sometimes get issues when downloading the license, I think thats when it has multiple licenses with the same url, we get throttles. I counter this by linking to a local file, for the known licenses. This will not be generated on each build, only on demand.

The way to run it is from the root:
> mvn -f sonar-plugin/sonar-javascript-plugin clean install -Dlicense-regeneration

[JS-799]: https://sonarsource.atlassian.net/browse/JS-799?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sonar-plugin/sonar-javascript-plugin/src/main/license-resources/missing-licenses-config-file.xml
```diff
@@ -0,0 +1,25 @@
...
+      <artifactId>xz</artifactId>
+
+      <matchLicenses>
+        <license>
+          <name>0BSD</name>
+          <distribution>repo</distribution>
+        </license>
+      </matchLicenses>
+      <licenses>
+        <!-- Manually fix the existing license nodes: -->
```
vdiez, line 16: why? the found one is not txt?

sonar-plugin/sonar-javascript-plugin/src/main/resources/licenses/THIRD_PARTY_LICENSES/com.google.code.findbugs.jsr305_apache_v2.txt
```diff
@@ -0,0 +1,73 @@
+Apache License
```
vdiez, line 1: I understand that all files in this folder have been automatically downloaded?

sonar-plugin/sonar-javascript-plugin/src/main/license-resources/missing-licenses-config-file.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0"?>
+
+<licenseSummary>
+  <dependencies>
+    <dependency>
+      <groupId>org.tukaani</groupId>
+      <artifactId>xz</artifactId>
```
vdiez, line 7: why do we need to specify an artifact? can't this be applied to all LICENSES named `0BSD`?

vdiez (approved):
very nice!!

sonar-plugin/sonar-javascript-plugin/src/main/license-resources/missing-licenses-config-file.xml
zglicz: This is for org.tukaani:xz <- in their pom they do not have a url for the license, hence here we set the url for it.

sonar-plugin/sonar-javascript-plugin/src/main/resources/licenses/THIRD_PARTY_LICENSES/com.google.code.findbugs.jsr305_apache_v2.txt
zglicz: They have been automatically generated. As for this license and URL where it was downloaded from, I was often getting 429 about too many requests, therefore, I've added it to the repo and this is just copied from the local license file I've added in: src/main/license-resources/Apache-2.0.txt

sonar-plugin/sonar-javascript-plugin/src/main/license-resources/missing-licenses-config-file.xml
zglicz: Answered above. And to add to that, many times licenses include the years and the author names, so linking to their local license is better. Bot sure that is the case here, but still.

PR #5531, opened 2025-07-09
JS-799 Perform the same for npm

zglicz:
[JS-799](https://sonarsource.atlassian.net/browse/JS-799)

Ehhh, this was so much easier in JS than in Java.
Points to take away:
1) We use 'license-checker' library to get the dependencies
2) We have ±1000 licenses we generate now. This takes up 1mb (when unpacked) of space.
3) Following the same naming convention as in Java: <package_name>-<license_name>.txt
4) Storing it in the root `licenses/` folder. This way, when we `npm pack` I can just reference it.
5) We have the `npm run license-regeneration` script if we want to refresh this. For NPM/JS/TS - this seems much more stable than the Java coounter part, so we might consider this to be on the main build path.

[JS-799]: https://sonarsource.atlassian.net/browse/JS-799?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vdiez (approved):
yeah, this looks much better than in java. I'm wondering if we really need to use package names in the license filenames, this way we have many times the same license. Wouldn't be enough with just the license name itself? do we care about what dependency has what license?

zglicz:
As per the task description: 

> Note that some of the information, in particular LICENSE.txt, may be redundant.

> That’s okay, in particular in view of the fact that some inner archives may be distributed separately.

PR #5534, opened 2025-07-10
JS-802 Automagically keep node versions in sync between package.json, pom.xml and NodeDeprecationWarnings

zglicz:
[JS-802](https://sonarsource.atlassian.net/browse/JS-802)

We keep getting into trouble when we update versions and don't always remember to manage the properties correctly.

[JS-802]: https://sonarsource.atlassian.net/browse/JS-802?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sonar-plugin/bridge/src/main/java/org/sonar/plugins/javascript/bridge/NodeDeprecationWarning.java
+++ b/sonar-plugin/bridge/src/main/java/org/sonar/plugins/javascript/bridge/NodeDeprecationWarning.java
@@ -44,19 +44,17 @@ public class NodeDeprecationWarning {
         NODE_PROPERTIES_FILE
       );
     ) {
-      if (inputStream != null) {
-        props.load(inputStream);
-      }
+      props.load(inputStream);
+      MIN_SUPPORTED_NODE_VERSION = Version.parse(props.getProperty("node.version.min"));
+      RECOMMENDED_NODE_VERSIONS = Arrays.asList(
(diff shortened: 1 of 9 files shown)


sonar-plugin/bridge/src/main/java/org/sonar/plugins/javascript/bridge/NodeDeprecationWarning.java
```diff
@@ -28,29 +32,46 @@
 public class NodeDeprecationWarning {
 
   private static final Logger LOG = LoggerFactory.getLogger(NodeDeprecationWarning.class);
+  private static final String NODE_PROPERTIES_FILE = "/node-info.properties";
+  static final Version MIN_SUPPORTED_NODE_VERSION;
```
saberduck, line 36: hey @zglicz , since you are touching this. I would avoid using `Version` from sonar API, it might change and it's out of your control. Imo it's better to have your own class for this.

zglicz: @saberduck - I'm not sure I understand. I want to be able to represent the version requirements in a similar way as this Version class operates (sans build number), it is versioned, it gives me a comparison. Why would I re-write it on my own?

vdiez (approved):
LGTM!

PR #5546, opened 2025-07-16
Avoid repeated extractions of embedded node runtime

dreis2211:
Greetings,

I noticed that our Sonar scans were showing the following debug logs
```
20:52:16.795 DEBUG Currently installed Node.js version: . Available version in analyzer: v22.11.0
20:52:16.795 DEBUG Lock acquired for extraction
20:52:16.803 DEBUG Extracting embedded node to /path/to/our/.sonar/js/node-runtime/node
```
As you can see the currently installed version seems to be empty and the scanner consequently doing the extraction of the node runtime. Unfortunately it does so for each scan.

On closer inspection, this seems to be caused by the `js/node-runtime/version.txt` file being empty. This is caused by the `InputStream` of the file from the JAR being already fully consumed in `EmbeddedNode::isDifferent` through `readAllBytes`. When the file should be then written via `Files.copy(...)` it's basically copying an "empty" stream.

For pragmatic reasons, this PR just opens the InputStream again before trying to copy things.

Would appreciate if this is picked up as it saves us from extracting ~110+ MB on each scan. Let me know if you need anything from me.

Cheers,
Christoph

### DIFF
--- a/sonar-plugin/bridge/src/main/java/org/sonar/plugins/javascript/bridge/EmbeddedNode.java
+++ b/sonar-plugin/bridge/src/main/java/org/sonar/plugins/javascript/bridge/EmbeddedNode.java
@@ -194,7 +194,7 @@ public class EmbeddedNode {
       if (Files.exists(targetVersion) && !isDifferent(versionIs, targetVersion)) {
         LOG.debug("Skipping node deploy. Deployed node has latest version.");
       } else {
-        extractWithLocking(is, versionIs, targetRuntime, targetDirectory);
+        extractWithLocking(is, targetRuntime, targetDirectory);
       }
       // we try to run 'node -v' to test that node is working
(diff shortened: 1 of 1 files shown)


vdiez:
Thank you for spotting and fixing this! I guess we did not spot it as the most normal use case is always running the scanner in CI. Can you explain your use case for this to be an actual issue?

dreis2211:
@vdiez Our scans run in a job on Jenkins, so CI as well. I guess you rather mean a containerized environment where every job run creates a fresh container anyway. That's not the case for us. We have dedicated persistent VMs (managed through KVM on bare-metal hypervisors).
So in other words, our environment can benefit from this because the `js/node-runtime/version.txt` file actually persists across job runs. Hope that gives you a rough idea :)

vdiez:
ok, that makes all the sense. Thanks again!

PR #5567, opened 2025-07-22
JS-810 Support sonar properties related to analysis scope when provided values are files instead of paths

vdiez:


packages/shared/src/helpers/filter/filter-path.ts
```diff
@@ -67,7 +67,7 @@ export function filterPathAndGetFileType(filePath: string): FileType | undefined
 
 function fileIsTest(filePath: string): boolean {
   const testPaths = getTestPaths();
-  if (!testPaths?.some(testPath => filePath.startsWith(testPath))) {
+  if (!testPaths?.some(testPath => filePath === testPath || filePath.startsWith(`${testPath}/`))) {
```
zglicz, line 70: do you need the / at the end of startsWith? Wouldn't that then cover both cases? Or would that then not adhere to the docs?
```suggestion
  if (!testPaths?.some(testPath => filePath.startsWith(`${testPath}`))) {
```

vdiez: that would detect files which are not in the list, for example:

`sonar.sources=src/foo`

would catch as source

`src/foo-bar.js`

Adding the slash to startsWith forces that the checked file is inside the given folder. The other part of the condition is that the sources property is using files instead of paths.

PR #5569, opened 2025-07-23
JS-815 Use same WebSocket connection on each Sensor execution

vdiez:
Part of

### DIFF
--- a/sonar-plugin/bridge/src/main/java/org/sonar/plugins/javascript/bridge/BridgeServerImpl.java
+++ b/sonar-plugin/bridge/src/main/java/org/sonar/plugins/javascript/bridge/BridgeServerImpl.java
@@ -252,8 +252,14 @@ public class BridgeServerImpl implements BridgeServer {
   }
 
   void establishWebSocketConnection() {
-    try {
+    if (client == null) {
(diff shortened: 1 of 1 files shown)


sonar-plugin/bridge/src/main/java/org/sonar/plugins/javascript/bridge/BridgeServerImpl.java
```diff
@@ -252,6 +252,10 @@ void startServer(BridgeServerConfig serverConfig) throws IOException {
   }
 
   void establishWebSocketConnection() {
+    if (client != null && client.isOpen()) {
+      LOG.debug("WebSocket connection already established");
+      return;
+    }
```
zglicz, line 258: Can we get to a case, like with dotnet, that they have concurrent calls to analyzeProject with different params? So that we actually have intertwining initLinter and actual analysis calls?

I think also you should be able to test this, no?

vdiez: in principle, there is no concurrency at all. All sensors and modules are executed serially. But just want to play safe here.

PR #5570, opened 2025-07-23
JS-496 Improve UI message for missing tsconfig file

vdiez:
[JS-496](https://sonarsource.atlassian.net/browse/JS-496)

[JS-496]: https://sonarsource.atlassian.net/browse/JS-496?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/jsts/src/analysis/projectAnalysis/analyzeWithProgram.ts
+++ b/packages/jsts/src/analysis/projectAnalysis/analyzeWithProgram.ts
@@ -96,7 +96,7 @@ async function analyzeProgram(
   }
   if (missingTsConfig) {
     const msg =
-      "At least one tsconfig.json was not found in the project. Please run 'npm install' for a more complete analysis. Check analysis logs for more details.";
+      "At least one referenced/extended tsconfig.json was not found in the project. Please run 'npm install' for a more complete analysis. Check analysis logs for more details.";
     warn(msg);
     results.meta.warnings.push(msg);
   }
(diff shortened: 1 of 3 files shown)


packages/jsts/src/analysis/projectAnalysis/analyzeWithProgram.ts
```diff
@@ -96,7 +96,7 @@ async function analyzeProgram(
   }
   if (missingTsConfig) {
     const msg =
-      "At least one tsconfig.json was not found in the project. Please run 'npm install' for a more complete analysis. Check analysis logs for more details.";
+      "At least one referenced/extended tsconfig.json was not found in the project. Please run 'npm install' for a more complete analysis. Check analysis logs for more details.";
```
zglicz, line 99: Can a user have many of these? Either way, we are still printing this one time per tsconfig.json found. Why not print also the path we are referring to?

Looking deeper, seems like we will also log this separately with referencing the tsconfig path here - https://github.com/SonarSource/SonarJS/blob/1ea58ff388600c1b0fb3054252aed778104b6a97/packages/jsts/src/program/program.ts#L81

Seems like we could consolidate both? Perhaps we wouldn't have to pass the `missingTsConfig` flag so much around. It could be part of the exception? And we would add it to the warnings a few lines above?

vdiez: this will not make the program creation fail. we added this to be more permissive even if node_modules was not present. So the error above is catched, but this one does not raise an exception. I agree that printint the problematic `extends` would be better, but it's not that simple because we would have the guess what was the value inside the `tsconfig.json`, instead of the resolved paths that Typescript will use (and those are the only ones we receive)

PR #5573, opened 2025-07-23
JS-818 tsconfigPaths should accept absolute paths

vdiez:
[JS-818](https://sonarsource.atlassian.net/browse/JS-818)

[JS-818]: https://sonarsource.atlassian.net/browse/JS-818?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/jsts/src/analysis/projectAnalysis/file-stores/tsconfigs.ts
+++ b/packages/jsts/src/analysis/projectAnalysis/file-stores/tsconfigs.ts
@@ -29,8 +29,7 @@ import {
   shouldClearFileToTsConfigCache,
   shouldClearTsConfigCache,
 } from '../../../../../shared/src/helpers/configuration.js';
-import { toUnixPath } from '../../../../../shared/src/helpers/files.js';
-import { join, basename, normalize } from 'node:path/posix';
+import { basename, normalize } from 'node:path/posix';
 import { Minimatch } from 'minimatch';
 import type { Dirent } from 'node:fs';
 import { FileStore } from './store-type.js';
@@ -234,11 +233,10 @@ export class TsConfigStore implements FileStore {
     } as TsConfigJson;
   }
 
-  setup(baseDir: string) {
+  setup() {
     this.foundPropertyTsConfigs = [];
     this.foundLookupTsConfigs = [];
-    this.providedPropertyTsConfigs = getTsConfigPaths().map(tsConfigPath => {
-      const tsConfig = toUnixPath(join(baseDir, tsConfigPath.trim()));
+    this.providedPropertyTsConfigs = getTsConfigPaths().map(tsConfig => {
       return {
         path: tsConfig,
         pattern: new Minimatch(tsConfig, { nocase: true, matchBase: true, dot: true }),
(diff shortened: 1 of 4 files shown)


packages/jsts/tests/analysis/tsconfigs.test.ts
```diff
@@ -68,6 +68,16 @@ describe('tsconfigs', () => {
     expect(tsConfigStore.getCacheOrigin()).toEqual('property');
   });
 
+  it('should validate the provided TSconfig files when absolute paths are used', async () => {
+    setGlobalConfiguration({ baseDir: fixtures });
+    await initFileStores(fixtures);
+    setTsConfigPaths(tsConfigStore.getTsConfigs().map(tsconfig => tsconfig));
```
zglicz: whats with the map(x => x)?

packages/jsts/src/analysis/projectAnalysis/file-stores/tsconfigs.ts
```diff
@@ -234,11 +233,10 @@ export class TsConfigStore implements FileStore {
     } as TsConfigJson;
   }
 
-  setup(baseDir: string) {
+  setup() {
```
zglicz, line 236: so this is overriding the base method, but we are not interested at all in the baseDir?

packages/jsts/tests/analysis/tsconfigs.test.ts
```diff
@@ -232,7 +242,9 @@ describe('tsconfigs', () => {
 
     expect(
       (console.error as Mock<typeof console.error>).mock.calls.map(call => call.arguments[0]),
-    ).toContain('Failed to find any of the provided tsconfig.json files: tsconfig.fake.json');
+    ).toContain(
+      `Failed to find any of the provided tsconfig.json files: ${join(fixtures, 'tsconfig.fake.json')}`,
```
zglicz, line 246: that means that we will always now provide the absolute path to the tsconfig we were looking for?

zglicz (commented):
So what do we support now? Is it only absolute paths or paths relative from the baseDir?

packages/jsts/tests/analysis/tsconfigs.test.ts
vdiez: just removed it 5mins ago

packages/jsts/src/analysis/projectAnalysis/file-stores/tsconfigs.ts
vdiez: correct, we don't need it any more. some of the other methods happen the same. this is valid in js/ts

packages/jsts/tests/analysis/tsconfigs.test.ts
vdiez: yep

vdiez:
we were only supporting relative paths

zglicz:
Now we support both?

vdiez:
yep

vdiez:
we were supporting both before: https://github.com/SonarSource/SonarJS/blob/f7176d6a190068d55354acf95a2bb9d9b6859749/sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/analysis/TsConfigProvider.java#L172

zglicz (approved):

PR #5576, opened 2025-07-24
ESLINTJS-18 Use ESLint built-in rule `no-unreachable-loop` instead of our own `no-one-iteration-loop`

vdiez:
[ESLINTJS-18](https://sonarsource.atlassian.net/browse/ESLINTJS-18)

[ESLINTJS-18]: https://sonarsource.atlassian.net/browse/ESLINTJS-18?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/jsts/src/rules/S1751/rule.ts
guilhermesimoes, line 1: Hey @vdiez, I think this breaking change is missing from the release notes 🤔

vdiez: indeed, and many others. I forgot to add the issues from the ESLint Jira project. Thanks for the heads up!

vdiez: I would have released a minor instead of a patch... :(

guilhermesimoes: I see the notes now on the [CHANGELOG.md](https://github.com/SonarSource/SonarJS/blob/master/packages/jsts/src/rules/CHANGELOG.md), thanks!

Btw I wonder if [the release](https://github.com/SonarSource/SonarJS/releases/tag/11.3.0.34350) should be updated as well 🙏

vdiez: the SonarJS releases correctly show those tickets, check previous releases. SonarJS release cadence is higher than the eslint plugin

guilhermesimoes: Ah, I see now! Once again, thanks!

PR #5639, opened 2025-08-18
JS-839 Clean up Maven dependencies

vdiez:
[JS-839](https://sonarsource.atlassian.net/browse/JS-839)

we were only using Apache commons lang in tests and there was a [CVE](https://sonarsource.atlassian.net/browse/SSF-863). For the usage we need I think we better don't have this dependency

Making sure that running `mvn dependency:analyze` now does not show any "used and undeclared" or "unused and declared" dependencies in any of the maven modules of the project.

Apache lang3 is not explicitly mentioned in our pom files any more. Using `mvn dependency:tree -Dverbose` it was visible which version of lang3 was picked up, and ordering of dependency declaration matters in Maven.

[JS-839]: https://sonarsource.atlassian.net/browse/JS-839?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -68,6 +68,7 @@
     <junit.version>5.13.4</junit.version>
     <mockito.version>5.19.0</mockito.version>
     <slf4j.version>2.0.17</slf4j.version>
+    <jetbrains.annotations.version>25.0.0</jetbrains.annotations.version>
     <sonar.version>25.8.0.112029</sonar.version>
     <sonar.api.version>13.0.0.3026</sonar.api.version>
     <sonar-orchestrator.version>5.6.2.2625</sonar-orchestrator.version>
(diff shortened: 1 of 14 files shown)


pom.xml
```diff
@@ -149,11 +149,6 @@
...
         <version>2.20.0</version>
       </dependency>
-      <dependency>
-        <groupId>commons-lang</groupId>
-        <artifactId>commons-lang</artifactId>
-        <version>2.6</version>
-      </dependency>
       <dependency>
         <groupId>org.apache.commons</groupId>
         <artifactId>commons-lang3</artifactId>
```
zglicz: Isn't the commons-lang3 the offending depedency, and you still have it, no?

zglicz (approved):

pom.xml
vdiez: yes, because there is a conflict version sonar-api which depends on 3.14 and commons-compress that depends on 3.17. I need to enforce what version to use. but I'm still looking at it

vdiez: https://github.com/SonarSource/sonar-plugin-api/blob/0546bdf694917fc3f21933cfe3c39c78cfb3765e/settings.gradle#L11

PR #5685, opened 2025-09-03
JS-854 Dont raise on arrow function as lambda argument

zglicz:
[JS-854](https://sonarsource.atlassian.net/browse/JS-854)

See community report - https://community.sonarsource.com/t/typescript-s2301-false-positive/147981

This rule shouldn't raise when this isn't a callable methods from outside.

[JS-854]: https://sonarsource.atlassian.net/browse/JS-854?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/its/ruling/src/test/expected/jsts/eigen/typescript-S2301.json
+++ b/its/ruling/src/test/expected/jsts/eigen/typescript-S2301.json
@@ -1,7 +1,4 @@
 {
-"eigen:src/app/Components/ArtworkFilter/Filters/useMultiSelect.ts": [
-76
-],
 "eigen:src/app/store/config/features.ts": [
 277
 ],
--- a/package-lock.json
+++ b/package-lock.json
@@ -116,13 +116,13 @@
     },
     "node_modules/@angular-eslint/bundled-angular-compiler": {
       "version": "20.2.0",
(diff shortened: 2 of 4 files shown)


vdiez (approved):
LGTM!

its/ruling/src/test/expected/jsts/eigen/typescript-S2301.json
```diff
@@ -1,7 +1,4 @@
 {
-"eigen:src/app/Components/ArtworkFilter/Filters/useMultiSelect.ts": [
-76
-],
```
zglicz, line 4: This is the snippet on which we won't raise anymore:
```
  const handleSelect = useCallback(
    (option: FilterData, updatedValue: boolean) => {
      if (updatedValue) {
        // Append the paramValue
        setNextParamValues((prev) => {
          return [...prev, getParamValue(option)]
        })
      } else {
        // Remove the paramValue
        setNextParamValues((prev) => {
          return prev.filter((value) => {
            return value !== getParamValue(option)
          })
        })
      }
    },
    [getParamValue]
  )
```

PR #5697, opened 2025-09-08
Js 757 Refactored package.json cache

vdiez:


packages/jsts/src/analysis/projectAnalysis/file-stores/package-jsons.ts
```diff
@@ -87,29 +83,27 @@ export class PackageJsonStore implements FileStore {
...
+    this.paths.get(parent)?.children.add(dir);
+    this.paths.set(dir, {
+      children: new Set(),
+      parent,
+    });
   }
 
   async postProcess() {
-    if (!this.packageJsons) {
+    if (!this.packageJsons || !this.baseDir) {
```
zglicz: package.Jsons is initialized to be a map in the initializer of the properties. Is it then necessary?

Also, you check `!this.baseDir` in 2 places now. Perhaps, you could have a helper to check `isInitialized` or even a guard that would perform the check and throw if it is not initialized

packages/jsts/src/linter/linter.ts
```diff
@@ -32,9 +32,11 @@ import globalsPkg from 'globals';
 import { APIError } from '../../../shared/src/errors/error.js';
 import { pathToFileURL } from 'node:url';
 import * as ruleMetas from '../rules/metas.js';
-import { extname } from 'node:path/posix';
+import { extname, dirname } from 'node:path/posix';
 import { defaultOptions } from '../rules/helpers/configs.js';
 import merge from 'lodash.merge';
+import { getDependencies } from '../rules/helpers/package-jsons/dependencies.js';
+import { getClosestPackageJSONDir } from '../rules/helpers/package-jsons/closest.js';
```
zglicz, line 39: logically these are both related to package-jsons, would you want to import them from: `.../helpers/package-jsons/index.js` ?

packages/jsts/src/rules/helpers/find-up/find-minimatch.ts
```diff
@@ -0,0 +1,67 @@
...
+import { ComputedCache } from '../cache.js';
+
+interface Stats {
+  isFile(): boolean;
+}
+
+export interface Filesystem {
+  readdirSync: (typeof fs)['readdirSync'];
+  readFileSync: (typeof fs)['readFileSync'];
+  statSync: (typeof fs)['statSync'];
```
zglicz, line 30: why the sync versions of fs?

packages/jsts/src/rules/helpers/find-up/find-minimatch.ts
```diff
@@ -0,0 +1,67 @@
...
+      for (const entry of filesystem.readdirSync(from)) {
+        const fullEntryPath = join(from, entry.toString());
+        if (matcher.match(basename(fullEntryPath))) {
+          let stats: Stats;
+
+          // the resource may not be available
+          try {
+            stats = filesystem.statSync(fullEntryPath);
+          } catch (error) {
+            // todo: this is testable and should be tested
```
zglicz, line 50: doesn't this throw if the file isn't present? Should be easy to test, no?

packages/jsts/src/rules/helpers/find-up/find-minimatch.ts
```diff
@@ -0,0 +1,67 @@
...
+
+          if (stats.isFile()) {
+            files.push({
+              path: fullEntryPath,
+              content: filesystem.readFileSync(fullEntryPath),
+            });
+          }
+        }
+      }
+    } catch {}
```
zglicz, line 64: hm? thats not nice

packages/jsts/src/rules/helpers/package-jsons/closest.ts
```diff
@@ -0,0 +1,31 @@
...
+import { closestPatternCache } from '../find-up/closest.js';
+
+export function getClosestPackageJSONDir(dir: string, topDir?: string) {
+  const closestPackageJSONDir = closestPatternCache
+    .get(PACKAGE_JSON)
+    .get(topDir ? toUnixPath(topDir) : '/')
+    .get(toUnixPath(dir))?.path;
+  if (closestPackageJSONDir) {
+    return dirname(closestPackageJSONDir);
+  }
```
zglicz, line 30: so an implicit undefined return? I thought you were against that?

packages/jsts/src/rules/helpers/package-jsons/dependencies.ts
```diff
@@ -0,0 +1,56 @@
...
+        result.add(dependency.name);
+      });
+    });
+  }
+  return result;
+});
+/**
+ * Retrieve the dependencies of all the package.json files available for the given file.
+ *
+ * @param dir context.filename
```
zglicz: is this a comment? is this correct?

packages/jsts/src/rules/helpers/package-jsons/dependencies.ts
```diff
@@ -0,0 +1,56 @@
...
+ * @param topDir working dir, will search up to that root
+ * @returns
```
zglicz: no explanation here as well? Not sure how useful this jsdoc is?

packages/jsts/src/rules/helpers/package-jsons/index.ts
```diff
@@ -0,0 +1,56 @@
...
+ * along with this program; if not, see https://sonarsource.com/license/ssal/
+ */
+
+import type { File, PathTree } from '../files.js';
+import { MinimatchCache } from '../find-up/find-minimatch.js';
+import { dependenciesCache } from './dependencies.js';
+import { closestPatternCache } from '../find-up/closest.js';
+import { patternInParentsCache } from '../find-up/all-in-parent-dirs.js';
+
+export const PACKAGE_JSON = 'package.json';
```
zglicz, line 24: how is this constant used in other places and should it be-renamed to sth like: PACKAGE_JSON_CACHE_PATTERN if its mainly or only used for the cache?

packages/jsts/src/rules/helpers/package-jsons/index.ts
```diff
@@ -0,0 +1,56 @@
...
+    const currentPackageJson = packageJsons.get(dir);
+    const allPackageJsons = [];
+    if (parent) {
+      closestCache.set(dir, closestCache.get(parent));
+      allPackageJsons.push(...allPackageJsonsCache.get(dir));
+    }
+    if (currentPackageJson) {
+      closestCache.set(dir, currentPackageJson);
+      allPackageJsons.push(currentPackageJson);
+    }
```
zglicz: I see here we are depending on the order of the allPaths, that an ancestor needs to be added before the children. I do not like that we might set in the map twice on line 38 and 42 if both `parent` and `currentPackageJson` are present.

packages/jsts/src/rules/helpers/cache.ts
```diff
@@ -0,0 +1,54 @@
...
+    this.cache.set(key, value);
+  }
+
+  has(key: K) {
+    return this.cache.has(key);
+  }
+
+  delete(key: K) {
+    return this.cache.delete(key);
+  }
```
zglicz, line 45: is this ever used?

packages/jsts/src/rules/S1607/rule.ts
```diff
@@ -34,7 +32,9 @@ import {
 } from '../helpers/index.js';
 import * as meta from './generated-meta.js';
 import type { TSESTree } from '@typescript-eslint/utils';
-import { dirname } from 'path/posix';
+import { dirname } from 'node:path';
```
zglicz, line 35: no posix?

packages/jsts/tests/analysis/package-jsons.test.ts
```diff
@@ -42,33 +50,46 @@ describe('files', () => {
...
+    expect(closestPackageJsonCache.size).toEqual(1);
+
+    //dependencies cache is filled on demand
+    expect(dependenciesCache.size).toEqual(0);
+    getDependencies(baseDir, baseDir);
+    expect(dependenciesCache.size).toEqual(1);
+    expect(dependenciesCache.has(baseDir)).toEqual(true);
   });
 
   it('should ignore malformed the package.json files', async ({ mock }) => {
```
zglicz: ehhh, this `it` shouldn't have `the`

zglicz (approved):
I've read through it. Some places are indeed cleaner.

packages/jsts/src/linter/linter.ts
vdiez: I wanted to leave in the index just the common data used by all helpers

packages/jsts/src/rules/helpers/find-up/find-minimatch.ts
vdiez: these may be run from rules, where we are in sync mode

packages/jsts/src/rules/helpers/find-up/find-minimatch.ts
vdiez: just left the same logic as before in find-up. need to be mocked as the file is listed by readdirsync but not stat-able

packages/jsts/src/rules/helpers/find-up/find-minimatch.ts
vdiez: just left the same logic as before in find-up. we stop when no access to reading the path

packages/jsts/src/rules/helpers/package-jsons/closest.ts
vdiez: nope, I was never against

packages/jsts/src/rules/helpers/package-jsons/index.ts
vdiez: yes, the Map keeps the ordering, that why i changed the implementation of the package-json stores.

packages/jsts/src/rules/helpers/cache.ts
vdiez: nope, can be removed

packages/jsts/src/rules/S1607/rule.ts
vdiez: no, needs to work in the ESLint plugin where the context.filename is set with backslashes in windows. path/posix.dirname will not work as expected in that scenario

PR #5736, opened 2025-09-18
Remove quickfix list

zglicz:
There was a list of Sonar rules that we allow a quick fix for. We are removing it for a few reasons:
1) This data can be managed in the rspec repo by the 'quickfix' property
2) We might easily forget to update this list and never even ship the quick fix to users


zglicz:
Should be done!

vdiez (approved):
LGTM!

PR #5785, opened 2025-10-02
JS-871 Add S7923

vdiez:
[JS-871](https://sonarsource.atlassian.net/browse/JS-871)

[JS-871]: https://sonarsource.atlassian.net/browse/JS-871?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/css/src/rules/S7923/rule.ts
```diff
@@ -0,0 +1,141 @@
...
+
+function isAngleInvalid(angle: number | undefined) {
+  if (angle === undefined) {
+    return false;
+  }
+  // Check if angle is within range of 90° (60° to 120°) or within range of 270° (240° to 300°)
+  return incorrectAngles.some(incorrectAngle => {
+    const lowerBound = incorrectAngle - tolerance;
+    const upperBound = incorrectAngle + tolerance;
+    return angle >= lowerBound && angle <= upperBound;
```
zglicz: Also, not sure if you want strict or not script to allow for the 60, 120 and 240 and 300
```suggestion
    return Math.abs(angle - incorrectAngle) <= tolerance
```

packages/css/src/rules/S7923/rule.ts
```diff
@@ -0,0 +1,141 @@
...
+        return getAngleFromString(node.nodes[0].value);
+      }
+      break;
+    }
+    case 'rotate3d': {
+      const nodes = getWordNodes(node.nodes);
+      if (
+        nodes?.length === 4 &&
+        // 3rd node refers to rotation in z-axis
+        Number.parseInt(nodes[2].value) === 1 &&
```
zglicz: 3rd as in nodes[2] ? Perhaps in a comment reference the index

packages/css/src/rules/S7923/rule.ts
```diff
@@ -0,0 +1,141 @@
...
+        angleRegex.test(nodes[3].value)
+      ) {
+        return getAngleFromString(nodes[3].value);
+      }
+      break;
+    }
+    case 'matrix':
+    // 90°: matrix(0, 1, -1, 0, tx, ty) where first param = cos(90°) = 0, second param = sin(90°) = 1
+    // 270°: matrix(0, -1, 1, 0, tx, ty)
```
zglicz, line 96: then what about the 3rd argument? Is it important -1 and 1? why even put in these comments when the important thing you want is to just get the angle from the matrix and that is abstracted to a method?

packages/css/src/rules/S7923/rule.ts
```diff
@@ -0,0 +1,141 @@
...
+  if (angle && unit) {
+    return normalize360Angle(unitConverter(Number.parseFloat(angle), unit as CSSUnits, 'deg'));
+  }
+}
+
+function getAngleFromMatrix(node: postcssValueParser.FunctionNode) {
+  const nodes = getWordNodes(node.nodes);
+  if (
+    (node.value.toLowerCase() === 'matrix3d' && nodes.length === 16) ||
+    (node.value.toLowerCase() === 'matrix' && nodes.length === 6)
```
zglicz, line 115: do you care if its 16 or 6? Wouldn't js just fill the other arguments as undefined and in your case you only care about the first 2?

packages/css/src/rules/S7923/rule.ts
```diff
@@ -0,0 +1,141 @@
...
+  ) {
+    const angleRad = Math.atan2(
+      Number.parseFloat(nodes[1].value),
+      Number.parseFloat(nodes[0].value),
+    );
+    return normalize360Angle(angleRad * (180 / Math.PI));
```
zglicz, line 121: just curios, why the parenthesis?

packages/css/src/rules/S7923/unit.test.ts
```diff
@@ -0,0 +1,160 @@
...
+ *
+ * You should have received a copy of the Sonar Source-Available License
+ * along with this program; if not, see https://sonarsource.com/license/ssal/
+ */
+import { StylelintRuleTester } from '../../../tests/tools/tester/index.js';
+import { rule, messages } from './rule.js';
+import { describe, it } from 'node:test';
+
+const ruleTester = new StylelintRuleTester(rule.ruleName);
+describe('S7423', () => {
```
zglicz: S7923?

packages/css/src/rules/S7923/unit.test.ts
```diff
@@ -0,0 +1,160 @@
...
+  it('detects valid angle with rotateZ', () =>
+    ruleTester.valid({
+      codeFilename: 'noRestriction.html',
+      code: `
+<html lang="en">
```
zglicz, line 29: that's a valid language :)

packages/css/src/rules/S7923/unit.test.ts
```diff
@@ -0,0 +1,160 @@
...
+					transform: rotateZ(1turn);
+				}
+			}
+		</style>
+	</head>
+	<body>
+		<main>
+			Page Content
+		</main>
+	</body>
```
zglicz: is this necessary?

packages/css/src/rules/S7923/unit.test.ts
```diff
@@ -0,0 +1,160 @@
...
+
+const ruleTester = new StylelintRuleTester(rule.ruleName);
+describe('S7423', () => {
+  it('detects valid angle with rotateZ', () =>
+    ruleTester.valid({
+      codeFilename: 'noRestriction.html',
+      code: `
+<html lang="en">
+	<head>
+		<title>Page with some content</title>
```
zglicz: is this necessary?

packages/css/src/rules/S7923/unit.test.ts
```diff
@@ -0,0 +1,160 @@
...
+ * See the Sonar Source-Available License for more details.
+ *
+ * You should have received a copy of the Sonar Source-Available License
+ * along with this program; if not, see https://sonarsource.com/license/ssal/
+ */
+import { StylelintRuleTester } from '../../../tests/tools/tester/index.js';
+import { rule, messages } from './rule.js';
+import { describe, it } from 'node:test';
+
+const ruleTester = new StylelintRuleTester(rule.ruleName);
```
zglicz, line 23: Can these testcases be shortened, as in the sample css code?
I'd be even more interested to have a nice test suite of the getAngle method and just a few tests where you check the orientation and rotations of the css.

Perhaps you can use AI to generate coverage of the angle, or maybe wait for some cool solution from the hackathon?

zglicz (approved):

packages/css/src/rules/S7923/rule.ts
vdiez: just checking it's a valid CSS function call.

packages/css/src/rules/S7923/rule.ts
vdiez: C/P

PR #5919, opened 2025-11-12
JS-824 Add rule S7639

loris-s-sonarsource:
[JS-824](https://sonarsource.atlassian.net/browse/JS-824)

Part of 

[JS-824]: https://sonarsource.atlassian.net/browse/JS-824?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/jsts/src/rules/S7639/rule.ts
```diff
@@ -0,0 +1,107 @@
...
+  meta: generateMeta(meta, {
+    messages: {
+      reviewBlockchainSeedPhrase: `Revoke and change this seed phrase, as it is compromised.`,
+    },
+  }),
+  create(context: Rule.RuleContext) {
+    let isBlockchainModuleImported = false;
+    const hardcodedVariables = new Map<string, estree.Node>();
+
+    function isHardcodedString(expr: estree.Expression): boolean {
```
loris-s-sonarsource, line 42: I intentionally removed concatenations, as this does not seem to make sense in the context of this type of code

zglicz:
Hi @loris-s-sonarsource - I think you need to sync rspec with your new rule in order for all of the build to work.
For this workflow, we have a script to help with generating the boilerplate `tools/new-rule.mts`. Although, I think you might get away with just syncing the rspec (as this drives the generated SXXXX java classes).

EDIT: As written in chat, the rspec metadata.json is missing the `compatibleLanguages` field

zglicz:
@loris-s-sonarsource  - you need to add the `compatibleLanguages` into the rspec file - https://github.com/SonarSource/rspec/blob/master/rules/S7639/javascript/metadata.json

The jsons in the SonarJS repo will be copied over from rspec automatically.

loris-s-sonarsource:
@zglicz ohhhh ok thank you !! I thought I could trick the system by directly uploading a correct json to this PR before merging my rspec PR

zglicz:
> @zglicz ohhhh ok thank you !! I thought I could trick the system by directly uploading a correct json to this PR before merging my rspec PR

No, sadly we cannot do that! Ideally, we wouldn't even store these json and htmls in git and just generate them on the fly. However, there are some tools like the rules.sonarsource.com or the rspec.sonar page that look into this exact folder from the git repo and expect these files to be present there. If they looked at the jar we would be fine, but they clone the repo and expect them tracked there.

loris-s-sonarsource:
A test is broken, but it seems that it's because of Next:

```
Error:  Error during SonarScanner Engine execution
java.lang.IllegalStateException: Fail to download plugin [dbdjavafrontend] into /home/runner/.sonar/_tmp/fileCache6122588177309986843.tmp
	at org.sonar.scanner.bootstrap.PluginFiles.downloadBinaryTo(PluginFiles.java:145)
	at org.sonar.scanner.bootstrap.PluginFiles.download(PluginFiles.java:107)
...
Caused by: okhttp3.internal.http2.StreamResetException: stream was reset: INTERNAL_ERROR
	at okhttp3.internal.http2.Http2Stream$FramingSource.read(Http2Stream.kt:398)
	at okhttp3.internal.connection.Exchange$ResponseBodySource.read(Exchange.kt:324)
...
```

packages/jsts/src/rules/S7639/cb.fixture.ts
```diff
@@ -0,0 +1,36 @@
...
+
+// TronWeb with hardcoded string
+const wallet = TronWeb.fromMnemonic("candy sugar pudding cream honey rich smooth crumble sweet treat"); // Noncompliant {{Revoke and change this seed phrase, as it is compromised.}}
+
+// Template literal without expressions
+const wallet = Wallet.fromPhrase(`abandon abandon abandon abandon abandon abandon abandon abandon about`); // Noncompliant {{Revoke and change this seed phrase, as it is compromised.}}
+
+// CommonJS require
+const ethers = require('ethers');
+const wallet = ethers.HDNodeWallet.fromPhrase("test test test test test test test test test test test junk"); // Noncompliant {{Revoke and change this seed phrase, as it is compromised.}}
```
loris-s-sonarsource:
```suggestion
const wallet = ethers.HDNodeWallet.fromPhrase("fake test test test test test test test test test test junk"); // Noncompliant {{Revoke and change this seed phrase, as it is compromised.}}
```

vdiez (approved):
LGTM, but I would recommend adding some sources for ruling tests in `its/sources/jsts/custom`, and adding the expected findings in `its/ruling/src/test/expected/jsts/file-for-rules` in lits format. Just to make sure the rule is executed during analysis

loris-s-sonarsource:
Noted, thank you! I am keeping that for my monday actions

PR #6139, opened 2026-01-05
JS-971 Auto-apply eslint-plugin label to Jira tickets

zglicz:
## Summary
- Add a new GitHub workflow that automatically applies the `eslint-plugin` label to Jira tickets when a PR modifies files in `packages/jsts/src/rules/`

## How it works
1. Triggers on PR opened, synchronized (new commits), or edited (title changed)
2. Only runs when files in `packages/jsts/src/rules/**` are changed (via path filter)
3. Extracts the Jira key (e.g., `JS-1234`) from the PR title
4. Uses Jira REST API to add the `eslint-plugin` label to the ticket

## Test plan
- [x] This PR itself modifies a file in `packages/jsts/src/rules/` (minor comment fix in S100/rule.ts)
- [ ] After the PR is created, verify the workflow runs
- [x] Check that JS-971 gets the `eslint-plugin` label added

### DIFF
--- a/sonar-plugin/javascript-checks/pom.xml
+++ b/sonar-plugin/javascript-checks/pom.xml
@@ -51,7 +51,7 @@
             </goals>
             <configuration>
               <ruleSubdirectory>javascript</ruleSubdirectory>
-              <vcsBranchName>rule/S1125-javascript-quickfix</vcsBranchName>
+              <vcsBranchName>dogfood-automerge</vcsBranchName>
               <targetDirectory>${project.basedir}/../../resources/rule-data/javascript</targetDirectory>
             </configuration>
           </execution>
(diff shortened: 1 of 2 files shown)


sonar-plugin/javascript-checks/pom.xml
```diff
@@ -51,7 +51,7 @@
             </goals>
             <configuration>
               <ruleSubdirectory>javascript</ruleSubdirectory>
-              <vcsBranchName>rule/S1125-javascript-quickfix</vcsBranchName>
+              <vcsBranchName>dogfood-automerge</vcsBranchName>
```
zglicz, line 54: yyyy, broke this from a previous PR. Forgot to revert

zglicz (commented):
I've verified that in a previous commit that the label is applied. I've reverted it and removed the label from the ticket now as well. But I swear it was there.

vdiez (approved):

PR #6146, opened 2026-01-06
JS-90 Fix S3801: Handle exhaustive switch statements over union types

zglicz:
## Summary
- Fixes [JS-90](https://sonarsource.atlassian.net/browse/JS-90): S3801 incorrectly flags functions with exhaustive switch statements over TypeScript union types as having inconsistent returns
- Uses TypeScript parser services to detect exhaustive switches over union types and enums
- Prevents false positives when all cases end with return or throw statements

## Technical Details
When a switch statement handles all variants of a TypeScript union type and all cases return or throw, TypeScript knows the code after the switch is unreachable. However, ESLint's code path analysis doesn't account for TypeScript's type narrowing, so it incorrectly thinks there's an implicit return at the end of the function.

The fix:
1. Tracks switch statements within each function context
2. When checking for implicit return, uses TypeScript type services to detect exhaustive switches
3. A switch is considered exhaustive if:
   - The discriminant is a union type or enum
   - All union members are covered by case clauses
   - All cases end with return or throw (not fall-through)

## Test Plan
- [x] Added unit tests for exhaustive switches over union types
- [x] Added unit tests for exhaustive switches over enums
- [x] Added unit tests for discriminated unions (the exact case from the Jira ticket)
- [x] Added unit tests for non-exhaustive switches (should still raise)
- [x] Added unit tests for switches with fall-through (should still raise)
- [x] Existing tests pass

[JS-90]: https://sonarsource.atlassian.net/browse/JS-90?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/jsts/src/rules/S3801/rule.ts
```diff
@@ -193,3 +225,130 @@ function isVoidType(typeNode: TSESTree.TypeNode) {
...
+    type.symbol?.flags === ts.SymbolFlags.EnumMember
+  );
+}
+
+/**
+ * Checks if a case clause ends with a return or throw statement.
+ * A case "ends" with return/throw if the last statement (ignoring trailing break) is return/throw,
+ * or if all paths through the case end with return/throw.
+ */
+function caseEndsWithReturnOrThrow(caseClause: estree.SwitchCase): boolean {
```
vdiez: not convinced by the whole login from this point onwards. It's an exhaustive list of Nodes to be checked. Why not for loop, while, do..while, etc... there could be returns in those as well. Hard to maintain with current approach, wdyt?

vdiez (commented):
Having some doubts about this approach

zglicz:
Good point! I've simplified the approach:

- Now only checks if the **last statement** (ignoring trailing `break`) is a direct `return`/`throw`
- For complex control flow (if/else, loops, nested switches, etc.), we conservatively return `false` and let ESLint's code path analysis handle those cases
- This is more maintainable and safer - we won't incorrectly suppress issues, just miss some optimization opportunities

The key insight is that we only need to eliminate the "no case matches" phantom path that ESLint creates because it doesn't know the switch is exhaustive. ESLint already correctly tracks paths through complex control flow within cases.

Changes:
- Removed `caseEndsWithReturnOrThrow`, `blockEndsWithReturnOrThrow`, `ifEndsWithReturnOrThrow`, `statementEndsWithReturnOrThrow`
- Added simpler `lastStatementExits` that only checks direct return/throw (with nested block support)
- Updated tests to remove complex control flow cases (those will be flagged, which is the conservative/safe behavior)

zglicz:
Simplified further. The key insight is that **all fall-throughs eventually reach the last case**. So we only need to check:

1. The switch is exhaustive (covers all union/enum values)
2. The last case exits (return/throw)

This correctly handles the fall-through pattern:
```typescript
case 'a': console.log('a'); // falls through
case 'b': return 2;  // last case exits
```
All paths return via the last case, so no FP.

The code now only checks `lastCaseExits()` instead of `every(case => exits)`.

vdiez (approved):

PR #6167, opened 2026-01-09
JS-1063 Fix S7727 false positives on non-array method calls

zglicz:
## Summary
- Fix S7727 decorator to check if the callee is actually an array-like type before reporting
- Filters out false positives on non-array objects with methods like `find`, `filter`, `map` (e.g., `tree.find(key)`, `query.filter('active')`)
- Add test cases for non-array method calls

## Context
The rule was incorrectly flagging calls like `tree.find(key)` in `simplified-regex-character-class.ts` and `finder.find(node)` in `S5843/rule.ts` because it only checked if the argument was a callable with >1 parameters, not whether the callee was actually an array.

## Test plan
- [x] Unit tests pass
- [x] Added test cases for non-array `.find()` and `.filter()` calls


packages/jsts/src/rules/S7727/unit.test.ts
```diff
@@ -52,6 +52,22 @@ describe('S7727', () => {
           // Inline functions are not flagged by the base rule
           code: `[1, 2, 3].map(x => x * 2);`,
         },
+        {
+          // Non-array .find() - should not be flagged (tree.find(key) is not an array method)
+          code: `
+            interface Tree<K, V> { find(key: K): { key: K; value: V } | undefined; }
+            declare const tree: Tree<number, string>;
+            const entry = tree.find(42);
```
vdiez: you are not passing function references in these unit tests

packages/jsts/src/rules/S7727/decorator.ts
```diff
@@ -48,32 +52,68 @@ export function decorate(rule: Rule.RuleModule): Rule.RuleModule {
...
+ * Checks if the callee object (e.g., `arr` in `arr.map(fn)`) is an array-like type.
+ */
+function isCalleeArrayLike(
+  node: estree.Node,
+  context: Rule.RuleContext,
+  services: RequiredParserServices,
+): boolean {
+  const callExpr = getParent(context, node);
+  if (callExpr?.type !== 'CallExpression') {
+    return true; // Can't determine, assume it's an array to be safe
```
vdiez: I know you have handled multiple PRs where if we did not know we went back to original. But not sure is the best, I think we should change this policy to ONLY raise when typechecker is available AND we know for sure it's an array. Otherwise we don't raise. I would rather raise only if we are sure. Specially now that we will shortly merge the PR where we analyze always with type information

vdiez (changes requested):

packages/jsts/src/rules/S7727/unit.test.ts
zglicz: This is a FP we are no longer raising on - https://next.sonarqube.com/sonarqube/project/issues?impactSoftwareQualities=RELIABILITY&issueStatuses=OPEN%2CCONFIRMED&id=org.sonarsource.javascript%3Ajavascript&open=8b9f812c-7e50-45c7-a204-032cc16e9aaf

packages/jsts/src/rules/S7727/decorator.ts
zglicz:
```suggestion
    return false; // Can't determine, lets not raise
```

packages/jsts/src/rules/S7727/decorator.ts
```diff
@@ -48,32 +52,68 @@ export function decorate(rule: Rule.RuleModule): Rule.RuleModule {
         return;
       }
 
-      const paramCount = getCallbackParameterCount(node, context);
+      const services = context.sourceCode.parserServices;
+      if (!isRequiredParserServices(services)) {
+        // Without type info, defer to base rule
+        context.report(reportDescriptor);
```
vdiez: we still report without type information

packages/jsts/src/rules/S7727/unit.test.ts
vdiez: so we should probably check that the parameter of the array method is a functionType

vdiez (changes requested):

vdiez (approved):

PR #6222, opened 2026-01-21
USER-1469 Fix S3735 false positive for Thenable with function property syntax

zglicz:
## Summary
- Fix `isThenable` helper to recognize `then` defined as function property (not just method signature)
- Resolves false positives for VSCode's `Thenable` interface and similar types
- Adds test cases for both Thenable styles

## Test plan
- [x] S3735 unit tests pass with new Thenable test cases
- [x] S4822 tests pass (also uses `isThenable`)
- [x] S7059 tests pass (also uses `isThenable`)

Fixes USER-1469

### DIFF
--- a/packages/jsts/src/rules/helpers/type.ts
+++ b/packages/jsts/src/rules/helpers/type.ts
@@ -144,7 +144,17 @@ export function isThenable(node: estree.Node, services: RequiredParserServices) 
   const mapped = services.esTreeNodeToTSNodeMap.get(node as TSESTree.Node);
   const tp = services.program.getTypeChecker().getTypeAtLocation(mapped);
   const thenProperty = tp.getProperty('then');
-  return Boolean(thenProperty && thenProperty.flags & ts.SymbolFlags.Method);
+  if (!thenProperty) {
+    return false;
+  }
+  // Check if it's declared as a method
+  if (thenProperty.flags & ts.SymbolFlags.Method) {
+    return true;
+  }
+  // Check if 'then' is a callable property (function type)
+  const checker = services.program.getTypeChecker();
+  const thenType = checker.getTypeOfSymbol(thenProperty);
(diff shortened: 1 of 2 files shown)


packages/jsts/src/rules/helpers/type.ts
```diff
@@ -144,7 +144,17 @@ export function isThenable(node: estree.Node, services: RequiredParserServices)
...
-  return Boolean(thenProperty && thenProperty.flags & ts.SymbolFlags.Method);
+  if (!thenProperty) {
+    return false;
+  }
+  // Check if it's declared as a method
+  if (thenProperty.flags & ts.SymbolFlags.Method) {
+    return true;
+  }
+  // Check if 'then' is a callable property (function type)
+  const checker = services.program.getTypeChecker();
```
vdiez: we are already using the checker on line 145, we should declare it there.

packages/jsts/src/rules/helpers/type.ts
```diff
@@ -144,7 +144,17 @@ export function isThenable(node: estree.Node, services: RequiredParserServices)
   const mapped = services.esTreeNodeToTSNodeMap.get(node as TSESTree.Node);
   const tp = services.program.getTypeChecker().getTypeAtLocation(mapped);
   const thenProperty = tp.getProperty('then');
-  return Boolean(thenProperty && thenProperty.flags & ts.SymbolFlags.Method);
+  if (!thenProperty) {
+    return false;
+  }
+  // Check if it's declared as a method
+  if (thenProperty.flags & ts.SymbolFlags.Method) {
```
vdiez: do we need to keep this check? I guess we can rely on the callSignatures as a unified solution for both methods and properties

vdiez (approved):
some minor improvements

zglicz:
good review @vdiez

PR #6226, opened 2026-01-21
JS-1135 Fix FP on S3735: Union types containing Promise and void/undefined

francois-mora-sonarsource:
## Summary

Fixes false positives in S3735 (`void-use` rule) for legitimate TypeScript union type patterns involving Promises.

**Jira Issue:** [JS-1135](https://sonarsource.atlassian.net/browse/JS-1135)

### Problem
The rule was incorrectly flagging `void` operator usage in these valid scenarios:
- Functions returning `void | Promise<void>` (optional async callbacks)
- Optional chaining with async methods: `obj?.asyncMethod()` → `Promise<T> | undefined`
- Union types combining Promises with "nothing" types (void, undefined, null)

### Solution
- Added new helper function `isThenableOrVoidUnion()` in `type.ts`
- Updated S3735 to use this new function instead of `isThenable()`
- The new function allows `void` when ALL union members are either:
  - Thenable (Promise-like), OR
  - "Nothing" types (void, undefined, null)
- Preserves existing `isThenable()` unchanged to avoid impacting other rules (S4822, S7059)

### Changes
- `packages/jsts/src/rules/helpers/type.ts` - Added `isThenableOrVoidUnion()` function
- `packages/jsts/src/rules/S3735/rule.ts` - Updated to use new helper
- `packages/jsts/src/rules/S3735/unit.test.ts` - Added comprehensive test cases

## Test plan

- [x] Added test cases for valid patterns (union types with Promise and void/undefined/null)
- [x] Added test cases for invalid patterns (unions with meaningful values like string/boolean)
- [x] All S3735 unit tests pass
- [x] Verified original false positive examples from the issue report

### Test coverage added:
**Valid (no false positives):**
- `void | Promise<void>` function returns
- Optional chaining: `maybeObject?.asyncMethod()`
- `Promise<void> | undefined/null`

**Invalid (correctly raises issues):**
- `string | Promise<void>` - meaningful value shouldn't be discarded
- `boolean | Promise<void>` - from logical expressions

[JS-1135]: https://sonarsource.atlassian.net/browse/JS-1135?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/jsts/src/rules/helpers/type.ts
+++ b/packages/jsts/src/rules/helpers/type.ts
@@ -143,8 +143,38 @@ export function isUndefinedOrNull(node: estree.Node, services: RequiredParserSer
 export function isThenable(node: estree.Node, services: RequiredParserServices) {
   const mapped = services.esTreeNodeToTSNodeMap.get(node as TSESTree.Node);
   const checker = services.program.getTypeChecker();
-  const tp = checker.getTypeAtLocation(mapped);
-  const thenProperty = tp.getProperty('then');
+  return hasThenMethod(checker.getTypeAtLocation(mapped), checker);
+}
+
+/**
+ * Checks if a node's type is either:
+ * - Thenable (Promise-like), OR
+ * - A union where ALL members are either thenable or "nothing" types (void, undefined, null)
+ *
+ * This is useful for rules like S3735 that allow voiding Promise operations,
+ * including cases like optional chaining (Promise<T> | undefined) or
+ * optional async callbacks (() => void | Promise<void>).
+ */
+export function isThenableOrVoidUnion(node: estree.Node, services: RequiredParserServices) {
+  const checker = services.program.getTypeChecker();
+  const type = getTypeFromTreeNode(node, services);
+  const unionTypes = type.isUnion() ? type.types : [type];
+  let hasThenable = false;
+  let allThenableOrVoid = true;
+
+  for (const unionType of unionTypes) {
+    const isThenable = hasThenMethod(unionType, checker);
+    const isNothingType =
+      (unionType.flags & (ts.TypeFlags.Void | ts.TypeFlags.Undefined | ts.TypeFlags.Null)) !== 0;
+    hasThenable ||= isThenable;
+    allThenableOrVoid &&= isThenable || isNothingType;
+  }
+
+  return hasThenable && allThenableOrVoid;
+}
+
+function hasThenMethod(type: ts.Type, checker: ts.TypeChecker): boolean {
+  const thenProperty = type.getProperty('then');
   if (!thenProperty) {
     return false;
   }
--- a/packages/jsts/src/rules/S3735/rule.ts
+++ b/packages/jsts/src/rules/S3735/rule.ts
@@ -22,7 +22,7 @@ import {
   generateMeta,
   isFunctionCall,
   isRequiredParserServices,
-  isThenable,
+  isThenableOrVoidUnion,
 } from '../helpers/index.js';
 import * as meta from './generated-meta.js';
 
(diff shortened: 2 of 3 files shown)


packages/jsts/src/rules/helpers/type.ts
```diff
@@ -142,14 +142,36 @@ export function isUndefinedOrNull(node: estree.Node, services: RequiredParserSer
...
+ */
+export function isThenableOrVoidUnion(node: estree.Node, services: RequiredParserServices) {
+  const type = getTypeFromTreeNode(node, services);
+  const types = type.isUnion() ? type.types : [type];
+
+  // All union members must be either thenable or "nothing" types
+  return types.every(t => {
+    // Check if it's a thenable (has 'then' method)
+    const thenProperty = t.getProperty('then');
+    const isThenableType = Boolean(thenProperty && thenProperty.flags & ts.SymbolFlags.Method);
```
zglicz: we have a new and improved way of checking isThenable - https://github.com/SonarSource/SonarJS/pull/6222

zglicz (commented):
please rebase

packages/jsts/src/rules/helpers/type.ts
francois-mora-sonarsource: Thanks for pointing this out, it should be aligned with the new way now.

packages/jsts/src/rules/helpers/type.ts
```diff
@@ -152,6 +152,39 @@ export function isThenable(node: estree.Node, services: RequiredParserServices)
...
+  // All union members must be either thenable or "nothing" types
+  return types.every(t => {
+    // Check if it's a thenable (has 'then' method with call signatures)
+    const thenProperty = t.getProperty('then');
+    if (thenProperty) {
+      const thenType = checker.getTypeOfSymbol(thenProperty);
+      if (thenType.getCallSignatures().length > 0) {
+        return true;
+      }
+    }
```
zglicz: there is a helper method isThenable right above this method, you should use it.

zglicz (changes requested):

zglicz:
Also, you can see the comment on your PR, this is generated by this part of CI: https://github.com/SonarSource/SonarJS/blob/8848d97bd3441f9f7fb34a17355ffb06506ce9d0/.github/workflows/build.yml#L820

This actually syncs the ruling jsons as well as show inline in the comment some of the snippets to better understand the nature of the changes. However, as part of the process, the github-user is pushing changes, and github doesn't run github actions. See the bottom of the comment on how to continue.

packages/jsts/src/rules/helpers/type.ts
francois-mora-sonarsource: Implemented, both functions share now a common private function.

packages/jsts/src/rules/helpers/type.ts
```diff
@@ -143,8 +143,38 @@ export function isUndefinedOrNull(node: estree.Node, services: RequiredParserSer
...
+  let hasThenable = false;
+  let allThenableOrVoid = true;
+
+  for (const unionType of unionTypes) {
+    const isThenable = hasThenMethod(unionType, checker);
+    const isNothingType =
+      (unionType.flags & (ts.TypeFlags.Void | ts.TypeFlags.Undefined | ts.TypeFlags.Null)) !== 0;
+    hasThenable ||= isThenable;
+    allThenableOrVoid &&= isThenable || isNothingType;
+  }
```
zglicz, line 171: I started writing out the suggestion below, but it wouldn't help much and potentially cause us to run the 'hasThenMethod' twice on all unionTypes. Leaving just for fun.

```suggestion
  const hasThenable = unionTypes.some(unionType => hasThenMethod(unionType, checker));
  const allThenableOrNothing = unionTypes.all(unionType => hasThenMethod(..) || isNothing(...))
```

zglicz (approved):
Looks good!

packages/jsts/src/rules/helpers/type.ts
francois-mora-sonarsource: I had the same question, but I ended up using a single loop to make it more efficient.

PR #6227, opened 2026-01-21
JS-1136 Fix SonarCloud quality gate code smells

zglicz:
## Summary
- S4721: Use `.at(-1)` instead of `[length - 1]` for array access
- S100: Remove unnecessary `any` cast, use proper optional chaining
- LabelEslintPlugin.yml: Fix workflow failure when PR has no Jira key in title

### LabelEslintPlugin workflow fix
The `fromJSON()` in the `env` block was being evaluated **before** the `if` condition was checked. When there's no Jira key, the secrets step is skipped, leaving `steps.secrets.outputs.vault` empty, causing `fromJSON()` to fail on the empty string.

Fix: Move JSON parsing from `env` block into `run` block using `jq`.

## Test plan
- [x] S4721 unit tests pass
- [x] S100 unit tests pass
- [x] Project compiles without type errors

### DIFF
--- a/.github/workflows/LabelEslintPlugin.yml
+++ b/.github/workflows/LabelEslintPlugin.yml
@@ -44,10 +44,11 @@ jobs:
       - name: Add eslint-plugin label
         if: steps.extract-jira-key.outputs.has_jira_key == 'true'
         env:
-          JIRA_USER: ${{ fromJSON(steps.secrets.outputs.vault).JIRA_USER }}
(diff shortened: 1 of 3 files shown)


zglicz:
**Note:** These issues appear on [SonarCloud](https://sonarcloud.io/project/issues?issueStatuses=OPEN%2CCONFIRMED&sinceLeakPeriod=true&id=SonarSource_SonarJS) but are not showing as open on [SonarQube NEXT](https://next.sonarqube.com/sonarqube/dashboard?id=org.sonarsource.javascript%3Ajavascript) 🤔

vdiez (approved):

PR #6228, opened 2026-01-21
JS-1137 Fix maven cache miss on PR branches

zglicz:
## Summary
Use `actions/cache@v5` consistently for both master and PR branches instead of mixing `gh-action_cache` with `actions/cache/restore`. This ensures cache compatibility and uses the latest action version.

- Branches: `actions/cache/restore@v5` (restore only)
- Master: `actions/cache@v5` (restore + save)

### Root cause
PR branches were using `actions/cache/restore@v5` while master used `gh-action_cache` which internally uses `actions/cache@v4.3.0`. This caused cache lookups to fail.

## Test plan
- [ ] Verify PR builds now hit the maven cache

### DIFF
--- a/.github/actions/maven-cache/action.yml
+++ b/.github/actions/maven-cache/action.yml
@@ -17,18 +17,14 @@ runs:
       if: github.ref_name != github.event.repository.default_branch
       uses: actions/cache/restore@v5
       with:
-        path: |
-          ~/.m2/repository
-          !~/.m2/repository/org/sonarsource/javascript
+        path: ~/.m2/repository
         key: maven-${{ runner.os }}-${{ inputs.cache-month }}-${{ inputs.maven-hash }}
         restore-keys: maven-${{ runner.os }}-${{ inputs.cache-month }}-
 
     # Full cache on default branch (save enabled)
-    # Note: Exclusion patterns don't work reliably with runs-on/cache used by gh-action_cache.
-    # Instead, we clean up SonarJS artifacts at the end of the job before post-job cache save.
     - name: Cache Maven dependencies (default branch)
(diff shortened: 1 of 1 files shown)


.github/actions/maven-cache/action.yml
```diff
@@ -17,18 +17,14 @@ runs:
       if: github.ref_name != github.event.repository.default_branch
       uses: actions/cache/restore@v5
       with:
-        path: |
-          ~/.m2/repository
-          !~/.m2/repository/org/sonarsource/javascript
```
zglicz, line 22: removing this as it doesn't correctly filter out these exclusions.

.github/actions/maven-cache/action.yml
```diff
@@ -17,18 +17,14 @@ runs:
...
+        path: ~/.m2/repository
         key: maven-${{ runner.os }}-${{ inputs.cache-month }}-${{ inputs.maven-hash }}
         restore-keys: maven-${{ runner.os }}-${{ inputs.cache-month }}-
 
     # Full cache on default branch (save enabled)
-    # Note: Exclusion patterns don't work reliably with runs-on/cache used by gh-action_cache.
-    # Instead, we clean up SonarJS artifacts at the end of the job before post-job cache save.
     - name: Cache Maven dependencies (default branch)
       if: github.ref_name == github.event.repository.default_branch
-      uses: SonarSource/gh-action_cache@v1
```
zglicz, line 31: womp womp - https://github.com/SonarSource/gh-action_cache/blob/master/action.yml#L75C7-L75C76
Still on 4.3 . 🤞 that this helps.

vdiez (approved):

PR #6249, opened 2026-01-26
JS-1160 Fix bump-versions workflow to update revision property

zglicz:
## Summary
Fix the `bump-versions` workflow to update the `<revision>` property instead of `<version>` tag.

The `pom.xml` uses:
```xml
<revision>11.8.0-SNAPSHOT</revision>
```

Not:
```xml
<version>11.8.0-SNAPSHOT</version>
```

The previous regex was not matching anything, causing the workflow to complete without making changes.

## Test plan
- [ ] Run the bump-versions workflow manually to verify it creates a PR with the version change


francois-mora-sonarsource (approved):
LGTM

PR #6291, opened 2026-01-29
JS-1194 feat(S2077): Add use getFullyQualifiedname & add sqlite3

loris-s-sonarsource:
Part of


zglicz:
@loris-s-sonarsource - could you fix up sq and re-run the CI, perhaps it will succeed

packages/jsts/src/rules/S5256/rule.ts
```diff
@@ -27,6 +27,18 @@ import { generateMeta, isPresentationTable, getElementType } from '../helpers/in
 import * as meta from './generated-meta.js';
 import type { JSXOpeningElement } from 'estree-jsx';
 
+/**
+ * Detects reusable table wrapper components that receive children via props spread.
+ * Returns true if the table has no JSX children AND has a JSXSpreadAttribute.
+ */
+function isReusableWrapperComponent(tree: TSESTree.JSXElement): boolean {
```
zglicz, line 34: this shouldn't be part of your PR no?

sonar-plugin/css/src/main/java/org/sonar/css/rules/AtRuleNoUnknown.java
```diff
@@ -27,7 +27,7 @@
 public class AtRuleNoUnknown implements CssRule {
 
   private static final String DEFAULT_IGNORED_AT_RULES =
-    "value,at-root,content,debug,each,else,error,for,function,if,include,mixin,return,warn,while,extend,use,forward,tailwind,apply,layer,container,theme,utility,custom-variant,source,plugin,config,reference,variant";
+    "value,at-root,content,debug,each,else,error,for,function,if,include,mixin,return,warn,while,extend,use,forward,tailwind,apply,layer,container,theme,utility,custom-variant,source,plugin,config,reference,variant,/^@.*/";
```
zglicz, line 30: as well as this?

zglicz (dismissed):
Seems like there was some issue with incorporating new changes.

Also, the logic in helpers/module.ts is quite convoluted. Would you consider using/testing helpers/module-ts.ts <- this is FQN computation using Typescript compiler to help with deep imports/requires. In theory it should be able to track import/requires cross files and you get more completeness. The downside is, if we can't create the Typescript program, we don't get the FQN. It's a trade-off, but more often now, we try to rely on TS for such things.

loris-s-sonarsource:
@zglicz Yeah, my rebasing added a tonne of new commits unfortunately. I am not very good at that it seems

edit: I just changed the bases to make it go away

vdiez:
good catch about the namespace import @loris-s-sonarsource !!

vdiez (approved):

loris-s-sonarsource:
thanks for everything @vdiez  !

PR #6298, opened 2026-01-30
JS-1198 Fix S2234 performance issue with complex destructuring patterns

zglicz:
## Summary

Fixes [USER-1508](https://sonarsource.atlassian.net/browse/USER-1508) - a performance issue where rule S2234 would hang when analyzing files with nested destructuring parameters where inner defaults reference outer parameters.

## Problem

TypeScript's `getResolvedSignature()` exhibits **exponential performance degradation** when resolving call expressions inside functions with this pattern:

```javascript
const Component = ({
  PropA: PropADefault,
  PropB: PropBDefault,
  // ...more properties
  Config: {
    PropA = PropADefault,  // default references outer param
    PropB = PropBDefault,
    // ...more properties
  },
  someFunction,
}) => {
  someFunction('test');  // getResolvedSignature is slow here
};
```

### Timing from standalone reproducer

| # of nested defaults | Time per `getResolvedSignature` call |
|---------------------|--------------------------------------|
| 1                   | ~4ms                                 |
| 3                   | ~6ms                                 |
| 5                   | ~67ms                                |
| 6                   | ~532ms                               |
| 7                   | **~5,756ms** (nearly 6 seconds!)     |

A real-world React component with 7 such properties caused analysis to hang for several minutes.

## Root Cause Analysis

The issue is in TypeScript's type inference engine, not specifically the `getResolvedSignature` API. Testing confirmed that the alternative approach of using `getTypeAtLocation` + `getCallSignatures` exhibits the **same exponential slowdown**:

| # of nested defaults | `getResolvedSignature` | `getTypeAtLocation` + `getCallSignatures` |
|---------------------|------------------------|-------------------------------------------|
| 7                   | ~5,829ms               | ~5,761ms                                  |

Both approaches go through the same underlying type inference code path when dealing with nested destructuring defaults that reference outer parameters.

### Why caching doesn't help

Each call expression only triggers a single signature resolution call. There's no redundant computation within our rule - the expensive work happens entirely inside TypeScript's type checker.

## Solution

Check if there are at least 2 identifier arguments **before** calling the expensive `resolveFunctionDeclaration()` (which internally calls `getResolvedSignature()`). 

Calls with 0-1 identifier arguments can never have swapped parameters anyway, so we skip type resolution entirely for those cases. This fix:

1. Is semantically correct - the rule can only detect issues with 2+ identifier arguments
2. Avoids triggering the TypeScript performance issue in most practical cases
3. Has minimal code change and no behavior change for valid inputs

## Reproducer (TypeScript issue)

This appears to be an underlying TypeScript performance bug. Standalone reproducer:

```js
// Save as reproducer.js, then: npm install typescript && node reproducer.js
const ts = require('typescript');

function generateCode(numProps) {
  const props = Array.from({ length: numProps }, (_, i) => String.fromCharCode(65 + i));
  const outerParams = props.map(p => `  Prop${p}: Prop${p}Default,`).join('\n');
  const nestedParams = props.map(p => `    Prop${p} = Prop${p}Default,`).join('\n');

  return `
const Component = ({
${outerParams}
  Config: {
${nestedParams}
  },
  doSomething,
}) => {
  doSomething('test');
};
`;
}

function measureGetResolvedSignature(code) {
  const fileName = 'test.ts';
  const sourceFile = ts.createSourceFile(fileName, code, ts.ScriptTarget.Latest, true, ts.ScriptKind.TS);
  const compilerHost = {
    getSourceFile: (name) => (name === fileName ? sourceFile : undefined),
    getDefaultLibFileName: () => 'lib.d.ts',
    writeFile: () => {},
    getCurrentDirectory: () => '/',
    getCanonicalFileName: (f) => f,
    useCaseSensitiveFileNames: () => true,
    getNewLine: () => '\n',
    fileExists: (f) => f === fileName,
    readFile: () => undefined,
  };
  const program = ts.createProgram([fileName], { strict: false, noEmit: true }, compilerHost);
  const checker = program.getTypeChecker();
  const source = program.getSourceFile(fileName);
  let callExpr;
  function visit(node) {
    if (ts.isCallExpression(node)) callExpr = node;
    ts.forEachChild(node, visit);
  }
  visit(source);
  const start = performance.now();
  checker.getResolvedSignature(callExpr);
  return performance.now() - start;
}

console.log('TypeScript version:', ts.version);
for (const n of [1, 2, 3, 4, 5, 6, 7]) {
  console.log(`${n} props: ${measureGetResolvedSignature(generateCode(n)).toFixed(2)} ms`);
}
```

## Test plan

- [x] Existing S2234 unit tests pass
- [x] Tested with the original reproducer file - completes in ~580ms instead of hanging

[JS-1198]: https://sonarsource.atlassian.net/browse/JS-1198?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[USER-1508]: https://sonarsource.atlassian.net/browse/USER-1508?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/jsts/src/rules/S2234/rule.ts
+++ b/packages/jsts/src/rules/S2234/rule.ts
@@ -46,16 +46,27 @@ export const rule: Rule.RuleModule = {
     const canResolveType = isRequiredParserServices(services);
 
     function checkArguments(functionCall: estree.CallExpression) {
+      // Extract argument names first (cheap operation)
+      const argumentNames = functionCall.arguments.map(arg => {
+        const argument = arg as TSESTree.Node;
+        return argument.type === 'Identifier' ? argument.name : undefined;
+      });
+
+      // Early exit: detecting swapped arguments requires at least 2 identifier arguments.
+      // Skip expensive type resolution if there aren't enough identifiers to check.
+      // This avoids a TypeScript performance issue where type inference for complex
+      // destructuring patterns with nested defaults can have exponential complexity.
+      const identifierCount = argumentNames.filter(name => name !== undefined).length;
+      if (identifierCount < 2) {
+        return;
+      }
+
       const resolvedFunction = resolveFunctionDeclaration(functionCall);
       if (!resolvedFunction) {
         return;
       }
 
       const { params: functionParameters, declaration: functionDeclaration } = resolvedFunction;
-      const argumentNames = functionCall.arguments.map(arg => {
-        const argument = arg as TSESTree.Node;
-        return argument.type === 'Identifier' ? argument.name : undefined;
-      });
 
       for (let argumentIndex = 0; argumentIndex < argumentNames.length; argumentIndex++) {
         const argumentName = argumentNames[argumentIndex];


vdiez (approved):
nice! how did you spot this issue?

zglicz:
> nice! how did you spot this issue?

It was a user ticket. I've added in the PR summary - [USER-1508](https://sonarsource.atlassian.net/browse/USER-1508)

[USER-1508]: https://sonarsource.atlassian.net/browse/USER-1508?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #6313, opened 2026-02-03
JS-1207 Skip minified/bundled JavaScript embedded in HTML/YAML files

zglicz:
## Summary

Apply minification detection to embedded JS snippets extracted from HTML and YAML files to skip analyzing minified libraries (e.g., jQuery) embedded in `<script>` tags.

Previously, minification filtering only applied to standalone `.js` files, causing false positives from rules like S7739 when analyzing minified code in HTML.

## Changes

- Added `acceptSnippet()` function that uses minification detection (average line length > 200) for embedded JS filtering
- Modified embedded JS builder to skip snippets that appear minified
- **Note**: Bundle detection is intentionally NOT applied to snippets, as the regex pattern `/* comment */ (function...` produces false positives on legitimate code with IIFEs and comments

### Why no bundle detection for snippets?

The bundle detection regex was matching normal code like:

```javascript
/* set theme color */
function setColor() {}
/* theme related */
(function() { ... })();  // IIFE - common JS pattern
```

This caused legitimate code in files like `ant-design/site/theme/static/template.html` to be incorrectly skipped. Minification detection (line length) is sufficient for embedded snippets.

## Test plan

- [x] Added unit tests for `acceptSnippet()` covering normal code and minified code
- [x] Added integration test with HTML file containing minified script
- [x] Verified existing HTML and YAML analyzer tests still pass
- [x] Verified ruling tests pass (no false positives on ant-design template.html)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -829,19 +829,10 @@ jobs:
           HEAD_REF: ${{ github.head_ref }}
           RULING_FAILED: ${{ steps.ruling.outcome == 'failure' }}
         run: |
-          # Generate report comparing against master
           git fetch origin master
-          node tools/ruling-report.js > ruling-report.md
 
-          # Check if there are ruling differences
-          if [ -s ruling-report.md ]; then
-            HAS_DIFFERENCES=true
-          else
-            HAS_DIFFERENCES=false
-          fi
-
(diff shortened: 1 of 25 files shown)


vdiez (approved):
LGTM

PR #6340, opened 2026-02-06
JS-1227 perf: Optimize S125 commented-out code detection

vdiez:
## Summary

- **Add cheap regex prefilter** (`/[;{}()=<>]/`) before the recognizer+parser pipeline — skips the entire expensive path for comments that can't possibly be JS code
- **Defer `new SourceCode()` construction** — only build it when `getLastToken` is actually needed, avoiding expensive token indexing for most comments
- **Short-circuit `couldBeJsCode`** — use `.some()` instead of `.filter().length` to bail on first matching line
- **Pre-compile regex patterns in `ContainsDetector`** — avoid `new RegExp()` on every `scan()` call per line (also fixed missing `g` flag in `replaceAll`)
- **Replace per-char regex whitespace check in `EndWithDetector`** with a `Set` char-code lookup
- **Fix `injectMissingBraces`** — proper min-balance algorithm that handles closing braces before opening braces (e.g. `}}}{{{`), plus single-pass loop instead of two `match()` calls + array allocations

## Benchmark (ESLint Linter with `stats: true`, 5 runs, best/avg)

| File | Master (S125) | PR (S125) | Speedup |
|------|--------------|-----------|---------|
| xqlint.js (100K lines) | 1014.9ms | 329.3ms | **3.1x** |
| typescript.js (88K lines) | 414.1ms | 109.9ms | **3.8x** |
| jshint.js (31K lines) | 288.1ms | 150.1ms | **1.9x** |
| ag-grid.js (29K lines) | 55.5ms | 15.3ms | **3.6x** |
| **Grand total S125** | **1772.6ms** | **604.6ms** | **2.9x** |

Overall S125 share of total lint time drops from ~23% to ~11%.

## Test plan

- [x] S125 unit tests pass
- [x] Verify no behavior regressions on ruling tests (CI green)

### DIFF
--- a/packages/jsts/src/rules/S125/rule.ts
+++ b/packages/jsts/src/rules/S125/rule.ts
@@ -177,17 +177,21 @@ function couldBeJsCode(input: string): boolean {
 
 function injectMissingBraces(value: string) {
   let balance = 0;
+  let minBalance = 0;
   for (let i = 0; i < value.length; i++) {
     if (value[i] === '{') {
       balance++;
     } else if (value[i] === '}') {
       balance--;
+      if (balance < minBalance) {
+        minBalance = balance;
+      }
     }
   }
-  if (balance > 0) {
-    return value + '}'.repeat(balance);
-  } else if (balance < 0) {
-    return '{'.repeat(-balance) + value;
+  const openToAdd = -minBalance;
+  const closeToAdd = balance - minBalance;
+  if (openToAdd > 0 || closeToAdd > 0) {
+    return '{'.repeat(openToAdd) + value + '}'.repeat(closeToAdd);
   }
   return value;
 }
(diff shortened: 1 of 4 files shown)


packages/jsts/src/rules/S125/rule.ts
```diff
@@ -158,28 +164,32 @@ function containsCode(value: string, context: Rule.RuleContext) {
...
+      balance--;
+    }
+  }
+  if (balance > 0) {
+    return value + '}'.repeat(balance);
+  } else if (balance < 0) {
+    return '{'.repeat(-balance) + value;
   }
+  return value;
 }
```
zglicz, line 197: :) this is a standard algorithmic problem. And if we'd actually want this to balance out parenthesis, we also need to check what was the minimal balance at any point when counting the parens. And that is how many open braces you need to add. This will in the end also impact how many you have open at the very end.

Like a snippet like:
  `}}}{{{`
To make it work, we need to add: `{{{` at the start and `}}}` at the end.

zglicz (commented):
What is the context of this? I don't see any ticket or anything in community

vdiez:
>What is the context of this? I don't see any ticket or anything in community

S125 has always been the slowest, we have a new comment on an old post complaining about it: https://community.sonarsource.com/t/eslint-plugin-sonarjs-performance-issues-on-large-codebase/138392/6

zglicz (approved):

PR #6350, opened 2026-02-09
JS-1234 Use new jacoco aggregate report paths property

leonardo-pilastri-sonarsource:
After the update of the jacoco plugin to properly support multiple aggregate report paths, the analysis parameter changed to a plural form, requiring the update of existing configurations.

Part of


zglicz:
@leonardo-pilastri-sonarsource - please rebase, we have fixed the master so CI should pass now

vdiez (approved):
Thanks @leonardo-pilastri-sonarsource !

PR #6368, opened 2026-02-11
JS-1317 Fix changelog workflow permissions by using default GITHUB_TOKEN

vdiez:
## Summary
- The `update-eslint-plugin-changelog` workflow was failing with a 403 error because it used a Vault-fetched Jira token (`hashicorp-vault-sonar-prod[bot]`) that lacks push permissions on the repo
- Removed the Vault secret step and `token:` override so the action uses the default `GITHUB_TOKEN`, which already has the required `contents:write` and `pull-requests:write` permissions
- Removed the now-unnecessary `id-token: write` permission from both the changelog workflow and its caller in `release_eslint_plugin.yml`

Failed run: https://github.com/SonarSource/SonarJS/actions/runs/21902454324/job/63233961431

## Test plan
- [x] Re-run the "Update ESLint Plugin Changelog" workflow manually and verify it succeeds

### DIFF
--- a/.github/workflows/release_eslint_plugin.yml
+++ b/.github/workflows/release_eslint_plugin.yml
@@ -66,7 +66,6 @@ jobs:
     needs: publish
     uses: ./.github/workflows/update-eslint-plugin-changelog.yml
     permissions:
-      id-token: write
       contents: write
       pull-requests: write
     with:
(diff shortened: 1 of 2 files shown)


vdiez:
seems to be working: https://github.com/SonarSource/SonarJS/pull/6369

zglicz (approved):

PR #6404, opened 2026-02-18
JS-1337 Use trie-based directory grouping for TypeScript program creation

zglicz:
## Summary
- Instead of creating a single massive TypeScript program for all pending files, group files by directory using a trie and create smaller, focused programs
- Top-down traversal: if a subtree has < 100 pending files, create one program for the whole subtree; otherwise process direct files at the current level and recurse into children
- Bonus files pulled in by TypeScript are consumed eagerly to avoid redundant programs later
- Pending counts are tracked on each trie node and decremented as files are analyzed, avoiding expensive re-collection
- Extracted `ProgressReport.logProgress()` and call it after each program batch for better visibility during long analyses

## Test plan
- [ ] Run ruling tests on large projects (e.g., Next.js) to verify correctness and improved memory behavior
- [ ] Verify progress logging appears between program batches


zglicz:
## Investigation of ruling differences

All ruling differences are caused by the trie splitting global JS files into smaller programs.

### Root cause: global JS codebases without imports (25 issues: S7728 + S6582 + S3403)

Affected projects: angular.js, jquery, p5.js — pre-module JS using globals (`angular.forEach(...)`, `window.jQuery`). No `import`/`require` or `/// <reference>` directives, so TypeScript resolves cross-file types **only** through `rootNames`.

- Old approach: all files as rootNames in one program → TypeScript sees globals from all directories
- Trie approach: only files in the same directory subtree are rootNames → globals defined in other directories are unresolved

### Options

- Accept the tradeoff (lost issues are in legacy JS codebases: angular 1.x, jquery, p5.js)
- Increase threshold so small projects stay in a single program
- Detect non-module files (no imports) and skip trie splitting for them

vdiez:
should we close this?

zglicz:
Closing for now. We have added a new flag to control TS program creation for orphan files: #6411

PR #6421, opened 2026-02-19
JS-1343 refactor: generate estree.proto at build time instead of loading at runtime

vdiez:
## Summary
- Replace runtime `protobufjs.loadSync()` of `estree.proto` with static code generation (`pbjs`/`pbts`), consistent with how `language_analyzer.proto` and `health.proto` are already handled
- Add `package estree;` to `estree.proto` to fix a `pbts` bug where interfaces are not generated for package-less protos (`java_package` option still overrides for Java — no impact)
- Remove `estree.proto` asset copy from esbuild — the generated JS is bundled directly, eliminating the need to ship the raw `.proto` file
- Use `NodeType.XxxType` dot notation instead of `NodeType['XxxType']` string literal bracket access
- Fix vacuous `TSSatisfiesExpression` test assertion where both sides evaluated to `undefined`

## Test plan
- [x] All 37 AST serialization/deserialization tests pass
- [x] TypeScript type-check passes (`tsc -p packages/tsconfig.app.json --noEmit`)
- [x] `npm run bridge:build:fast` completes successfully
- [x] `npm run bridge:bundle` bundles without errors, `estree.proto` no longer in `bin/`
- [ ] CI build (Java + Node) passes
- [ ] Ruling tests unaffected


zglicz:
> Fix vacuous TSSatisfiesExpression test assertion where both sides evaluated to undefined

some words I don't understand 🎩

its/ruling/src/test/expected/jsts/TypeScript/typescript-S7728.json
```diff
@@ -42,5 +42,10 @@
 "TypeScript:src/harness/test262Runner.ts": [
 113,
 118
+],
+"TypeScript:src/services/formatting/rulesMap.ts": [
+32,
+46,
+47
```
zglicz, line 49: I don't like this :(

package.json
```diff
@@ -21,7 +21,7 @@
...
+    "grpc:generate-proto": "pbjs -t static-module -w packages/grpc/src/proto/es6-wrapper.js --es6 -o packages/grpc/src/proto/language_analyzer.js packages/grpc/src/proto/language_analyzer.proto && pbts -o packages/grpc/src/proto/language_analyzer.d.ts packages/grpc/src/proto/language_analyzer.js && pbjs -t static-module -w packages/grpc/src/proto/es6-wrapper.js --es6 -o packages/grpc/src/proto/health.js packages/grpc/src/proto/health.proto && pbts -o packages/grpc/src/proto/health.d.ts packages/grpc/src/proto/health.js && pbjs -t static-module -w packages/grpc/src/proto/es6-wrapper.js --es6 -o packages/jsts/src/parsers/estree.js packages/jsts/src/parsers/estree.proto && pbts -o packages/jsts/src/parsers/estree.d.ts packages/jsts/src/parsers/estree.js",
```
zglicz: at which point does this warrant its own separate script/file ?

zglicz (approved):

its/ruling/src/test/expected/jsts/TypeScript/typescript-S7728.json
vdiez: not related to this PR

package.json
vdiez: agreed, i'll extract this from package.json

vdiez:
> some words I don't understand 🎩

"Vacuous" as in "trivially true / testing nothing". On master, [line 131 of `ast.test.ts`](https://github.com/SonarSource/SonarJS/blob/master/packages/jsts/tests/parsers/ast.test.ts#L131):

```typescript
expect(literalNode.type).toEqual(NODE_TYPE_ENUM.values['Literal']);
```

`NODE_TYPE_ENUM.values['Literal']` is `undefined` because the enum only has `'LiteralType'`, not `'Literal'`. And `literalNode.type` is also `undefined` because `literalNode` is the inner shape object `{ raw, valueNumber }` (not the Node wrapper that has `.type`). So the assertion was `expect(undefined).toEqual(undefined)` — always passes, checks nothing.

PR #6437, opened 2026-02-23
SC-41129 Refactor customForConfiguration from Java expression to JS function

vdiez:
## Summary
- Moved `customForConfiguration` from a Java expression string to a JS function, so parameter transformations apply in both HTTP bridge and gRPC paths
- S1441 (quotes) now correctly transforms boolean `true`/`false` to `"single"`/`"double"` in gRPC path
- S6418 (hard-coded secrets) `randomnessSensibility` string-to-number conversion now handled in JS
- Removed Java expression inlining from `generate-java-rule-classes.ts`; generated Java classes now send raw field values

## Test plan
- [x] S1441 gRPC test now passes (was previously failing)
- [x] S6418 and S7718 gRPC tests still pass
- [x] Full gRPC server test suite passes (31/31)
- [x] Type-check passes (`npx tsc -b packages`)
- [ ] Verify Java integration tests pass in CI

### DIFF
--- a/packages/grpc/src/transformers/rule-configurations/jsts.ts
+++ b/packages/grpc/src/transformers/rule-configurations/jsts.ts
@@ -53,6 +53,8 @@ type FieldDef = {
   field: string;
   displayName?: string;
   default: unknown;
+  customDefault?: unknown;
+  customForConfiguration?: (value: unknown) => unknown;
 };
 
 /**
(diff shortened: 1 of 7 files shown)


packages/grpc/src/transformers/rule-configurations/jsts.ts
```diff
@@ -225,7 +238,13 @@ function buildPrimitiveConfiguration(
...
     const paramValue = paramsLookup.get(sqKey);
     if (paramValue !== undefined) {
-      return parseParamValue(paramValue, element.default);
+      // When customForConfiguration exists, parse against customDefault type
+      // so the linter receives the raw SQ value in the right type for transformation.
+      const parseTarget =
+        element.customForConfiguration && element.customDefault !== undefined
+          ? element.customDefault
+          : element.default;
+      return parseParamValue(paramValue, parseTarget);
```
zglicz: this looks like a copy-paste from the lines 171-175, add a helper for that

zglicz (approved):

PR #6480, opened 2026-02-27
JS-1384 feat: Add sonar.javascript.disableTypeChecking property

vdiez:
## Summary
- Add new `sonar.javascript.disableTypeChecking` boolean property (defaults to `false`)
- When enabled, skips **all** TypeScript program creation (both tsconfig-based and orphan-based), so every file is analyzed without type information
- Unlike the existing `createTSProgramForOrphanFiles` which only affects orphan files, this completely disables type-checking for faster analysis when type-aware rules are not needed

## Test plan
- [x] Java unit tests: property definition exposed correctly (`JavaScriptPluginTest`)
- [x] Java unit tests: property sent to bridge correctly (`WebSensorTest`)
- [x] TypeScript unit tests: SonarQube mode skips all programs (`analyzeProject-sonarqube.test.ts`)
- [x] TypeScript unit tests: SonarLint mode skips all programs (`analyzeProject-sonarlint.test.ts`)
- [ ] Integration test: type-aware issue found by default, not found when disabled (`DisableTypeCheckingTest`)

### DIFF
--- a/packages/jsts/src/analysis/projectAnalysis/analyzeProject.ts
+++ b/packages/jsts/src/analysis/projectAnalysis/analyzeProject.ts
@@ -95,7 +95,11 @@ export async function analyzeProject(
 
   const progressReport = new ProgressReport(pendingFiles.size);
   if (pendingFiles.size) {
-    if (sonarlint && rules.length) {
+    if (jsTsConfigFields.disableTypeChecking) {
+      info(
+        'Type checking is disabled (sonar.javascript.disableTypeChecking=true). All files will be analyzed without type information.',
+      );
+    } else if (sonarlint && rules.length) {
       await analyzeWithIncrementalProgram(
         filesToAnalyze,
         results,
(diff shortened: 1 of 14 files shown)


packages/jsts/src/analysis/projectAnalysis/analyzeProject.ts
zglicz, line 132: could you rework this comment a bit?

zglicz (approved):
1 small comment about a log

packages/jsts/src/analysis/projectAnalysis/analyzeProject.ts
vdiez: Fixed — suppressed the misleading "not part of any tsconfig" log when disableTypeChecking is true, since in that case files may well be part of a tsconfig but type-checking was explicitly disabled.

vdiez: good catch!

PR #6521, opened 2026-03-04
JS-1406 Improve S2068: Fix FPs due to transition from hotspot to vuln

loris-s-sonarsource:
Part of 

This PR aligns S2068 (hardcoded passwords, heuristics in the variable name) with S6418 (hardcoded secrets, same heuristics), 

S2068 was a hotspot and was recently moved to a vulnerabilty while keeping a "Blocker" state, and it definitely hurt people's quality gates :(


packages/jsts/src/rules/helpers/entropy.ts
loris-s-sonarsource, line 1: This was already implemented for S6418 so I made it a shared helper for both rules

packages/jsts/src/rules/S2068/rule.ts
```diff
@@ -19,12 +19,21 @@
...
+} from '../helpers/index.js';
 import path from 'node:path';
 import type { FromSchema } from 'json-schema-to-ts';
 import * as meta from './generated-meta.js';
 
 const DEFAULT_NAMES = ['password', 'pwd', 'passwd', 'passphrase'];
+const ENTROPY_THRESHOLD = 3.0;
+const MIN_PASSWORD_LENGTH = 5;
+const NON_CREDENTIAL_CHARS = /[\s/\[\]"'<>]/;
+const TEST_FILE_PATTERN = /\.(spec|test|mock)\.[jt]sx?$/;
```
loris-s-sonarsource: These additional heuristics aim to avoid over-raising issues. We have hundreds of secret/password detection rules so I don't think our False Negative rates will highly suffer from them.

packages/jsts/src/rules/S2068/rule.ts
```diff
@@ -73,33 +88,90 @@ export const rule: Rule.RuleModule = {
...
+function extractDefaultOperatorIfNeeded(node: estree.AssignmentExpression): estree.Node {
+  if (
+    isLogicalExpression(node.right as TSESTree.Node) &&
+    ['??', '||'].includes((node.right as estree.LogicalExpression).operator)
+  ) {
+    return (node.right as estree.LogicalExpression).right;
+  }
+  return node.right;
+}
+
```
loris-s-sonarsource: I also took the opportunity to add features that S6418 had but not S2068, such as default operators

packages/jsts/src/rules/S2068/rule.ts
```diff
@@ -73,33 +88,90 @@ export const rule: Rule.RuleModule = {
...
+      hasHighEntropy(value)
+    );
+  }
+  if (node.type === 'ConditionalExpression') {
+    return findValueSuspect(node.consequent) || findValueSuspect(node.alternate);
+  }
+  return false;
+}
+
 function checkLiteral(context: Rule.RuleContext, patterns: RegExp[], literal: estree.Literal) {
```
loris-s-sonarsource, line 131: I split this function into small chunks after getting flagged by the cognitive complexity rule

packages/jsts/src/rules/S2068/rule.ts
vdiez: I don't think this function provides any value, you are always returning the right side of the AssignmentExpression as estree.Node. I think it can be removed.

packages/jsts/src/rules/S2068/rule.ts
```diff
@@ -92,14 +118,68 @@ function checkAssignment(
...
+    return (
+      value.length >= MIN_PASSWORD_LENGTH &&
+      !NON_CREDENTIAL_CHARS.test(value) &&
+      hasHighEntropy(value)
+    );
+  }
+  if (node.type === 'ConditionalExpression') {
+    return findValueSuspect(node.consequent) || findValueSuspect(node.alternate);
+  }
+  return false;
```
vdiez, line 128: this function would be the place to handle `LogicalExpression` if that's what you want to do, the same way you handle `Literal` type sting and `ConditionalExpression`. I think it's better to keep here all the handlers to the specific Nodes/Expressions you plan to support.

vdiez (changes requested):

vdiez (approved):
nice, thanks! you will still need to rebase :)

loris-s-sonarsource:
Thanks a lot for your reactivity @vdiez haha I was about to double check and ask for another look but you were fast

loris-s-sonarsource:
I had tested locally before pushing.... let me check why it's red

loris-s-sonarsource:
Ok, this was a TP, thankfully

PR #6552, opened 2026-03-09
JS-1425 fix(grpc): prevent stale analysis across requests

vdiez:
## Why
We are tackling 2 issues:
1. Request contents must always have precedence over cache.
2. gRPC requests come from different projects, so gRPC should not reuse analysis caches between requests.

## What Changed
- Reset gRPC caches at the start of each Analyze request (source file store, package.json store, tsconfig store, and source-file content/parsed AST cache).
- Ensure compiler host reads prefer current request context over shared cache entries.
- Ensure getSourceFile syncs request content before parsed SourceFile cache lookup.
- Add a gRPC regression test for same file path analyzed twice with different content.

## Validation
- cmd /c npx tsx --tsconfig packages/tsconfig.test.json --test --experimental-test-isolation=none --test-reporter=spec --test-reporter-destination=stdout packages/grpc/tests/server.test.ts
- Result: 34 passed, 0 failed.

### DIFF
--- a/packages/grpc/src/service.ts
+++ b/packages/grpc/src/service.ts
@@ -22,11 +22,28 @@ import {
 } from './transformers/request.js';
 import { transformProjectOutputToResponse } from './transformers/response.js';
 import { analyzeProject } from '../../jsts/src/analysis/projectAnalysis/analyzeProject.js';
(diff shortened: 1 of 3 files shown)


henryju (approved):
LGTM

PR #6567, opened 2026-03-11
JS-1429 Fix S4030 crash on Svelte use: directives

francois-mora-sonarsource:
## Summary

- S4030 (`no-unused-collection`) crashed with `TypeError: Cannot read properties of null (reading 'type')` when linting Svelte files containing a `use:` action directive
- Svelte-specific AST nodes can match `type === 'ExpressionStatement'` but have `expression: null`, which `isElementWrite` and `isWritingMethodCall` did not guard against
- Fix: add optional chaining (`?.`) on `statement.expression` in both functions

## Test plan

- Integration test added to `its/eslint10-plugin-sonarjs`: lints the exact repro from the community report using `svelte-eslint-parser` and asserts no crash
- Lifecycle: `npm run bbf` → `npm run bridge:test` → `npm run eslint-plugin:build` → `cd its/eslint10-plugin-sonarjs && npm install && node index.test.js`

Fixes https://sonarsource.atlassian.net/browse/JS-1429
Community report: https://community.sonarsource.com/t/javascript-s4030-no-unused-collection-crashes-on-svelte-use-directive/179385

### DIFF
--- a/its/eslint10-plugin-sonarjs/fixtures/file.svelte
+++ b/its/eslint10-plugin-sonarjs/fixtures/file.svelte
@@ -1,0 +1,6 @@
+<script lang="ts">
+  let x = 0
+  function action() {}
+</script>
+
+<div use:action={x}></div>
--- a/its/eslint10-plugin-sonarjs/index.test.js
+++ b/its/eslint10-plugin-sonarjs/index.test.js
@@ -62,3 +62,17 @@ test('should work with TSESLint config', async t => {
   );
   verifyErrors(result.stdout);
 });
+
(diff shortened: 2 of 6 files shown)


its/eslint10-plugin-sonarjs/fixtures/file.svelte
```diff
@@ -0,0 +1,6 @@
+<script lang="ts">
```
francois-mora-sonarsource, line 1: The problem is raised when using the TypeScript ESLint parser because the node doesn't respect the AST structures. Therefor, the solution is to add defensive checks to prevent the analyzer from failing.

vdiez (approved):

PR #6620, opened 2026-03-17
JS-1439 Fix S2068: detect hardcoded credentials in static template literals

zglicz:
## Summary
This PR fixes `S2068` to detect hardcoded credentials in static template literals, matching existing behavior for string literals.

## What changed
- Added `TemplateLiteral` handling in S2068 for static templates (no expressions).
- Extended assignment-value checks so static template literals are evaluated for entropy and filtering.
- Refactored shared string scanning logic to reuse the same detection path for literals and static template literals.

## Tests
- Added reproducer test for static template literal password assignment (failed before fix).
- Added static template literal URL case with `password=`.
- Added dynamic template literal case to assert we do not report non-static values.
- Ran: `npx tsx --test packages/jsts/src/rules/S2068/**/*.test.ts`

### DIFF
--- a/packages/jsts/src/rules/S2068/rule.ts
+++ b/packages/jsts/src/rules/S2068/rule.ts
@@ -20,7 +20,7 @@ import type { Rule } from 'eslint';
 import type estree from 'estree';
 import type { TSESTree } from '@typescript-eslint/utils';
 import { generateMeta } from '../helpers/generate-meta.js';
-import { isLogicalExpression, isStringLiteral } from '../helpers/ast.js';
+import { isLogicalExpression, isStaticTemplateLiteral, isStringLiteral } from '../helpers/ast.js';
 import { shannonEntropy } from '../helpers/entropy.js';
 import path from 'node:path';
 import type { FromSchema } from 'json-schema-to-ts';
@@ -71,6 +71,10 @@ export const rule: Rule.RuleModule = {
         const literal = node as estree.Literal;
         checkLiteral(context, literalRegExp, literal);
       },
+      TemplateLiteral: (node: estree.Node) => {
+        const templateLiteral = node as estree.TemplateLiteral;
+        checkTemplateLiteral(context, literalRegExp, templateLiteral);
+      },
       PropertyDefinition: (node: estree.Node) => {
         const property = node as TSESTree.PropertyDefinition;
         checkAssignment(
@@ -108,6 +112,15 @@ function findValueSuspect(node: estree.Node | undefined | null): boolean {
   if (!node) {
     return false;
   }
+  if (isStaticTemplateLiteral(node)) {
+    const value = node.quasis[0].value.cooked;
+    return (
+      value != null &&
+      value.length >= MIN_PASSWORD_LENGTH &&
+      !NON_CREDENTIAL_CHARS.test(value) &&
+      hasHighEntropy(value)
+    );
+  }
   if (isStringLiteral(node)) {
     const value = node.value as string;
     return (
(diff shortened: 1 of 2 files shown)


vdiez (approved):

joao-dias-tma:
@zglicz I don't know how but the latest version of `eslint-plugin-sonarjs` (4.0.2) broke this rule and I'm guessing this is the reason, since nothing else seems to have changed

zglicz:
Hi @joao-dias-tma - I'm sorry to hear that. Would you be as kind as to write a post in community - https://community.sonarsource.com/c/rules-and-languages/51 
If you provide us with a repro we can get it patched up in a jiffy.

Cheers,
Michal

joao-dias-tma:
@zglicz here is a repro with instructions: https://github.com/alfaproject/repro-sonarjs

loris-s-sonarsource:
Perfect, thanks for this PR guys (I was out sick)

zglicz:
Thanks again for the repro, it made it easy to verify.

I verified it locally with your repo:

eslint-plugin-sonarjs@4.0.1 -> 3 issues
eslint-plugin-sonarjs@4.0.2 -> 1 issue
So your report is valid.

The drop is caused by S2068 now skipping files matching *.spec|*.test|*.mock (added in PR #6521), which suppresses findings in __tests__/index.spec.ts. This was the exact objective of #6521

If I recall correctly @loris-s-sonarsource added that as we were getting very many reports of this rule triggering in test files which is not desired. That being said, in SonarJS and in SonarQube as a whole, we have a concept of MAIN or TEST files and we can specify on a rule, whether it applies to one or the other. In the eslint plugin, we don't have this concept.

I think from the JS/TS team @SonarSource/quality-web-squad  we could add a similar heuristic of detecting test rules on rules marked as MAIN, so that we don't need to do this check on the rule level. I'll bring it up in our weekly and see if we can improve it.

loris-s-sonarsource:
> That being said, in SonarJS and in SonarQube as a whole, we have a concept of MAIN or TEST files and we can specify on a rule, whether it applies to one or the other. In the eslint plugin, we don't have this concept.

This feature exists, but our main problem is that a ton of users do not use it... So we need to filter them out ourselves at the rule level to avoid drowning under FP feedbacks

joao-dias-tma:
Isn't that something that should be controlled by the users of the plugin?
Perhaps that should be documented in case people are not aware?

For instance, if I don't want that rule in spec files, then I just do this:
```js
export default defineConfig({
  ...sharedConfigWithAllTheOptionsYouGetTheIdea,
  {
    files: [''**/*.spec.ts'],
    rules: {
      // Assume hardcoded passwords in spec files are dummy
      'sonarjs/no-hardcoded-passwords': 'off'
    }
  }
});
```

I'n not sure the rule itself should dictate where or where it doesn't apply, that seems weird to me.

loris-s-sonarsource:
In a perfect world, users should control that themselves.

In this concrete world, we have hundreds of users setting their issues as false positives and using the sonarcloud feedback box to say "this is a test file". As a results, our internal graphs show bad false positive rates because of that, and this is the KPI monitored by leadership

joao-dias-tma:
I mean, if the problem is those KPIs, then fair enough, it's your product ofc.

I will say tho, that in terms of security it doesn't make much sense. Someone can leak a password or secret by accident in a test file, and then what? It kind of defeats the point. I know it's not your fault. We actually have a policy that all disabled rules should have a comment why regardless even if it's a false positive, because you never know when a bad day will happen.

Thank you for following up on my report, much appreciated. If you are not interested in making the rule configurable to include spec files or not (default not), I guess we'll just fork it internally for now.

loris-s-sonarsource:
Honestly I am 100% aligned with you, secrets are also likely to be leaked in test files too.

On my side I am on the AppSec team here and I think @zglicz from the JSTS team will actually discuss about it so I would say you should not deem this problem as WONTFIX for now. In the past we considered automatically setting up "main" and "test" files in the secret detection analyzer, I am going to ask them if it was done in the end, because they are the ones who provided the file patterns I used in my PR

loris-s-sonarsource:
cc @zglicz, I am writing this here for continuity.
FTR: I asked the Cloud Security squad, who said that in the end they also do it themselves IF users did not set sonar.tests. This could be the nice solution (the nicer being setting it maybe in sonar-scanner)

vdiez:
The override of MAIN/TEST files should not be happening in the rule themselves, as this is affecting users of the ESLint plugin as we can see here. The problem we are trying to solve is only applicable in SQS, SQC context, so the logic to treat files as TEST instead of MAIN in the scanner context (NOT in ESLint context) should happen on another layer, most problably in the [Linter class](https://github.com/SonarSource/SonarJS/blob/0c8da7510105302e90850c625a6a575e41192935/packages/jsts/src/linter/linter.ts#L186), without polluting the ESLint user settings.

vdiez:
I created a [ticket](https://sonarsource.atlassian.net/browse/JS-1501) to fix this

joao-dias-tma:
Thank you so much!

PR #6626, opened 2026-03-18
JS-1460 chore: evaluate TypeScript 6.0.1-rc

vdiez:
## Summary
- Evaluate SonarJS compatibility with `typescript@6.0.1-rc`.
- Keep CI install path working during RC phase (`npm ci --legacy-peer-deps`) while `@typescript-eslint` peer ranges still exclude TS6.
- Stabilize TS6-related CI and test behavior, including ESLint plugin packaging and TS fixtures.
- Add strictness-compatibility logic during TS program creation to avoid forcing TS6 semantics on projects targeting older TS behavior.

### Policy A: Use TS6 default strictness everywhere (no override)
- Description: Let TS6 implicit defaults apply in all program creation paths when `strict` is not set.
- Outcome: Produced ruling noise/regressions in type-sensitive rules, especially in projects that do not explicitly set strictness and are effectively authored against pre-TS6 behavior.

### Policy B: Force non-strict only for synthesized programs (orphan/single-file)
- Description: Set `defaultCompilerOptions.strict = false` for synthesized programs (no project tsconfig context).
- Outcome: Significant FP reduction compared to Policy A; this was a strong first mitigation but did not address tsconfig-backed programs with implicit strictness.

### Policy C: Compatibility-aware defaults for tsconfig-backed programs when user TS is `<6` and strictness is implicit
- Description:
  - Detect user TS signal from `package.json` dependencies.
  - Apply legacy strictness defaults only when version range is **definitely** `<6`.
  - Inspect strictness options across `tsconfig` + `extends` chain.
  - If `strict` is explicitly set, do not override.
  - If only sub-options are explicitly set, preserve them and default only missing strictness options.
- Outcome:
  - Restored CI stability (Linux/Windows build, unit tests, knip green on latest run).
  - Latest ruling differences are improved or neutral on overlapping files versus previous artifact; key observed improvement is `S3403` in TypeScript sources (`10 -> 3`, delta `-7`), with no regressions on the tracked strictness-sensitive set (`S6606`, `S4325`, `S1154`, `S6551`, `S3757`) in overlapping files.

## Recommendation for TS6 GA
- Keep Policy C as the default initially to preserve backward-compatible analysis for projects effectively on TS<6 semantics when strictness is implicit.
- Always respect explicit user strictness configuration (`strict` or sub-options), including through `extends`.
- After TS6 is GA and ecosystem support is fully aligned, consider an explicit migration path toward TS6-style implicit strictness (opt-in first, then revisit default), gated by ruling quality and FP/FN monitoring.

## What I checked
1. Verified TS6 compile paths in CI and local targeted checks.
2. Verified strictness behavior with focused tests in `packages/jsts/tests/program/options.test.ts`.
3. Verified latest ruling artifact deltas after strictness-policy updates.

This PR remains an evaluation branch to capture TS6 readiness and migration strategy.


vdiez:
﻿Follow-up on latest ruling evaluation after pinning synthesized compiler defaults to `strict: false`.

Scope of this estimate:
- Compared latest ruling-differences artifact (`5987302279`, run `23249204802`) against the previous baseline (`5984169036`).
- Focused on the TS6-sensitive changed-rule set we reviewed.
- As discussed, **S1874 is excluded** from FP estimation (deprecation diagnostics are strongly tied to TS/ecmascript lib evolution).

FP estimate (excluding S1874):
- Strict FP estimate: **3.17%** (`94 / 2966`)
- FP + 50% of mixed cases: **3.89%**
- Worst-case (all mixed treated as FP): **4.62%**

Main residual FP drivers:
- **S3403**: assignability-based strict comparison checks can still flag defensive `null/undefined` idioms.
- **S1154**: broad StringLike expression-statement heuristic.
- **S6551**: generic/dynamic stringification heuristics.
- **S3757**: occasional coarse typing around arithmetic/object flows.

Conclusion:
- With path-1 (`strict: false` for synthesized programs), the ruling changes now look **mostly acceptable** and significantly cleaner than before.
- I’m comfortable merging this TS6 upgrade **once TS6 is GA**, provided CI remains green and we keep follow-up cleanup under **JS-1469** for intentional strict-by-default adoption and targeted rule tuning (`S3403`, `S1154`, `S6551`, `S3757`).

packages/jsts/src/program/tsconfig/options.ts
```diff
@@ -56,8 +60,34 @@ type CustomParseConfigHost = {
...
+  'noImplicitAny',
+  'noImplicitThis',
+  'strictBindCallApply',
+  'strictBuiltinIteratorReturn',
+  'strictFunctionTypes',
+  'strictNullChecks',
+  'strictPropertyInitialization',
+  'useUnknownInCatchVariables',
+];
+
```
francois-mora-sonarsource, line 90: This could be merged into a single list like this:
```typescript
export const STRICTNESS_COMPILER_OPTIONS = [
  'strict',
  'alwaysStrict',
  'noImplicitAny',
  'noImplicitThis',
  'strictBindCallApply',
  'strictBuiltinIteratorReturn',
  'strictFunctionTypes',
  'strictNullChecks',
  'strictPropertyInitialization',
  'useUnknownInCatchVariables',
] as const;

export type StrictnessCompilerOption =
  (typeof STRICTNESS_COMPILER_OPTIONS)[number];
```

PR #6629, opened 2026-03-18
JS-1193 Escape backslashes in generated Java rule classes

zglicz:
## Summary
- escape Java string literals when generating JS/TS Java check classes
- remove now-unnecessary `customDefault` workaround values from S139 and S7718 configs
- update related gRPC test descriptions to reflect escaped-default handling

## Validation
- npm run grpc:generate-proto
- npm run generate-meta
- npm run generate-java-rule-classes
- npx tsx --tsconfig packages/tsconfig.test.json --test packages/grpc/tests/server.test.ts --test-name-pattern "escaped regex string defaults|escaped regex defaults in arrays"

### DIFF
--- a/tools/generate-java-rule-classes.ts
+++ b/tools/generate-java-rule-classes.ts
@@ -150,12 +150,10 @@ function generateBody(
       return;
     }
 
-    const getSQDefault = () => {
-      return property.customDefault ?? property.default;
-    };
+    const sqDefault = property.customDefault ?? property.default;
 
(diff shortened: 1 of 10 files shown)


tools/generate-java-rule-classes.ts
vdiez: this can be removed right?

vdiez (approved):
just a small change, line 154 is no longer needed

PR #6635, opened 2026-03-19
JS-776 Remove in-rule React detection and rely on requiredDependency

zglicz:
## Summary
- remove per-rule React detection logic (`detectReactRule`) from S6440 and S6441
- remove redundant `react` dependency guards from S6477 and S6747
- delete unused `rule-detect-react` helper
- keep fixtures focused on React scenarios only
- add linter tests verifying `requiredDependency` filtering for a React rule (disabled without `react`, enabled with `react`)

## Validation
- `npx tsx --tsconfig packages/tsconfig.test.json --test packages/jsts/src/rules/S6440/cb.test.ts packages/jsts/src/rules/S6441/cb.test.ts packages/jsts/tests/linter/index.test.ts`


packages/jsts/src/rules/S6441/rule.ts
```diff
@@ -18,51 +18,20 @@
...
-
 export const rule: Rule.RuleModule = {
   meta: generateMeta(meta, {
     ...noUnusedClassComponentMethod.meta,
     messages: {
-      ...noUnusedClassComponentMethod.meta!.messages,
       unused:
         'Remove this property or method or refactor this component, as "{{name}}" is not used inside component body',
       unusedWithClass:
         'Remove this property or method or refactor "{{className}}", as "{{name}}" is not used inside component body',
```
zglicz, line 33: Here we override the builtin messages from eslint

vdiez (changes requested):
this needs to be more exhaustive, there are more rules trying to detect react aside of the rule-detect-react.ts helper. some are using custom esquery selectors. check S6439 or S6443 for example.

zglicz:
Both of those rule can't immediately benefit from simplification. 
- S6439 - additionaly checks if react-native is present and has a little bit different logic
- S6443 - checks fqn to make sure its a react hook call

I did update rspec with required dependencies for a few of these rules - https://github.com/SonarSource/rspec/pull/6379

vdiez (approved):

PR #6652, opened 2026-03-20
JS-1494 Fix S6418 legacy decimal parameter crash and add scanner smoke tests

zglicz:
## Summary
- fix S6418 scanner-side parameter shape to accept persisted legacy value `randomnessSensibility="5.0"` without crashing
- add scanner smoke coverage in `JsTsChecksTest` for all built-in JS/TS checks with reflected `@RuleProperty` defaults
- add legacy-parameter smoke coverage (including S6418 `5.0`) to catch upgrade regressions

## Validation
- `npx tsx --test packages/jsts/src/rules/S6418/unit.test.ts`
- `npx tsx --test packages/grpc/tests/server.test.ts --test-name-pattern "S6418"`
- `mvn -pl sonar-plugin/sonar-javascript-plugin -am -Dtest=JsTsChecksTest -Dsurefire.failIfNoSpecifiedTests=false test`

## Regression proof
- with temporary pre-fix S6418 config (`default: 5`, no conversion), `should_initialize_all_builtin_checks_with_legacy_rule_properties` fails with `NumberFormatException: For input string: "5.0"`
- after restoring the fix, the same test command passes

### DIFF
--- a/packages/jsts/src/rules/S6418/config.ts
+++ b/packages/jsts/src/rules/S6418/config.ts
@@ -28,7 +28,8 @@ export const fields = [
     {
       field: 'randomnessSensibility',
       description: 'Minimum shannon entropy threshold of the secret',
-      default: '5.0',
+      default: 5,
+      customDefault: '5.0',
       customForConfiguration: Number,
     },
   ],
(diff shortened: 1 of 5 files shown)


packages/jsts/src/rules/S6418/config.ts
```diff
@@ -29,6 +29,8 @@ export const fields = [
       field: 'randomnessSensibility',
       description: 'Minimum shannon entropy threshold of the secret',
       default: 5,
+      customDefault: '5.0',
```
zglicz, line 32: we need this for SQ

vdiez (approved):

PR #6695, opened 2026-03-26
JS-1505 Fix S1874 crash on TSX files with react-jsx in monorepos

francois-mora-sonarsource:
## Summary

- **Root cause fix**: Add `jsx` and `importHelpers` to the `parsedSourceFileCache` key in `IncrementalCompilerHost`. Both options affect `SourceFile.imports` structure; without them in the key, a stale SourceFile from a non-jsx program was shared with a jsx program, causing `getSuggestionDiagnostics` (called by S1874) to fire a TypeScript internal assertion: _"Debug Failure. False expression: Expected sourceFile.imports[0] to be the synthesized JSX runtime import"_.
- **Defensive fix**: Wrap S1874's `getSuggestionDiagnostics` call in try/catch. The method is already non-public (`@ts-ignore`), so internal TypeScript assertion failures must not propagate to the user.
- **Reproducer**: New ITS project `s1874-tsx-multi-tsconfig` with `reproduce-crash.mjs` that demonstrates the crash with unfixed code and passes with the fix applied.

## Test plan

- [x] `sourceFileCache.test.ts` — 20 tests covering new cache behavior for `jsx` and `importHelpers` options (separate entries per option combination, cache miss on mismatch)
- [x] `S1874/cb.test.ts` — existing rule tests still pass
- [x] `reproduce-crash.mjs` — crashes with unfixed `lib/`, succeeds with fixed `lib/`
- [ ] `S1874TsxMultiTsconfigTest.java` — ITS test asserting no "Debug Failure" log when non-jsx tsconfig is listed first in `sonar.typescript.tsconfigPaths`

### DIFF
--- a/packages/analysis/tests/jsts/program/compilerHost.test.ts
+++ b/packages/analysis/tests/jsts/program/compilerHost.test.ts
@@ -17,6 +17,8 @@
 import { describe, it, beforeEach } from 'node:test';
 import { expect } from 'expect';
 import ts from 'typescript';
+import path from 'node:path';
+import { readFileSync } from 'node:fs';
 import { IncrementalCompilerHost } from '../../../src/jsts/program/compilerHost.js';
 import {
   setSourceFilesContext,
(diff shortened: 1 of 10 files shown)


francois-mora-sonarsource:
### Code review

Found 1 issue:

1. `reproduce-crash.mjs` imports from the `lib/` build artifact directory, which is listed in `.gitignore` and not version-controlled. The script cannot function in a clean checkout (the import target won't exist), and no other IT project in `its/plugin/projects/` contains a similar debug helper. This file appears to be a local development aid that was not intended for the repository.

https://github.com/SonarSource/SonarJS/blob/7925e265246bd942dd8cae2e0e4f0e4073b923c7/its/plugin/projects/s1874-tsx-multi-tsconfig/reproduce-crash.mjs#L33-L40

<sub>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

francois-mora-sonarsource:
Addressed the review note by removing its/plugin/projects/s1874-tsx-multi-tsconfig/reproduce-crash.mjs in commit 10c2a9595, since it depended on ignored lib/ build artifacts and was only a local debug helper.

packages/analysis/tests/jsts/program/compilerHost.test.ts
```diff
@@ -412,4 +414,86 @@ describe('IncrementalCompilerHost', () => {
...
   });
+
+  describe('cross-program SourceFile cache with different jsx options (JS-1505)', () => {
+    // Fixture: server.ts (jsx:preserve tsconfig) imports Component.tsx, which is also
+    // included in a jsx:react-jsx tsconfig. This reproduces the cross-batch crash scenario:
+    //
+    // Batch 1: server.ts is in the analysis context. Program A (jsx:preserve) is created.
+    //   TypeScript follows the import and calls getSourceFile(Component.tsx). Component.tsx
+    //   is NOT in the context → read from disk → SourceFile cached without synthesized import.
+    //
```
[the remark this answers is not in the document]
francois-mora-sonarsource: Done

packages/analysis/src/jsts/rules/S1874/rule.ts
```diff
@@ -38,9 +38,15 @@ export const rule: Rule.RuleModule = {
         const program = services.program;
         const checker = program.getTypeChecker();
         const sourceFile = program.getSourceFile(context.filename);
-        const diagnostics: ts.DiagnosticWithLocation[] =
+        let diagnostics: ts.DiagnosticWithLocation[];
+        try {
           // @ts-ignore: TypeChecker#getSuggestionDiagnostics is not publicly exposed
-          checker.getSuggestionDiagnostics(sourceFile);
+          diagnostics = checker.getSuggestionDiagnostics(sourceFile);
+        } catch {
```
vdiez: let's remove this catch, otherwise we don't know if your cache fix is working

francois-mora-sonarsource: Done — removed the try-catch so the cache key fix (including `jsx` and `importHelpers` in the SourceFile cache key) is the sole guard against the crash.

vdiez (approved):

PR #6748, opened 2026-04-08
JS-1551 Fix S7785 false positives for non-Promise .catch()/.finally() calls

nathsou:
Part of [JS-1525](https://sonarsource.atlassian.net/browse/JS-1525)

The unicorn prefer-top-level-await rule matches .then(), .catch(), and .finally() by method name alone, without checking the receiver's type. This caused false positives for libraries like Zod whose fluent APIs use .catch() as a synchronous schema method, not a Promise chain.

The decorator now intercepts 'promise' reports and suppresses them when the receiver is not thenable, using two modes:

With type checker: checks whether the receiver has a callable .then() (i.e. is PromiseLike) via isThenable().
Without type checker (fallback): suppresses if the chain root is imported from an allowlisted package (zod).

[JS-1525]: https://sonarsource.atlassian.net/browse/JS-1525?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/analysis/src/jsts/rules/S7785/decorator.ts
+++ b/packages/analysis/src/jsts/rules/S7785/decorator.ts
@@ -17,27 +17,131 @@
 // https://sonarsource.github.io/rspec/#/rspec/S7785/javascript
 
 import type { Rule } from 'eslint';
+import type estree from 'estree';
+import { interceptReport } from '../helpers/decorators/interceptor.js';
 import { generateMeta } from '../helpers/generate-meta.js';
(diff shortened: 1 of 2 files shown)


nathsou:
Closing as this is a duplicate of https://github.com/SonarSource/SonarJS/pull/6753, I used an invalid branch name, meaning that it's not automatically linked with the Jira task.

PR #6753, opened 2026-04-09
JS-1525 Fix FP S7785 : Suppress zod catch warnings

nathsou:
Part of JS-1525

Suppresses warnings suggesting top-level await if the chain root is imported from an allowlisted package (zod).

Zod (and other packages to be identified) defines a custom interface with a  synchronous`catch` method that does not behave like a Promise. The underlying unicorn rule simply checks for `catch`/`then`/`finally` calls by name, triggering a false positive.

Fix: Walk up the call chain to reach the root identifier, if the root comes from an identified module (zod for now), then we can assume that the warning can be suppressed.


nathsou:
Duplicate of https://github.com/SonarSource/SonarJS/pull/6760

PR #6765, opened 2026-04-10
JS-1462 Fix S6582 over-suppression in boolean guard contexts

francois-mora-sonarsource:
Relates to [JS-1462](https://sonarsource.atlassian.net/browse/JS-1462)

This adjusts the `S6582` interceptor so nested logical-expression reports are not suppressed when they only appear inside boolean-coercion contexts such as `if`, `while`, ternary tests, or `!expr`.

The previous logic relied on the immediate reported node's contextual type. That worked for typed return/assignment false positives, but it also treated right-hand `&&` branches inside boolean guards as having contextual type `boolean` and suppressed them, for example:

- `!outerLoop || (outerLoop.labels && outerLoop.labels.get(labelText))`
- `!editScript || (lastMappingResult && lastMappingResult['_countWaitingForRemove'])`
- `!curve.hasHandles() || (clearLookup && clearLookup[getId(curve)])`

Those are legitimate rule hits and should still be reported.

Changes:
- add `isInBooleanCoercionContext(...)` in `rule.ts`
- keep existing typed-context suppressions intact
- add regression tests for the chained boolean-guard case and the ITS-derived repros above

Validation:
- `npx tsx --tsconfig packages/tsconfig.test.json --test packages/analysis/src/jsts/rules/S6582/cb.test.ts packages/analysis/src/jsts/rules/S6582/unit.test.ts`

[JS-1462]: https://sonarsource.atlassian.net/browse/JS-1462?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


zglicz (commented):
I want to understand the history and what is happening. The rule went from a simple decorated rule with 20 lines of logic to a 300 line decorator, where we first introduced supperssion in #6637 and we suppressed too much, and now we need to unsuppress some in this PR.

The full rule implementation is 200 lines of code, which makes me quite uneasy about introducing such convoluted logic. This rule does have a few options, perhaps if its too complicated on booleans, we can use the flag and not enable. it on booleans?

francois-mora-sonarsource:
> I want to understand the history and what is happening. The rule went from a simple decorated rule with 20 lines of logic to a 300 line decorator, where we first introduced supperssion in #6637 and we suppressed too much, and now we need to unsuppress some in this PR.
> 
> The full rule implementation is 200 lines of code, which makes me quite uneasy about introducing such convoluted logic. This rule does have a few options, perhaps if its too complicated on booleans, we can use the flag and not enable. it on booleans?

The original PR was #6637. Basically, the change aimed at adding cases that answered the question "Is the reported issue by the upstream rule compiling or creating a compiler error?" as in `function getLen(arr: string[] | null): number | null { return arr && arr.length; }` that becomes `function getLen(arr: string[] | null): number | null { return arr?.length; }`. Different cases were implemented:
1. `!a && !a.b` -> negation makes the result undefined value safely converted to a boolean value.
2. `(a && a.b) ?? c` -> null coalescing operator also handles safely the undefined value.
3. `(a && a.b) !== c` -> any binary operator also handle undefined safely.
4. Otherwise, use the contextual type and verify undefined is part of it.

Then, checking the impact of this on peach (which happened after the merge) showed that some issues were wrongly suppressed because the contextual type doesn't contain 'undefined', even though we know that statements like `if` are safe: `if (!outerLoop || (outerLoop.labels && outerLoop.labels.get(labelText)))`

I guess the expected outcome for this review, is to determine if we want to patch the rule with this additional case or simply revert the initial commit. Alternatively, we could also decide to reverse the implementation strategy and identify the false positives to suppress instead of the true positives to report. I am trying to see if it's possible to reverse the logic in this PR https://github.com/SonarSource/SonarJS/pull/6830

francois-mora-sonarsource:
Replaced by https://github.com/SonarSource/SonarJS/pull/6830

PR #6777, opened 2026-04-13
JS-1576 Upgrade to TypeScript 6

vdiez:
## Summary
- upgrade the repository TypeScript version to `6.0.2`
- port strictness-compatibility behavior from #6626 onto current `master`
- port non-rule TS6 compatibility updates needed on current layout (`tsconfig-plugin.json` + analyzer module-resolution fixtures)
- keep the scope intentionally limited: no rule implementation changes are ported

## TS5/TS6 strictness handling details
TypeScript 6 changed how implicit strictness options are defaulted when `strict` is not explicitly configured. To keep pre-TS6 behavior for TS5 projects while preserving TS6 behavior for TS6+ projects, we do the following:

1. Detect project TypeScript version signal
- read `typescript` from the nearest project `package.json` chain (`getTypeScriptVersionSignal`).

2. Decide whether to apply legacy defaults
- apply legacy behavior only when the version range is definitely `< 6` (intersects `<6.0.0-0` and does not intersect `>=6.0.0-0`).

3. Respect explicit configuration (including `extends`)
- collect strictness options explicitly set in the root tsconfig and all extended configs via `extendedConfigCache`.

4. Preserve explicit `strict`
- if `strict` is explicitly configured anywhere, we do not override strict sub-options.

5. Apply legacy implicit defaults only for TS < 6
- for strictness options not explicitly set, force them to `false` to match pre-TS6 implicit behavior.

6. Keep TS6+ defaults untouched
- for TS6+ projects, we do not inject these `false` values; TypeScript 6 defaults are preserved.

7. Avoid TypeScript internal API reliance
- strictness options are listed explicitly (instead of using `ts.optionDeclarations` internals) to reduce fragility.

## Additional TS6 compatibility ports (non-rule)
- `tsconfig-plugin.json`
- add `types` for Node and JSX helper package
- add `ignoreDeprecations: "6.0"` to handle `moduleResolution=node` deprecation surfaced by `eslint-plugin:emit`
- analyzer module-resolution fixtures
- set explicit `moduleResolution` for `commonjs`, `esnext`, and `nodenext` fixture tsconfigs to keep test expectations stable with TS6

## Validation
- `npm run eslint-plugin:check`
- `npm run eslint-plugin:emit`
- `npx tsx --tsconfig packages/tsconfig.test.json --test --test-reporter=spec --test-reporter-destination=stdout packages/analysis/tests/jsts/analysis/analyzer.test.ts`
- `npx tsx --tsconfig packages/tsconfig.test.json --test --test-reporter=spec --test-reporter-destination=stdout packages/analysis/tests/jsts/program/options.test.ts packages/analysis/tests/jsts/program/factory.test.ts`

## Unit test expectation updates (after TS6 rebase)
The unit-test updates in `584be19d10` are expected for the TS6 + updated upstream baseline. They are not rule-implementation ports.

- `S2259` (`packages/analysis/src/jsts/rules/S2259/unit.test.ts`): the previous valid snippet relied on known FN behavior (`x.foo()` under `x == null` / `x == undefined` branches). With the current typed baseline, those dereferences are now reportable. Classification: expected improvement (fewer FNs).
- `S3403` (`packages/analysis/src/jsts/rules/S3403/unit.test.ts`): `str === null` is now reported as always false in typed mode, so the valid sample uses `str == null`. Classification: expected improvement (precision).
- `S3735` (`packages/analysis/src/jsts/rules/S3735/unit.test.ts`): TS flow narrowing on `const ... = undefined/null` made those `void` examples effectively non-thenable. Tests were switched to `declare const` to preserve declared union intent. Classification: expected test-harness correction (neutral).
- `S4325` sentinel (`packages/analysis/src/jsts/rules/S4325/unit.test.ts`): upstream `@typescript-eslint/no-unnecessary-type-assertion` behavior changed (plain `querySelector(...) as HTMLElement` no longer reported; `querySelector(...)! as HTMLElement` still reported once). Classification: expected upstream baseline drift.
- `S4325` non-null assertions under implicit non-strict defaults (`packages/analysis/src/jsts/rules/S4325/unit.test.ts`): upstream now does not report the previously expected nullable `!` cases in this path; since S4325 decorates upstream, we inherit that. Classification: expected upstream-driven detection drop in this specific non-strict path (potential FN increase vs previous baseline).

Overall assessment:
- expected: yes
- improvements: `S2259`, `S3403`
- neutral/test intent alignment: `S3735`
- potential regression (upstream-driven, non-strict path): part of `S4325`


packages/analysis/src/jsts/rules/helpers/package-jsons/dependencies.ts
```diff
@@ -245,6 +245,23 @@ export function getNodeVersionSignal(baseDir: NormalizedAbsolutePath): string |
...
+ * @returns raw version string from typescript dependency, or null if not found
+ */
+export function getTypeScriptVersionSignal(baseDir: NormalizedAbsolutePath): string | null {
+  for (const packageJson of getManifests(baseDir, baseDir, fs)) {
+    const typescriptVersion = getDependencyVersionSignal(packageJson, 'typescript');
+    if (isValidDependencySignal(typescriptVersion)) {
+      return typescriptVersion;
+    }
+  }
+  return null;
```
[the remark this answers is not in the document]
vdiez: Good catch, thanks. I refactored the duplicated loop into `getVersionSignalFromManifests(...)` and delegated both `getTypeScriptVersionSignal` and `getNodeVersionSignal` to it, while preserving the `engines.node` fallback behavior for Node signals. Included in 094c15b95e.

packages/analysis/src/jsts/program/tsconfig/options.ts
```diff
@@ -58,8 +62,34 @@ type CustomParseConfigHost = {
...
+  'alwaysStrict',
+  'noImplicitAny',
+  'noImplicitThis',
+  'strictBindCallApply',
+  'strictBuiltinIteratorReturn',
+  'strictFunctionTypes',
+  'strictNullChecks',
+  'strictPropertyInitialization',
+  'useUnknownInCatchVariables',
+];
```
zglicz: This can probably be simplified, so that we only list it once and derive the type from it.

packages/analysis/src/jsts/program/tsconfig/options.ts
```diff
@@ -58,8 +62,34 @@ type CustomParseConfigHost = {
 export const defaultCompilerOptions: ts.CompilerOptions = {
   allowJs: true,
   noImplicitAny: true,
+  strict: false,
```
zglicz, line 67: oh, thats the default ? Shouldn't this also moving forward be default = true, or even not present, as in ts6 it will be nice and strict

packages/analysis/src/jsts/program/tsconfig/options.ts
```diff
@@ -110,6 +140,81 @@ export function nodeVersionToEs(major: number): number {
...
+  const options = compilerOptions as Record<string, unknown>;
+  for (const optionName of STRICTNESS_COMPILER_OPTIONS) {
+    if (Object.hasOwn(options, optionName)) {
+      explicitOptions.add(optionName);
+    }
+  }
+}
+
+function getExplicitStrictnessCompilerOptions(
+  rootConfig: any,
```
zglicz: perhaps start with unknown and get to what types you want

packages/analysis/src/jsts/rules/helpers/package-jsons/dependencies.ts
```diff
@@ -178,6 +178,24 @@ function getDependencyVersionSignal(
   return typeof dependencyVersion === 'string' ? dependencyVersion : null;
 }
 
+function getVersionSignalFromManifests(
+  baseDir: NormalizedAbsolutePath,
+  dependencyName: string,
+  fallbackSignal?: (packageJson: PackageJson) => string | null,
+): string | null {
+  for (const packageJson of getManifests(baseDir, baseDir, fs)) {
```
zglicz: I feel like we should have a helper for getManifests(dir, dir) <- cause there is at most 1 we can get back, no?

packages/analysis/src/jsts/rules/S4325/unit.test.ts
```diff
@@ -213,12 +201,11 @@ describe('S4325', () => {
     });
   });
 
-  it('should flag non-null assertions on declared nullable types without strictNullChecks', () => {
-    ruleTester.run('S4325 non-null assertions without strictNullChecks', rule, {
-      valid: [],
-      invalid: [
+  it('should not flag non-null assertions on declared nullable types with implicit non-strict defaults', () => {
+    ruleTester.run('S4325 non-null assertions with implicit non-strict defaults', rule, {
```
zglicz: is this still necessary?

zglicz (approved):
few cosmetic nits

packages/analysis/src/jsts/program/tsconfig/options.ts
vdiez: Kept strict=false intentionally. This object is the fallback compiler-options baseline for program creation paths without project tsconfig strictness context; TS6 strictness behavior is preserved in the tsconfig parsing path via applyLegacyStrictnessDefaults.

packages/analysis/src/jsts/program/tsconfig/options.ts
vdiez: Applied in c38129f8ec: switched rootConfig from any to unknown and added a small getCompilerOptions helper for explicit narrowing.

packages/analysis/src/jsts/rules/S4325/unit.test.ts
vdiez: Agreed. I removed that additional upstream-behavior assertion block in c38129f8ec. It was only documenting an upstream non-strict path and was adding maintenance churn without strengthening decorator coverage.

packages/analysis/src/jsts/rules/helpers/package-jsons/dependencies.ts
vdiez: Applied in c38129f8ec: agreed on the simplification for getManifests(dir, dir). getVersionSignalFromManifests(...) now reads only the single local manifest entry (const [packageJson] = ...) and handles the no-manifest case explicitly.

packages/analysis/src/jsts/program/tsconfig/options.ts
vdiez: Applied in c38129f8ec: I removed the duplicated strictness-option union and now derive the type from the single STRICTNESS_COMPILER_OPTIONS const list via (typeof STRICTNESS_COMPILER_OPTIONS)[number].

packages/analysis/src/jsts/program/tsconfig/options.ts
vdiez: Correction: kept strict=false intentionally. defaultCompilerOptions is a fallback baseline when program options are created without project tsconfig strictness context; TS6 strictness behavior remains preserved via applyLegacyStrictnessDefaults.

packages/analysis/src/jsts/program/tsconfig/options.ts
vdiez: Correction: updated in c38129f8ec by replacing rootConfig:any with rootConfig:unknown and narrowing through getCompilerOptions.

vdiez: Plain-text clarification: c38129f8ec replaces rootConfig:any with rootConfig:unknown and narrows it through getCompilerOptions helper.

packages/analysis/src/jsts/program/tsconfig/options.ts
vdiez: Plain-text clarification: strict=false is intentional for fallback compiler options. TypeScript 6 strictness defaults are still preserved in the tsconfig parsing path through applyLegacyStrictnessDefaults gating.

PR #6795, opened 2026-04-14
JS-1536 Save orchestrator cache in post-job phase

zglicz:
## Summary
- replace the default-branch split restore/save flow in orchestrator-cache with a single `actions/cache@v5` step
- keep branch and `save: false` callers on restore-only behavior
- ensure default-branch cache writes happen in the post-job phase instead of immediately when the composite action runs

## Testing
- ruby -e "require 'yaml'; YAML.load_file('.github/actions/orchestrator-cache/action.yml'); puts 'YAML OK'"
- git diff --check -- .github/actions/orchestrator-cache/action.yml

Supersedes #6783.


zglicz:
This is what the fix should've been from the start. I didn't check the AI code properly in #6783

vdiez (approved):

PR #6807, opened 2026-04-14
JS-1589 Apply monorepo grouping rule to maven updates

vdiez:
## Summary
- apply the existing monorepo/fallback grouping rule to both 
pm and maven
- stop inheriting the single global Maven grouping behavior from the shared preset

## Why
Align Maven behavior with the npm strategy so updates are grouped by suite/monorepo when possible, and otherwise split per dependency.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -26,8 +26,8 @@
       "matchDepNames": ["/^org\\.sonarsource\\.javascript:/"]
     },
     {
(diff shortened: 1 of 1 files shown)


zglicz (approved):

vdiez:
Closing: we keep monorepo grouping for npm only, not maven.

PR #6812, opened 2026-04-14
JS-366: detect commented JSX attributes in S125

vdiez:
﻿## Summary
- reproduce JS-366: S125 misses commented JSX attributes like `// views={["year"]}`
- keep `containsCode` parser-driven and add a parser fallback that wraps comment text as JSX attributes (`<_S125Wrapper ... />`) when standalone parse fails
- validate fallback AST shape to only accept parsed JSX attributes, avoiding `TODO:` namespace-like prose matches
- remove extension-based gating so JSX-in-`.js` is also handled by parser outcome

## Tests
- `npx tsx --tsconfig packages/tsconfig.test.json --test packages/analysis/src/jsts/rules/S125/unit.test.ts`

### DIFF
--- a/its/ruling/src/test/expected/console/typescript-S125.json
+++ b/its/ruling/src/test/expected/console/typescript-S125.json
@@ -11,6 +11,9 @@
 "console:src/views/FunctionsView/FunctionLogs/Log.tsx": [
 6
 ],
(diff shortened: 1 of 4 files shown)


francois-mora-sonarsource (approved):
LGTM

PR #6830, opened 2026-04-15
JS-1462 Fix S6582 over-suppression by suppressing only known false positives

francois-mora-sonarsource:
Relates to [JS-1462](https://sonarsource.atlassian.net/browse/JS-1462)

S6582 flags `a && a.prop` and `!a || !a.prop` as replaceable with optional chaining without checking whether the replacement is type-safe at the usage site. When these expressions are used in typed contexts — assigned to a typed variable, returned from a typed function, or passed as a typed argument — the rewrite changes `T | null` to `T | undefined`, breaking the surrounding type contract.

The fix intercepts reports and suppresses them only in contexts where `undefined` is not assignable to the expected type, using a centralized `isKnownFalsePositive()` predicate that covers five pattern categories.

- **Strategy reversal**: report by default, suppress only known false positives — eliminates the previous fallback-to-suppress fragility
- **Bounded suppression**: boolean/void contexts (`if`, `while`, ternary test) impose no contextual type and still report correctly
- Ruling test JSON updates reflect issues that were previously over-suppressed now correctly reported

[JS-1462]: https://sonarsource.atlassian.net/browse/JS-1462?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


francois-mora-sonarsource:
Clarification only: I do not think this point requires any change in the PR.

My conclusion is that the current false-positive remediations should be kept. The problem is not only that rewriting `a && a.b` to `a?.b` may introduce compilation errors by replacing `null` with `undefined`, but also that in these cases the rule stops being a simple readability improvement and starts pushing against an explicit `null`-based type contract chosen by the developer.

When the rewrite changes `T | null` into `T | undefined`, the fix is no longer straightforward: it may require changing a variable type, return type, property type, or parameter contract, or adding compensating code such as `?? null`.

For S6582, that tradeoff is not worth it. We cannot assume that `null` and `undefined` are interchangeable, so these suppressions are aligned with the intended policy and should remain as they are.

zglicz (approved):
I'll accept but this so hard to grasp for me :( I'm not sure if its good or bad. I do like the changes in the ruling though and agree with them...

francois-mora-sonarsource:
Plan before merge: I created  and will dogfood this change on Peach first to measure its impact. We’ll review the resulting deltas on Monday before deciding whether to merge.

francois-mora-sonarsource:
Plan before merge: I created the dogfood/JS-1462 branch and will dogfood this change on Peach first to measure its impact. We’ll review the resulting deltas on Monday before deciding whether to merge.

PR #6847, opened 2026-04-16
JS-736 - Search for deno.json files during lookup

nathsou:
Part of [JS-736](https://sonarsource.atlassian.net/browse/JS-736)

[JS-736]: https://sonarsource.atlassian.net/browse/JS-736?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Scope of this PR:

- [x] Detect `npm:` dependencies declared inside `deno.json` and `deno.jsonc` files
- [x] Invalidate caches when a deno manifest file is changed
- [x] Deno workspaces

To be handled later:
- inline import declarations: `import React from 'npm:react@18';`
- non-npm imports: `jsr:`, `https:`
- Add deno projects to ruling and peachee-js
- [Workspace path patterns](https://docs.deno.com/runtime/fundamentals/workspaces/#workspace-path-patterns)
- importMap property


packages/analysis/src/file-stores/dependency-manifests.ts
vdiez, line 1: we should rename this file to `depedency-manifests.ts`

packages/analysis/src/jsts/rules/helpers/dependency-manifests/dependencies.ts
```diff
@@ -66,11 +69,11 @@ export const dependenciesCache = new ComputedCache(
  */
 export const moduleTypeCache = new ComputedCache(
   (dir: NormalizedAbsolutePath, topDir?: NormalizedAbsolutePath): ModuleType | undefined => {
-    const closestPackageJSONDirName = getClosestPackageJSONDir(dir, topDir);
-    if (!closestPackageJSONDirName) {
+    const closestDependencyManifestDirName = getClosestDependencyManifestDir(dir, topDir);
```
vdiez, line 74: I think in this function we should not be affected by `deno.json` files. With current logic, if we find a `deno.json` and there is a `package.json` file in a parent folder, we will return `commonjs`. We either search for only `package.json` if we want to continue to rely on the "type" field, or if we find a deno.json we shortcircuit to "module" as I think all deno projects are "esm"

packages/analysis/src/jsts/rules/helpers/package-jsons/parse.ts
```diff
@@ -73,6 +102,86 @@ function addDependencies(
...
+  }
+}
+
+function parseDenoManifest(file: File): DenoManifest | undefined {
+  try {
+    // ts.parseConfigFileTextToJson handles JSON with comments and trailing commas
+    const parsed = ts.parseConfigFileTextToJson(file.path, stripBOM(file.content.toString()));
+    if (parsed.error) {
+      const message = ts.flattenDiagnosticMessageText(parsed.error.messageText, '\n');
+      throw new Error(message);
```
vdiez: as you are catching in the very same function, i would rather just log directly here and return undefined

packages/analysis/src/telemetry.ts
```diff
@@ -312,7 +312,7 @@ function resolveTypeScriptVersion(typeScriptSignal: string): string | undefined
 
 function getPackageJsonFiles(): Iterable<{ content: string | Buffer }> {
   try {
-    return packageJsonStore.getPackageJsons().values();
+    return dependencyManifestStore.getPackageJsons().values();
```
vdiez, line 315: do we want to add `deno.json` files to telemetry @zglicz ?

vdiez (commented):
Looking good! I would be curious to see how we merge from child path to parent paths. I see you have a test where both are present at the same level (probably would be nice to log if both manifest declare different versions).

packages/analysis/src/jsts/rules/helpers/dependency-manifests/parse.ts
```diff
@@ -0,0 +1,209 @@
...
+  }
+}
+
+// https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/script/type/importmap
+type ImportMap = Record<string, unknown>;
+
+export type DenoManifest = {
+  imports?: ImportMap;
+  workspace?: string[] | { members?: string[] };
+};
```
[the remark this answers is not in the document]
nathsou: Out of scope for this task, I will add it to the description

vdiez (approved):
LGTM!

PR #6849, opened 2026-04-16
Recognize test.prop and it.prop from @fast-check/vitest in `S2187`

pavel-kalmykov:
`S2187` flags test files that use `test.prop` from [`@fast-check/vitest`](https://github.com/dubzzz/fast-check/tree/main/packages/vitest) as empty. The rule knows about `test.each`, `test.only`, `test.for`, etc., but not `test.prop`.

```ts
import { test, fc } from '@fast-check/vitest';

// S2187 reports: "Add some tests to this file or delete it."
test.prop([fc.string(), fc.string(), fc.string()])('should detect substrings', (a, b, c) => {
  expect((a + b + c).includes(b)).toBe(true);
});
```

This PR adds `test.prop` and `it.prop` to the `APIs` set in the `S2187` rule. They follow the same curried-call pattern as `test.each`, so just adding the names to the set is enough. Two test cases added to `unit.test.ts`.

References:
- [`@fast-check/vitest` on GitHub](https://github.com/dubzzz/fast-check/tree/main/packages/vitest)
- [Property-based testing with Vitest tutorial](https://fast-check.dev/docs/tutorials/setting-up-your-test-environment/property-based-testing-with-vitest/)

### DIFF
--- a/packages/analysis/src/jsts/rules/S2187/rule.ts
+++ b/packages/analysis/src/jsts/rules/S2187/rule.ts
@@ -78,6 +78,9 @@ const APIs = new Set([
   'it.runIf',
   'test.for',
   'it.for',
+  // @fast-check/vitest (property-based testing)
+  'test.prop',
+  'it.prop',
   // eslint rule tester
   'ruleTester.run',
 ]);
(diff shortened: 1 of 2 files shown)


vdiez:
Thanks for the contribution. I opened #6900 with your commit cherry-picked into this repository so CI can run there, since pull requests from external forks do not trigger our CI.

I am closing this PR in favor of #6900.

PR #6891, opened 2026-04-21
JS-1027 Migrate runtime analyze-project transport to gRPC

vdiez:
## Summary
- supersede and fold together #6871 and #6878 into a single `master`-targeted branch
- replace the runtime analyze-project HTTP/WebSocket transport with a dedicated internal typed gRPC contract while keeping `LanguageAnalyzerService` unchanged
- remove the remaining JSON/Gson analyze-project bridge handling and migrate Java/Node consumers to generated protobuf messages
- replace heartbeat/HTTP-close lifecycle management with a long-lived `Lease` ownership stream plus explicit `CancelAnalysis`
- keep production worker isolation, restore no-worker analyze-project server mode for tests/debugging, and delete the obsolete `packages/http` stack
- add follow-up hardening for typed request handling, worker cloning/memory behavior, and Java-side cancellation semantics

## Context
This PR is the master-targeted consolidation of the earlier gRPC migration work from #6871 and the typed-contract follow-up from #6878, together with the latest hardening done on this branch.

### Contract and runtime
- Treat `packages/grpc/src/proto/analyze-project.proto` as an internal Java <-> Node contract, intentionally independent from `language_analyzer.proto`
- Model the request after the actual analyze-project workflow: `configuration`, `files`, `rules`, `css_rules`, `bundles`, optional `rules_workdir`
- Keep responses fully typed:
  - unary: file map + meta
  - streaming: `oneof` file_result / meta / cancelled
  - AST payloads as embedded typed `estree.Node` protobuf messages
- Surface fatal request/runtime failures as gRPC status errors instead of legacy streamed JSON error messages
- Keep runtime single-flight and reject concurrent analyses with `RESOURCE_EXHAUSTED`
- Keep the worker for production isolation while still supporting in-process server mode for tests/debugging

### Node boundary
- Remove the analyze-project JSON transport path entirely: no JSON envelope, no `request_json` / `response_json`, no raw-object validation for this workflow
- Add a typed normalization layer in `packages/grpc/src/analyze-project-normalize.ts` for the semantic work protobuf cannot do for us:
  - validate and normalize `base_dir`, bundles, `rules_workdir`, tsconfig paths, fs events, and file paths
  - require absolute `base_dir`
  - initialize stores from disk when `files` is omitted and filesystem access is allowed
  - read file contents from disk when a file entry omits inline content
  - infer file type when left unspecified
  - default file status when left unspecified
  - preserve configuration defaulting, ignore filtering, and store initialization behavior before analysis
- Add typed response conversion in `packages/grpc/src/analyze-project-convert.ts`
- Preserve sparse/manual-client requests instead of requiring a fully-populated Java-style payload
- Handle `max_file_size` correctly across worker structured-clone boundaries
- Convert stream and unary responses inside the worker so request-scoped `pathMap`s and final accumulated project outputs stay local to the worker thread instead of being cloned back unnecessarily

### Java boundary
- Build analyze-project protobuf messages directly from Java using generated types and a thin adapter instead of bridge-local DTOs + Gson
- Centralize Java-side protobuf construction in `AnalyzeProjectMessages` for:
  - `ProjectConfiguration`
  - `ProjectFileInput`
  - JS/TS rules
  - CSS rules
  - protobuf `Value` conversion for rule configurations
- Update the main Java consumers to use generated protobuf messages directly:
  - `BridgeServerImpl`
  - `WebSensor`
  - `AnalysisProcessor`
  - `PluginTelemetry`
  - `QuickFixSupport`
  - cache / CPD serializers and deserializers
  - `ExternalIssueRepository`
  - `StandaloneParser`
- Consume AST directly as typed `estree.Node` protobuf messages instead of base64-encoded serialized payloads
- Keep generated Java protobuf/gRPC sources build-time only under `target/generated-sources`

### Runtime lifecycle and cancellation
- Replace the old heartbeat ping and HTTP close path with a long-lived `Lease` stream
- Use the lease as a Java -> Node ownership signal for the runtime spawned by `server.mjs`
- Skip lease ownership entirely for externally-managed debug runtimes attached through `SONARJS_EXISTING_NODE_PROCESS_PORT`
- Keep a startup grace timeout only before lease acquisition so a spawned runtime does not live forever if Java never connects
- Keep explicit `CancelAnalysis` for in-flight analysis cancellation
- Harden Java-side streaming cancellation so a cancelled `SensorContext` interrupts a blocked gRPC stream read instead of waiting for the next message to arrive
- Translate transport-side cancellation back to the existing plugin `CancellationException`
- Add focused deadlines only for control/unary calls; the main streaming analyze-project RPC remains intentionally unbounded

### Transport and dependency cleanup
- Delete the legacy `packages/http` runtime package and its tests
- Remove HTTP/WebSocket-specific Java bridge classes and transport handling
- Add gRPC runtime dependencies / generated artifacts / third-party licenses needed for the new transport
- Use `grpc-netty-shaded` for the Java analyze-project channel instead of `grpc-okhttp`
- Refresh the plugin third-party license inventory for the updated Java dependency graph and normalize license-regeneration URL replacements so the generated gRPC-related licenses remain `.txt`
- Update build/docs/tooling to reflect the new gRPC-based runtime path

### Follow-up hardening captured from review and branch work
- Guard optional proto setters such as `rulesWorkdir` so nullable Java state does not break request construction
- Map null `InputFile` metadata to `_UNSPECIFIED` proto values instead of throwing during request construction
- Preserve the previous Gson-era rule-config behavior by serializing existing bean/POJO configs into protobuf `Value` / `Struct`
- Keep user-visible consumer behavior stable after the transport switch (typed location handling, optional-field presence checks, AST handling, cache consumers)
- Add regression coverage around:
  - `AnalyzeProjectMessages`
- typed request normalization, including protobuf value decoding and omitted-files vs empty-files semantics
  - bridge startup/runtime wiring
  - no-worker + worker gRPC server modes
  - path restoration through `pathMap`
  - worker protocol forwarding / cancellation
  - Java-side streaming cancellation while blocked on the next message

## Accepted Design Decisions / Remaining Edge Case
- The `Lease` stream is an ownership/lifecycle mechanism, not a health probe
- The unary analyze-project path is intentionally simple because it is used by `StandaloneParser`, not by the production scanner / SonarLint analysis flow
- gRPC message sizes remain effectively unbounded by design because a single request may legitimately carry a whole project worth of files and contents; memory is the practical limit unless the protocol is redesigned around batching or client streaming
- The remaining known corner case is the ambiguity between an explicitly empty `files` map and an omitted `files` map when filesystem access is allowed

## Validation
Validation carried over from #6871 / #6878 and the current branch work includes:

- `cmd /c node_modules\\.bin\\tsx.cmd --tsconfig packages/tsconfig.test.json --test --test-reporter=spec --test-reporter-destination stdout packages/grpc/tests/analyze-project-server.test.ts`
- `SONARSOURCE_QA=true mvn -pl its/plugin/tests -am verify -DskipTests=false`
- `SONARSOURCE_QA=true mvn -pl its/plugin/fast-tests -am verify -DskipTests=false`
- `mvn -pl sonar-plugin/bridge -Dlicense.skip=true -Dtest=BridgeServerImplTest test`
- `mvn -pl sonar-plugin/bridge -Dlicense.skip=true -Dtest=AnalyzeProjectMessagesTest,BridgeServerImplTest test`
- `mvn -pl sonar-plugin/sonar-javascript-plugin -am -Dlicense.skip=true -Dsurefire.failIfNoSpecifiedTests=false -Dtest=WebSensorTest,AnalysisProcessorTest,PluginTelemetryTest,QuickFixSupportTest test`
- `mvn -pl sonar-plugin/standalone -am -Dlicense.skip=true -Dsurefire.failIfNoSpecifiedTests=false -Dtest=StandaloneParserTest test`
- `npm.cmd run bridge:compile`
- `C:\www\projects\SonarJS\node_modules\.bin\tsx.cmd --tsconfig packages/tsconfig.test.json --test --test-concurrency=1 --test-reporter=spec --test-reporter-destination stdout packages/grpc/tests/analyze-project-worker.test.ts`
- `C:\www\projects\SonarJS\node_modules\.bin\tsx.cmd --tsconfig packages/tsconfig.test.json --test --test-concurrency=1 --test-reporter=spec --test-reporter-destination stdout packages/grpc/tests/analyze-project-server.test.ts`
- `mvn --% -pl sonar-plugin/bridge -am -Dtest=BridgeServerImplTest#should_cancel_streaming_analysis_while_waiting_for_the_next_message -Dsurefire.failIfNoSpecifiedTests=false test`
- `mvn --% -pl sonar-plugin/sonar-javascript-plugin -am -Dtest=WebSensorTest#stop_analysis_if_cancelled -Dsurefire.failIfNoSpecifiedTests=false test`


vdiez:
Review follow-up after discussion:

- Transport readiness / lease semantics: accepted by design.
  The lease is being used as a Java -> Node ownership signal, especially for SonarLint, so the owned Node runtime exits when the JVM / IDE goes away. It is not intended to be an analyzer-health probe. In that sense the current lease behavior is correct, and this is not a blocker for the branch.

- Empty `files` map vs omitted `files`: still the only known unresolved corner case.
  With protobufjs map decoding, `files: {}` and omitted `files` remain ambiguous when filesystem access is allowed, so an explicitly empty request can still fall back to disk discovery. This is the main protocol edge case that still looks worth addressing.

- Unary cancellation hook: accepted by design.
  The unary RPC is used by `StandaloneParser` as an internal/build-time tool to parse a single file and return its AST. It is not used in production analysis, and there is no expectation that this path needs a separate cancellation policy.

- Unlimited gRPC message size: accepted by design.
  `AnalyzeProjectRequest` may legitimately carry a whole project worth of files and contents, so any hard message-size cap would be arbitrary. In the current protocol shape, memory is the intended practical limit unless the request model is redesigned to batch or stream inputs.

Separate from those points, the worker-side response conversion / cloning reduction is still a useful improvement because it lowers avoidable memory pressure without changing any of the accepted-by-design transport choices above.

docs/BUILD.md
```diff
@@ -1,509 +1,167 @@
...
-2. javascript-checks
-3. bridge
-4. css
-5. sonar-javascript-plugin
-6. standalone
-```
+## Default workflow
 
-Maven builds modules respecting their dependencies - modules without dependencies between them may run in parallel when using `-T` flag. The declaration order matters when modules have no explicit dependencies but rely on side effects (like generated files).
+Use `mvn install` for normal development.
```
zglicz: npm ci <- is a prerequisite

docs/BUILD.md
```diff
@@ -1,509 +1,167 @@
...
+```bash
+mvn -pl sonar-plugin/sonar-javascript-plugin -am -Dskip-nodejs -Dskip-java-generation test-compile
+```
 
-Enables JaCoCo coverage aggregation.
+Bridge-only Node build:
 
 ```bash
-mvn clean install -P coverage-report
+npm run bridge:build:fast
```
zglicz, line 40: npm ci beforehand

docs/BUILD.md
```diff
@@ -1,509 +1,167 @@
...
-- Enables `its/ruling`
-- Runs against SonarQube orchestrator
-- Note: `-DskipTests=false` is required because integration test modules have `skipTests=true` by default in their pom.xml
+## Fast Java iteration flags
 
----
+The fast path is implemented with property-activated profiles in `sonar-plugin/pom.xml`.
 
-## Key Maven Plugins
+### `-Dskip-nodejs`
```
zglicz, line 75: haven't seen this before

packages/analysis/src/common/input-sanitize.ts
```diff
@@ -111,39 +113,41 @@ export async function sanitizeProjectAnalysisInput(
...
+
 type SanitizedInputFiles = {
   files: AnalyzableFiles;
   pathMap: Map<string, string>;
 };
 
-export async function sanitizeRawInputFiles(
-  rawFiles: Record<string, unknown> | undefined,
+export async function sanitizeInputFiles(
+  inputFiles: Record<string, ProjectAnalysisFileInput> | undefined,
```
zglicz, line 133: thats nice

packages/grpc/src/proto/analyze-project.proto
```diff
@@ -0,0 +1,270 @@
...
+}
+
+message LeaseRequest {}
+
+message LeaseResponse {}
+
+message ProjectConfiguration {
+  string base_dir = 1;
+  optional bool sonarlint = 2;
+  repeated string fs_events = 3;
```
zglicz, line 55: why not a StringList? or what is the difference? why no consistency?

packages/grpc/src/proto/analyze-project.proto
```diff
@@ -0,0 +1,270 @@
...
+  optional bool report_ncloc_for_test_files = 33;
+}
+
+message StringList {
+  repeated string values = 1;
+}
+
+message ProjectFileInput {
+  optional string file_content = 1;
+  FileType file_type = 2;
```
zglicz, line 97: if instead of adding the _UNPECIFIED, perhaps we could make these optional? I'm not a fan of adding unnecessary constants. I don't know.

packages/grpc/src/proto/analyze-project.proto
```diff
@@ -0,0 +1,270 @@
...
+  TEXT_TYPE_CONSTANT = 1;
+  TEXT_TYPE_COMMENT = 2;
+  TEXT_TYPE_STRUCTURED_COMMENT = 3;
+  TEXT_TYPE_KEYWORD = 4;
+  TEXT_TYPE_STRING = 5;
+}
+
+message JsTsRule {
+  string key = 1;
+  repeated google.protobuf.Value configurations = 2;
```
zglicz, line 153: so this can be anything?

packages/grpc/src/proto/analyze-project.proto
```diff
@@ -0,0 +1,270 @@
...
+}
+
+message Issue {
+  int32 line = 1;
+  int32 column = 2;
+  optional int32 end_line = 3;
+  optional int32 end_column = 4;
+  string message = 5;
+  string rule_id = 6;
+  AnalysisLanguage language = 7;
```
zglicz, line 197: surprising that issue need to have a language. I would think it would be derived from the file

packages/grpc/src/proto/analyze-project.proto
```diff
@@ -0,0 +1,270 @@
...
+  string text = 1;
+  IssueLocation loc = 2;
+}
+
+message IssueLocation {
+  optional int32 line = 1;
+  optional int32 column = 2;
+  optional int32 end_line = 3;
+  optional int32 end_column = 4;
+  optional string message = 5;
```
zglicz, line 222: everything is optional? Sounds excessive optionality

packages/grpc/src/analyze-project-handle-request.ts
```diff
@@ -45,31 +59,36 @@ export async function handleRequest(
...
+          result: {
+            output,
+            pathMap: sanitizedInput.pathMap,
+          },
+        };
       }
       case 'on-cancel-analysis': {
         cancelAnalysis();
-        return { type: 'success', result: 'OK' };
+        return { type: 'success', result: undefined };
```
zglicz, line 77: is that a helpful result? Isn't optional result good enough?

zglicz, line 77: isn't the result supposed to be `{ cancelled: true; }`?

packages/grpc/src/analyze-project-normalize.ts
```diff
@@ -0,0 +1,426 @@
...
+
+function createConfigurationFromProto(configuration: ProjectConfiguration | null | undefined) {
+  if (!configuration?.baseDir) {
+    throw new InvalidAnalyzeProjectRequestError('configuration.base_dir is required');
+  }
+
+  const input: ConfigurationInput = {
+    baseDir: configuration.baseDir,
+    sonarlint: optionalBoolean(configuration.sonarlint),
+    fsEvents: arrayOrUndefined(configuration.fsEvents),
```
zglicz: I don't understand why this is different to stringListValues

packages/grpc/src/analyze-project-normalize.ts
```diff
@@ -0,0 +1,426 @@
...
+      ? value
+      : isLongLike(value)
+        ? value.toNumber()
+        : isClonedLongLike(value)
+          ? toNumberFromClonedLong(value)
+          : Number(value);
+  if (!Number.isFinite(numberValue) || !Number.isSafeInteger(numberValue)) {
+    throw new InvalidAnalyzeProjectRequestError(`${path} must be a safe integer`);
+  }
+  return numberValue;
```
zglicz: what sort of magic is happening here? LongLike? high,low

packages/grpc/src/analyze-project-request.ts
```diff
@@ -14,34 +14,47 @@
...
 };
 
 type CancellationRequest = {
   type: 'on-cancel-analysis';
 };
 
-type SerializedError = {
+export type SerializedError = {
   code: ErrorCode;
   message: unknown;
```
zglicz: still an error message is unknown?

packages/grpc/src/analyze-project-server.ts
```diff
@@ -0,0 +1,760 @@
...
+    closeLeaseCall();
+    unregisterGarbageCollectionObserver();
+    await cancelAnalysisOnShutdown(handleRequestInCurrentThread, newWorkerRequestId, state, worker);
+    await closeWorker(worker);
+
+    server.forceShutdown();
+    resolveClosed();
+  };
+
+  const armStartupShutdownTimeout = () => {
```
zglicz: why arm? what does that acronym mean?

zglicz (commented):
I went through the js/ts files just with some small questions for my understanding. I'm getting info, that you are still updating the PR, so I'll continue with the Java part after lunch.

docs/BUILD.md
vdiez: Good catch. I updated `BUILD.md` to call out `npm ci` as the checkout prerequisite before the normal Maven flow.

docs/BUILD.md
vdiez: Added that note as well under the bridge-only Node build command, since `npm run bridge:build:fast` also assumes dependencies are already installed.

docs/BUILD.md
vdiez: It is project-specific, not a Maven builtin. The `-Dskip-nodejs` property activates profiles in `sonar-plugin/pom.xml` that skip the Node/protobuf generation path for faster Java-only loops; I clarified that in `BUILD.md`.

packages/grpc/src/proto/analyze-project.proto
vdiez: The difference is presence semantics. I use `StringList` only where `unset` means "use runtime defaults" and `[]` means "explicitly override to empty"; for `fs_events` and the path lists, omitted and empty collapse to the same behavior, so plain repeated strings are enough. I added comments in the proto and renamed the normalization helper to `repeatedStringValues` to make that distinction clearer.

packages/grpc/src/proto/analyze-project.proto
vdiez: I kept `_UNSPECIFIED` on purpose. Optional enum presence would only help on singular fields like `ProjectFileInput`, but the same enums are also used in repeated fields (`file_type_targets`, `analysis_modes`) and proto3 still needs a zero value anyway, so `_UNSPECIFIED` gives one consistent "not provided" representation across the contract.

packages/grpc/src/proto/analyze-project.proto
vdiez: Not literally arbitrary, but intentionally JSON-like: `google.protobuf.Value` covers null/bool/number/string/list/struct, which matches the existing heterogeneous rule-config shape. We also still need to support Java-side POJO configs that are serialized into `Value`/`Struct`, so a narrower schema would regress current behavior. I added a comment on the field.

packages/grpc/src/proto/analyze-project.proto
vdiez: Often it matches the file language, but not always. A single analyzed file can produce issues from different analyzers / embedded languages (JS/TS/CSS), so the file path alone is not a reliable discriminator; I added a short comment on the field.

packages/grpc/src/proto/analyze-project.proto
vdiez: Only `IssueLocation` is this permissive, and that is deliberate. Some secondary / quick-fix locations can arrive without a full range (for example placeholder/default instances), and the Java side already ignores incomplete locations instead of inventing coordinates; tightening the schema would force synthetic data or dropped entries. I added that rationale in the proto.

packages/grpc/src/analyze-project-handle-request.ts
vdiez: Here the result is only an internal ack from the worker/in-process runtime: success means "the cancel signal was delivered", failure means it was not. The outward-facing `{ cancelled: boolean }` response is computed one layer up by `CancelAnalysis` in `analyze-project-server.ts`, because only the server knows whether an analysis was actually active.

vdiez: Yes, that shape exists one layer up on the public `CancelAnalysisResponse`. This internal handler returns only success/failure; the server wraps that into `{ cancelled: hadActiveAnalysis && requestCancel() }`. I added a comment here to make the layering explicit.

packages/grpc/src/analyze-project-normalize.ts
vdiez: It was just schema-shape leakage from the proto. `stringListValues` reads the wrapped `StringList` message, while this helper reads plain repeated-string fields; I renamed it to `repeatedStringValues` so the difference is visible in the code now.

packages/grpc/src/analyze-project-normalize.ts
vdiez: This is protobufjs `int64` handling. Before the worker hop, `max_file_size` can be a Long-like object with `toNumber()`; after worker structured clone, that prototype is gone and only `{ low, high, unsigned }` remains, so we reconstruct the integer manually. I renamed the type to `StructuredCloneLongLike` and documented that inline.

packages/grpc/src/analyze-project-request.ts
vdiez: Good point. I tightened `SerializedError.message` back to `string` and now normalize non-Error throws during serialization instead of carrying `unknown` through the stack.

packages/grpc/src/analyze-project-server.ts
vdiez: No acronym here, just the "arm a timer" idiom. I renamed it to `scheduleStartupShutdownTimeout` because that is clearer.

packages/grpc/src/proto/analyze-project.proto
vdiez: adding myself instead of codex: since we merged sensor, we need to put language on the issues, because an html file can have js and css issues

sonar-plugin/bridge/src/main/java/org/sonar/plugins/javascript/bridge/BridgeServerImpl.java
```diff
@@ -479,20 +577,99 @@ public void stop() {
...
+  private void closeChannel() {
+    if (channel == null) {
+      return;
+    }
     try {
-      return new URI("ws", null, hostAddress, port, "/ws", null, null);
-    } catch (URISyntaxException e) {
-      throw new IllegalStateException("Invalid URI: " + e.getMessage(), e);
+      channel.shutdownNow();
+      channel.awaitTermination(1, TimeUnit.SECONDS);
```
zglicz, line 673: both needed?

sonar-plugin/bridge/src/main/java/org/sonar/plugins/javascript/bridge/BridgeServerImpl.java
```diff
@@ -479,20 +577,99 @@ public void stop() {
...
+      Thread.currentThread().interrupt();
+    } finally {
+      channel = null;
+      leaseObserver = null;
+      leaseTerminated = true;
     }
   }
 
-  private URI url(String endpoint) {
+  private synchronized void startLease() {
```
zglicz, line 683: what is the meaning of the "Lease"?

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/analysis/cache/CacheAnalysis.java
```diff
@@ -18,7 +18,7 @@
 
 import java.util.List;
 import javax.annotation.Nullable;
-import org.sonar.plugins.javascript.bridge.BridgeServer.CpdToken;
+import org.sonar.plugins.javascript.analyzeproject.grpc.CpdToken;
```
zglicz, line 21: these are things we cache somehwer for PPR analysis? These types didn't change?

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/analysis/WebSensor.java
```diff
@@ -289,27 +299,32 @@ public ProjectAnalysisRequest getRequest() {
...
       }
-      configuration.setSkipAst(context.skipAst(consumers));
-      var request = new BridgeServer.ProjectAnalysisRequest(
-        files,
-        checks.enabledEslintRules(),
-        configuration
-      );
-      request.setCssRules(cssRules.getStylelintRules());
-      return request;
+      configurationBuilder.setSkipAst(context.skipAst(consumers));
```
zglicz, line 310: for the future, can this skipAst be part of the configuration? It shouldn't be a special flag floating around

sonar-plugin/sonar-javascript-plugin/src/test/java/org/sonar/plugins/javascript/analysis/cache/CacheTestUtils.java
```diff
@@ -53,7 +54,20 @@ public static CpdSerializer.SerializationResult getSerializedCpdTokens(List<CpdT
   }
 
   public static List<CpdToken> getCpdTokens() {
-    return new Gson().fromJson(CPD_TOKENS, CpdData.class).getCpdTokens();
```
zglicz, line 56: aah, they were already gson structs, carry on

sonar-plugin/standalone/src/main/java/org/sonar/plugins/javascript/standalone/StandaloneParser.java
zglicz, line 1: Did you notify the folks from Jasmin/Taint/Security - so they are aware?

sonar-plugin/pom.xml
```diff
@@ -31,4 +36,33 @@
...
+      <id>skip-java-generation</id>
+      <activation>
+        <property>
+          <name>skip-java-generation</name>
+        </property>
+      </activation>
+      <properties>
+        <java.generation.phase>none</java.generation.phase>
+      </properties>
+    </profile>
```
zglicz, line 66: this could've should've been a separate PR

zglicz (approved):
Its okay, or maybe not. Overall looks fine. A bit cleaner, but I see we carry over all of the error handling/special casing we had before. Perhaps, as you mentioned, it'd be good to have a peachy run to see how it behaves performance wise before pushing to master.

sonar-plugin/bridge/src/main/java/org/sonar/plugins/javascript/bridge/BridgeServerImpl.java
vdiez: Not strictly both for initiating shutdown. `shutdownNow()` is the part that actually starts teardown and cancels active RPCs; `awaitTermination(1s)` is only a bounded wait so the transport can quiesce before we clear the channel reference and potentially recreate it later. I kept the wait to avoid making close fire-and-forget.

sonar-plugin/bridge/src/main/java/org/sonar/plugins/javascript/bridge/BridgeServerImpl.java
vdiez: Here lease means ownership/lifecycle, not analysis payload. Java opens this long-lived stream only when it owns the Node child process. If the JVM / IDE disappears or the channel drops, Node sees the lease terminate and exits instead of staying orphaned. For externally managed runtimes (`SONARJS_EXISTING_NODE_PROCESS_PORT`) we do not start the lease at all.

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/analysis/cache/CacheAnalysis.java
vdiez: Yes, this is the PPR cache payload for unchanged JS/TS files: CPD tokens plus optional AST. The persisted cache format did not change here; this import switch is only the in-memory type moving from the old `BridgeServer.CpdToken` record to the generated gRPC `CpdToken`, which carries the same `location` / `image` data.

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/analysis/WebSensor.java
vdiez: Agreed. In this PR it already moved under `ProjectConfiguration` (`skip_ast` in the proto), and this setter is just filling that configuration field. The only remaining special logic is deciding on the Java side whether any consumer actually needs ASTs for the current analysis.

sonar-plugin/standalone/src/main/java/org/sonar/plugins/javascript/standalone/StandaloneParser.java
vdiez: Not specifically yet. I treated this as an internal transport change because `StandaloneParser`'s public API is unchanged: it still takes source text and returns ESTree, only the Java <-> Node hop became unary gRPC. Security is the main external consumer of this path though, so a heads-up is reasonable.

sonar-plugin/pom.xml
vdiez: Fair. I kept `skip-nodejs` in this branch because it made the Java-side migration/review loop tractable while the transport was in flux, but in hindsight it is build tooling and could have been split.

sonar-plugin/bridge/src/main/java/org/sonar/plugins/javascript/bridge/BridgeServerImpl.java
vdiez: this is the replacement for the old heartbeat mechanism. it's a long-lived stream (like a websocket connection). if nodejs sees this stream disconnecting, it's because java side disappeared (user closed IDE), so it will exit.

vdiez:
Final HTTP vs gRPC compare

Compared `origin/master` (`28213ccc2`, merge-base `1db0e513f`) against current `HEAD` (`e81561724`) with a clean worktree.

**Summary**
- For the production `analyze-project` flow used by SonarQube/SonarLint, this branch is very close to functionally equivalent. The analyzer engine, per-file plus final-meta result model, downstream Java consumers, process topology, and standalone parser feature contract all remain in place.
- The biggest deltas are transport-level, not analyzer-level: gRPC/protobuf replaces HTTP + WebSocket + JSON/base64, lifecycle ownership moved from heartbeat/inactivity timeout to an explicit lease, error propagation moved to gRPC statuses, and request normalization is now strict and typed.
- I did not find a production gRPC analysis deadline. `BridgeServerImpl.analyzeProject(ProjectAnalysisHandler)` uses the streaming stub without `withDeadlineAfter(...)`, and `packages/grpc/src/analyze-project-server.ts` uses `WORKER_RESPONSE_TIMEOUT_MS = 15_000` only for worker control-flow waits (cancel/shutdown/completion), not to cap analysis duration.
- The one remaining semantic gap I can confirm in the typed request model is the `files: {}` vs omitted `files` ambiguity when `canAccessFileSystem = true`.

**Confirmed Functional Equivalence**
- Same launcher/process/port model: `server.mjs` still takes `port`, `host`, `debugMemory`, `timeoutSeconds`; `BridgeServerImpl` still binds loopback, calls `findOpenPort()`, and starts one Node child. `SONARJS_EXISTING_NODE_PROCESS_PORT` is still supported. Parallel analyses still isolate by separate Node process + loopback port, not by sharing one transport endpoint.
- Same analysis engine: master HTTP sanitized input and then called `analyzeProject()`. gRPC now normalizes the proto request in `packages/grpc/src/analyze-project-normalize.ts` / `packages/grpc/src/analyze-project-handle-request.ts` and then calls the same `analyzeProject()` implementation.
- Same result application on the Java side: `WebSensor` still consumes per-file results plus final metadata, saves issues/highlights/metrics/CPD, caches JS/TS AST + CPD, and feeds AST consumers through `ESTreeFactory`.
- Same standalone parser feature contract: one request, one AST-bearing response, then `ESTreeFactory.from(...)`.
- Same high-level timeout shape: long-running streamed analysis is still not deadline-bound; only the unary parser/tooling path is bounded by `timeoutSeconds`.

**Intentional / Accepted Transport Differences**
- Direct proto payloads instead of JSON/base64: on master the AST crossed HTTP as a base64-encoded protobuf string inside JSON (`BridgeServer.AnalysisResponseDTO.ast` -> `AstProtoUtils.parseProtobuf(...)`); now `ProjectAnalysisFileResult.ast` is a direct `estree.Node` field in `packages/grpc/src/proto/analyze-project.proto` and Java consumes it directly from generated protobuf messages.
- Rule configs now cross the wire as `google.protobuf.Value` / `Struct` (`AnalyzeProjectMessages.java`, `analyze-project.proto`) instead of ad hoc JSON. Existing heterogeneous config shapes are preserved, but the transport is typed.
- Request normalization is materially stricter: invalid `base_dir`, empty file keys, invalid enums, and unsafe `max_file_size` now fail fast instead of being silently defaulted/dropped.
- Worker IPC is now request-scoped (`requestId`) and the server rejects overlapping analyses with `RESOURCE_EXHAUSTED`. The old HTTP/WS path used shared listener arrays and had no equivalent transport-level single-flight guard.
- Streaming cancellation is stronger: Java watches `SensorContext`, cancels the streaming RPC, and Node maps `call.cancelled` to `requestCancel()`. That is a real reliability improvement over the old WebSocket-side cancel path.
- Ownership is explicit: heartbeat/inactivity timeout has been replaced by the `Lease` RPC. This is a real behavior change, but it matches the �owned Node dies when owning Java dies� model more directly.
- Worker-side protobuf conversion (`packages/grpc/src/analyze-project-worker.ts`) removes the old JSON/base64 boundary and reduces how much raw analysis object data crosses the worker-parent boundary.

**Non-Equivalences / Residual Risk**
- `isAlive()` is not the same probe anymore. Master did a real `/status` round-trip. Current code only checks gRPC channel readiness plus lease state. That is cheaper and consistent with the ownership model, but it is not an application-level health check.
- Message sizing is intentionally different. Master capped unary HTTP bodies at `50mb` and left WebSocket unlimited; current gRPC sets send/receive message limits to unlimited on both sides. That supports very large requests/responses, but it also removes the old transport backstop and is the clearest transport-level candidate for higher peak memory if large payloads are involved.
- Error surface changed. Master could emit JSON / in-band WebSocket error payloads; gRPC now reports failures as statuses (`INVALID_ARGUMENT`, `INTERNAL`, `RESOURCE_EXHAUSTED`, etc.). Cleaner overall, but less verbose than the old WS error frame. Java still collapses some non-special-cased statuses to the generic �bridge server is unresponsive� path.
- Wire-level path spelling changed for direct consumers: gRPC preserves the caller�s original file keys in unary and streamed responses; master HTTP returned normalized absolute paths internally.
- Unary cancellation is still not wired from call-cancel to `requestCancel()`. That only affects `AnalyzeProjectUnary` / standalone parser, not the production streamed `analyze-project` flow.
- Remaining semantic corner case: protobufjs map decoding still makes `files: {}` indistinguishable from omitted `files` when `canAccessFileSystem = true`, so an explicitly empty map still behaves like disk discovery.

**Performance / Reliability Readout**
- Code-level improvements that should help: no HTTP JSON serialization, no base64 AST wrapping, and worker-side protobuf conversion before crossing the worker boundary.
- Code-level risks that remain: unbounded gRPC message sizes can push peak memory to transport buffering + payload size, and `isAlive()` no longer proves that the runtime can actually answer work, only that the channel/lease still look alive.
- I did not find evidence that production streamed analysis can hang because of a new gRPC deadline; the only explicit deadline is on the unary parser path, and the only 15s timeout in the Node server is for worker coordination, not analysis runtime.

**Verification / Coverage**
- Bounded reruns on the current tree passed:
  - `mvn --% -pl sonar-plugin/bridge -am -Dtest=BridgeServerImplTest,AnalyzeProjectMessagesTest -Dsurefire.failIfNoSpecifiedTests=false -DfailIfNoTests=false test` -> 53/53 passed.
  - `node node_modules/tsx/dist/cli.mjs --tsconfig packages/tsconfig.test.json --test packages/grpc/tests/analyze-project-normalize.test.ts packages/grpc/tests/analyze-project-server.test.ts packages/grpc/tests/analyze-project-worker.test.ts` -> 45/45 passed.
- Coverage is stronger than the old HTTP transport around normalization, lease lifecycle, concurrency rejection, large-message handling, cancel RPC, and worker-exit handling.
- The main remaining coverage gaps I found are: no direct replacement for the old HTTP memory diagnostics tests, limited direct coverage of worker failure serialization / runtime -> `INTERNAL` Node-side mapping, and no end-to-end IT specifically asserting HTTP-vs-gRPC transport equivalence.

Bottom line: for the production analysis path this looks functionally equivalent enough to move from �migration� to �polishing�, with the real remaining gap still being the empty-vs-omitted `files` case. The notable accepted-by-design differences are lease-based ownership, transport-readiness `isAlive()`, status-based errors, and unlimited message sizing.

PR #6904, opened 2026-04-24
JS-1315 lookup for pnpm-workspaces.yaml

guillemsarda:
Part of task JS-1315.


vdiez:
Review findings from local worktree inspection:

1. [P1] Cache key ignores `topDir`

`dependenciesCache` and `moduleTypeCache` pass `topDir` as the optional `ComputedCache` context, but `ComputedCache` only keys on the first argument. That means the same manifest directory can return stale results when queried under a different root. I reproduced this with the existing `child-parent-merge` fixture: after caching the subdir under the repo root, a later lookup with `topDir=subdir` still leaked the parent dependency until the cache was cleared.

`packages/analysis/src/jsts/rules/helpers/dependency-manifests/dependencies.ts` (lines 46-78)

2. [P1] React version lookup still bypasses pnpm resolution

`getReactVersion()` still iterates over raw `package.json` files from `getPackageJsonManifests()` instead of using the resolved manifests produced by `getDependencyManifests()`. As a result, pnpm catalog values are still seen as `catalog:*` here and the React version remains unknown. On the branch fixture for pnpm catalog resolution, this path still returns `null`, so the Jira item’s explicit `getReactVersion()` use case is not actually covered.

`packages/analysis/src/jsts/rules/helpers/dependency-manifests/dependencies.ts` (lines 141-157)

3. [P1] pnpm workspace edits do not clear the dependency cache

The invalidation check compares each fs event entry against the bare string `pnpm-workspace.yaml`, but `fsEvents` are normalized to absolute paths before they reach this store. In practice the condition never matches, so editing `pnpm-workspace.yaml` leaves the dependency cache alive and can serve stale catalog resolutions in IDE mode.

`packages/analysis/src/file-stores/dependency-manifests.ts` (lines 73-75)

4. [P2] `workspace:` protocol support from the Jira is still missing

The new npm resolver only resolves `catalog:` references. The Jira task also calls out `workspace:*`, `workspace:^`, and `workspace:~`, but there is no corresponding implementation here, no fixture for it, and the current helper tests still treat those inputs as unresolved. That leaves a documented part of the ticket scope unimplemented.

`packages/analysis/src/jsts/rules/helpers/dependency-manifests/resolvers/npm.ts` (lines 96-107)

vdiez:
Re-reviewed the latest changes on the branch head.

- `getReactVersion()` now goes through `getDependencyManifests()`, so the pnpm `catalog:` case is covered.
- `pnpm-workspace.yaml` edits now invalidate the dependency cache correctly.

Two notes remain:

1. `workspace:` versions are still unresolved. The resolver now imports `packages` from `pnpm-workspace.yaml` into `workspaces`, but it still only rewrites `catalog:` values. `workspace:*`, `workspace:^`, and `workspace:~` are still left as-is, so version-sensitive paths like `getReactVersion()` still return `null` for a common pnpm workspace setup. This still looks like missing Jira scope.

2. The `ComputedCache` / `topDir` mismatch is still there, but this is a corner case. It only matters when one project baseDir is nested inside another and both hit the same cached subtree; in normal non-overlapping project layouts it does not show up. If we consciously accept that tradeoff to maximize reuse of the precomputed caches and avoid extra FS reads, I think that part is acceptable.

So the only remaining actionable gap I still see against the Jira ticket is `workspace:` protocol resolution.

vdiez (approved):

PR #6910, opened 2026-04-24
JS-1560 Update existing testing rules to cover more libs

nathsou:
Part of [JS-1560](https://sonarsource.atlassian.net/browse/JS-1560)

[rspec PR](https://github.com/SonarSource/rspec/pull/6691)

## Summary

This PR expands existing JavaScript testing rules to cover more testing/assertion libraries and modern test-definition shapes. It does not introduce new rules; it updates S1607, S2187, S2699, and S6426, plus the shared helpers those rules rely on.

## Rule changes

- S1607 ("Tests should not be skipped without providing a reason"): extend skipped-test detection to Vitest, Cypress, Playwright, and Karma setups backed by `karma-jasmine` / `karma-mocha`. The rule now recognizes the skip aliases those ecosystems expose (`test.skip`, `suite.skip`, `specify.skip`, `x*`, `test.describe.skip`, etc.), so skipped tests are treated consistently across supported frameworks. Playwright runtime annotations such as `test.skip(condition, description)` are intentionally excluded because they are conditional runtime annotations rather than silently disabled test definitions. References: https://vitest.dev/api/ ; https://mochajs.org/#inclusive-tests ; https://docs.cypress.io/app/core-concepts/writing-and-organizing-tests ; https://playwright.dev/docs/api/class-test ; https://nodejs.org/docs/latest/api/test.html#skipping-tests
- S2187 ("Test files should contain at least one test case"): broaden test-case recognition so `.cy.*` files and builder/modifier chains are treated as real tests. The rule now relies on shared framework-aware call detection, which lets it recognize APIs such as `specify.only`, `xtest.each`, `it.only.each`, `test.skipIf`, and Playwright test definitions. This reduces false positives on non-empty test files and catches empty Cypress specs as intended. References: https://vitest.dev/api/ ; https://docs.cypress.io/app/core-concepts/writing-and-organizing-tests ; https://playwright.dev/docs/api/class-test ; https://nodejs.org/api/test.html
- S2699 ("Tests should include assertions"): broaden the rule entry condition from imported assertion libraries only to framework/assertion signals that also cover Jest, Jasmine, Cypress, Playwright, and Node `assert` / `assert/strict`. Test-case extraction now goes through a shared helper instead of a Mocha-specific path, so global-assertion frameworks participate in the same analysis. While doing that, the implementation keeps counting helper-delegated assertions and `$httpBackend.expect*` expectations to avoid regressions. References: https://jestjs.io/docs/expect ; https://jasmine.github.io/api/edge/global ; https://playwright.dev/docs/test-assertions ; https://docs.cypress.io/app/references/assertions ; https://nodejs.org/api/assert.html ; https://sinonjs.org/releases/latest/assertions/ ; https://github.com/ladjs/supertest
- S6426 ("Exclusive tests should not be committed to version control"): extend `.only()` detection beyond the previous Mocha/Jest-style shapes so it also catches Vitest, Cypress, Playwright suites, and wrapped builders such as `it.only.each(...)`, `describe.only.each(...)`, and `test.describe.only(...)`. This keeps the rule aligned with the focused-test APIs developers actually use while preserving the existing quick fix that removes `.only`. References: https://vitest.dev/api/test.html#test-only ; https://docs.cypress.io/app/references/bundled-libraries#Mocha ; https://playwright.dev/docs/test-annotations#focus-a-test ; https://mochajs.org/#exclusive-tests ; https://jestjs.io/docs/api

## Validation

- Added new comment-based fixtures for S1607 covering Cypress, Vitest, Playwright, Karma + Jasmine, and Karma + Mocha.
- Expanded S2187 unit tests with `.cy` files, parameterized builders, conditional builders, and Playwright cases.
- Expanded S2699 unit tests with Jest, Jasmine, Playwright, Cypress, `assert/strict`, helper-function assertions, and `$httpBackend.expect*` cases.
- Extended S6426 quick-fix coverage for `context.only`, `suite.only`, `test.describe.only`, and `.only.each(...)`.
- Added helper tests for dependency and import detection.

[JS-1560]: https://sonarsource.atlassian.net/browse/JS-1560?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/analysis/src/jsts/rules/helpers/ast.ts
```diff
@@ -533,21 +527,28 @@ export function getProperty(
   }
 }
 
+// Resolves either a named function declaration or a variable with a single function-valued write.
+function resolveFromVariableReference(
```
vdiez: function name should clarify that this is specific for functions

packages/analysis/src/jsts/rules/helpers/ast.ts
```diff
@@ -165,18 +171,6 @@ export function isMethodInvocation(
...
-  functionName: string,
-  minArgs: number,
-): boolean {
-  return (
-    callExpression.callee.type === 'Identifier' &&
-    isIdentifier(callExpression.callee, functionName) &&
-    callExpression.arguments.length >= minArgs
-  );
-}
-
```
vdiez, line 179: this wasn't used? I'm surprised knip didn't detect it

packages/analysis/src/jsts/rules/helpers/test-frameworks.ts
```diff
@@ -14,59 +14,268 @@
...
+}
+
+export function hasTestFrameworkSignal(
+  context: Rule.RuleContext,
+  frameworks: TestFramework[],
+): boolean {
+  return (
+    hasTestFrameworkDependency(context, frameworks) || hasTestFrameworkImport(context, frameworks)
+  );
+}
```
vdiez: this is too expensive, each element in the array will traverse ast. I would rather find a way to detect `frameworks` is single pass

packages/analysis/src/jsts/rules/helpers/module.ts
```diff
@@ -56,6 +56,38 @@ export function getRequireCalls(context: Rule.RuleContext) {
...
+  const modules = new Set(moduleNames);
+
+  return (
+    getRequireCalls(context).some(
+      call =>
+        call.arguments[0].type === 'Literal' &&
+        typeof call.arguments[0].value === 'string' &&
+        modules.has(call.arguments[0].value),
+    ) ||
+    context.sourceCode.ast.body.some(
```
vdiez: `getRequireCalls` should already get all `require`'s,  why do we need to go through the body here. However I think we need a `getDynamicImportCalls` with a similar approach traversing the scopes

packages/analysis/src/jsts/rules/helpers/test-frameworks.ts
```diff
@@ -14,59 +14,268 @@
...
+      return objectPath ? [...objectPath, node.property.name] : null;
+    }
+    case 'CallExpression':
+      return getCallPath(node.callee);
+    case 'TaggedTemplateExpression':
+      return getCallPath(node.tag);
+    default:
+      return null;
+  }
+}
```
vdiez: this seems very related to the `getFullyQualifiedName` helper in module.ts. I would rather try to use that and complete if some cases are not covered there

vdiez (changes requested):

packages/analysis/src/jsts/rules/helpers/ast.ts
nathsou: This was code introduced by the first commit that wasn't used, so I removed it after the knip step failed.

packages/analysis/src/jsts/rules/helpers/module.ts
nathsou: This covers the case where a module is required without being affected to a variable like a raw `require('baz');`, not `const baz = require('baz');` which probably does not make sense as imports cannot be used so I will remove this, good catch!

packages/analysis/src/jsts/rules/helpers/test-frameworks.ts
nathsou: Indeed, I refactored to direct calls to `importsModule`  and `hasDependency` with the full list of dependencies.

packages/analysis/src/jsts/rules/helpers/module.ts
nathsou: I also added support for dynamic imports (only when assigned to a variable), in a single pass that collects `require` and `import` calls by traversing the scopes.

PR #6913, opened 2026-04-24
JS-1654 Align S6842 with upstream jsx-a11y recommended allowlist

francois-mora-sonarsource:
This follow-up aligns S6842 with the upstream `jsx-a11y` recommended configuration by reading the allowlist from the plugin instead of duplicating it locally.

- adds `upstream.ts` to extract and validate the `jsx-a11y/no-noninteractive-element-to-interactive-role` allowlist from `eslint-plugin-jsx-a11y`
- updates `config.ts` to derive S6842 default `fields` from that upstream allowlist
- expands the unit test to cover extraction failure, default-options wiring, every upstream `ul`/`ol`/`li` element-role pair, and representative allowed composite widgets plus invalid boundary cases
- updates the local `eslint-plugin-jsx-a11y` typings so `configs.recommended.rules` can be accessed safely
- removes one desktop ruling expectation for `vertical-segmented-control.tsx:215` because it is `ul[role="radiogroup"]`, which is allowed by the upstream preset

Verification:
- `./node_modules/.bin/tsx --tsconfig packages/tsconfig.test.json --test packages/analysis/src/jsts/rules/S6842/unit.test.ts`
- `npm run bbf`
- `npx knip`


francois-mora-sonarsource:
Following up on the remaining ruling question: in the `its/sources/projects` submodule commit used by this branch (`81a52e2c0180fc6e555b2c3efc44729fb46539d9`), `desktop/app/src/ui/lib/vertical-segmented-control/vertical-segmented-control.tsx:215` is `<ul role="radiogroup">`.

That matches the upstream jsx-a11y recommended allowlist now used by S6842 (`ul -> radiogroup`), so removing that expected issue is intentional. The sibling retained expectation at `desktop/app/src/ui/lib/vertical-segmented-control/segmented-item.tsx:71` is still `<li role="radio">`, so we are not suppressing the child issue.

erwan-leforestier-sonarsource (approved):
LGTM ✅ Thanks @francois-mora-sonarsource !

PR #6925, opened 2026-04-27
JS-1560 Update existing testing rules to cover more libs

nathsou:
Part of [JS-1560](https://sonarsource.atlassian.net/browse/JS-1560)

[rspec PR](https://github.com/SonarSource/rspec/pull/6691)

[JS-1560]: https://sonarsource.atlassian.net/browse/JS-1560?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

# Cover more testing/assertion libraries in test-related rules

Extends four test-related rules so they cover the full set of frameworks listed in their RSPEC.

## Rules updated

**[S1607](packages/analysis/src/jsts/rules/S1607/rule.ts) — Tests should not be skipped without a reason**
- Added Cypress, Playwright, and Vitest dispatch arms.
- Mocha/Cypress: handles `specify.skip` and the x-prefixed aliases (`xit`, `xdescribe`, `xcontext`, `xspecify`).
- Vitest: flags `it.skip`, `test.skip`, `describe.skip`, `suite.skip`.
- Playwright: flags declaration-style skips (`test.skip('title', ...)`, `test.describe.skip`, plus `parallel` / `serial` modifier chains) while ignoring runtime annotations like `test.skip(callback, description)`.
- Title argument now accepts template literals in addition to string literals.

**[S2187](packages/analysis/src/jsts/rules/S2187/rule.ts) — Test files should contain at least one test case**
- Test-file regex extended from `.test.|.spec.` to also match `.cy.`.
- Cypress (`specify`, `xspecify`) and Vitest (`skipIf`/`runIf`/`fails`/`sequential`/`for`) APIs added.
- New `APIS_WITH_TEST_TITLE` set distinguishes Playwright's overloaded `test.fail` / `test.fixme`: only the declaration form (string title) counts as a test, runtime annotations do not.

**[S2699](packages/analysis/src/jsts/rules/S2699/rule.ts) — Tests should include assertions**
- Detection now triggers for the full RSPEC list: chai, sinon, vitest, supertest, jasmine, jest, cypress, Playwright, and node.js `assert` / `node:assert`.
- Imported libraries: `chai`, `sinon`, `vitest`, `supertest`, `@playwright/test`, `assert`, `node:assert`.
- Global runners detected via `package.json`: `jasmine`, `jest`, `cypress`, `@playwright/test`.
- New [helpers/cypress.ts](packages/analysis/src/jsts/rules/helpers/cypress.ts) recognizes Cypress `.should()` / `.and()` chains rooted at the `cy` global (estree + TypeScript paths) so Cypress-only tests don't false-positive.

**[S6426](packages/analysis/src/jsts/rules/S6426/rule.ts) — Exclusive tests should not be committed**
- Added `context` and `specify` (Mocha/Cypress) plus Playwright `test.describe.only` with `parallel` / `serial` modifier chains.

## Helper refactor

[helpers/module.ts](packages/analysis/src/jsts/rules/helpers/module.ts):
- `getRequireCalls` → `getRequireAndDynamicImportCalls`, which also picks up `import('x')` and `await import('x')`.
- New `importsModule(context, names)` unifies require/import detection. The four assertion-library helpers (chai/sinon/vitest/supertest) collapse to a single one-liner.
- `reduceTo` handles `TaggedTemplateExpression` so FQN resolution covers tagged-template chains.

## Tests

Per-framework comment-based fixtures added for each new arm: S1607 (cypress/playwright/vitest), S2699 (jasmine/jest/cypress/playwright), S2187 fixture cases for the new APIs, and S6426 fixture extended with Cypress + Playwright cases.


packages/analysis/src/jsts/rules/helpers/module.ts
```diff
@@ -29,31 +29,117 @@ export function isESModule(context: Rule.RuleContext): boolean {
...
+
+  return (
+    getRequireAndDynamicImportCalls(context).some(module =>
+      moduleNames.includes(module.moduleName),
+    ) ||
+    getImportDeclarations(context).some(
+      declaration =>
+        typeof declaration.source.value === 'string' &&
+        moduleNames.includes(declaration.source.value),
+    )
```
vdiez: compute this part before in the OR, it's cheaper

packages/analysis/src/jsts/rules/S2187/rule.ts
```diff
@@ -17,18 +17,23 @@
 // https://sonarsource.github.io/rspec/#/rspec/S2187/javascript
 
 import type { Rule } from 'eslint';
-import type { Node } from 'estree';
+import type { CallExpression, Node } from 'estree';
 import { generateMeta } from '../helpers/generate-meta.js';
+import * as Playwright from '../helpers/playwright.js';
 import * as meta from './generated-meta.js';
 
+const TEST_FILE_PATTERN = /\.(?:spec|test|cy)\./;
```
vdiez, line 25: I would rather remove this filtering. We need to handle this globally https://sonarsource.atlassian.net/browse/JS-1501

packages/analysis/src/jsts/rules/S2699/rule.ts
```diff
@@ -69,6 +76,15 @@ export const rule: Rule.RuleModule = {
...
 
+function hasSupportedAssertionLibrary(context: Rule.RuleContext): boolean {
+  if (importsModule(context, ASSERTION_LIBRARIES)) {
+    return true;
+  }
+
+  const dependencies = getDependenciesSanitizePaths(context);
+  return GLOBAL_ASSERTION_DEPENDENCIES.some(dependency => dependencies.has(dependency));
+}
+
```
vdiez, line 87: very nice

vdiez (approved):
LGTM! just some small comments for you to evaluate :)

vdiez:
I still see two uncovered cases here:

- `S2699`: `packages/analysis/src/jsts/rules/S2699/rule.ts` still goes through `Mocha.extractTestCase()`, and `packages/analysis/src/jsts/rules/helpers/mocha.ts` still only extracts `it` / `specify`. So `test(...)` / `test.only(...)` cases are still skipped, which leaves the common Playwright/Jest/Vitest `test` style uncovered. The added Playwright fixture only uses `it(...)`, so this path is not exercised yet.

- `S2187`: `test.skip(...)` is still in the generic `APIs` set, so runtime/file-level annotations like `test.skip(browserName === 'webkit', 'reason')` still count as "this file contains a test". That suppresses the rule for otherwise-empty Playwright files that only contain a skip annotation, even though `test.fail(...)` / `test.fixme(...)` now have dedicated title-based filtering.

packages/analysis/src/jsts/rules/S2187/rule.ts
nathsou: As discussed, to be handled as part of JS-1501

PR #6935, opened 2026-04-29
JS-1677: Cover JS for rule S5914: No trivial test assertions

nathsou:
Part of [JS-1677](https://sonarsource.atlassian.net/browse/JS-1677)

Implements S5914 for JS.

Corresponding [rspec PR](https://github.com/SonarSource/rspec/pull/6705)

[JS-1677]: https://sonarsource.atlassian.net/browse/JS-1677?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

**Scope**

Detect trivially true/false assertions that are not testing anything interesting, this includes:

1. Assertions on constant expressions, including const variables initialized with constant values.
2. Strict equality checks on expressions which always produce a new reference.

**Out of scope**
* Do not require TypeScript type information.
* Do not handle type-dependent incompatible comparisons. Those belong to S5845

**Approach**

Add helper functions to recognise assertions in multiple popular testing/assertion libraries and represent them in a unified way, these will be useful in other rules which cover test assertion issues.


packages/analysis/src/jsts/rules/helpers/assertions.ts
```diff
@@ -0,0 +1,698 @@
...
+      return { predicate: 'truthy', negated: false };
+    case 'false':
+      return { predicate: 'falsy', negated: false };
+    case 'null':
+      return { predicate: 'null', negated: false };
+    case 'undefined':
+      return { predicate: 'undefined', negated: false };
+    default:
+      return null;
+  }
```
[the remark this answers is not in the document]
nathsou: addressed

packages/analysis/src/jsts/rules/helpers/assertions.ts
```diff
@@ -0,0 +1,709 @@
...
+  | 'isUndefined'
+  | 'isDefined'
+  | 'exists'
+  | 'notExists'
+  | 'equal'
+  | 'notEqual'
+  | 'strictEqual'
+  | 'notStrictEqual';
+
+export function extractTestAssertion(
```
vdiez, line 109: [P2] Missing Cypress `should`/`and` assertions

The new shared assertion dispatcher never routes Cypress chains through `helpers/cypress.ts`, so the rule only sees Cypress's global `expect`/`assert` forms and misses the dominant `cy.should(...)` / `.and(...)` assertions. That leaves cases like `cy.wrap(true).should('be.true')` unreported even though the companion RSPEC advertises Cypress coverage.

packages/analysis/src/jsts/rules/S5914/rule.ts
```diff
@@ -0,0 +1,201 @@
...
+
+export function isConstantPrimitiveValue(
+  context: Rule.RuleContext,
+  node: estree.Node,
+  visited?: Set<Scope.Variable>,
+): boolean {
+  // resolve `const` bindings (other than `undefined`) through their initializer so that
+  // identifiers bound to a constant primitive are treated as constants too. `visited` guards
+  // against mutually-recursive const references like `const a = b; const b = a;` (TDZ at runtime).
+  if (node.type === 'Identifier' && node.name !== 'undefined') {
```
vdiez, line 116: [P2] Shadowed `undefined` is treated as a constant

Any identifier named `undefined` bypasses scope resolution and is treated as the primitive value, but JavaScript legally allows shadowing such as `function t(undefined) { ... }` or `const undefined = expr;`. Those cases will now be reported as trivial assertions even though the asserted value comes from runtime, which also contradicts the RSPEC note that parameters and other non-const bindings are not resolved.

packages/analysis/src/jsts/rules/S5914/rule.ts
```diff
@@ -0,0 +1,201 @@
...
+      if (isConstantPrimitiveValue(context, assertion.actual)) {
+        return { reportNode: assertion.actual, messageId: 'issue' };
+      }
+      if (isFreshReferenceExpression(assertion.actual)) {
+        return { reportNode: assertion.actual, messageId: 'freshPredicate' };
+      }
+      return null;
+    case 'comparison':
+      if (assertion.comparison === 'identity') {
+        // a freshly-created value on either side makes the identity check trivially fail (or trivially succeed when negated)
```
vdiez, line 80: [P3] Fresh-reference issues highlight the wrong node

The companion RSPEC says fresh-reference identity assertions should highlight the assertion call or matcher, but this logic reports the freshly created expression itself (`{}`, `[]`, `new Foo()`, etc.). The extractors already populate `reportNode`, so this looks accidental and makes the implementation diverge from the documented issue location.

vdiez (changes requested):
Requesting changes for three issues in the new S5914 implementation. Details are inline.

packages/analysis/src/jsts/rules/S5914/rule.ts
nathsou: Handled, though shadowing undefined should be handled by another rule as this is not a great practice.

packages/analysis/src/jsts/rules/S5914/rule.ts
nathsou: Updated rspec to specify the current behaviour.

packages/analysis/src/jsts/rules/helpers/assertions.ts
```diff
@@ -0,0 +1,794 @@
...
+    case 'null':
+      return { predicate: 'null', negated };
+    case 'undefined':
+      return { predicate: 'undefined', negated };
+    case 'exist':
+      return { predicate: 'defined', negated };
+    default:
+      return null;
+  }
+}
```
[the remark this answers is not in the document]
nathsou: Cypress does not define `exists`: https://docs.cypress.io/app/references/assertions

packages/analysis/src/jsts/rules/helpers/module.ts
```diff
@@ -36,53 +38,48 @@ export function getImportDeclarations(context: Rule.RuleContext): estree.ImportD
...
- * import('lodash').then(...); // not stored in variable
+ * Unsupported patterns (not stored in a variable):
+ * require('lodash');
+ * import('lodash').then(lodash => ...);
  */
-function getRequireAndDynamicImportCalls(context: Rule.RuleContext): RequireOrDynamicImport[] {
-  const imports: RequireOrDynamicImport[] = [];
-  const { scopeManager } = context.sourceCode;
-  for (const scope of scopeManager.scopes) {
+const importedModulesCache = new ComputedCache((sourceCode: Rule.RuleContext['sourceCode']) => {
```
vdiez: clear this cache after linter is done verifying, as discussed

vdiez (approved):
Rule looking good and ruling changes make sense. Just a correction needed in the cache

PR #6937, opened 2026-04-29
JS-1682 Fix S6324 false negatives in printable ranges

francois-mora-sonarsource:
## Summary
- restrict the S6324 character-class exemption to ranges that start in the control-character block
- add regression tests for standalone control characters mixed with printable ranges like `a-z`, `A-Z`, and `0-9`
- preserve the existing compliant YAML, Jest, and asciify-style patterns already covered by the S6324 suite

## Root cause
`isInCharacterClassWithRanges` treated any character class containing any `CharacterClassRange` as intentional character-set construction. That incorrectly exempted standalone control characters when the only range in the class was printable.

## Impact
S6324 now reports standalone control characters again when they are mixed with printable ranges, while still allowing intentional control-character range patterns.

## Validation
- `npx tsx --tsconfig packages/tsconfig.test.json --test --test-reporter=spec --test-reporter-destination stdout packages/analysis/src/jsts/rules/S6324/unit.test.ts`
- `npx prettier --check packages/analysis/src/jsts/rules/S6324/rule.ts packages/analysis/src/jsts/rules/S6324/unit.test.ts`

### DIFF
--- a/packages/analysis/src/jsts/rules/S6324/rule.ts
+++ b/packages/analysis/src/jsts/rules/S6324/rule.ts
@@ -40,15 +40,17 @@ function isCharacterClassRangeBoundary(character: AST.Character): boolean {
 }
 
 /**
- * Control characters inside character classes that contain ranges indicate intentional
- * character set construction (e.g., [\x00-\x08\x0b\x0c] for YAML/CSS non-printables).
+ * Standalone control characters are exempt only when the character class contains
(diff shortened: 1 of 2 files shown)


vdiez (approved):
Approved.

The S6324 exemption narrowing matches the false-negative fix, and the targeted unit test and formatting check passed.

PR #6938, opened 2026-04-29
JS-1680 Prevent S7780 quickfix from introducing S4624 issues

francois-mora-sonarsource:
## Summary

- decorate `S7780` so the upstream `String.raw` quick fix is dropped when it would create a nested template literal
- keep the quick fix for safe `S7780` cases
- add regression coverage for both the safe and unsafe quick-fix paths

## Verification

- `./node_modules/.bin/tsx --tsconfig packages/tsconfig.test.json --test packages/analysis/tests/jsts/linter/quickfixes/transform.test.ts`
- `./node_modules/.bin/tsx --tsconfig packages/tsconfig.test.json --test packages/analysis/tests/jsts/linter/quickfixes/transform.test.ts packages/analysis/tests/jsts/linter/issues/decode.test.ts packages/analysis/tests/jsts/linter/issues/message.test.ts packages/analysis/tests/jsts/linter/issues/transform.test.ts`

## Notes

- The focused tests passed before moving the identical diff off `master`.
- A rerun from the dedicated worktree is currently blocked because that worktree's shared `node_modules` does not resolve the test dependencies yet.


vdiez (approved):
LGTM!

PR #6940, opened 2026-04-30
JS-1674 SonarJS expose logic to map Protobuf AST representation to Java ESTree

quentin-jaquier-sonarsource:


quentin-jaquier-sonarsource:
In the end, we should be able to rely on the standalone parser, that already does this logic. So this step is not required.

PR #6951, opened 2026-04-30
JS-1685 [codex] Fix S6759 readonly props inherited from Readonly base

francois-mora-sonarsource:
## What changed
- added an `S6759` regression test for props interfaces extending a `Readonly<...>` alias
- updated the rule to recover readonly information from inherited mapped base types when `ts-api-utils` loses it on transient symbols

## Why
`S6759` incorrectly raised on props that were already effectively readonly when an interface inherited members from a `Readonly<...>` alias.

## Root cause
The rule relied on `ts-api-utils.isPropertyReadonlyInType(...)`. That helper handles direct mapped `Readonly<T>` types, but for inherited interface members it can see only a transient property symbol whose declaration no longer carries the mapped readonly modifier. As a result, inherited props such as `className` were treated as mutable even though the base type came from `Readonly<...>`.

## Impact
This removes a false positive for TypeScript React components using readonly utility types through interface inheritance, without changing the behavior for explicit mutable overrides.

## Validation
- `npm run bbf`
- `git submodule update --init --recursive`
- `npm run ruling`
- `node --import tsx --test --test-reporter=spec packages/analysis/src/jsts/rules/S6759/unit.test.ts packages/analysis/src/jsts/rules/S6759/cb.test.ts`

### DIFF
--- a/packages/analysis/src/jsts/rules/S6759/rule.ts
+++ b/packages/analysis/src/jsts/rules/S6759/rule.ts
@@ -173,6 +173,51 @@ function isReadOnly(props: Node, services: RequiredParserServices) {
 
   /* All properties must be read-only */
   return properties.every(property =>
-    isPropertyReadonlyInType(type, property.getEscapedName(), checker),
+    isPropertyReadonlyIncludingBaseTypes(type, property, checker),
   );
+}
+
+function isPropertyReadonlyIncludingBaseTypes(
+  type: ts.Type,
+  property: ts.Symbol,
+  checker: ts.TypeChecker,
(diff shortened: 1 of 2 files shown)


vdiez (approved):
Note: there is a shorter alternative here if we are willing to rely on TypeScript internals.

The resolved inherited property symbol already carries the readonly information via internal check flags (`getCheckFlags(symbol) & CheckFlags.Readonly`), so this could be implemented as a symbol-level fallback instead of recursively walking base types.

I think the current version is still the better tradeoff for this rule, because those checker internals are runtime exports but not part of the supported typed API.

PR #6956, opened 2026-05-01
JS-1688 extend dependencies cache with versions

guillemsarda:
Part of the context awareness epic. Relates to Jira task 1688.

### DIFF
--- a/packages/analysis/src/jsts/rules/helpers/dependency-manifests/dependencies.ts
+++ b/packages/analysis/src/jsts/rules/helpers/dependency-manifests/dependencies.ts
@@ -82,7 +82,7 @@ export const moduleTypeCache = new ComputedCache(
  *
  * @param dir dirname of the context.filename
  * @param topDir working dir, will search up to that root
- * @returns Set with the dependency names
+ * @returns Map of dependency names to their version strings
  */
 export function getDependencies(
   dir: NormalizedAbsolutePath,
--- a/packages/analysis/src/jsts/rules/helpers/dependency-manifests/dependencies.ts
+++ b/packages/analysis/src/jsts/rules/helpers/dependency-manifests/dependencies.ts
@@ -104,10 +104,7 @@ export function getDependencies(
  * take precedence. Otherwise, package.json#type from the closest manifest only
  * is used. If that closest manifest exists but omits "type", default to CommonJS.
  */
-export function getModuleType(
-  filePath: NormalizedAbsolutePath,
-  topDir: NormalizedAbsolutePath,
-): ModuleType | undefined {
+export function getModuleType(filePath: NormalizedAbsolutePath, topDir: NormalizedAbsolutePath) {
   const extensionSignal = MODULE_TYPE_BY_EXTENSION[extname(filePath).toLowerCase()];
   if (extensionSignal) {
     return extensionSignal;
--- a/packages/analysis/src/jsts/rules/helpers/dependency-manifests/dependencies.ts
+++ b/packages/analysis/src/jsts/rules/helpers/dependency-manifests/dependencies.ts
@@ -16,7 +16,6 @@
  */
 import type { Rule } from 'eslint';
 import { ComputedCache } from '../cache.js';
-import type { Minimatch } from 'minimatch';
 import fs from 'node:fs';
 import { extname } from 'node:path/posix';
 import { minVersion } from 'semver';
@@ -27,11 +26,11 @@ import {
   dirnamePath,
   stripBOM,
 } from '../files.js';
-import { getDependenciesFromManifest } from './parse.js';
 import { getClosestDependencyManifestDir } from './closest.js';
-import { getDependencyManifests } from './all-in-parent-dirs.js';
+import { getDependencyManifests, getPackageJsonManifests } from './all-in-parent-dirs.js';
+import { DependenciesList, type ModuleType } from './resolvers/types.js';
 
-export type ModuleType = 'module' | 'commonjs';
+export type { ModuleType } from './resolvers/types.js';
 
 const MODULE_TYPE_BY_EXTENSION: Readonly<Record<string, ModuleType>> = {
   '.mjs': 'module',
@@ -46,22 +45,20 @@ const MODULE_TYPE_BY_EXTENSION: Readonly<Record<string, ModuleType>> = {
 export const dependenciesCache = new ComputedCache(
   (dir: NormalizedAbsolutePath, topDir?: NormalizedAbsolutePath) => {
     const closestDependencyManifestDir = getClosestDependencyManifestDir(dir, topDir);
-    const result = new Set<string | Minimatch>();
+    const result: DependenciesList = new Map();
 
     if (!closestDependencyManifestDir) {
       return result;
     }
 
     for (const manifest of getDependencyManifests(closestDependencyManifestDir, topDir, fs)) {
-      const manifestDependencies = getDependenciesFromManifest(manifest);
-      for (const dependency of manifestDependencies) {
-        result.add(dependency.name);
-        if (dependency.alias) {
-          // Also add the alias as a dependency, so it can be resolved in rules, for instance S4328
-          result.add(dependency.alias);
+      for (const [name, version] of manifest.getDependencies()) {
+        if (!result.has(name)) {
+          result.set(name, version);
         }
       }
     }
+
     return result;
   },
 );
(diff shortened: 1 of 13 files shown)


packages/analysis/src/jsts/rules/helpers/dependency-manifests/resolvers/deno.ts
```diff
@@ -15,25 +15,80 @@
...
     } else if (denoJson) {
-      return [{ type: 'deno', manifest: parseDenoManifest(denoJson) ?? {} }];
+      return [
+        {
+          type: 'deno',
+          getDependencies: buildDependencies(parseDenoManifest(denoJson) ?? {}),
+          getModuleType: () => denoModuleType,
+        },
+      ];
     }
```
zglicz: as you are already touching this, refactor this the opposite to show precedence:
```
  if (denoJson) {
    ...
  } else if (denoJsonc) {
    ...
  }
```
Or better yet, seems like it can be rewritten in:
```
  const effectiveDenoJson = denoJson ?? denoJsonc
  if (effectiveDenoJson) {
    ...
  }
```

packages/analysis/src/jsts/rules/helpers/dependency-manifests/resolvers/npm.ts
```diff
@@ -16,11 +16,17 @@
  */
 import type { PackageJson } from 'type-fest';
 import yaml from 'yaml';
-import { type ManifestResolver, type DependencyManifest } from './types.js';
+import {
+  type ManifestResolver,
+  type DependencyManifest,
+  type ModuleType,
+  DependenciesList,
```
zglicz, line 23: this is a type, no? It can be in front of the {}

packages/analysis/src/jsts/rules/helpers/dependency-manifests/resolvers/npm.ts
```diff
@@ -47,10 +53,52 @@ export const npmManifestResolver: ManifestResolver = {
...
+        Array.isArray(packageJson[field])
+          ? packageJson[field]
+          : (packageJson[field]?.packages ?? []),
+      );
+      continue;
+    }
+    addDependencies(dependencies, packageJson[field] as PackageJson.Dependency);
+  }
+
+  return () => dependencies;
```
zglicz: I must've missed it, but why are we making this return a function instead of the computed dependencies?

packages/analysis/src/jsts/rules/helpers/dependency-manifests/dependencies.ts
```diff
@@ -27,11 +26,11 @@ import {
   dirnamePath,
   stripBOM,
 } from '../files.js';
-import { getDependenciesFromManifest } from './parse.js';
 import { getClosestDependencyManifestDir } from './closest.js';
-import { getDependencyManifests } from './all-in-parent-dirs.js';
+import { getDependencyManifests, getPackageJsonManifests } from './all-in-parent-dirs.js';
+import { DependenciesList, type ModuleType } from './resolvers/types.js';
```
zglicz: type outside. Also, seems like we really need to include this rule - https://typescript-eslint.io/rules/consistent-type-imports

packages/analysis/src/jsts/rules/helpers/dependency-manifests/dependencies.ts
```diff
@@ -27,11 +26,11 @@ import {
...
 
-export type ModuleType = 'module' | 'commonjs';
+export type { ModuleType } from './resolvers/types.js';
```
zglicz: do you need to re-export this? It's not being re-named, whats the value?

packages/analysis/src/jsts/rules/helpers/dependency-manifests/dependencies.ts
```diff
@@ -102,12 +87,12 @@ export const moduleTypeCache = new ComputedCache(
 export function getDependencies(
   dir: NormalizedAbsolutePath,
   topDir: NormalizedAbsolutePath,
-): Set<string | Minimatch> {
+): DependenciesList {
```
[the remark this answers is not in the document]
zglicz: address this

packages/analysis/src/jsts/rules/helpers/dependency-manifests/dependencies.ts
```diff
@@ -117,15 +102,18 @@ export function getDependencies(
  * take precedence. Otherwise, package.json#type from the closest manifest only
  * is used. If that closest manifest exists but omits "type", default to CommonJS.
  */
-export function getModuleType(filePath: NormalizedAbsolutePath, topDir: NormalizedAbsolutePath) {
+export function getModuleType(
+  filePath: NormalizedAbsolutePath,
+  topDir: NormalizedAbsolutePath,
+): ModuleType | undefined {
```
zglicz: why add return types? These are derived and less verbose.

zglicz (commented):
As this is a wip, here are some preliminary comments.

packages/analysis/src/jsts/rules/helpers/dependency-manifests/dependencies.ts
```diff
@@ -140,25 +128,13 @@ export function getDependenciesSanitizePaths(context: Rule.RuleContext): Set<str
...
-    const reactVersion = packageJson.dependencies?.react ?? packageJson.devDependencies?.react;
-    if (reactVersion) {
-      const parsed = parseReactVersion(reactVersion);
-      if (parsed) {
-        return parsed;
-      }
-      // Continue searching in parent package.json files if parsing fails
-    }
+  const dependencies = getDependencies(dir, normalizeToAbsolutePath(context.cwd));
+  const reactVersion = dependencies.get('react');
```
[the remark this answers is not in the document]
guillemsarda: We agree with this assumption, `deno.json` should take precedence against a same-level `package.json`.

packages/analysis/src/jsts/rules/helpers/dependency-manifests/resolvers/npm.ts
```diff
@@ -47,10 +53,52 @@ export const npmManifestResolver: ManifestResolver = {
       manifest = injectWorkspacePackages(manifest, parsedPnpmWorkspace);
       manifest = resolveCatalogReferences(manifest, parsedPnpmWorkspace);
     }
-    return [{ type: 'npm', manifest }];
+    const moduleType: ModuleType = manifest.type === 'module' ? 'module' : 'commonjs';
```
zglicz, line 56: Just to be sure, I'll mention how module vs commonjs are determined:
1. if .mjs -> module
2. if .cjs -> commonjs
3. determine based on closest package.json
    - if type == "moduje" -> module
    - if missing or module == commonjs => commonjs

Not sure if this comment is relevant here or not :)

packages/analysis/src/jsts/rules/helpers/dependency-manifests/dependencies.ts
```diff
@@ -140,25 +130,13 @@ export function getDependenciesSanitizePaths(context: Rule.RuleContext): Set<str
...
+  // Deno npm: imports reflect the actual runtime version and are intentionally included here, unlike the TypeScript/Node.js version signals
+  if (!reactVersion) {
+    return null;
   }
-  return null;
+  return parseReactVersion(reactVersion);
 }
```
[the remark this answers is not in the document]
zglicz: As we discussed offline, seems like an odd edge-case that really shouldn't occur.

packages/analysis/src/jsts/rules/helpers/dependency-manifests/parse.ts
```diff
@@ -154,23 +49,22 @@ type AddDependencyParameters = {
...
+        ? dependency.substring(DEFINITELY_TYPED.length)
         : dependency,
       version,
-      alias,
-    });
+    );
+  }
+  if (alias) {
+    // Also add the alias as a dependency, so it can be resolved in rules, for instance S4328
+    result.set(alias, version);
```
zglicz, line 68: Did you create some ticket about handling aliases separately from dependencies? Keeping it as a dependency is helpful to detect if we have unimported dependencies, but if we want to have a real list of dependencies, it isn't helpful and can be used to "trick" the analyzer to think that an alias is a dependency for whatever reason.

zglicz (approved):
Image

packages/analysis/src/jsts/rules/helpers/dependency-manifests/parse.ts
guillemsarda: Yes, I have. [Here](https://sonarsource.atlassian.net/browse/JS-1713) is the task.

PR #6992, opened 2026-05-08
JS-1649 Enable rules dynamically based on per-file import declarations

nathsou:
Part of [JS-1649](https://sonarsource.atlassian.net/browse/JS-1649)

[JS-1649]: https://sonarsource.atlassian.net/browse/JS-1649?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/analysis/src/jsts/linter/linter.ts
```diff
@@ -265,34 +279,75 @@ export class Linter {
...
+    if (dependencySensitiveRules.length === 0) {
+      // Note: denoImportCounts telemetry is not recorded for files with no dependency-sensitive rules,
+      // since mergedInlineNpmDependencies (where counting happens) is only called below.
+      return baseRules;
+    }
+
+    // if there are rules with required dependencies, we need to check if the dependencies are present in the file imports before enabling them
+    const context: RuleFilterContext = {
+      ...baseContext,
+      dependencies: mergedInlineNpmDependencies(manifestDependencies, sourceCode),
```
vdiez: This currently enables React-only rules from inline `npm:` imports during rule selection, but several of those rules still resolve dependency and version data from manifests only via `getReactVersion()` / `getDependenciesSanitizePaths()`. That leaves activation and execution out of sync.

Concrete examples:
- `S6774` calls `getReactVersion(context)` and treats a missing version as React 19+, so an inline-only import such as `npm:react@18` can suppress issues that should still be reported.
- `S6957` falls back to the linter-wide `settings.react.version = 999.999.999`, which can make an inline-only `npm:react@18` file look like a much newer React version and overreport deprecations.

I think we need the current file's inline dependencies to be available from the dependency helpers too, not only from `getRulesForFile()`. A reasonable shape would be a small singleton in the dependency helpers that stores the current file-specific dependencies for the file being linted, and then have `getDependenciesSanitizePaths()` / `getReactVersion()` merge that singleton with the shared manifest cache.

packages/analysis/src/telemetry.ts
```diff
@@ -93,6 +93,7 @@ export type ProjectAnalysisTelemetry = {
   programCreation: ProgramCreationTelemetry;
   esmFileCount: number;
   cjsFileCount: number;
+  denoImportCounts: Record<string, number>;
```
vdiez, line 96: `denoImportCounts` is added to the analysis-side telemetry object, but it never leaves the analysis package: the gRPC contract still ends at `cjs_file_count`, and `toTelemetry()` does not map the new field.

That means the analysis process records the counts locally, but every gRPC caller silently receives telemetry without them. Please either wire this through `packages/grpc/src/proto/analyze-project.proto` and `packages/grpc/src/analyze-project-convert.ts`, or drop the field from this PR until the transport is updated.

vdiez (approved):
very nice!

PR #6994, opened 2026-05-08
JS-1721 Materialize upstream defaults during linter rule loading

zglicz:
## Summary
This PR makes wrapped and external rule options consistent across the two runtime paths that were still diverging in JS-1721:
- quality-profile rule loading in `linter.ts`
- inline `eslint` pragma remapping in `pragmas.ts`

The core change is that rule options are now materialized in one place with the same precedence everywhere:
`upstream meta.defaultOptions -> Sonar field defaults -> explicit rule configuration`.

## What changed
- `materializeRuleOptions()` in `configs.ts` now builds the effective option set for normal rule loading.
- `mergeRuleOptions()` keeps object merging but replaces array values, so inline overrides do not inherit stale array entries.
- `createRulesRecord()` in `linter.ts` now uses the materialized option set for every rule, including wrapped rules without Sonar `fields`.
- `pragmas.ts` now remaps inline configs against the already-active Sonar rule options, so severity-only pragmas keep Sonar defaults and inline array overrides replace instead of merge.

## Regressions covered
- Normal loading: `S106`, `S108`, `S1186`, `S2430`, `S6325`, `S6535`, `S905`, `S878`
- Inline pragmas: severity-only default preservation for `S106`, array override replacement for `S2068`, plus inline coverage for `S108`, `S1186`, `S2430`, and `S905`

## Validation
- `TMPDIR=/private/tmp npx tsx --tsconfig packages/tsconfig.test.json --test packages/analysis/tests/jsts/linter/index.test.ts`

### DIFF
--- a/packages/analysis/src/jsts/linter/linter.ts
+++ b/packages/analysis/src/jsts/linter/linter.ts
@@ -228,7 +228,7 @@ export class Linter {
       files: [`**/*${path.posix.extname(normalizePath(filePath))}`],
     };
 
-    const messages = Linter.linter.verify(sourceCode, config, createOptions(filePath));
+    const messages = Linter.linter.verify(sourceCode, config, createOptions(filePath, rules));
     clearFileCaches();
     return transformMessages(messages, language, {
       sourceCode,
--- a/packages/analysis/src/jsts/linter/linter.ts
+++ b/packages/analysis/src/jsts/linter/linter.ts
@@ -35,9 +35,8 @@ import { APIError } from '../../contracts/error.js';
 import { pathToFileURL } from 'node:url';
 import * as ruleMetas from '../rules/metas.js';
 import { extname } from 'node:path/posix';
-import { defaultOptions, applyTransformations } from '../rules/helpers/configs.js';
+import { materializeRuleOptions } from '../rules/helpers/configs.js';
 import type { SonarMeta } from '../rules/helpers/generate-meta.js';
-import merge from 'lodash.merge';
 import {
   getDependencies,
   getModuleType,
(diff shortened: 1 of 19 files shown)


vdiez:
rebase?

zglicz:
Rebased on fresh `master` and force-pushed.

There was overlap with `#6911`, so I dropped the already-merged source changes from this branch during the rebase:
- `packages/analysis/src/jsts/linter/pragmas.ts`
- `packages/analysis/src/jsts/rules/helpers/validate-version.ts`

That means `#6911` still owns the inline ESLint pragma/default propagation hardening and the malformed `package.json#engines.node` hardening for `S6661`.

What remains in this PR is the separate normal rule-loading fix in `packages/analysis/src/jsts/linter/linter.ts`: when we build the quality-profile rule record, we now materialize upstream `meta.defaultOptions` before applying Sonar defaults and explicit parameters.

That remaining fix affects these JS-1721 buckets:
- `S2430` (`reading reduce`)
- `S6325` (`disallowRedundantWrapping`)
- `S6535` (`allowRegexCharacters`)
- and the same default propagation path for `S106`, `S108`, `S1186`, `S905`, plus `S878` coverage in tests.

I kept the JS-1721 regression coverage for the already-on-master `#6911` cases as well, so the ticket coverage stays explicit in one PR.

vdiez (changes requested):
Requesting changes because the PR fixes the profile-driven rule-loading path but leaves the inline `eslint` pragma path inconsistent.

1. In `packages/analysis/src/jsts/linter/pragmas.ts`, `normalizeInlineRuleOptions()` still rebuilds severity-only pragmas from the alias rule's `meta.defaultOptions` instead of the Sonar-materialized options. Repro: S106 is clean with the profile/default path for `console.info()`, but `/* eslint no-console: "error" */` still reports it because the Sonar `allow` list is dropped.

2. The same helper still uses `lodash.merge`, so array-valued inline overrides are merged instead of replaced. Repro: S2068 with profile config `{ passwordWords: [] }` suppresses the issue, but `/* eslint S2068: ["error", { "passwordWords": [] }] */` still reports because the default password-word list survives the merge.

The new tests cover scalar/boolean inline options, but not these two pragma cases.

vdiez (approved):

PR #6997, opened 2026-05-08
JS-1717 Cache parsed dependency manifests and preload pnpm-workspace.yaml

guillemsarda:
Part of [JS-733](https://sonarsource.atlassian.net/browse/JS-733) epic

[JS-733]: https://sonarsource.atlassian.net/browse/JS-733?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[JS-733]: https://sonarsource.atlassian.net/browse/JS-733?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/analysis/src/file-stores/dependency-manifests.ts
+++ b/packages/analysis/src/file-stores/dependency-manifests.ts
@@ -25,28 +25,30 @@ import {
 import type { Configuration } from '../common/configuration.js';
 import {
   clearDependenciesCache,
-  DENO_JSON,
-  DENO_JSONC,
-  getDependencyManifestName,
-  isDependencyManifestPath,
+  getPreloadableDependencyManifestName,
   PACKAGE_JSON,
   fillManifestCaches,
-  PNPM_WORKSPACE_YAML,
+  PRELOADABLE_DEPENDENCY_MANIFESTS,
+  type PreloadableDependencyManifestName,
+  isPreloadableDependencyManifestPath,
 } from '../jsts/rules/helpers/dependency-manifests/index.js';
-import { basename } from 'node:path/posix';
 
 export const UNINITIALIZED_ERROR =
   'dependency manifest cache has not been initialized. Call loadFiles() first.';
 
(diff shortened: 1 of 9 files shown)


vdiez (approved):
## Review

Looks good to me. I checked this against JS-1717 and the PR covers the expected behavior:

- `pnpm-workspace.yaml` is now discovered and preloaded through the dependency manifest store path.
- Parsed npm, pnpm, and Deno manifests are cached on demand.
- Failed parses are cached as well.
- Dependency cache invalidation also clears the parsed-manifest caches.

Validation run:

- Focused dependency-manifest tests passed.
- Focused tsconfig tests passed.

## Future Improvement

Non-blocking cleanup idea: this PR already moves `pnpm-workspace.yaml` into upfront file-store discovery, which is the right direction. As a follow-up, we may want to make the terminology and ownership more explicit by separating **discoverable dependency files** from **standalone dependency manifests**.

The file store probably should not care about the semantic difference between `pnpm-workspace.yaml` and files like `package.json`, `deno.json`, or `deno.jsonc`. From the store perspective, they are all dependency-related files worth discovering, reading, and preloading into lookup caches. The distinction that `package.json` / Deno files produce manifests while `pnpm-workspace.yaml` is supporting metadata for npm resolution should live only in the dependency resolvers.

That would also improve readability by unifying how we handle discovery and lookup-cache preloading for all dependency-related files. For example, the shared discovery/preload layer could expose all discoverable dependency files uniformly, while resolver-specific code decides whether a file contributes a standalone manifest, enriches another manifest, or needs a parent-list lookup cache. The current implementation works, but making that split explicit could make future package-manager support easier to reason about.

PR #7004, opened 2026-05-11
JS-1723 Scope Node.js version signal to the relevant pkg in monorepos

nathsou:
Part of [JS-1723](https://sonarsource.atlassian.net/browse/JS-1723)

[JS-1723]: https://sonarsource.atlassian.net/browse/JS-1723?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


its/ruling/src/test/expected/fireact/javascript-S7781.json
nathsou, line 1: The flagged file is fireact/functions/index.js, which sits inside functions/ whose package.json declares "engines": { "node": "10" }. The root fireact/package.json has  
  no node signal.                                                                                                                                                          
                                                                                                                                                                           
  - Before: getNodeVersionSignal(baseDir) looked only at the root package.json → no signal → lib fell back to esnext. String.prototype.replaceAll (ES2021) was assumed     
  available, so S7781 suggested rewriting .replace(/…/g, …) as replaceAll.                                                                                                 
  - After: functions/index.js now correctly resolves the closest package, finds engines.node: "10" → lib: es2018. replaceAll is not in lib.es2018, so the rule no longer   
  fires on a file that runs on Node 10 where replaceAll doesn't exist.                                                                                                     
                                                                                                                                                                           
  That's a real false-positive removal driven by per-package signal resolution.

vdiez (approved):
﻿Approving.

I no longer consider the remaining point a blocker. Once analysis is attached to a configured `tsconfig` project, deriving a single fallback `lib` at the project boundary is consistent with TypeScript's own program model: `lib` is a compiler-option/program setting, not a per-file setting.

This PR fixes the monorepo cases that were clearly wrong for JS-1723:
- nested `tsconfig` projects no longer inherit a single root-level Node.js signal
- orphan files are grouped so different packages no longer share the same fallback program

The only limitation left is a repo-root `tsconfig` that spans several nested packages with different package-level Node.js signals. I think that should be treated as follow-up work rather than a reason to block this PR, because addressing it would require a different analysis granularity inside one configured TS project, not just a different signal lookup.

Local validation looked good on my side with `npm ci`, `npm run bbf`, and the focused analysis tests around the new scenarios.

vdiez:
Adding the follow-up note here so we keep the design boundary explicit.

I think the current grouping is needed only for synthetic orphan programs, not for programs created from tsconfig files.

Why the orphan grouping is needed:
- a TypeScript program still has a single `compilerOptions.lib`
- when we synthesize one fallback program for orphan files in a monorepo, files from different packages otherwise get flattened onto the same Node.js signal
- in the incremental SonarLint path this is even more important because the cache can reuse any program that already contains the file, so one overly broad orphan program can be replayed for another package
- grouping by resolved `lib` is the smallest useful split: coarser than per-file, less aggressive than per-package, but still enough to avoid cross-package signal leakage

Why I would not extend grouping to tsconfig-owned programs:
- once files are claimed by a tsconfig, they already belong to a configured TS project
- `lib` is a compiler-option/program setting, not a per-file setting
- keeping one tsconfig -> one program stays aligned with TypeScript's model and with the typechecker
- splitting a single tsconfig project into multiple grouped programs would be a different analysis model, not a small extension of this fix

That is also why I am happy with the current PR boundary:
- nested tsconfig projects now get package-scoped fallback lib resolution
- orphan files now get isolated fallback programs when needed
- tsconfig-created programs stay untouched

Future consideration:
- the remaining unsupported shape is a repo-root tsconfig spanning several nested packages with different package-level Node.js signals and no explicit `compilerOptions.lib`
- if we ever want to support that case, it should be separate design work. At minimum we should add a dedicated fixture, keep explicit tsconfig `lib` as highest priority, and decide explicitly whether we want to preserve "one tsconfig = one program" or introduce synthetic subprojects inside a configured TS project
- if we go down the synthetic-subproject route, the current grouping key (resolved `lib`) is probably not enough on its own anymore and caching/program identity would need another pass

PR #7008, opened 2026-05-11
JS-1727 Fix dependency manifest cache warmup ordering

vdiez:
Part of

## Summary

Fix dependency manifest cache warmup so the preload pass seeds parent caches from already-warmed parent entries instead of consulting the lazy child lookup while the cache is still being built.

This keeps warmup independent from directory registration order and avoids sync filesystem fallbacks during preload.

## Testing

- `npx tsx --tsconfig packages/tsconfig.test.json --test --test-reporter=spec --test-reporter-destination=stdout packages/analysis/tests/dependency-manifests.test.ts`
- `npm run bbf`

### DIFF
--- a/packages/analysis/src/jsts/rules/helpers/dependency-manifests/index.ts
+++ b/packages/analysis/src/jsts/rules/helpers/dependency-manifests/index.ts
@@ -60,20 +60,21 @@ export function fillManifestCaches(
       ? undefined
       : patternInParentsCache.get(manifestName).get(topDir);
 
-  // We depend on the order of the paths, from parent-to-child paths (guaranteed by the use of a Map in the package-json store)
-  for (const [dir, parent] of dirnameToParent) {
+  const sortedDirs = Array.from(dirnameToParent.entries()).sort(
+    ([leftDir], [rightDir]) =>
+      leftDir.split('/').length - rightDir.split('/').length || leftDir.localeCompare(rightDir),
+  );
+
+  for (const [dir, parent] of sortedDirs) {
     const currentManifest = manifests.get(dir);
-    closestCache.set(dir, currentManifest ?? (parent ? closestCache.get(parent) : undefined));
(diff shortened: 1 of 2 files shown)


packages/analysis/src/jsts/rules/helpers/dependency-manifests/index.ts
```diff
@@ -60,20 +60,21 @@ export function fillManifestCaches(
...
-    closestCache.set(dir, currentManifest ?? (parent ? closestCache.get(parent) : undefined));
+    const inheritedClosestManifest = parent ? closestCache.get(parent) : undefined;
+    closestCache.set(dir, currentManifest ?? inheritedClosestManifest);
     if (!manifestsInParentsCache) {
       continue;
     }
-    const manifestsInParents: File[] = [];
-    if (parent) {
-      // Read the current dir cache to preserve closest-first manifest precedence.
-      manifestsInParents.push(...manifestsInParentsCache.get(dir));
```
vdiez, line 73: this line was hitting disk during cache warm up, but we can fill manifestInParents deterministically from current information.

vdiez:
Note on the `push` -> `unshift` change in `fillManifestCaches`:

Before this patch, warmup was reading `manifestsInParentsCache.get(dir)`, so the returned list already contained the current directory first. Appending `currentManifest` with `push` only made sense because that lazy `get(dir)` was doing the real ordering work, and that was also the path that could fall through to sync filesystem access on a cache miss.

After the fix, warmup seeds from `manifestsInParentsCache.get(parent)`, which returns `[parent, grandparent, ...]`. To preserve the cache contract as closest-first, the current manifest has to be prepended, producing `[current, parent, grandparent, ...]`. Keeping `push` there would build `[parent, grandparent, ..., current]`, which breaks precedence.

The child-first regression test added in `dependency-manifests.test.ts` checks that ordering explicitly.

guillemsarda (approved):
Looks good to me!

PR #7009, opened 2026-05-11
JS-1728 Resolve root Bun catalogs in package.json

vdiez:
## Summary

Use the current `package.json` as a Bun catalog source before falling back to parent `package.json` files or `pnpm-workspace.yaml`.

This fixes the case where the workspace root defines Bun catalogs and also consumes its own `catalog:` dependencies.

## Testing

- `npm ci`
- `npx tsx --tsconfig packages/tsconfig.test.json --test --test-reporter=spec --test-reporter-destination=stdout packages/analysis/tests/dependency-manifests.test.ts`
- `npm run bbf`

### DIFF
--- a/packages/analysis/src/jsts/rules/helpers/dependency-manifests/resolvers/package-json.ts
+++ b/packages/analysis/src/jsts/rules/helpers/dependency-manifests/resolvers/package-json.ts
@@ -40,30 +40,27 @@ export const packageJsonManifestResolver: ManifestResolver = {
       return [];
     }
     let parsedPackageJson = parsePackageJson(packageJson) ?? {};
+    const pnpmWorkspaceFile = closestPatternCache
+      .get(PNPM_WORKSPACE_YAML, fileSystem)
+      .get(topDir)
+      .get(dir);
+    const parsedPnpmWorkspace = pnpmWorkspaceFile
+      ? parsePnpmWorkspace(pnpmWorkspaceFile)
+      : undefined;
 
-    let catalogSource: CatalogSource | undefined = undefined;
-    const closestParent = findClosestParentPackageJsonWithCatalogs(dir, topDir, fileSystem);
+    if (parsedPnpmWorkspace) {
+      parsedPackageJson = injectWorkspacePackages(parsedPackageJson, parsedPnpmWorkspace);
+    }
 
-    if (closestParent) {
-      // If the closest parent package.json has catalogs defined, we use it as the catalog source for resolving catalog references.
-      const workspaces = Array.isArray(closestParent.workspaces)
-        ? undefined
-        : closestParent.workspaces;
-      catalogSource = {
-        catalog: workspaces?.catalog ?? closestParent.catalog,
-        catalogs: workspaces?.catalogs ?? closestParent.catalogs,
-      };
-    } else {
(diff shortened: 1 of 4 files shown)


erwan-leforestier-sonarsource:
After trying manually with the following root `package.json`:
```json
{
  "name": "bun-monorepo",
  "private": true,
  "catalog": {
    "typescript": "^5.8.3",
    "bun-types": "latest"
  },
  "catalogs": {
    "testing": {
      "vitest": "^3.0.0",
      "@vitest/coverage-v8": "^3.0.0"
    }
  },
  "workspaces": {
    "packages": ["."]
  },
  "scripts": {
    "dev": "bun run --filter='@bun-monorepo/hello-world' dev",
    "build": "bun run --filter='*' build"
  },
  "dependencies": {
    "typescript": "catalog:"
  }
}

```

And running `bun install`, I got the following output:
```
  > bun install
bun install v1.3.13 (bf2e2cec)
error: ConnectionClosed downloading package manifest typescript
error: typescript@catalog: failed to resolve
```
Similar result when `catalog` and `catalogs` are defined under `workspaces`.

Therefore, having self-reference to catalogs inside the root `package.json` is not supported by bun.

Closing this PR now.

haydenbleasel:
This was closed on the basis that Bun doesn't support `catalog:` self-references in the root `package.json`, but I don't think that holds. The `bun install` failure quoted above is a network error (`ConnectionClosed downloading package manifest typescript`), not a catalog resolution error.

I ran the exact `package.json` from the comment above on Bun 1.4.0:

```
$ bun install
Resolved, downloaded and extracted [1]
Saved lockfile

+ typescript@5.9.3 (v7.0.2 available)

1 package installed [121.00ms]
```

And a more typical layout where the root defines the catalog and consumes it while workspaces also reference it:

```json
{
  "name": "repro",
  "workspaces": ["apps/*"],
  "devDependencies": { "hono": "catalog:", "vitest": "catalog:" },
  "catalog": { "hono": "4.13.7", "vitest": "5.0.0" }
}
```

```
$ bun install
Saved lockfile

+ hono@4.13.7
+ vitest@5.0.0

39 packages installed [620.00ms]
```

The resulting `bun.lock` records the root's `catalog:` entries and the `catalog` block. Bun's docs also state that `catalog` and `catalogs` work at the top level of `package.json`, and root self-references are a common pattern in Bun monorepos.

With `eslint-plugin-sonarjs` 4.2.0, `findClosestParentPackageJsonWithCatalogs` returns early when `dir === topDir`, so the root manifest's own `catalog:` references are never resolved and every lint run prints `Dependency "<name>" could not be resolved for catalog "default"` per root dep. Users are hitting this through oxlint (oxc-project/oxc#26346, haydenbleasel/ultracite#796). Could this be reopened? The change here looks like the right fix.

erwan-leforestier-sonarsource:
Hi @haydenbleasel,

Thank you for your comment. I've checked again and you're right, it was due to a network issue. Bun is indeed capable of resolving versions from self-referenced catalog.
I cannot reopen this PR since the source branch has been deleted so I've created a new copy of this PR: https://github.com/SonarSource/SonarJS/pull/7892


Thank you for helping us improving SonarJS.

Best Regards,
Erwan Le Forestier

PR #7013, opened 2026-05-11
JS-1736 Clarify /tests skill guidance for false-positive fixes

francois-mora-sonarsource:
## Summary

This updates the `/tests` skill for SonarJS to better reflect how false-positive fixes are represented in rule tests.

- add guidance for converting existing FP reproducers from `invalid` to failing `valid` cases
- clarify that autofix expectations such as `output` and `suggestions` should be removed during that conversion while keeping setup fields like `filename` and `options` when needed
- add decorator upstream sentinel guidance and explicit guidance to move suppressed cases from `invalid` to `valid` instead of deleting them

## Why

The SonarJS skill and the mirrored Nigel guidance had drifted.
SonarJS already had the richer explanation for `errors` objects used in quickfix assertions, while Nigel had accumulated useful guidance for FP reproducers and decorator sentinels.
This change merges those pieces so the SonarJS-side skill is the more complete reference again.

## Validation

- `git diff --check`
- no automated SonarJS test currently validates `.claude/skills/tests/SKILL.md`

### DIFF
--- a/.claude/skills/tests/SKILL.md
+++ b/.claude/skills/tests/SKILL.md
@@ -126,8 +126,8 @@ When `meta.ts` contains `implementation = 'decorated'`, add an upstream sentinel
        const ruleTester = new DefaultParserRuleTester();
        ruleTester.run('{ESLINT_ID}', upstreamRule, {
          valid: [],
-          { code: `...`, errors: 1, output: `...` }, // include output if the upstream rule has autofix; omit if the rule is not fixable
-          // one entry per distinct FP pattern
+         invalid: [
+           { code: `...`, errors: 1, output: `...` }, // include output if the upstream rule has autofix; omit if the rule is not fixable
            // one entry per distinct FP pattern
          ],
        });
--- a/.claude/skills/tests/SKILL.md
+++ b/.claude/skills/tests/SKILL.md
@@ -63,6 +63,30 @@ problematicCode();
 
 When fixing false positives, add test cases to the `valid` array - the false positive is code that should be treated as valid.
 
+### Converting FP Reproducers to Failing Tests (TDD Style)
+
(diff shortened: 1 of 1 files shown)


.claude/skills/tests/SKILL.md
```diff
@@ -80,3 +104,44 @@ This isolation technique helps identify exactly which test cases are failing whe
...
+3. Read `index.ts` to find the exact import pattern already in use:
+   - ESLint core rules: `import { getESLintCoreRule } from '../external/core.js'` then `getESLintCoreRule('{ESLINT_ID}')`
+   - Plugin rules: `import { rules } from '../external/{PLUGIN}.js'` then `rules['{ESLINT_ID}']`
+4. Add the same import to the test file and obtain a reference to the upstream rule the same way
+   - Never import directly from `eslint-plugin-*` packages or `eslint/use-at-your-own-risk` — always use the `external/` shims
+5. Add a comment above the sentinel block explaining its purpose, then the block itself:
+   ```typescript
+   // Sentinel: verify that the upstream ESLint rule still raises on the patterns our decorator fixes.
+   // If this test starts failing (i.e., the upstream rule no longer reports these patterns),
+   // it signals that the decorator can be safely removed.
```
erwan-leforestier-sonarsource, line 123: If I understand correctly, these comments will be added on each test every time we fix a FP. Is it really something we want ?

francois-mora-sonarsource: Yes, this is a behavior that has been requested for SonarJS to detect when an upstream rule changes and becomes better. The failing tests sends the signal that the decorator logic is now obsolete and can be removed without affecting the rule performance. This helps keep the analyzer implementation as lean as possible. The comments help make the difference between tests that are just "signals" from tests that are caused by bugs.

erwan-leforestier-sonarsource (approved):

PR #7033, opened 2026-05-15
JS-1499 Harden parser options and propagate ES/module signals (fixes JS-674)

erwan-leforestier-sonarsource:
### Summary
Split parser-option construction by parser (TypeScript / Babel / Vue) and feed the analyzer's detected ECMAScript year, module type, and JSX intent into parsing. As a result, [JS-674](https://sonarsource.atlassian.net/browse/JS-674) is fixed: TypeScript angle-bracket type assertions inside `.vue` files no longer break parsing.

### Changes
- Replace the single `buildParserOptions` with one builder per parser (TS / Babel / Vue).
- Feed the analyzer's detected ES year, module type, and JSX intent (from `tsconfig.compilerOptions.jsx`) into parser options.
- Vue+TS parse failures retry with the opposite JSX setting; the original error wins if both fail.

### Validation

Download `js-1499-validation.zip` (attached), unzip it, then run:

```bash
./validate.sh /path/to/SonarJS
```

- On `master`: prints `FAILED — Identifier expected.`
- On this branch: prints `OK — JS-674 reproducer parses cleanly. Fix verified.`

The companion `jsts-test-sources` PR adds the same scenario as a ruling-test fixture under `it-tools/src/components/`.

[js-1499-validation.zip](https://github.com/user-attachments/files/27790617/js-1499-validation.zip)

[JS-674]: https://sonarsource.atlassian.net/browse/JS-674?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vdiez (changes requested):
Requesting changes because the PR does fix the JS-674 Vue+TS ambiguity, but it still introduces a module-type regression on ordinary `.js` files.

`packages/analysis/src/jsts/rules/helpers/dependency-manifests/resolvers/package-json.ts` now returns `undefined` when `package.json#type` is absent, but `packages/analysis/src/jsts/parsers/options.ts` still falls back to `sourceType: 'module'`, and `packages/analysis/src/jsts/linter/filters/filter-module-type.ts` keeps module-gated rules enabled when the signal is unknown.

On the current head (`25fd3da6a6`), I reproduced this with a plain `file.js` inside a package whose `package.json` has no `type`:

```js
// package.json
{ "name": "tmp" }

// file.js
Promise.resolve(42).then(console.log);
```

That file is parsed as `sourceType: "module"` and reports `S7785`, even though Node treats that setup as CommonJS by default, so top-level `await` is not available there.

I think the fix needs to preserve CommonJS semantics for unspecified `.js` files, or at minimum avoid treating an unknown module signal as ESM in both parsing and rule activation.

erwan-leforestier-sonarsource:
Thanks again for the review. To address [your comment](https://github.com/SonarSource/SonarJS/pull/7033#pullrequestreview-4319551629) @vdiez and after digging into it more, I went the opposite way on the parser side.

  Using the detected module type as `sourceType` did fix the S7785 false positive, but it also caused **scope-aware rules to silently stop firing** on top-level declarations in CJS files. ESLint's
  `sourceType: 'commonjs'` puts top-level `const`/`let` into a `global`-typed scope, and many rules (S4158 and others) explicitly skip the global scope to avoid false positives on real browser/Node
  globals. So any top-level empty array / unused collection / etc. in a `.cjs` file or a file without a `package.json#type` would just stop being reported — no parse error, no warning, just a missed
  bug.

  The IT `TypeScriptRuleTest` actually caught this: 6 issues in `S4158.ts` disappeared because the file lives in a project without `"type": "module"` in its `package.json`.

  So I reverted the `sourceType` coupling and kept `sourceType: 'module'` for parsing. Rationale:
  - `sourceType` controls both parsing **and** scope-tree construction. The `'module'` scope model is more semantically faithful to how top-level CJS vars actually behave at runtime (they're locally
  scoped to the module wrapper, not real globals).
  - `detectedModuleType` is still fully consumed by the **rule activation** layer via `filter-module-type` — that's the correct place for the runtime CJS/ESM distinction.

  To address what you originally flagged: I kept the `commonjs` fallback at the **resolver** level, so when `package.json#type` is missing, `detectedModuleType` resolves to `'commonjs'` and
  `filter-module-type` correctly disables module-only rules (S7785 et al.) for those files.

vdiez (approved):
The blocker I raised earlier is resolved on the current head.

I re-checked the latest changes in an isolated worktree and re-ran the focused validation, including the plain `.js` / missing `package.json#type` repro that previously reported `S7785`. The current split looks correct:

- parser `sourceType` stays `module` so scope-aware rules keep their expected behavior;
- missing `package.json#type` now resolves to CommonJS for module-type detection;
- module-only rules are filtered via `detectedModuleType`, so `S7785` is no longer raised on ordinary CommonJS `.js` files.

I also re-ran the targeted test suites locally and they are green.

vdiez:
One non-blocking follow-up I would still add: a focused regression test for the exact plain `.js` + missing `package.json#type` + `S7785` scenario.

That behavior is now covered indirectly by the dependency-manifest and linter tests, and I manually rechecked it on this head, but a single end-to-end test would make the original regression harder to reintroduce.

PR #7047, opened 2026-05-18
JS-1394 Refactor S6767 false-positive escape helpers

francois-mora-sonarsource:
## Summary
Refactor the existing S6767 false-positive escape logic into focused helpers without changing behavior.

## Included
- extract whole-props usage detection
- extract forwardRef callback prop-usage detection
- extract custom superclass props forwarding detection
- keep the decorator wiring behavior unchanged

## Verification
- node --import=tsx --test --test-reporter=spec packages/analysis/src/jsts/rules/S6767/unit.test.ts packages/analysis/src/jsts/rules/S6767/unit.typescript.test.ts packages/analysis/src/jsts/rules/S6767/unit.upstream.test.ts

### DIFF
--- a/packages/analysis/src/jsts/rules/S6767/decorator.ts
+++ b/packages/analysis/src/jsts/rules/S6767/decorator.ts
@@ -34,8 +34,8 @@ export function decorate(rule: Rule.RuleModule): Rule.RuleModule {
       const componentNode = findComponentNode(node, context);
       if (
         componentNode &&
-        (hasSupportedWholePropsUsage(componentNode, context, undefined) ||
-          hasOwnCustomSuperclassPropsForwarding(componentNode, context, undefined))
+        (hasSupportedWholePropsUsage(componentNode, context) ||
+          hasOwnCustomSuperclassPropsForwarding(componentNode))
       ) {
         return;
       }
--- a/packages/analysis/src/jsts/rules/S6767/decorator.ts
+++ b/packages/analysis/src/jsts/rules/S6767/decorator.ts
@@ -16,33 +16,15 @@
  */
 // https://sonarsource.github.io/rspec/#/rspec/S6767/javascript
 
-import type { Rule, Scope, SourceCode } from 'eslint';
+import type { Rule } from 'eslint';
 import type estree from 'estree';
-import type { JSXSpreadAttribute } from 'estree-jsx';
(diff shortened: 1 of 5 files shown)


packages/analysis/src/jsts/rules/S6767/decorator.ts
```diff
@@ -63,172 +45,11 @@ export function decorate(rule: Rule.RuleModule): Rule.RuleModule {
...
-        return true;
-      }
-    }
-    prev = current;
-    current = getNodeParent(current);
-  }
-  return false;
-}
-
-function hasPropsCall(root: estree.Node, keys: SourceCode.VisitorKeys): boolean {
```
erwan-leforestier-sonarsource, line 185: Found by Claude.
There are still mentions of `hasPropsCall` function inside comments in `unit.typescript.test.ts`. They must be removed.

packages/analysis/src/jsts/rules/S6767/forward-ref-indirect-prop-usage.ts
```diff
@@ -0,0 +1,109 @@
...
+ *   forwardRef(() => props.propName);
+ * }
+ */
+export function hasForwardRefCallbackPropUsage(
+  componentNode: estree.Node,
+  context: Rule.RuleContext,
+  propName: string | undefined,
+): boolean {
+  if (!propName || !isFunctionNode(componentNode)) {
+    return false;
```
erwan-leforestier-sonarsource, line 47: From what I can see, you only call `hasForwardRefCallbackPropUsage` from one place where you check if propName is defined. I suggest to simplify the signature of this function to allow only type string for propName and remove the check here. Wdyt ?

```suggestion
 */
export function hasForwardRefCallbackPropUsage(
  componentNode: estree.Node,
  context: Rule.RuleContext,
  propName: string,
): boolean {
  if (!isFunctionNode(componentNode)) {
    return false;
```

erwan-leforestier-sonarsource (approved):
Thanks @francois-mora-sonarsource for your work. Good overall, some minor things to change but other than that, good job 👍

packages/analysis/src/jsts/rules/S6767/forward-ref-indirect-prop-usage.ts
francois-mora-sonarsource: Thanks. Changes implemented as suggested.

PR #7048, opened 2026-05-18
JS-1394 Fix S6767 React owner resolution

francois-mora-sonarsource:
## What changed
This PR fixes the S6767 React owner-resolution path used by the false-positive escapes.

It introduces a more precise `findComponentNodes()` flow that:
- resolves all owners from shared reported types instead of assuming a single owner
- handles contextually typed function components whose parameter type shows up as `any` or `unknown`
- applies the existing S6767 remediations only when all matching owners satisfy the escape

It also removes the temporary legacy single-owner fallback from `packages/analysis/src/jsts/rules/S6767/decorator.ts` and deletes the supporting legacy owner-by-type path from `packages/analysis/src/jsts/rules/helpers/react.ts`.

## Why it changed
The branch originally mixed two different concerns:
- the owner-resolution fix in the React helpers
- temporary compatibility behavior for later false-positive escapes

This version keeps the merged history clean by landing the multi-owner resolution change in its final form instead of shipping an intermediate legacy bridge and removing it later.

That means fallback-dependent cases are no longer suppressed in this PR:
- the real-project `AnchorState` non-props case is re-raised in ruling
- the synthetic anonymous `React.memo(function (props) { ... })` unit fixture is intentionally flipped back to invalid

The dedicated follow-up escapes for those cases stay in later stacked PRs and are not part of this branch.

## Impact
The S6767 ruling delta for this PR is `24` changes across `ant-design`, `desktop`, and `eigen`.

Manual audit of that delta:
- `13` confirmed suppressed false positives from corrected owner resolution
- `8` additional suppressions explained by the already accepted coarse limitation of the existing whole-props remediation
- `2` confirmed raised true positives
- `1` known temporary false positive re-raised after removing the legacy fallback

Confirmed owner-resolution wins:
- `ant-design`: `ClearableLabeledInput.tsx:43,44`
- `desktop`: `app-menu.tsx:41`, `branch-list.tsx:30,35,45,50`, `changes/sidebar.tsx:49,54`, `commit-message-dialog.tsx:37,77`
- `eigen`: `SelectMaxBid.tsx:25`, `AllEvents.tsx:15`

Confirmed raised true positives:
- `eigen`: `CommercialInformation.tsx:34,37`
  - `tracking` is unused by any owner in the file.
  - `refetchArtwork` is used by the wrapper owner, but not by the inner `CommercialInformation` owner, so the report should stay once owners are resolved correctly.

Accepted coarse whole-props suppressions now reached by the corrected owner lookup:
- `desktop/app/src/ui/branches/pull-request-list.tsx:44`
- `desktop/app/src/ui/changes/changes-list.tsx:119`
- `desktop/app/src/ui/diff/side-by-side-diff.tsx:79`
- `desktop/app/src/ui/diff/text-diff.tsx:147`
- `ant-design/components/input/ClearableLabeledInput.tsx:32,40,42,45`

These are not caused by a new PR-specific suppression path. They come from the existing `hasSupportedWholePropsUsage(...)` remediation, which intentionally treats calls such as `resolveSelectedItem(props)`, `getSelectedRowsFromProps(props)`, `highlightParametersEqual(this.props, ...)`, or `hasAddon(this.props)` as sufficient evidence to suppress all props on that owner. `whole-props-usage.ts` is unchanged in this PR; the new owner lookup simply lets that pre-existing coarse remediation apply to the correct owner.

Known temporary regression intentionally accepted in this PR:
- `ant-design/components/anchor/Anchor.tsx:70`
  - This is the old `AnchorState` non-props false positive that was previously hidden by the legacy fallback.
  - The ruling update for this re-raised issue is folded into this branch.

## Root cause
The previous helper could mis-resolve React owners when a typed function component was contextually typed and the function parameter surfaced as `any` / `unknown`. That prevented some existing escapes from applying consistently.

The temporary single-owner fallback also kept old suppression behavior alive after the new lookup. Removing it makes this PR describe the actual steady state of the owner-resolution change rather than an intermediate compatibility step.

## How owner resolution works
For TypeScript-reported props, `findComponentNodes()` now resolves owners in two modes:
- If it can recover a specific reported member such as `relay` in `interface SharedProps { relay: string }`, it uses member-level matching. For each candidate component, it recovers the component props type and keeps the component only if that props type either:
  - contains the same reported member declaration, or
  - contains a compatible member and still uses the reported enclosing type through `extends`, aliases, intersections, or similar type composition
- If it cannot recover a usable reported member identity, it falls back to whole-type matching and keeps only components whose props type is mutually assignable with the reported enclosing props type.
- The result of `findComponentNodes()` is the full owner set for that report node.
- The decorator then applies the existing false-positive escapes to that owner set and suppresses only when every resolved owner satisfies one of those escapes.

Example:

```ts
interface SharedProps {
  relay: string;
}

interface ChildProps extends SharedProps {
  title: string;
}

const Child: React.FC<ChildProps> = props => <div>{props.title}</div>;
const Wrapper: React.FC<SharedProps> = props => <div>{props.relay}</div>;
```

For the report on `relay`, `findComponentNodes()` returns both `Child` and `Wrapper`. The decorator keeps the issue unless both owners justify suppression, which is why the rule now continues to report shared props that are only used by some owners.

## Validation
- `npx tsx --tsconfig packages/tsconfig.test.json --test --test-concurrency=1 packages/analysis/tests/jsts/rules/helpers/react.test.ts packages/analysis/src/jsts/rules/S6767/unit.test.ts packages/analysis/src/jsts/rules/S6767/unit.typescript.test.ts`
- `npx tsx --tsconfig packages/tsconfig.test.json --test --test-concurrency=1 packages/ruling/projects/ant-design.ruling.test.ts`
- full PR check suite green on GitHub, including `Sonar Review`, `SonarQube Code Analysis`, `JS/TS Ruling`, and `Ruling Test`


francois-mora-sonarsource:
Verified on the current head (`2a4ed6bae`):

- all Sonar Review threads are resolved
- `SonarQube Code Analysis` is green with 0 new issues
- the full PR check suite is green

So the earlier review-summary comments saying some issues remain open are stale relative to the latest revision and analysis.

packages/analysis/tests/jsts/rules/helpers/react.test.ts
```diff
@@ -0,0 +1,336 @@
...
+    return parent.id.name;
+  }
+
+  if (hasIdentifierId(componentNode)) {
+    return componentNode.id.name;
+  }
+
+  return parent?.type === 'VariableDeclarator' && parent.id.type === 'Identifier'
+    ? parent.id.name
+    : undefined;
```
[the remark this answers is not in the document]
francois-mora-sonarsource: Keeping the current helper shape.

The first branch is intentional and mirrors `getComponentIdentifier` in `react.ts`. If `hasIdentifierId(componentNode)` runs first, a named expression assigned to a variable changes meaning:

```ts
const Foo = function Bar() {};
const Baz = class Qux extends React.Component {};
```

In those cases, `componentNode.id.name` would be `Bar` / `Qux`, but the component owner name we want for these tests is the variable binding: `Foo` / `Baz`.

So the `FunctionExpression` / `ClassExpression` + `VariableDeclarator` branch is not equivalent to the later generic `VariableDeclarator` fallback. It exists specifically to prefer the binding name over the inner expression name for named expressions.

packages/analysis/src/jsts/rules/helpers/react.ts
```diff
@@ -17,11 +17,23 @@
...
+  ReportedTypeDetails,
+  type ReportedEnclosingType,
+} from './reported-type.js';
+import { areMutuallyAssignableTypes, areSameTypeDeclarations, getTypeFromTreeNode } from './type.js';
+
+type TypeMemberNode = TSESTree.TSPropertySignature | TSESTree.TSMethodSignature;
+type ReportedTypeMember = ReportedTypeDetails<TypeMemberNode, ts.TypeElement>;
+type SourceCache = {
+  componentNodes: estree.Node[] | undefined;
+  ownersByReportNode: WeakMap<estree.Node, estree.Node[] | null>;
```
nathsou: This cache only memoizes by the exact reported node (property level), so the expensive owner search is repeated once per prop member. S6767 reports each unused property separately, which means one shared props interface with many members now re-runs the full component scan, re-resolves component props types, and re-traverses the enclosing-type graph for every member. The previous implementation cached at the type-declaration level, so this is a real hot-path regression on large React files.

francois-mora-sonarsource: You're right that the earlier version cached by reported member, so the TypeScript owner lookup could be repeated for each unused prop.

The cache refactor was meant to move that expensive work up to the props-type level instead:
- cache the component analysis once per component
- cache the broad candidate-owner set once per reported props type

After that, each reported member only does the last filtering step on top of that cached candidate set.

I did not cache the final owner list per props type, because that result is intentionally member-specific. Two members from the same props type can still resolve to different owner sets, so only the broad candidate phase is shared.

nathsou (approved):
This rule has a high FP rate at around 20% so I think the added complexity is worth it in this case, as we discussed I think it would be interesting to see if we get feedback about the change from: suppress if at least one component uses the prop, to: suppress only if all the components consume the prop.

francois-mora-sonarsource:
> This rule has a high FP rate at around 20% so I think the added complexity is worth it in this case, as we discussed I think it would be interesting to see if we get feedback about the change from: suppress if at least one component uses the prop, to: suppress only if all the components consume the prop.

Thanks for the review. I agree with you, the handling of multiple components sharing the type definition of reported property is conservative and still reports code with the idea that types should not be shared. This could be still considered as noise. And at least it should be documented in the follow-up rspec PR. The change is required though as the current behavior is not exactly 'suppress if at least one component is a false positive' but 'suppress if A component is a false positive' as the current mechanism elects a component without a clear logic.

PR #6492, opened 2026-03-02
JS-1398 Replace rspec-maven-plugin with sync-rspec script

zglicz:
## Summary
- Add `tools/sync-rspec.ts` that reads the SonarSource/rspec repo, merges parent + language-specific `metadata.json` files, and writes them to `resources/rule-data/`
- The script auto-clones the rspec repo (sparse checkout, `rules/` only) when no local clone exists, and fetches latest changes on subsequent runs
- New `sync_rspec` CI prep job runs the script once and uploads deployed sonar-plugin resource dirs as artifact
- All dependent jobs download this artifact instead of using the maven plugin
- Remove `rspec-maven-plugin` and `deploy-rule-data` exec from `pom.xml`
- Remove RSPEC token from `build` and `build_win` jobs

## Test plan
- [ ] CI `sync_rspec` job completes and uploads artifact
- [ ] All dependent jobs (`build`, `build_win`, `build_eslint_plugin`, `test_js`, `test_js_win`, `js_ts_ruling`) download and use the artifact correctly
- [ ] Maven build succeeds without `rspec-maven-plugin`
- [ ] `npm run generate-meta` works with artifact-provided rule data
- [ ] Local dev: `npm run sync-rspec -- --language javascript` auto-clones and produces correct output

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #7059, opened 2026-05-19
JS-1764 Re-enable rspec-maven-plugin for rule data generation

zglicz:
## Summary
- re-enable `rspec-maven-plugin` for JS/CSS rule-data generation and make `npm run generate-meta` invoke the Maven-backed RSPEC refresh only when the generated JS/CSS outputs or `rspec.sha` are missing
- consume `rspec-maven-plugin` `1.3.0.115`, built from SonarSource/rspec-maven-plugin#75, for GitHub token forwarding and RSPEC SHA pinning support
- remove the obsolete JS `sync-rspec` HTML/JSON generation path and drop the now-unused `@asciidoctor/core` dependency
- make Maven reuse existing generated RSPEC outputs automatically, so local `mvn install` reuses them and `mvn clean install` refreshes them
- make CI prepare rule data once, publish it as an artifact, and reuse it from downstream jobs without a dedicated skip env var
- keep the artifact download target at `sonar-plugin/`, because the uploaded zip strips that leading path and needs to be re-expanded there to recreate `sonar-plugin/javascript-checks/...` and `sonar-plugin/css/...`
- make `generate-rule-data:maven` bootstrap `sonar-plugin/api` first, then run the `javascript-checks` lifecycle with RSPEC generation enabled
- make `deploy-rule-data` publish `sonar-plugin/javascript-checks/src/main/resources/rspec.sha` from the rule-api cache clone so the prepared artifact is complete
- include `sonar-plugin/javascript-checks/src/main/resources/rspec.sha` in the prepared artifact and make `clean` remove the generated JS/CSS plugin resource copies as part of the refresh path
- update release/docker workflows and developer docs for the Maven-backed RSPEC flow and token usage

## Validation
- clean regeneration path:
  - `rm -rf sonar-plugin/javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript sonar-plugin/css/src/main/resources/org/sonar/l10n/css/rules/css sonar-plugin/javascript-checks/src/main/resources/rspec.sha resources/rule-data sonar-plugin/javascript-checks/target`
  - `GITHUB_TOKEN="$(gh auth token 2>/dev/null)" npm run generate-rule-data:maven`
  - verified this recreated the JS rule-data folder, the CSS rule-data folder, and `sonar-plugin/javascript-checks/src/main/resources/rspec.sha`
- prepared-data fast path:
  - `env -u GITHUB_TOKEN npm run generate-meta`
  - verified this reused prepared outputs and ran only `generate-meta:raw`
- downstream Maven reuse:
  - `env -u GITHUB_TOKEN mvn -B -ntp -pl sonar-plugin/sonar-javascript-plugin -am -DskipTests package`
  - verified there were no `generate-rspec`, `generate-rule-data`, `deploy-rule-data`, or `org.sonarsource.rspec` executions in the Maven log
- jar contents:
  - verified the produced plugin jar contains `org/sonar/l10n/javascript/rules/javascript/**`, `org/sonar/l10n/css/rules/css/**`, and `rspec.sha`
  - expected counts: `531` JS json, `528` JS html, `32` CSS json, `30` CSS html
- artifact layout:
  - downloaded the prepared artifact from GitHub Actions and verified that extracting it under `sonar-plugin/` recreates the expected `sonar-plugin/javascript-checks/...`, `sonar-plugin/css/...`, and `sonar-plugin/javascript-checks/src/main/resources/rspec.sha` paths
- parsed `.github/workflows/build.yml` successfully after the workflow refactor
- after the `1.3.0.115` plugin bump:
  - `mvn -q -DskipTests -Drevision=1.3.0.115 install` in rspec-maven-plugin
  - `mvn -q -ntp org.sonarsource.rspec:rspec-maven-plugin:1.3.0.115:help -Ddetail=false` in SonarJS
  - `node --check tools/generate-rule-data.mjs` in SonarJS
  - `mvn -q -ntp -pl sonar-plugin/javascript-checks -am -Dskip-nodejs -DskipTests -Drule.data.build.phase=none -Drspec.sha=0123456789abcdef0123456789abcdef01234567 validate` in SonarJS

## Notes
- SonarSource/sonar-rule-api#306 and SonarSource/rspec-maven-plugin#75 provide the upstream token and RSPEC SHA plumbing consumed by this branch
- on a fresh checkout, or after `mvn clean`, local RSPEC refresh still needs GitHub authentication through `GITHUB_TOKEN` or the plugin/rule-api fallback chain
- there is a separate pre-existing workflow failure on `QA on Windows (2)` around `.github/workflows/build.yml` line `931` / `fromJSON(...licenses_token)`; that also reproduces on `master`


vdiez (commented):
I dug into why `sync-rspec.ts` existed on `master`, and I do not think this PR is functionally equivalent yet for local developer workflows.

Historical context first.

The TypeScript `sync-rspec.ts` tooling was introduced in PR #6492 (commit `c164606ee45891c9174946437d97db99a5b325e3`). The point of that change was not just to move logic around. It was added specifically to replace the `rspec-maven-plugin` path because the Maven flow was not robust enough for day-to-day development:

- it depended on `sonar-rule-api` auth behavior that required `GITHUB_TOKEN` in the environment;
- it was not as straightforward to run on developer machines as the existing Node-based workflow;
- SonarJS wanted the generated rule-data that feeds JS/TS builds to be reproducible locally with the normal repo tooling, without special Maven/plugin auth setup.

That history matters here, because the bar for removing `sync-rspec.ts` is not only “the build passes in CI” or “the generated files look mostly the same”. The replacement also needs to preserve the developer ergonomics we had on `master`: on a clean machine/worktree, a developer should be able to regenerate the RSPEC-derived outputs with the normal build flow as easily as before.

I tested that explicitly.

On this branch, using a clean worktree, `npm ci` + `npm run bbf` does regenerate the expected outputs, and the repository shape is aligned with the intended direction:

- RSPEC JSON/HTML and `rspec.sha` are generated at build time and are not git-tracked;
- `generated-meta.ts` files are generated at build time and are not git-tracked;
- generated Java check classes are generated at build time and are not git-tracked.

So the repository-layout goal is mostly correct.

However, there is still one important regression compared to `master`: if I delete the generated RSPEC outputs in a clean worktree and run the generation flow without exporting `GITHUB_TOKEN`, the Maven path fails. In practice that means the replacement is still not as easy to run on dev machines as the `sync-rspec.ts` flow it is removing.

The reason is upstream:

1. `rspec-maven-plugin` still does not accept any explicit GitHub auth parameter. The plugin currently only exposes `ruleSubdirectory`, `targetDirectory`, and `vcsBranchName`, and then constructs `GitHubRuleMaker.create(branch)` internally.
2. The released plugin version used here (`1.3.0.46`) depends on `sonar-rule-api` `2.18.0.5734`, where `GitHubRuleMaker` only looks at `System.getenv(GITHUB_TOKEN)`.
3. Even on current `sonar-rule-api` `master`, while `GitHubRuleMaker` now falls back to `gh auth token`, `CoverageLoader` still hard-requires `GITHUB_TOKEN` from the environment.

So even if SonarJS switches back to the Maven-based generation flow, we are still missing the pieces needed to make that flow dev-machine-friendly in the same way `master` was.

For me, the minimum bar for functional equivalence is:

- a clean checkout/worktree can regenerate the RSPEC-derived outputs without requiring contributors to pre-export `GITHUB_TOKEN` in their shell;
- the build can use existing developer auth state (for example `gh auth login`) or an explicit parameter, instead of only ambient environment variables;
- the generated HTML/JSON output matches the current TypeScript flow closely enough that we do not introduce content regressions.

Concretely, I think this means the upstream work is still required before SonarJS can safely remove `sync-rspec.ts`:

1. In `sonar-rule-api`, introduce a shared token resolution mechanism used everywhere GitHub access happens.
   Precedence should be something like: explicit token parameter -> `GITHUB_TOKEN` env var -> `gh auth token`.
2. In `sonar-rule-api`, expose APIs that allow callers to pass the token explicitly instead of forcing env lookup internally (for example `GitHubRuleMaker.create(branch, token)` and a corresponding `CoverageLoader` constructor/entry point).
3. In `rspec-maven-plugin`, expose an optional auth parameter (or Maven settings-based credential lookup) and pass it explicitly to `sonar-rule-api`.
4. After that, SonarJS can bump to the new plugin/rule-api versions and keep this PR’s general repository policy: generated RSPEC outputs, generated TS meta, and generated Java checks should all be build-time artifacts and not be tracked in git.

There is also one output-level issue worth addressing before we treat the Maven flow as a drop-in replacement: I found a real HTML rendering delta on `S7724.html` where the generated output contains stray `<strong>` markup around the phrase after ``/* eslint-disable */``. The JSON differences I saw were only ordering/newline noise, but this HTML difference looks substantive.

So my conclusion is:

- the repository cleanup direction in this PR is good;
- but the TypeScript `sync-rspec.ts` tooling was originally added for a concrete developer-workflow reason, not just implementation preference;
- and we have not fully closed that gap yet, because the Maven path still depends on `GITHUB_TOKEN` being exported in the environment in scenarios where `master` was easier to use.

I would be comfortable with this switch once the Maven/plugin/rule-api stack is updated so that local generation works smoothly on developer machines without requiring manual env-token setup, and once the HTML rendering delta is understood/fixed.

zglicz:
Added RSPEC SHA pinning support on this branch.

Dependency PRs:
- SonarSource/sonar-rule-api#306 adds exact RSPEC revision checkout support and is merged.
- SonarSource/rspec-maven-plugin#75 passes `rspec.sha` through to rule-api and produced build `1.3.0.115`.

SonarJS now:
- uses a Node wrapper for `generate-rule-data:maven` so a root `rspec.sha` file is passed as `-Drspec.sha=<commit>` cross-platform;
- configures `<rspecSha>${rspec.sha}</rspecSha>` for direct Maven invocations;
- consumes `rspec-maven-plugin` `1.3.0.115` instead of the provisional build version;
- documents local pinning in `docs/DEV.md`.

Verification:
- `mvn -q -Dtest=GitCloneTest,GitHubRuleMakerTest,MainTest test` in sonar-rule-api
- `mvn -q -DskipTests install` in sonar-rule-api
- `mvn -q -DskipTests package` in rspec-maven-plugin
- `mvn -q -DskipTests -Drevision=1.3.0.115 install` in rspec-maven-plugin
- `node --check tools/generate-rule-data.mjs` in SonarJS
- Prettier check for `package.json`, `tools/generate-rule-data.mjs`, `docs/DEV.md`
- fake-Maven smoke test verifying `-Drspec.sha` propagation from root `rspec.sha`
- `mvn -q -ntp org.sonarsource.rspec:rspec-maven-plugin:1.3.0.115:help -Ddetail=false` in SonarJS
- `mvn -q -ntp -pl sonar-plugin/javascript-checks -am -Dskip-nodejs -DskipTests -Drule.data.build.phase=none -Drspec.sha=0123456789abcdef0123456789abcdef01234567 validate` in SonarJS

vdiez (commented):
I re-checked this with the upstream changes in `sonar-rule-api#306` and `rspec-maven-plugin#75` in place.

Main remaining finding:

- `tools/generate-rule-data.mjs`: fresh-checkout generation on Windows still goes through `mvn.cmd` via `spawnSync(...)` without `shell: true`. CI stays green because `prepare_rspec_rule_data` runs on Linux and the Windows jobs only download the prepared artifact, so the broken cold-start path is never exercised there. But a developer on a fresh Windows checkout hitting `npm run generate-meta` / `npm run generate-rule-data:maven` would still fail.

On the broader migration question, my previous concern about local developer workflow is mostly resolved now:

- the original reason for `sync-rspec.ts` on `master` was not just implementation preference, it was that the Maven path was not convenient enough for day-to-day local use;
- that gap looks closed upstream now: `sonar-rule-api#306` centralizes token resolution (`explicit -> GITHUB_TOKEN -> gh auth token`) and `CoverageLoader` uses the same path, while `rspec-maven-plugin#75` exposes the SHA/token plumbing consumed here;
- on a clean worktree I was able to run `npm ci` + `npm run bbf` successfully with `GITHUB_TOKEN` unset.

I still see a couple of HTML deltas vs `master` (`S5852`, `S7724`), but I probably would not block on those. The extra HTML-fidelity fixes were local to the SonarJS `sync-rspec.ts` path, and the more important direction here is to normalize back onto the shared `sonar-rule-api` / `rspec-maven-plugin` flow used across analyzers rather than keep a SonarJS-specific renderer fork.

vdiez (approved):
Approving.

The main concern I had on the earlier iteration was the local developer workflow gap that originally motivated `sync-rspec.ts`, and that now looks resolved with `sonar-rule-api#306` and `rspec-maven-plugin#75` in place.

I still see the fresh-Windows-checkout issue in the `mvn.cmd` wrapper path, but I do not want to block the more important direction here, which is converging back onto the shared `sonar-rule-api` / `rspec-maven-plugin` flow across analyzers.

I can tackle the Windows fix separately.

PR #7067, opened 2026-05-20
JS-1715 Implement sonar-resolve support in the SonarJS plugin

zglicz:
## Summary
- extract sonar-resolve comment candidates from JS/TS/CSS analysis results, including embedded content
- persist issue resolutions in the SonarJS plugin through `AnalysisProcessor` using analyzer-commons parsing
- add a plugin integration test covering sonar-resolve in embedded JavaScript inside HTML

## Testing
- mvn -pl sonar-plugin/sonar-javascript-plugin -am -DskipTests -Dskip-java-generation package
- mvn -pl its/plugin/tests -am -DskipTests=false -Dtest=SonarResolveTest -Dsurefire.failIfNoSpecifiedTests=false -Dskip-java-generation test

### DIFF
--- a/its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/SonarResolveTest.java
+++ b/its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/SonarResolveTest.java
@@ -59,6 +59,10 @@ class SonarResolveTest {
         assertThat(issue.getIssueStatus()).isEqualTo("ACCEPTED");
         assertThat(issue.hasResolution()).isTrue();
         assertThat(issue.getResolution()).isNotBlank();
+        assertThat(issue.hasComments()).isTrue();
+        assertThat(issue.getComments().getCommentsList()).anySatisfy(comment ->
+          assertThat(comment.getMarkdown()).contains("accepted in HTML")
+        );
       });
   }
 
--- a/its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/SonarResolveTest.java
+++ b/its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/SonarResolveTest.java
@@ -1,0 +1,94 @@
+/*
+ * SonarQube JavaScript Plugin
+ * Copyright (C) SonarSource Sàrl
+ * mailto:info AT sonarsource DOT com
+ *
+ * You can redistribute and/or modify this program under the terms of
+ * the Sonar Source-Available License Version 1, as published by SonarSource Sàrl.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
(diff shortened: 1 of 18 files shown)


its/plugin/tests/src/test/java/com/sonar/javascript/it/plugin/SonarResolveTest.java
```diff
@@ -0,0 +1,94 @@
...
+    String fileKey = projectKey + ":index.html";
+
+    assertThat(searchIssues(fileKey))
+      .singleElement()
+      .satisfies(issue -> {
+        assertThat(issue.getRule()).isEqualTo("javascript:S1116");
+        assertThat(issue.getLine()).isEqualTo(5);
+        assertThat(issue.getIssueStatus()).isEqualTo("ACCEPTED");
+        assertThat(issue.hasResolution()).isTrue();
+        assertThat(issue.getResolution()).isNotBlank();
```
zglicz, line 61: can we check the provided reason as well? for the exact phrase that was used?

packages/analysis/src/common/sonar-resolve.ts
```diff
@@ -0,0 +1,155 @@
...
+  const orderedComments = [...comments].sort(
+    (left, right) => left.line - right.line || left.column - right.column,
+  );
+  let lineCommentRun: CommentSource[] = [];
+
+  for (const comment of orderedComments) {
+    if (comment.kind === 'line') {
+      if (
+        lineCommentRun.length > 0 &&
+        lineCommentRun[lineCommentRun.length - 1].endLine + 1 !== comment.line
```
zglicz: lets use .at(-1)

packages/analysis/src/common/sonar-resolve.ts
```diff
@@ -0,0 +1,157 @@
...
+      if (
+        lineCommentRun.length > 0 &&
+        previousLineComment != null &&
+        previousLineComment.endLine + 1 !== comment.line
+      ) {
+        pushLineCommentRun(lineCommentRun, extracted);
+        lineCommentRun = [];
+      }
+
+      lineCommentRun.push(comment);
```
vdiez, line 104: This groups every consecutive line comment into one continuation candidate. That means a case like:

```js
// sonar-resolve javascript:S1116 "first
// sonar-resolve javascript:S1121 "second"
```

gets serialized as a single `SonarResolveComment`, so `AnalysisProcessor` consumes the second directive as part of the first justification and never parses it separately. I reproduced this locally with `extractSonarResolveCommentsFromJsTsComments()`. `sonar-cpp` handles the same case by stopping continuation when the next line starts with a fresh `sonar-resolve`, and I think we need the same guard here (plus a regression test), otherwise adjacent directives or unrelated consecutive line comments can be swallowed.

zglicz: Good catch!

vdiez (approved):
Rechecked the latest changes on May 27, 2026. The adjacent sonar-resolve directive handling issue is fixed, the regression test covers the swallowed-directive case, and the validation I reran passed.

PR #7085, opened 2026-05-22
JS-1773 Fix npm 11 lockfile consistency

jdkandersson:
## Summary
- add missing optional @emnapi lockfile entries required by newer npm 11
- keep the change limited to package-lock.json

## Validation
- npx npm@11.12.1 ci --ignore-scripts --dry-run

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2195,6 +2195,29 @@
         "postcss-selector-parser": "^7.1.1"
       }
     },
+    "node_modules/@emnapi/core": {
+      "version": "1.10.0",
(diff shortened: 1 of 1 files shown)


zglicz (changes requested):
Not sure if this is necessary? We currently ship our plugin with node version v24.11. It is shipped with npm@11.6.2 and I haven't had to change the lock file.
Perhaps synchronize the version of node you are using to the plugin's node version?

jdkandersson:
This was subsequently fixed, I think there was some issue with the lock file

PR #7103, opened 2026-05-26
JS-537 Fix LCOV resolution for duplicate suffix paths

vdiez:
## Summary

- resolve LCOV `SF:` paths in this order: exact analyzed path, path relative to the report directory, then suffix-based lookup
- fix the community case where package-local paths such as `src/index.ts` were incorrectly matched across modules by trying report-relative resolution before suffix matching
- walk report ancestor directories up to the analysis base dir so monorepo layouts like `<package>/coverage/lcov.info` still resolve package-local `SF:` paths strictly before guessing
- preserve the historical suffix fallback for ambiguous duplicates as a last resort by deterministically picking the first match and emitting a `DEBUG` log when we are explicitly guessing
- retry absolute `SF:` paths via the analysis base dir and its real path so Windows directory aliases such as junctions or symlinks still map to the indexed file before we ever fall back to guessing
- extend the reproducer coverage tests for package-relative reports, ambiguous duplicate suffix fallback, absolute-path control, and aliased analysis base directories

## Testing

```bash
mvn -pl sonar-plugin/sonar-javascript-plugin -am "-Dtest=org.sonar.plugins.javascript.lcov.CoverageSensorTest,org.sonar.plugins.javascript.lcov.FileLocatorTest,org.sonar.plugins.javascript.lcov.CoverageSensorDuplicateSuffixPathReproducerTest" "-Dsurefire.failIfNoSpecifiedTests=false" test
```

### DIFF
--- a/sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/lcov/LCOVParser.java
+++ b/sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/lcov/LCOVParser.java
@@ -215,8 +215,15 @@ class LCOVParser {
 
   @CheckForNull
   private InputFile inputFileByExactPath(String filePath) {
+    String sanitizedPath = PathUtils.sanitize(filePath);
+    if (sanitizedPath == null) {
+      return null;
+    }
     FilePredicates predicates = context.fileSystem().predicates();
-    return context.fileSystem().inputFile(predicates.hasPath(filePath));
+    if (new File(sanitizedPath).isAbsolute()) {
+      return context.fileSystem().inputFile(predicates.hasAbsolutePath(sanitizedPath));
+    }
+    return context.fileSystem().inputFile(predicates.hasPath(sanitizedPath));
   }
 
   @CheckForNull
--- a/sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/lcov/LCOVParser.java
+++ b/sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/lcov/LCOVParser.java
@@ -188,18 +188,30 @@ class LCOVParser {
    * Resolves the SF path to an indexed input file.
    * <p>
    * Resolution order:
-   * 1) exact filesystem predicate match (absolute or project-relative path),
+   * 1) exact absolute-path match,
    * 2) relative-to-report lookup for package-local LCOV paths,
-   * 3) suffix-based lookup through {@link FileLocator}, preferring unique matches and falling back
+   * 3) exact project-relative match,
+   * 4) suffix-based lookup through {@link FileLocator}, preferring unique matches and falling back
    * to the historical deterministic guess when several analyzed files share the same suffix.
    */
   @CheckForNull
   private InputFile inputFileForSourceFile(File reportFile, String line) {
     // SF:<absolute path to the source file>
     String filePath = line.substring(SF.length());
-    InputFile inputFile = inputFileByExactPath(filePath);
-    if (inputFile == null) {
-      inputFile = inputFileRelativeToReport(reportFile, filePath);
+    String sanitizedPath = PathUtils.sanitize(filePath);
+    if (sanitizedPath == null) {
+      unresolvedPaths.add(filePath);
+      return null;
+    }
+
+    InputFile inputFile;
+    if (new File(sanitizedPath).isAbsolute()) {
+      inputFile = inputFileByAbsolutePath(context.fileSystem().predicates(), sanitizedPath);
+    } else {
+      inputFile = inputFileRelativeToReport(reportFile, sanitizedPath);
+      if (inputFile == null) {
+        inputFile = inputFileByProjectRelativePath(sanitizedPath);
+      }
     }
     if (inputFile == null) {
       FileLocator.Resolution resolution = fileLocator.resolve(filePath);
@@ -220,25 +232,12 @@ class LCOVParser {
   }
 
   @CheckForNull
-  private InputFile inputFileByExactPath(String filePath) {
-    String sanitizedPath = PathUtils.sanitize(filePath);
-    if (sanitizedPath == null) {
-      return null;
-    }
-    FilePredicates predicates = context.fileSystem().predicates();
-    if (new File(sanitizedPath).isAbsolute()) {
-      return inputFileByAbsolutePath(predicates, sanitizedPath);
-    }
-    return context.fileSystem().inputFile(predicates.hasPath(sanitizedPath));
+  private InputFile inputFileByProjectRelativePath(String relativePath) {
+    return context.fileSystem().inputFile(context.fileSystem().predicates().hasPath(relativePath));
   }
 
   @CheckForNull
   private InputFile inputFileRelativeToReport(File reportFile, String filePath) {
-    String sanitizedPath = PathUtils.sanitize(filePath);
-    if (sanitizedPath == null || new File(sanitizedPath).isAbsolute()) {
-      return null;
-    }
-
     File reportDirectory = reportFile.getParentFile();
     if (reportDirectory == null) {
       return null;
@@ -252,7 +251,7 @@ class LCOVParser {
     ) {
       InputFile inputFile = inputFileByAbsolutePath(
         predicates,
-        currentDirectory.resolve(sanitizedPath).normalize().toString()
+        currentDirectory.resolve(filePath).normalize().toString()
       );
       if (inputFile != null) {
         return inputFile;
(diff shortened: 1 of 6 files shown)


sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/lcov/LCOVParser.java
```diff
@@ -191,25 +183,62 @@ private void logWrongDataWarning(String dataType, int reportLineNum, Exception e
...
+    InputFile inputFile = inputFileByExactPath(filePath);
+    if (inputFile == null) {
+      inputFile = inputFileRelativeToReport(reportFile, filePath);
+    }
     if (inputFile == null) {
-      inputFile = fileLocator.getInputFile(filePath);
+      FileLocator.Resolution resolution = fileLocator.resolve(filePath);
+      inputFile = resolution.inputFile();
+      if (inputFile != null && resolution.guessed() && LOG.isDebugEnabled()) {
+        LOG.debug(
```
erwan-leforestier-sonarsource, line 220: Isn't `warning` level more appropriate here for this message ?

erwan-leforestier-sonarsource (commented):
Thanks @vdiez for the quick fix ! Looks mostly good to me. One concern regarding the lcov location and from where we should resolve relative file paths (discussed in person) and one minor comment about the log level. Other than that, lgtm !

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/lcov/LCOVParser.java
vdiez: it's not a failure and we already warn when we truly cannot resolve paths, so i prefer to keep it as debug tbh. aside of that, changes on relative paths have been pushed according your proposal. pls check. i will retrigger CI when github actions is back online

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/lcov/LCOVParser.java
```diff
@@ -233,10 +233,37 @@ private InputFile inputFileRelativeToReport(File reportFile, String filePath) {
...
+        currentDirectory.resolve(sanitizedPath).normalize().toString()
+      );
+      if (inputFile != null) {
+        return inputFile;
+      }
+      if (
+        currentDirectory.equals(analysisBaseDir) || !currentDirectory.startsWith(analysisBaseDir)
+      ) {
+        break;
+      }
```
erwan-leforestier-sonarsource, line 261: This break could be relocated inside the loop exit condition so you're sure that we never read a inputFile that is outside of the base dir.
```java
for (
      Path currentDirectory = reportDirectory.toPath().toAbsolutePath().normalize();
      currentDirectory != null || (currentDirectory.equals(analysisBaseDir) || !currentDirectory.startsWith(analysisBaseDir));
      currentDirectory = currentDirectory.getParent()
    ) {
      InputFile inputFile = inputFileByAbsolutePath(
        predicates,
        currentDirectory.resolve(sanitizedPath).normalize().toString()
      );
      if (inputFile != null) {
        return inputFile;
      }
}
```

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/lcov/LCOVParser.java
```diff
@@ -191,25 +189,120 @@ private void logWrongDataWarning(String dataType, int reportLineNum, Exception e
...
+  private String relativePathFromAnalysisBaseDir(String absolutePath) {
+    Path absoluteFilePath = new File(absolutePath).toPath().toAbsolutePath().normalize();
+
+    String relativePath = PATH_RESOLVER.relativePath(analysisBaseDir, absoluteFilePath);
+    if (relativePath != null) {
+      return relativePath;
+    }
+
+    if (!realAnalysisBaseDir.equals(analysisBaseDir)) {
+      return PATH_RESOLVER.relativePath(realAnalysisBaseDir, absoluteFilePath);
```
erwan-leforestier-sonarsource, line 318: I think this still misses the “same file, different path spelling” case. For example, if analysis runs from `/repo-link `where `/repo-link` -> `/repo-real`, and the LCOV report contains `SF:/repo-real/src/index.ts`, we canonicalize the base dir but not `absoluteFilePath`, so `relativePath(realAnalysisBaseDir,absoluteFilePath)` still returns null and the file stays unresolved.

vdiez: Good catch, this was still missing. Applied in 8e33aaef8: after the cheap absolute and relative checks miss, we canonicalize the deepest existing parent of the LCOV absolute path and retry the relative-path mapping. That covers the alias vs real-path spelling case in both directions. I also added a reproducer for the alias-spelling scenario, and the targeted LCOV suite passes.

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/lcov/LCOVParser.java
vdiez: Applied in 8e33aaef8. I moved the guard into the loop condition so we stop before probing a directory outside the analysis roots before calling inputFileByAbsolutePath(). I reran the targeted LCOV tests after the change and they pass.

sonar-plugin/sonar-javascript-plugin/src/main/java/org/sonar/plugins/javascript/lcov/LCOVParser.java
vdiez: I rechecked this while doing the follow-up and I still prefer to keep it at DEBUG. We already emit WARN when we cannot resolve LCOV paths; this message is only about the deterministic fallback when we did resolve a file, so I think it should stay diagnostic rather than user-facing noise.

vdiez:
Checked the latest gitar-bot note. The helper-level finding is valid: ReversePathTree.getFirstLeaf() could skip a file stored on the current suffix node and keep descending to a deeper match.

The practical impact in LCOVParser is limited because exact project-relative matches are resolved before the suffix fallback, so the bot example is usually masked there. Still, the suffix tree behavior itself was wrong, so I fixed it in 06098bcbc by stopping the descent when the current node already has a file, and I added a regression test for lib/src/index.ts vs packages/a/lib/src/index.ts.

Validation: mvn -Dskip-nodejs -pl sonar-plugin/sonar-javascript-plugin -am -Dtest=org.sonar.plugins.javascript.lcov.FileLocatorTest,org.sonar.plugins.javascript.lcov.CoverageSensorTest,org.sonar.plugins.javascript.lcov.CoverageSensorDuplicateSuffixPathReproducerTest -Dsurefire.failIfNoSpecifiedTests=false test

zglicz:
One case still looks broken to me.

`inputFileForSourceFile` now does:
1. exact path lookup
2. report-relative lookup
3. suffix fallback

For relative `SF:` entries, that still prefers a repo-root exact match over the report-local file. So if the analysis contains both `src/index.ts` and `packages/a/src/index.ts`, and `packages/a/coverage/lcov.info` contains `SF:src/index.ts`, coverage is attached to the root file instead of `packages/a/src/index.ts`.

I reproduced this locally with a focused test on this branch: the package file stays uncovered and the root file gets the hit.

I think report-relative resolution needs to win for relative `SF:` paths, or the exact lookup needs to be split so repo-relative matching happens after report-relative matching. It would also be worth adding a regression test for the `src/index.ts` + `packages/a/src/index.ts` shape, since the current reproducer tests do not cover it.

That said, if the intended precedence is to keep project-relative exact matches ahead of report-relative matches, I’m fine with that decision as long as it is explicit. I mainly wanted to call out the behavioral tradeoff and the missing regression coverage for this shape.

zglicz (approved):

vdiez:
@zglicz Thanks, good catch. For relative `SF:` entries we were still checking repo-relative `hasPath(...)` before report-relative resolution, so `packages/a/coverage/lcov.info` with `SF:src/index.ts` could still attach coverage to a root `src/index.ts`.

Fixed in `35d6b1c386` by resolving relative paths in this order: report-relative, then repo-relative, then suffix fallback. Absolute paths keep the existing absolute-path flow. I also added a regression test for the `src/index.ts` + `packages/a/src/index.ts` shape.

Validation:
- `mvn -pl sonar-plugin/sonar-javascript-plugin -am -Dskip-java-generation "-Dtest=org.sonar.plugins.javascript.lcov.CoverageSensorDuplicateSuffixPathReproducerTest#should_prefer_report_relative_match_over_repo_root_relative_match" "-Dsurefire.failIfNoSpecifiedTests=false" test`
- `mvn -pl sonar-plugin/sonar-javascript-plugin -am -Dskip-nodejs -Dskip-java-generation "-Dtest=org.sonar.plugins.javascript.lcov.CoverageSensorTest,org.sonar.plugins.javascript.lcov.FileLocatorTest,org.sonar.plugins.javascript.lcov.CoverageSensorDuplicateSuffixPathReproducerTest" "-Dsurefire.failIfNoSpecifiedTests=false" test`

PR #7120, opened 2026-05-28
JS-1791 Handle deep AST protobuf serialization

vdiez:
## Summary

- raise the protobuf recursion limit while encoding and decoding ESTree ASTs
- align the JS-side recursion limit with the Java bridge decoder limit at `300`
- add a parser-level regression using the reproduced `commander.js` chain from `sonar-armor`
- keep the `commander.js` sample under `packages/analysis/tests/jsts/parsers/fixtures/ast`

## Root cause

Regenerating `packages/analysis/src/jsts/parsers/estree.js` with `protobufjs-cli >= 2.4.1` adds `_depth` checks in generated `encode` and `decode` methods. `protobufjs` defaults both recursion limits to `100`, while the reproduced `commander.js` protobuf AST reaches depth `148`, so AST serialization started failing with `max depth exceeded`.

The JS serializer now raises its local recursion limit to `300`, matching `AstProtoUtils` on the Java bridge side. That keeps both protobuf runtimes aligned while still leaving headroom above the reproduced depth.

## Testing

- `npm run bbf`
- `npx tsx --tsconfig packages/tsconfig.test.json --test packages/analysis/tests/jsts/parsers/ast.test.ts`
- `npx tsx --tsconfig packages/tsconfig.test.json --test packages/analysis/tests/analyzeProject-sonarqube.test.ts`

### DIFF
--- a/packages/analysis/src/jsts/parsers/ast.ts
+++ b/packages/analysis/src/jsts/parsers/ast.ts
@@ -15,6 +15,7 @@
  * along with this program; if not, see https://sonarsource.com/license/ssal/
  */
 import base64 from '@protobufjs/base64';
+import protobuf from 'protobufjs/minimal.js';
 import type { TSESTree } from '@typescript-eslint/utils';
 
 import { debug } from '../../../../shared/src/helpers/logging.js';
@@ -25,6 +26,7 @@ const NODE_TYPE = estree.Node;
 const NodeType = estree.NodeType;
 export { NodeType };
 const unsupportedNodeTypes = new Map<string, number>();
+const AST_PROTOBUF_RECURSION_LIMIT = 1000;
 
 export function serializeInProtobuf(
   ast: TSESTree.Program,
@@ -40,8 +42,10 @@ export function serializeInProtobuf(
           .join(', '),
     );
   }
-  const binaryArray = NODE_TYPE.encode(NODE_TYPE.create(protobufAST)).finish();
-  return base64.encode(binaryArray, 0, binaryArray.length);
+  return withRaisedProtobufRecursionLimit(() => {
+    const binaryArray = NODE_TYPE.encode(NODE_TYPE.create(protobufAST)).finish();
+    return base64.encode(binaryArray, 0, binaryArray.length);
+  });
 }
 
 /**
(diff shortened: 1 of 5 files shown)


packages/analysis/src/jsts/parsers/ast.ts
```diff
@@ -59,7 +63,26 @@ export function deserializeProtobuf(serialized: string): estree.Node {
...
+    AST_PROTOBUF_RECURSION_LIMIT,
+  );
+
+  protobuf.util.recursionLimit = targetLimit;
+  protobuf.Reader.recursionLimit = targetLimit;
+  try {
+    return callback();
+  } finally {
+    protobuf.util.recursionLimit = previousUtilLimit;
+    protobuf.Reader.recursionLimit = previousReaderLimit;
```
quentin-jaquier-sonarsource, line 84: Do we really need to bother to set the previous limit? I have the feeling that we can match the same behavior than what is done on the Java side: https://github.com/SonarSource/SonarJS/blob/989c6a70e0d334f912eea144d341e785d4022bc6/sonar-plugin/bridge/src/main/java/org/sonar/plugins/javascript/bridge/AstProtoUtils.java#L55

Always setting the limit to the value, and keeping it this way afterwards.

vdiez: Makes sense to compare them, but I think the scope is different here so I would keep the restore.

On the Java side, `setRecursionLimit(300)` is applied to the `CodedInputStream` instance used for that parse only.

On the JS side, `protobuf.util.recursionLimit` and `protobuf.Reader.recursionLimit` are global `protobufjs` values shared by the whole Node process. If we leave them at `300`, we also change the behavior of unrelated protobuf encode/decode paths in the same runtime.

So I aligned the numeric limit with the Java side, but kept the temporary override to scope the change to ESTree serialization/deserialization only.

quentin-jaquier-sonarsource: Thanks for clarifying. The pattern looked strange, but indeed, the API does not provide anything better.
So it works for me.

quentin-jaquier-sonarsource (approved):
I was able to confirm that it works on Armor ITs.
LGTM

PR #7125, opened 2026-05-29
JS-1597 add generated-source suppression API

francois-mora-sonarsource:
Part of JS-1597

## Summary
- add generated-source suppression metadata sourced from the RSPEC `editable-source` tag
- add a linter-level generated-source predicate hook and rule filter without introducing detector/store dependencies
- guard the generated-source predicate when `detectGeneratedCode` is disabled
- replace the brittle source-string generator test with a behavior-level helper test

## Testing
- npx tsx --tsconfig packages/tsconfig.test.json --test packages/analysis/tests/jsts/linter/index.test.ts packages/analysis/tests/generate-eslint-meta-source.test.ts
- npx prettier --check packages/analysis/src/jsts/linter/linter.ts packages/analysis/tests/jsts/linter/index.test.ts packages/analysis/tests/generate-eslint-meta-source.test.ts packages/analysis/src/jsts/rules/helpers/generated-source.ts tools/generate-eslint-meta.ts

### DIFF
--- a/packages/analysis/src/jsts/linter/filters/filter-generated-source.ts
+++ b/packages/analysis/src/jsts/linter/filters/filter-generated-source.ts
@@ -1,0 +1,21 @@
+/*
+ * SonarQube JavaScript Plugin
+ * Copyright (C) SonarSource Sàrl
+ * mailto:info AT sonarsource DOT com
+ *
+ * You can redistribute and/or modify this program under the terms of
+ * the Sonar Source-Available License Version 1, as published by SonarSource Sàrl.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the Sonar Source-Available License for more details.
+ *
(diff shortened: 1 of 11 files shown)


vdiez (approved):
Looks good to me.

This is a clean first split of `JS-1597`. It keeps generated-source handling at the linter boundary, adds RSPEC-driven `skipOnGeneratedSource` metadata, and makes the linter cache key distinguish generated vs regular contexts, so rule activation state should not bleed across files.

I validated this in an isolated worktree with `npm ci`, `npm run bbf`, and the focused tests for the touched behavior:
- `packages/analysis/tests/jsts/linter/index.test.ts`
- `packages/analysis/tests/jsts/linter/filter-generated-source.test.ts`
- `packages/analysis/tests/generate-eslint-meta-source.test.ts`

The guard to avoid calling `isGeneratedSourceFile` when `detectGeneratedCode` is disabled is also a good detail. That keeps this slice cheap and self-contained until the detector/store side lands.

Non-blocking follow-up for the next PRs: once a real detector is wired, I would keep it behind the analyzer stores/caches and avoid ad hoc filesystem probing. It may also be worth adding one end-to-end test that proves a generated file really drops `skipOnGeneratedSource` rules while an otherwise equivalent regular file keeps them.

PR #7126, opened 2026-05-29
JS-1803 Fix S6478 false positives for react-table headers

francois-mora-sonarsource:
## Summary

Fixes a React false positive in S6478 by treating react-table column `Header` functions as render callbacks, not nested component definitions.

The exemption is scoped to column-like objects containing `Header` plus `accessor`, `columns`, or `id`, and a regression test keeps generic `Header` renderer objects noncompliant.

## Validation

- `./node_modules/.bin/tsx --tsconfig packages/tsconfig.test.json --test packages/analysis/src/jsts/rules/S6478/unit.test.ts`
- `git diff --check`

`npm run bridge:compile` was attempted, but this fresh worktree lacks the ignored generated rule metadata/index artifacts expected by the command, so it fails before branch-specific checking.

### DIFF
--- a/packages/analysis/src/jsts/rules/S6478/decorator.ts
+++ b/packages/analysis/src/jsts/rules/S6478/decorator.ts
@@ -37,6 +37,8 @@ const REACT_INTL_COMPONENTS = new Set([
   '$t', // common alias
 ]);
 
+const REACT_TABLE_COLUMN_PROPERTIES = new Set(['accessor', 'columns', 'id']);
+
 /**
  * Checks if a JSXExpressionContainer is the `values` attribute of a react-intl component.
  */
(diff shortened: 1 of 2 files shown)


francois-mora-sonarsource:
### Rule Profile

| Field | Value |
|-------|-------|
| Rule | `S6478` - React components should not be nested |
| Severity / type | Major CODE_SMELL |
| Sonar Way | Yes |
| Scope | Main |

```jsx
function Parent() {
  function Header() {
    return <div />;
  }

  return <Header />;
}
```

### False Positive Pattern

The false positives happen in `react-table` v7 column definitions where the `Header` property is a render callback. S6478 normally treats functions returning JSX inside another component as nested component definitions, but `react-table` consumes `Header` as part of its column configuration API, not as a React component declared and rendered by the parent component.

```jsx
function Parent() {
  const columns = [
    {
      Header: () => <CustomHeader />,
      accessor: 'name',
    },
  ];

  useTable({ columns });
  return <Table columns={columns} />;
}
```

### False Negative Risk

The risk is bounded because the fix does not suppress all nested JSX-returning functions or all object properties named like components. It suppresses only static `Header` property values inside objects that look like `react-table` column definitions, identified by column-shape properties such as `accessor`, `id`, or nested `columns`. Computed properties and ordinary nested component definitions are still reported.

### Fix Summary

The implementation recognizes `react-table` column header renderers before delegating to the underlying nested-component rule. When a function is used as a column `Header` renderer, S6478 skips the report because that callback is a library configuration hook rather than an unstable nested React component.

- Detects static `Header` function values in column-like object expressions.
- Requires a `react-table` column shape, such as `accessor`, `id`, or nested `columns`.
- Keeps computed `Header` properties and non-column object shapes reportable.
- Adds regression coverage for valid react-table headers and invalid computed-header cases.

francois-mora-sonarsource:
Closed as this is covered by a change that allows users to configure the rule to ignore these type of issues.

PR #7127, opened 2026-05-29
JS-1802 Fix S3801 false positives for React effect cleanups

francois-mora-sonarsource:
## Summary

Fixes a React false positive in S3801 by skipping callbacks passed to React effect hooks. Effect callbacks may return a cleanup function or return nothing, so an early `return;` mixed with `return () => cleanup()` is valid React code rather than inconsistent API behavior.

The exemption covers `useEffect`, `useLayoutEffect`, and `useInsertionEffect` imported from `react`, including aliased named imports and `React.useEffect`-style calls.

## Validation

- `./node_modules/.bin/tsx --tsconfig packages/tsconfig.test.json --test packages/analysis/src/jsts/rules/S3801/unit.test.ts`
- `git diff --check`

`npm run bridge:compile` was attempted, but this fresh worktree lacks the ignored generated rule metadata/index artifacts expected by the command, so it fails before branch-specific checking.

### DIFF
--- a/its/ruling/src/test/expected/ant-design/typescript-S3801.json
+++ b/its/ruling/src/test/expected/ant-design/typescript-S3801.json
@@ -5,9 +5,6 @@
 "ant-design:components/button/button.tsx": [
 32
 ],
-"ant-design:components/checkbox/Checkbox.tsx": [
-77
-],
 "ant-design:components/drawer/index.tsx": [
 199
 ],
--- a/its/ruling/src/test/expected/eigen/typescript-S3801.json
+++ b/its/ruling/src/test/expected/eigen/typescript-S3801.json
@@ -19,8 +19,7 @@
 97
 ],
 "eigen:src/app/Components/ArtsyWebView.tsx": [
-101,
-298
+101
 ],
 "eigen:src/app/Components/ArtworkFilter/FilterModal.tests.tsx": [
 520
@@ -103,9 +102,6 @@
 "eigen:src/app/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tests.tsx": [
 56,
 77
-],
-"eigen:src/app/Scenes/Artwork/Components/CommercialButtons/InquiryModal.tsx": [
-177
 ],
 "eigen:src/app/Scenes/Artwork/Components/ImageCarousel/FullScreen/useDoublePressCallback.ts": [
 12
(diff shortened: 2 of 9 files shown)


francois-mora-sonarsource:
### Rule Profile

| Field | Value |
|-------|-------|
| Rule | `S3801` - Functions should use `return` consistently |
| Severity / type | Major CODE_SMELL |
| Sonar Way | No |
| Scope | Main |

```javascript
function getValue(condition) {
  if (condition) {
    return 42;
  }
  return;
}
```

### False Positive Pattern

The false positives happen in React effect callbacks that return a cleanup function on one path and return nothing on another path. S3801 normally treats this as inconsistent API behavior, but React gives effect callback returns a special meaning: returning a function registers cleanup, while returning nothing means there is no cleanup for that effect run.

```javascript
useEffect(() => {
  if (!enabled) {
    return;
  }

  const subscription = subscribe();
  return () => subscription.unsubscribe();
}, [enabled]);
```

### False Negative Risk

The risk is bounded because the fix does not suppress all functions with mixed returns, nor all callbacks. It suppresses only the first argument passed to `useEffect`, `useLayoutEffect`, or `useInsertionEffect` when those hooks are imported from `react`, including named aliases and `React.useEffect`-style calls. Normal functions, callbacks passed to other APIs, and untracked custom wrappers are still checked by S3801.

### Fix Summary

The implementation tracks React imports, recognizes React effect hook calls, and skips S3801 reporting for the effect callback itself. This matches the React contract where cleanup returns are optional and should not be interpreted as inconsistent function return behavior.

- Tracks named React effect hook imports, aliased imports, default imports, and namespace imports.
- Recognizes direct hook calls and `React.useEffect`-style member calls.
- Adds regression tests for `useEffect`, `useLayoutEffect`, `useInsertionEffect`, `React.useEffect`, and aliased `useEffect` imports.
- Updates ruling snapshots for the 20 React effect false positives no longer reported.

nathsou:
For TypeScript, maybe this FP could be more general, I think it applies to any function that returns a union of (either `undefined` or `void`) and another type.

See the signature for the useEffect callback argument: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L1665

Here the return type is `void | Destructor`, so an empty `return;` or `return undefined;` should be considered valid even if another return statement returns something else (granted we have type information)

francois-mora-sonarsource:
Replaced by a FP ticket for Nigel: https://sonarsource.atlassian.net/browse/JS-1805

PR #7134, opened 2026-05-29
JS-1790 Restrict S5725 to statically versioned URLs

pierre-loup-tristant-sonarsource:
## Summary

- Rule S5725 now only raises issues for remote scripts whose URL contains a recognizable version path segment (`/3.7.1/`, `/jquery@3.7.1/`, `/bootstrap@5.3.0`, etc.), eliminating false positives on dynamic CDN URLs (Google Tag Manager, Stripe, etc.)
- Checks both `integrity` and `crossorigin="anonymous"` attributes with distinct messages for each missing case
- Conservatively suppresses the `crossorigin` check when the assigned value cannot be statically resolved, avoiding false positives

## Test plan

- [ ] Unit tests updated with versioned/non-versioned URL cases, unresolved `crossOrigin`, and all three message variants
- [ ] Ruling ITs pass (59/59): `custom-jsts` and `ant-design` expected files updated

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


vdiez (approved):
Looks good overall.

Non-blocking follow-ups if we want to tighten a couple of edge cases later:
- `crossOrigin` is currently treated order-insensitively, so `script.crossOrigin = anonymous; script.crossOrigin = use-credentials;` no longer raises even though the final value is not compliant.
- The new version regexes do not match prerelease semver URLs such as `https://cdn.jsdelivr.net/npm/react@19.0.0-rc.1/umd/react.production.min.js`, so those pinned URLs are currently skipped.

I would keep UUID-shaped URLs out of the heuristic for now unless we can point to a concrete JS/CDN pattern we want to support. In the current ruling corpus the UUID-looking URLs I checked are mostly opaque asset paths rather than versioned script artifacts, so matching them generically looks likely to reintroduce false positives.

PR #7135, opened 2026-05-29
JS-1789 - S4782 resolve type alias

guillemsarda:
Part of JS-1789

----
## Summary by Gitar

- **Rule logic enhancements:**
  - Updated `S4782` to resolve type aliases using `ts.TypeChecker`, enabling detection of `undefined` within aliased types.
  - Added `classifyTypesByOrigin` helper to differentiate between internal and external type declarations, preventing false positives on library-defined types (e.g., `React.ReactNode`).
  - Implemented semantic analysis that respects `strictNullChecks` for accurate `undefined` tracking.
- **Rule metadata updates:**
  - Added `S1244` configuration to `Sonar way` and `Sonar agentic AI` profiles.
  - Updated `S2245` metadata, changing type from `SECURITY_HOTSPOT` to `VULNERABILITY` and adjusting impact levels.
- **Test suite updates:**
  - Added new `external-library.test.ts` to verify type origin classification logic.
  - Added extensive test fixtures and cases for aliased, recursive, and nested types in `unit.test.ts`.
- **Baseline updates:**
  - Updated expected ruling results for `ant-design`, `eigen`, `vitest`, and `vuetify` to account for improved resolution logic.


packages/analysis/tests/jsts/tools/testers/fixtures/tsconfig.json
```diff
@@ -3,7 +3,7 @@
     "allowJs": true,
     "noImplicitAny": true,
     "strict": false,
-    "strictNullChecks": false,
+    "strictNullChecks": true,
```
zglicz: so this is for all unittests, interesting. If all the unittests stay green, then I guess

packages/analysis/src/jsts/rules/S4782/rule.ts
```diff
@@ -69,13 +72,38 @@ export const rule: Rule.RuleModule = {
   },
 };
 
-function getUndefinedTypeAnnotation(tsTypeAnnotation?: TSESTree.TSTypeAnnotation) {
-  if (tsTypeAnnotation?.typeAnnotation.type === 'TSUnionType') {
+function getUndefinedTypeAnnotation(
+  tsTypeAnnotation: TSESTree.TSTypeAnnotation | undefined,
+  services: RequiredParserServices,
+) {
```
zglicz: These method names are atrocious. Perhaps think if they can be made a bit more readable.

zglicz (approved):

packages/analysis/tests/jsts/tools/testers/fixtures/tsconfig.json
guillemsarda: I've rolled it back because I added two specific `tsconfig` for the unit test of the rule.

PR #7156, opened 2026-06-02
JS-1810 S6281: Adjust detection logic for AWS CDK BlockPublicAccess changes

daniel-teuchert-sonarsource:
## Summary

The behavior of AWS CDK was changed with [aws-cdk/pull/33702](https://github.com/aws/aws-cdk/pull/33702):

- Omitting `blockPublicAccess` now defaults to fully blocked (secure) — no longer an issue
- `BLOCK_ACLS` now sets all four attributes to `true` (secure) — no longer an issue
- `BLOCK_ACLS_ONLY` is the new insecure preset — now raises an issue
- `new s3.BlockPublicAccess()` with no args defaults to all blocked (secure) — no longer an issue

Messages are updated to the vulnerability format consistent with other CDK rules.

This mirrors the equivalent change in sonar-python-enterprise (PR #1158).

## Test plan

- [ ] Run `npm ci` then `npx tsx --test packages/analysis/src/jsts/rules/S6281/**/*.test.ts`


zglicz (changes requested):
Could you update the ruling, so that at least once, this issue is triggered? Seeing it as it is already quite niche and now, seems like none of the custom-jsts sources trigger it.

daniel-teuchert-sonarsource:
> Could you update the ruling, so that at least once, this issue is triggered? Seeing it as it is already quite niche and now, seems like none of the custom-jsts sources trigger it.

Good point!
I added some tests and adjusted the existing code that is now compliant.

zglicz (approved):

PR #7158, opened 2026-06-02
JS-1650 Move WebSensor to project-level analysis

vdiez:
Part of JS-1650

## Summary
- migrate `WebSensor` to project-level analysis
- collect per-module web sensor configuration so module-relative `tsconfig` paths still resolve correctly
- preserve module-based ESLint report imports and cover the behavior with unit and multi-module integration tests

## Verification
- `mvn -pl sonar-plugin/sonar-javascript-plugin -am -DskipTests clean package`
- `mvn -pl its/plugin/tests -am "-DskipTests=false" "-Dtest=CoverageTest#LCOV_report_outside_module" "-Dsurefire.failIfNoSpecifiedTests=false" test`
- `mvn -pl sonar-plugin/sonar-javascript-plugin -am "-Dtest=WebSensorTest,EslintReportImporterTest" "-Dsurefire.failIfNoSpecifiedTests=false" test`
- `mvn -pl its/plugin/fast-tests -am "-DskipTests=false" "-Dtest=WebSensorMultiModuleTest" "-Dsurefire.failIfNoSpecifiedTests=false" test`

## Note
- `sonar.javascript.lcov.reportPaths` is intentionally not part of the new module property collection.
- JS-1650 only collects module-scoped properties now needed by the project-level `WebSensor`: `tsconfig` paths and ESLint report paths.
- LCOV import still runs in `CoverageSensor`, which remains module-scoped and reads `sonar.javascript.lcov.reportPaths` directly from its own `SensorContext`.
- If LCOV import moves to project-level later, it will need the same kind of module property aggregation.

### DIFF
--- a/its/plugin/fast-tests/src/test/java/com/sonar/javascript/it/plugin/SonarScannerIntegrationHelper.java
+++ b/its/plugin/fast-tests/src/test/java/com/sonar/javascript/it/plugin/SonarScannerIntegrationHelper.java
@@ -38,7 +38,7 @@ import org.sonar.plugins.javascript.TypeScriptLanguage;
 
 public final class SonarScannerIntegrationHelper {
 
-  static final String LITS_VERSION = "0.11.0.2659";
+  static final String LITS_VERSION = "0.12.0.5847";
 
   /**
    * Get the engine version based on the sonar.runtimeVersion system property.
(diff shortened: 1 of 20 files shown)


zglicz (approved):
Looks good overall.

One non-blocking thing I’d still consider augmenting is the new integration test. Right now it covers the case where the relevant JS/TS properties are defined only in sub-modules, which is the critical regression path.

If we want a bit more confidence, it would be useful to also cover a mixed root + sub-module setup for `sonar.typescript.tsconfigPaths` and `sonar.eslint.reportPaths`, just to ensure project-level values and collected module-level values coexist correctly end to end.

PR #7195, opened 2026-06-05
JS-1843 Add invalid declaration position rule

erwan-leforestier-sonarsource:
## Summary
Add CSS rule S8767 for stylelint's `no-invalid-position-declaration`.
RSPEC PR: https://github.com/SonarSource/rspec/pull/7044

## Changes
- add S8767 to the CSS rule metadata, exposing the `ignoreAtRules` parameter

## Functional Validation
Artifact: [JS-1843-fv.zip](https://github.com/user-attachments/files/28801580/JS-1843-fv.zip)

    ./run.sh


vdiez (changes requested):
I have one blocking test gap here.

- The rspec highlights top-level invalid declarations as a primary noncompliant case, but the current SonarJS tests only cover the nested `@media` variant. Please add a top-level invalid testcase too.

vdiez (approved):
Rechecked the latest commit. The tests now cover the top-level invalid declaration case and I do not have remaining blocking issues.

PR #7196, opened 2026-06-05
JS-1835 Add selector-type-case rule

erwan-leforestier-sonarsource:
## Summary
Adds CSS support for Stylelint `selector-type-case` as Sonar rule S8763 with a fixed `lower` primary option.
RSPEC PR: https://github.com/SonarSource/rspec/pull/7026

## Changes
- add S8763 to the CSS rule metadata with a fixed lowercase primary option
- extend CSS configuration generation so rules can emit fixed non-boolean primary options without exposing a Sonar parameter
- cover the new rule in the linter, gRPC, and generated Java CSS rule tests

## Functional Validation
Artifact: `JS-1835-fv.zip`

    ./run.sh


erwan-leforestier-sonarsource:
Closing without merging. On reflection this rule is purely stylistic — CSS type selectors are case-insensitive, so the before/after states behave identically. That doesn't meet the "not subjective" bar set by the parent epic JS-1826. Branch left in place for now; ticket JS-1835 closed as Won't Do.

PR #7198, opened 2026-06-05
JS-1828 Add stylelint rule "at-rule-descriptor-no-unknown"

erwan-leforestier-sonarsource:
## Summary
Add CSS rule S8770 for stylelint's `at-rule-descriptor-no-unknown`.
RSPEC PR: https://github.com/SonarSource/rspec/pull/7047

## Changes
- add S8770 to the CSS rule metadata, restricted to plain CSS via `css-only-rules.ts`

## Functional Validation
Artifact: [JS-1828-fv.zip](https://github.com/user-attachments/files/28793121/JS-1828-fv.zip)

    ./run.sh

----
## Summary by Gitar

- **Rule management:**
  - Integrated S8770 into CSS metadata, `css-only-rules.ts`, and the Sonar way profile.
  - Updated `RULE_CONFIG_PATTERNS.md` to reflect the total CSS rule count increase to 30.
- **Infrastructure updates:**
  - Refactored `transform.ts` to improve type safety and add filtering logic for non-CSS embedded blocks.
  - Added a new quick-fix decoration mechanism for S7743 in `decorator.ts`.
- **Rule metadata:**
  - Updated `quickfix` status for S7743 and multiple other JavaScript rules (S2255, S4784, etc.) to `covered`.
- **Testing:**
  - Added comprehensive test suite for `at-rule-descriptor-no-unknown` and unit tests for the S7743 quick-fix.
  - Verified proper mapping of S8770 to `at-rule-descriptor-no-unknown` in `transformers.test.ts`.


vdiez (approved):
Apart from the known metadata-sync gap with the unmerged rspec changes, I don't have blocking issues here.

vdiez (approved):
Rechecked the latest head and I do not have new concerns.

vdiez (changes requested):
The new `S7743` quick-fix still looks unsafe in one edge case.

When the concise arrow body is wrapped in more than one pair of parentheses, the decorator replaces only the innermost pair and can generate invalid syntax. For example, `(bar => ((x)))(getBar())` would become `(bar => ({ return x; }))(getBar())`.

I do not have new concerns on the CSS-rule changes in this update, but this added quick-fix should either handle that case correctly or skip the suggestion when extra wrapping parentheses are present.

erwan-leforestier-sonarsource:
Thanks for your review @vdiez I've pushed one commit to address your comment.

vdiez:
The latest head is still failing CI on the new `S7743` quick-fix code, so this is not ready to merge yet.

The current run fails in `Build SonarJS on Linux`, `Build SonarJS on Windows`, `Unit tests JavaScript/TypeScript`, `Unit tests JavaScript on Windows`, and `Knip`. The underlying blocker is the TypeScript compilation error in `packages/analysis/src/jsts/rules/S7743/decorator.ts`:

- `TS7022`: `nextOpeningParenthesis` implicitly has type `any`
- `TS2345`: `Token | null` is passed where `null` is not allowed
- `TS2322`: `Token | null` is assigned where `Token` is expected

So the original review comment looks addressed, but the current implementation still needs a typing fix before I can approve it.

erwan-leforestier-sonarsource:
Thanks @vdiez — pushed a typing fix in 2bad80478. Replaced the single-arg type predicate `isParenthesisPair` with `getParenthesisPair` that returns the validated pair, so both opening and closing tokens are narrowed to `Token` together (the previous predicate left `closingParenthesis` as `Token | null`, which is what cascaded into TS7022/TS2345/TS2322 in the loop). `tsc --noEmit -p packages/tsconfig.app.json` is now clean locally and the S7743 unit test still passes.

vdiez (approved):
Rechecked the latest head. The S7743 quick-fix now handles the nested-parentheses case correctly, the typing issue is fixed, and I do not have remaining blocking concerns here.

PR #7202, opened 2026-06-05
JS-1827 Add annotation-no-unknown CSS rule

erwan-leforestier-sonarsource:
## Summary
Add CSS rule S8757 for stylelint's `annotation-no-unknown`.
RSPEC PR: https://github.com/SonarSource/rspec/pull/7019

## Changes
- add S8757 to the CSS rule metadata, exposing the `ignoreAnnotations` parameter
- add the `sonar/annotation-no-unknown` wrapper to suppress Sass `!default` and `!global` warnings
- cover the new rule in unit tests, the CSS linter, and gRPC transformer tests

## Functional Validation
Artifact: [JS-1827-fv.zip](https://github.com/user-attachments/files/28784679/JS-1827-fv.zip)

    ./run.sh


vdiez (changes requested):
This should use the standard metadata-only integration instead of introducing a `sonar/annotation-no-unknown` wrapper.

For imported CSS rules, the common pattern in this repo is to map the Sonar rule directly to the upstream Stylelint key in `packages/analysis/src/css/rules/metadata.ts` and let the existing CSS pipeline handle configuration and reverse-mapping. `ignoreAnnotations` already fits that model via `listParam`, and HTML/Vue blocks are already handled by the Stylelint config with `postcss-html`.

With the current wrapper we are forking Stylelint behavior for this rule and adding maintenance cost without a clear need. It also changes semantics in Sass by suppressing `!default` / `!global` warnings in the wrapper, which is broader than a pure import and already creates a false negative for invalid declarations such as `a { color: red !default; }` in SCSS.

Please simplify this to the metadata approach used by most imported CSS rules:

- map `S8757` directly to upstream `annotation-no-unknown`
- keep `ignoreAnnotations` as metadata-driven configuration
- remove the custom `packages/analysis/src/css/rules/S8757/rule.ts` wrapper and its registration in `packages/analysis/src/css/rules/index.ts`

If we really want to diverge from upstream Stylelint behavior for Sass, that should be justified explicitly as a product decision. The default here should be parity and the simpler integration path.

erwan-leforestier-sonarsource:
@vdiez This one is special because I unfortunately, stylelint is raising FP on sass files due to `!default` and `!global` being valid annotations for sass (work for both `.sass` and `.scss` files).
That's why I've overridden the rule to silence the rule on sass code. I'm afraid that we would have a negative impact on customers if we rely only on stylelint rule implementation. Wdyt ?

vdiez (changes requested):
I re-checked this on a clean checkout of `127c37b`.

I agree with the motivation here: upstream `annotation-no-unknown` does report `!default` / `!global` in `.scss`, `.sass`, and `<style lang="scss">`, so a pure metadata-only import would create false positives.

The current wrapper still needs changes though. In `packages/analysis/src/css/rules/S8757/rule.ts`, `removeSassAnnotationWarnings()` filters by warning text only and applies to every warning emitted in `.scss/.sass`. That also suppresses invalid declaration-level annotations such as `a { color: red !default; }` in SCSS, which should still be reported. It also couples the behavior to Stylelint's current message wording.

Please narrow the exemption to actual Sass-variable cases using AST context, or make the rule CSS-only if the product decision is to skip annotation analysis on preprocessors entirely.

vdiez (approved):
My earlier concerns are addressed. Apart from the known metadata-sync gap with the unmerged rspec changes, I don't have blocking issues here.

PR #7208, opened 2026-06-05
JS-1837 Add nesting-selector-no-missing-scoping-root CSS rule

erwan-leforestier-sonarsource:
## Summary
Add CSS rule S8776 for stylelint's `nesting-selector-no-missing-scoping-root`.
RSPEC PR: https://github.com/SonarSource/rspec/pull/7053

## Changes
- add S8776 to the CSS rule metadata, exposing the `ignoreAtRules` parameter

## Functional Validation
Artifact: [JS-1837-fv.zip](https://github.com/user-attachments/files/28799880/JS-1837-fv.zip)

    ./run.sh


vdiez (changes requested):
I have one blocking coverage issue here.

- `ignoreAtRules` is exposed in metadata, but the current coverage never exercises that mapping end to end. Please add a transformer/server-style test with a non-empty `ignoreAtRules` value.

erwan-leforestier-sonarsource:
Thanks for the review @vdiez, I've pushed one commit to address your comment.

vdiez (approved):
Rechecked the latest changes. The end-to-end coverage for a non-empty `ignoreAtRules` value is now in place, so my previous blocker is addressed.

PR #7209, opened 2026-06-05
JS-1830 Add CSS at-rule prelude validation

erwan-leforestier-sonarsource:
## Summary
Add CSS rule S8777 for stylelint's `at-rule-prelude-no-invalid`.
RSPEC PR: https://github.com/SonarSource/rspec/pull/7054

## Changes
- add S8777 to the CSS rule metadata, restricted to plain CSS via `css-only-rules.ts` and exposing the `ignoreAtRules` parameter

## Functional Validation
Artifact: [JS-1830-fv.zip](https://github.com/user-attachments/files/28794977/JS-1830-fv.zip)

    ./run.sh


vdiez (changes requested):
I have one blocking issue here.

- `packages/analysis/src/css/rules/metadata.ts` documents `ignoreAtRules` as plain at-rule names only, but stylelint accepts both strings and regular expressions for this option. Please align the SonarQube parameter description with the actual contract.

erwan-leforestier-sonarsource:
Thanks for the review @vdiez, I've pushed one comment to address your comment.

vdiez (approved):
Rechecked the latest changes. The `ignoreAtRules` parameter description is now aligned with the actual stylelint contract, so my previous blocker is addressed.

PR #7211, opened 2026-06-05
JS-1841 Add invalid @import position rule

erwan-leforestier-sonarsource:
## Summary
Add CSS rule S8778 for stylelint's `no-invalid-position-at-import-rule`.
RSPEC PR: https://github.com/SonarSource/rspec/pull/7055

## Changes
- add S8778 to the CSS rule metadata, exposing the `ignoreAtRules` parameter

## Functional Validation
Artifact: [JS-1841-fv.zip](https://github.com/user-attachments/files/28800487/JS-1841-fv.zip)

    ./run.sh

----
## Summary by Gitar

- **New CSS Rule:**
  - Implemented `S8778` (`no-invalid-position-at-import-rule`) to validate `@import` placement.
- **Embedded Block Handling:**
  - Added `isInNonCssEmbeddedBlock` to prevent CSS-only rules from flagging non-CSS languages in HTML/Vue files.
- **Testing and Infrastructure:**
  - Added unit tests for `S8778` and updated gRPC transformers to handle the `ignoreAtRules` parameter.
  - Updated `RULE_CONFIG_PATTERNS.md` to include `S8778` in documentation metrics.


vdiez (changes requested):
I have one blocking coverage issue here.

- `ignoreAtRules` is exposed in metadata, but the transformer coverage only checks the empty/default path. Please add a test with a non-empty `ignoreAtRules` value so the parameter mapping is covered end to end.

erwan-leforestier-sonarsource:
Thanks for your review @vdiez 🙏 I've pushed one commit to address your comment. Let me know if it's ok for you

vdiez (approved):
Rechecked the latest changes. The non-empty `ignoreAtRules` mapping is now covered end to end, and I do not have remaining blocking issues here.

PR #7213, opened 2026-06-05
JS-1597 narrow generated-source cache invalidation

francois-mora-sonarsource:
## What changed

This PR narrows generated-source cache reuse in `generatedSourceStore`.

It keeps the lightweight match-cache refresh for true source-scope changes and explicit request-file-set changes, but it now hard-invalidates the detector cache when project-file discovery inputs change.

It also updates the file-store documentation to explain the detector cache vs match cache split in terms of the corresponding Sonar properties.

## Why it changed

The previous reuse logic treated every analyzable-file configuration change as match-cache-only.
That was too broad for `sonar.javascript.exclusions` and `sonar.typescript.exclusions`, which are merged into `jsTsExclusions` and are applied during the filesystem walk.

That means they can change which helper files, especially `package.json`, are discovered. When that happened, `generatedSourceStore` could incorrectly reuse a stale detector cache built from incomplete project metadata.

## Impact

- Keeps the intended optimization for source-scope-only changes such as `sonar.exclusions`
- Fixes stale generated-source metadata when JS/TS exclusions change helper-file discovery
- Documents the real refresh boundaries for the generated-source file store

## Validation

- `tsx --tsconfig packages/tsconfig.test.json --test --test-reporter=spec --test-reporter-destination stdout packages/analysis/tests/jsts/project-metadata/generated-sources.test.ts`
- `tsx --tsconfig packages/tsconfig.test.json --test --test-reporter=spec --test-reporter-destination stdout packages/analysis/tests/file-stores.test.ts`

### DIFF
--- a/docs/file-stores.md
+++ b/docs/file-stores.md
@@ -195,34 +195,43 @@ It maintains two caches:
 - a **detector cache** with generated-file to family mappings, config paths used by generated-source detectors, and watched output paths
 - a **match cache** with the subset of detector matches visible to the current analysis request
 
-It depends on a mix of the other stores:
+The most useful way to reason about this store is by the corresponding Sonar property families.
 
-- **project helper files** because the detector cache derives metadata from files such as `package.json`
-- **source-file selection** because the match cache should only surface analyzable source files to the linter
+- **Source-scope properties**: `sonar.sources`, `sonar.tests`, `sonar.inclusions`, `sonar.exclusions`, `sonar.test.inclusions`, and `sonar.test.exclusions`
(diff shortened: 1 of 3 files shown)


vdiez (approved):
Looks good to me. The cache split in `generatedSourceStore` improves reuse rather than just refactoring: source-scope changes now refresh only the filtered match cache, while detector metadata is retained until helper-file discovery inputs actually change.

One non-blocking follow-up: the docs drift slightly now. `docs/file-stores.md` describes the new detector-cache vs match-cache split, but `docs/generated-sources.md` still says analyzable-file-selection changes trigger full recomputation. It would be good to align that document with the new behavior in a follow-up.

PR #7214, opened 2026-06-05
JS-1597 simplify configuration parsing

francois-mora-sonarsource:
Part of JS-1597

## What
- simplify `createConfiguration` by extracting helpers for required and optional values
- keep the shared JS/TS extension lists in `file-kinds.ts` and reuse them when building the configuration defaults
- retain the existing JS default shape by building it from the shared JS list plus `.vue`

## Why
- reduce duplication in configuration parsing without changing behavior
- avoid duplicating the shared JS/TS extension lists across `configuration.ts` and `file-kinds.ts`
- keep the existing distinction that `.vue` is part of the configuration defaults but not part of JS/TS code-file detection

## Impact
- no intended functional change; `.vue` handling and the default suffixes stay the same
- the shared JS/TS extension lists remain defined in one place, which reduces drift risk

## Validation
- `npm run grpc:generate-proto && tsx --tsconfig packages/tsconfig.test.json --test --test-concurrency=4 --test-reporter=spec --test-reporter-destination stdout packages/analysis/tests/common/configuration.test.ts packages/analysis/tests/common/file-kinds.test.ts packages/analysis/tests/common/filter-path.test.ts`

### DIFF
--- a/packages/analysis/src/common/configuration.ts
+++ b/packages/analysis/src/common/configuration.ts
@@ -29,7 +29,6 @@ import {
   isStringArray,
   isObject,
 } from '../../../shared/src/helpers/sanitize.js';
-import { JAVASCRIPT_CODE_FILE_EXTENSIONS, TYPESCRIPT_CODE_FILE_EXTENSIONS } from './file-kinds.js';
 
 /**
  * A discriminator between JavaScript and TypeScript languages. This is used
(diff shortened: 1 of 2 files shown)


packages/analysis/src/common/configuration.ts
```diff
@@ -128,8 +127,8 @@ export type ConfigurationInput = {
 // Patterns enforced to be ignored no matter what the user configures on sonar.properties
 const IGNORED_PATTERNS = ['.scannerwork'];
 
-const DEFAULT_JS_EXTENSIONS = [...JAVASCRIPT_CODE_FILE_EXTENSIONS, '.vue'];
-const DEFAULT_TS_EXTENSIONS = TYPESCRIPT_CODE_FILE_EXTENSIONS.slice();
+const DEFAULT_JS_EXTENSIONS = ['.js', '.mjs', '.cjs', '.jsx', '.vue'];
```
vdiez: Could we keep the shared JS/TS defaults in one place and only add `.vue` here? The `.vue` distinction makes sense, but after this change `configuration.ts` and `file-kinds.ts` both define the overlapping extension lists, which makes drift easy the next time one of them changes.

francois-mora-sonarsource: Fixed. `file-kinds.ts` owns the shared JS/TS extension lists again, and `configuration.ts` now only builds the JS defaults from that shared list plus `.vue`. That keeps the `.vue` distinction while removing the duplicated base lists and the drift risk.

vdiez (approved):
Reviewed the latest changes. The shared JS/TS extension defaults are back to a single source of truth in file-kinds.ts, with configuration.ts only layering the .vue-specific config default on top. I also validated the updated head in a clean worktree with npm ci, npm run bbf, and the targeted configuration/file-kinds/filter-path tests.

PR #7217, opened 2026-06-05
JS-1597 wire detectGeneratedCode configuration

francois-mora-sonarsource:
## Summary
- wire the `detectGeneratedCode` scanner setting through the Java plugin, bridge protobuf payload, and Node analysis configuration
- default generated-code detection to enabled while allowing it to be explicitly disabled
- add coverage for the configuration, protobuf normalization, and WebSensor/bridge wiring paths

## Validation
- `npx tsx --tsconfig packages/tsconfig.test.json --test packages/analysis/tests/common/configuration.test.ts packages/grpc/tests/analyze-project-normalize.test.ts`
- `mvn -pl sonar-plugin/bridge,sonar-plugin/sonar-javascript-plugin -Dtest=AnalyzeProjectMessagesTest,WebSensorTest test`

### DIFF
--- a/packages/analysis/src/analyzeProject.ts
+++ b/packages/analysis/src/analyzeProject.ts
@@ -88,6 +88,7 @@ export async function analyzeProject(
     globals,
     bundles,
     baseDir,
+    detectGeneratedCode: configuration.detectGeneratedCode,
     isGeneratedSourceFile: filePath => generatedSourceStore.getFamily(filePath) !== undefined,
     rulesWorkdir,
     testFileExtensions,
--- a/packages/analysis/src/common/configuration.ts
+++ b/packages/analysis/src/common/configuration.ts
@@ -82,6 +82,7 @@ export type Configuration = {
   testInclusions: Minimatch[] /* sonar.test.inclusions - WILDCARD to narrow down sonar.tests */;
   testExclusions: Minimatch[] /* sonar.test.exclusions - WILDCARD to narrow down sonar.tests */;
   detectBundles: boolean /* sonar.javascript.detectBundles - whether files looking like bundled code should be ignored */;
+  detectGeneratedCode: boolean /* sonar.javascript.detectGeneratedCode - whether generated-source detection should affect analysis behavior */;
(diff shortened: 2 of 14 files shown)


vdiez (changes requested):
The wiring is still incomplete. With `sonar.javascript.detectGeneratedCode=false`, the flag currently disables only the linter-side generated-source check in `analyzeProject.ts`, but `generatedSourceStore` is still initialized and still derives/caches metadata during `initFileStores()` / `GeneratedSourceStore.postProcess()`.

I verified this on the PR branch with the graphql-codegen fixture: after `initFileStores(createConfiguration({ baseDir, detectGeneratedCode: false }))`, `generatedSourceStore.getFamily(...)` still returns `@graphql-codegen/cli`. If the property is set to `false`, the generated-source store should not run at all.

Please gate the store off entirely when the flag is disabled, and add coverage for that path. The current tests only prove that the linter callback is skipped, not that detection/caching work is skipped.

francois-mora-sonarsource:
Addressed the generated-source wiring gap. When sonar.javascript.detectGeneratedCode=false, initFileStores() now clears and skips generatedSourceStore entirely instead of still deriving and caching metadata. Added a regression test covering the disabled path and the re-enable path in generated-sources.test.ts.

Local verification:
- generated-source store regression suite
- analyze-project normalize tests
- AnalyzeProjectMessagesTest
- WebSensorTest
- SonarLintIntegrationTest#should_report_issues

vdiez (approved):
﻿Approved.

The wiring gap is fixed now: when `sonar.javascript.detectGeneratedCode=false`, `initFileStores()` clears and skips `generatedSourceStore`, and the regression test covers both the disabled and re-enabled paths.

One non-blocking follow-up I'd still like to see: if generated-code detection is enabled and it actually matches files, please emit a single INFO log similar to bundle detection, mentioning `sonar.javascript.detectGeneratedCode=false` as the opt-out. Right now that behavior is silent, so users do not get an explicit explanation that generated-source detection kicked in.

PR #7241, opened 2026-06-09
JS-1879 Forbid rspec.sha on master branch

nathsou:
This PR adds a lightweight CI guard to prevent merging `rspec.sha` into `master`.

It introduces a small check for a root-level `rspec.sha` override and runs it only for `master`-targeted PRs, merge queue runs, and pushes to `master`, so temporary RSPEC pins can still be used during development but must be removed before merge.


vdiez:
why is not everything self-contained in a single tools script?

vdiez:
@nathsou I may be missing some repo-specific branch protection, but with `promote` as the required status check I don't think this PR makes `rspec.sha` a hard merge blocker yet. `check_rspec_sha` is a separate job, and `promote` doesn't list it in `needs`, so `promote` can still run and succeed even if `Reject root rspec.sha on master` fails.

What's the expected behaviour of this PR?

1. Just surface a visible failing job on `master`-targeted PRs when a root `rspec.sha` is present, or
2. Make merge impossible whenever a root `rspec.sha` is present?

If it's meant to enforce the second case, I think `promote` needs to depend on `check_rspec_sha`, or this new check itself needs to be required.

nathsou:
Closing as this PR is stale and I won't be able to work on it in the following weeks

nathsou:
Re-opened as https://github.com/SonarSource/SonarJS/pull/7615

PR #7252, opened 2026-06-10
JS-1877: Port CleartextProtocolFilter safe-URL logic to S5332

daniel-teuchert-sonarsource:
## Summary

- Replaces the ad-hoc `LOOPBACK_PATTERN`, `EXCEPTION_FULL_HOSTS`, and `EXCEPTION_TOP_HOSTS` constants in `rule.lib.ts` with a TypeScript port of `CleartextProtocolFilter` from `sonar-analyzer-commons`
- Adds coverage for cloud IMDS endpoints (AWS/Azure/GCP/Alibaba link-local range, named endpoints), Docker-internal hostnames, and Kubernetes `.svc.cluster.local` DNS
- Expands namespace URI authorities to include Spring, Maven, HL7, Schema.org, OGP, and more
- Fixes the `DOCUMENTATION_HOSTS` pattern to align with RFC 6761 (removes `test.com` which is a real public domain; adds `example.net` and `.localhost`)
- Adds a lenient fallback for URLs that fail strict `URL` parsing (e.g. template placeholders like `http://localhost:${port}`)
- Updates unit tests to cover all new categories and adds prefix-spoofing attack cases

Note: `CleartextProtocolFilter` is a Java class and cannot be called directly from sonar-js's Node.js bridge — this PR ports the logic to TypeScript so the filtering behaviour is consistent with the centralised implementation.

## Test plan

- [ ] `npx tsx --test packages/analysis/src/jsts/rules/S5332/**/*.test.ts` passes

### DIFF
--- a/packages/analysis/src/jsts/rules/S5332/rule.lib.ts
+++ b/packages/analysis/src/jsts/rules/S5332/rule.lib.ts
@@ -64,7 +64,7 @@ const DOCUMENTATION_HOSTS =
 // Lenient authority extractor for URLs that fail strict URL parsing (e.g. template placeholders, underscores in hostnames).
 // Derived from CLEARTEXT_PROTOCOL_ALTERNATIVES — no manual update needed when adding protocols.
 const LENIENT_AUTHORITY = new RegExp(
-  `^(?:${Object.keys(CLEARTEXT_PROTOCOL_ALTERNATIVES).join('|')}):\\/\\/(?:[^@\\s/?#]+@)?([^\\s/?#]+)`,
+  String.raw`^(?:${Object.keys(CLEARTEXT_PROTOCOL_ALTERNATIVES).join('|')}):\/\/(?:[^@\s/?#]+@)?([^\s/?#]+)`,
   'i',
 );
 
--- a/packages/analysis/src/jsts/rules/S5332/rule.lib.ts
+++ b/packages/analysis/src/jsts/rules/S5332/rule.lib.ts
@@ -23,26 +23,30 @@ import { getFullyQualifiedName } from '../helpers/module.js';
 import { getProperty, getValueOfExpression } from '../helpers/ast.js';
 import { normalizeFQN } from '../helpers/aws/cdk.js';
 
-// Mirrors CleartextProtocolFilter.CLEARTEXT_SCHEMES from sonar-analyzer-commons
-const INSECURE_PROTOCOLS = [
-  'http://',
-  'ftp://',
-  'ws://',
-  'telnet://',
-  'rtmp://',
-  'tftp://',
-  'gopher://',
-  'irc://',
-  'smtp://',
-  'ldap://',
-  'amqp://',
-  'mqtt://',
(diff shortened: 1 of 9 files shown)


packages/analysis/src/jsts/rules/S5332/rule.lib.ts
```diff
@@ -159,21 +176,13 @@ export const rule: Rule.RuleModule = {
...
-    }
-
     return {
       Literal: (node: estree.Node) => {
         const literal = node as estree.Literal;
         if (typeof literal.value === 'string') {
           const value = literal.value.trim().toLocaleLowerCase();
           const insecure = INSECURE_PROTOCOLS.find(protocol => value.startsWith(protocol));
-          if (insecure && !isExceptionUrl(value, node)) {
+          if (insecure && !hasExceptionHost(value)) {
```
vdiez: Removing `isExceptionUrl()` makes bare scheme literals like `"http://"` reportable in every AST context, not just when they behave like URLs. That no longer matches the merged `CleartextProtocolFilter` in `sonar-analyzer-commons#395`: `isSafeWithoutTls("http://")` returns `true` there because `CLEARTEXT_AUTHORITY` requires an authority after `://`. The ruling diff shows the regression on placeholders and prefix checks such as `pathLowerCase.startsWith('http://')`, `'http://'` cursor placeholders, and `'http://'` upload sentinels. I think we still need analyzer-side context filtering here.

vdiez (changes requested):
One blocking issue remains.

Removing the old `isExceptionUrl()` guard broadens S5332 beyond the `JS-1877` scope: bare cleartext-scheme literals like `"http://"` now get reported in any AST context, not just when they behave like URLs. That no longer matches the merged `CleartextProtocolFilter` in `sonar-analyzer-commons#395`, where `isSafeWithoutTls("http://")` returns `true` because the fallback authority regex requires a host after `://`.

This shows up as new false positives in the ruling diff for placeholders and prefix checks, for example in `Ghost/core/client/app/mixins/ed-editor-shortcuts.js`, `desktop/app/src/main-process/main.ts`, `vitest/packages/vitest/src/node/test-run.ts`, and `fresh/docs/check_images_test.ts`.

I would keep the new safe-host parity logic, but restore analyzer-side context filtering for bare scheme sentinels and other non-URL-use sites.

packages/analysis/src/jsts/rules/S5332/rule.lib.ts
daniel-teuchert-sonarsource: This is a good point! I should have checked the ruling changes more thoroughly.
I will re-add this logic.

vdiez (approved):
﻿The follow-up commit fixes the issue I raised.

Restoring `isExceptionUrl()` brings S5332 back in line with `CleartextProtocolFilter` for bare protocol strings, and the updated ruling diff removes the false positives on placeholders and prefix checks. I re-ran the focused S5332 tests and a clean `npm run bbf` on the current head.

PR #7257, opened 2026-06-10
JS-1825 S5852: only report in case of exponential backtracking + add rule S8786

pierre-loup-tristant-sonarsource:
Part of JS-1825

----
## Summary by Gitar

- **Rule Changes:**
  - Modified `S5852` to report only in cases of exponential backtracking.
- **New Rule:**
  - Added `S8786` to detect regular expressions with super-linear performance due to backtracking.
- **Profile Updates:**
  - Added `S8786` to `Sonar agentic AI` and `Sonar way` quality profiles.


francois-mora-sonarsource (commented):
[P2] Keep S8786 exclusive to non-exponential cases

`S8786` currently reports some regexes that already belong to `S5852`. The issue is that `scslre.analyse()` can return a mixed `reports` array for one regexp, with both `{ exponential: true }` and `{ exponential: false }` entries. In that case:

```ts
// S5852
if (reports.some(r => r.exponential)) { ... }

// S8786
if (reports.some(r => !r.exponential)) { ... }
```

both rules raise on the same regex.

Reproducer: `/(a+)+$/`. `scslre` returns mixed reports for it, so `S8786` reports it even though it is an exponential/ReDoS case and should stay under `S5852` only. I verified that with a unit-test reproducer by adding `/(a+)+$/` to the `valid` cases of `packages/analysis/src/jsts/rules/S8786/unit.test.ts`; the test fails because `S8786` raises one issue on that pattern.

This is more than duplicate noise: it misclassifies an exponential backtracking regex under the `S8786` message ("super-linear performance due to backtracking"), while the stronger and more accurate classification is `S5852` ("vulnerable to exponential backtracking ... denial of service").

The fix should make the split disjoint by reporting `S8786` only for pure non-exponential cases, e.g.:

```ts
const hasExponential = reports.some(r => r.exponential);
const hasNonExponential = reports.some(r => !r.exponential);

if (hasNonExponential && !hasExponential) {
  context.report(...);
}
```

That avoids the false positive on `S8786` without introducing a real false negative, because mixed cases are still covered by `S5852`.

francois-mora-sonarsource (approved):
Validated the S8786/S5852 overlap fix. S8786 now stays exclusive to pure non-exponential cases, the original reproducer is covered in unit tests, and the updated ruling expectations remove the duplicated S8786 reports from the reviewed projects.

PR #7268, opened 2026-06-12
JS-1867 S4790: suppress when hash output is truncated/sliced

teemu-rytilahti-sonarsource:
## Summary
- Suppress S4790 when the result of `crypto.createHash(...).update(...).digest(...)` flows into `.slice` / `.substring`, either directly on the chain or via a `const`-bound variable.
- Truncated weak digests cannot serve password storage, signing, or integrity — those need the full output — so the use is non-cryptographic (short identifier, cache key, ETag, k-anonymity prefix).

### DIFF
--- a/packages/analysis/src/jsts/rules/S4790/rule.ts
+++ b/packages/analysis/src/jsts/rules/S4790/rule.ts
@@ -202,9 +202,11 @@ function isIdentifierTruncated(identifier: estree.Node): boolean {
   );
 }
 
-// Reject no-op slices that don't actually shorten the digest: `.slice()`, `.slice(0)`, `.substring(0)`.
-// Length-bound checks (e.g. `.slice(0, 64)` on a 40-char sha1 hex) are out of scope — that needs
-// digest-length and encoding awareness.
+// Conservative truncation check: only treat the call as truncation when at least one argument is a
+// non-zero numeric literal (positive or negative). Rejects all the obvious no-ops — `.slice()`,
+// `.slice(0)`, `.slice(-0)`, `.slice(0, undefined)`, `.slice(0, Infinity)`, etc.
+// Length-bound checks against the digest's own size (e.g. `.slice(0, 64)` on a 40-char sha1 hex)
+// remain out of scope — that needs digest-length and encoding awareness.
 function isTruncationCall(call: estree.CallExpression): boolean {
   if (call.callee.type !== 'MemberExpression' || call.callee.property.type !== 'Identifier') {
     return false;
@@ -212,12 +214,15 @@ function isTruncationCall(call: estree.CallExpression): boolean {
   if (!TRUNCATION_METHODS.has(call.callee.property.name)) {
     return false;
   }
-  if (call.arguments.length === 0) {
-    return false;
+  return call.arguments.some(isNonZeroNumericLiteral);
+}
+
+function isNonZeroNumericLiteral(node: estree.Node | estree.SpreadElement): boolean {
+  if (node.type === 'Literal') {
+    return typeof node.value === 'number' && node.value !== 0;
   }
-  const [first] = call.arguments;
-  if (call.arguments.length === 1 && first.type === 'Literal' && first.value === 0) {
-    return false;
+  if (node.type === 'UnaryExpression' && (node.operator === '-' || node.operator === '+')) {
+    return isNonZeroNumericLiteral(node.argument);
   }
-  return true;
+  return false;
 }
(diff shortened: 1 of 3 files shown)


packages/analysis/src/jsts/rules/S4790/rule.ts
```diff
@@ -86,3 +88,136 @@ export const rule: Rule.RuleModule = {
...
+// Length-bound checks (e.g. `.slice(0, 64)` on a 40-char sha1 hex) are out of scope — that needs
+// digest-length and encoding awareness.
+function isTruncationCall(call: estree.CallExpression): boolean {
+  if (call.callee.type !== 'MemberExpression' || call.callee.property.type !== 'Identifier') {
+    return false;
+  }
+  if (!TRUNCATION_METHODS.has(call.callee.property.name)) {
+    return false;
+  }
+  if (call.arguments.length === 0) {
```
vdiez: `isTruncationCall()` still treats several full-output/no-op forms as truncation, so this change can suppress `S4790` even when the full weak digest is still being used. I verified locally that `slice(0, undefined)`, `slice(0, Infinity)`, `substring(0, undefined)`, and `slice(-0)` now produce 0 diagnostics. The helper needs to inspect argument values, not just reject the zero-arity and single-literal-0 cases.

vdiez (changes requested):
One blocking issue remains in the new truncation suppression logic. See the inline comment.

packages/analysis/src/jsts/rules/S4790/rule.ts
```diff
@@ -86,3 +88,141 @@ export const rule: Rule.RuleModule = {
...
+// Length-bound checks against the digest's own size (e.g. `.slice(0, 64)` on a 40-char sha1 hex)
+// remain out of scope — that needs digest-length and encoding awareness.
+function isTruncationCall(call: estree.CallExpression): boolean {
+  if (call.callee.type !== 'MemberExpression' || call.callee.property.type !== 'Identifier') {
+    return false;
+  }
+  if (!TRUNCATION_METHODS.has(call.callee.property.name)) {
+    return false;
+  }
+  return call.arguments.some(isNonZeroNumericLiteral);
```
vdiez: Requiring a non-zero numeric literal here is too narrow: common truncated forms like `const PREFIX = 8; digest('hex').slice(0, PREFIX)` or `const h = digest(...); h.slice(0, PREFIX)` now report again, even though they are still clearly truncated and the previous revision suppressed them. I verified this locally for both `crypto.createHash(...).digest(...)` and `await crypto.subtle.digest(...)`. This change fixes the no-op cases, but it also reintroduces false positives for named-constant bounds and other non-literal arguments that still shorten the digest.

vdiez (changes requested):
The previous false-negative is fixed, but the new positive truncation check now regresses some common truncated forms. See the inline comment.

packages/analysis/src/jsts/rules/S4790/rule.ts
```diff
@@ -86,3 +88,149 @@ export const rule: Rule.RuleModule = {
...
+function isTruncationCall(call: estree.CallExpression): boolean {
+  if (call.callee.type !== 'MemberExpression' || call.callee.property.type !== 'Identifier') {
+    return false;
+  }
+  const method = call.callee.property.name;
+  if (!TRUNCATION_METHODS.has(method)) {
+    return false;
+  }
+  const allowNegatives = method === 'slice';
+  return call.arguments.some(arg => isTruncatingArg(arg, allowNegatives));
```
vdiez: This is still too narrow for the original suppression goal. The new `substring` no-op cases are correctly rejected now, but common truncated forms like `const PREFIX = 8; digest('hex').slice(0, PREFIX)`, `const h = digest(...); h.slice(0, PREFIX)`, and `(await crypto.subtle.digest(...)).slice(0, LEN)` still report because only literal numeric bounds count as truncation. I rechecked those locally on this commit and they all still produce `S4790`, while the previous `ebbd864` revision suppressed them.

vdiez (changes requested):
The latest changes fix the new `substring` no-op cases, but the main regression from the previous revision is still present. See the inline comment.

vdiez (approved):
Pushed a fix for the truncation gate. The rule now rejects only obvious full-output  /  forms while still suppressing named-constant truncation bounds, and the added fixture coverage exercises both sides of that behavior.

vdiez (approved):
Pushed a fix for the truncation gate. The rule now rejects only obvious full-output slice / substring forms while still suppressing named-constant truncation bounds, and the added fixture coverage exercises both sides of that behavior.

vdiez:
Added a code comment to make the scope explicit: this PR intentionally models only the existing `slice` / `substring` truncation idiom. We are not expanding it here to other shortening helpers such as `substr`, `at`, `charAt`, `subarray`, or transform-based extraction patterns, because that would broaden the PR beyond the false-positive fix we are making for `slice` / `substring`.

PR #7271, opened 2026-06-12
Cleanup generated-source file-store plumbing after JS-1597

vdiez:
## Summary

This PR is the cleanup pass after the JS-1597 generated-source rollout.

The feature is functionally complete, but the implementation still reflected an intermediate rollout design where generated-source detection was split across:

- a project-derived detector cache
- a request-sensitive filtered match cache
- request-scoped telemetry/logging state
- special handling in `file-stores/index.ts`

That intermediate split made the cache model harder to reason about, pushed request-specific plumbing into the file-store API, and left the docs describing a more complicated lifecycle than the one we actually want to keep.

This PR simplifies that architecture and aligns the implementation with the long-term model:

- `generatedSourceStore` now caches only project-derived detector output
- the current tagged subset is computed at analysis time from the current analyzable files
- generated-source observability/logging also happens at analysis time
- file stores now operate on plain sanitized `inputFiles` instead of a request-context object
- generated-source detection reuses the shared project walk instead of doing detector-time filesystem probing

## Design Constraints

The refactor is driven by a few constraints that are important to make explicit, because they explain why some of the code now looks the way it does.

### 1. Generated-source detection must reuse the shared project walk

We do not want a second filesystem walk just for generated-source detection.

The project is already traversed once to initialize the pending file stores. This PR keeps generated-source derivation on that same path and avoids a second scan to discover config files or generated outputs.

### 2. Detectors should not go back to disk after the walk

Once the walk is over, detectors should derive from in-memory data only.

That means:

- no ad hoc `fs` access from the detectors
- no fallback directory scanning from detector helpers
- no post-walk probing for config files or output files

If a file's content may be needed during derivation, it has to be captured during the walk.

### 3. The generated-source cache should not depend on request-specific provenance

The cache should answer a project-level question:

> which files did the detectors resolve as generated for this project?

The current request should only decide which of those resolved files are currently visible to the current analysis.

That is why the store now keeps project-derived detector output only, while tagging and telemetry are computed later from the current analyzable files.

### 4. Telemetry should stay coarse: `resolved` vs `tagged`

For the real production path, Node cannot reliably distinguish all the reasons why a resolved file is not currently analyzed.

In particular, when `inputFiles` come from the Java-side scanner engine, "missing from the current analyzable set" can mean several different things upstream:

- filtered in Java
- omitted because Java reused cache state
- excluded by current scope settings
- filtered later in Node

We do not want Node to publish fake precision here. The telemetry contract should stay coarse:

- `resolved`: the detector derived this file from project metadata
- `tagged`: the file is present in the current analyzable set

### 5. The `canAccessFileSystem=false` path should stay simple

When filesystem access is not available, we do not want generated-source detection to maintain a parallel cache lifecycle.

In that mode the store is skipped entirely and behaves as a no-op. That keeps the no-FS path explicit and easy to reason about.

### 6. Ownership should stay clear across stores

This cleanup also tries to make ownership explicit instead of letting generated-source detection silently duplicate other stores' state.

The intended ownership is now:

- `sourceFileStore`: analyzable file contents and current request-visible files
- `dependencyManifestStore`: raw `package.json` ownership
- `generatedSourceStore`: project-derived generated-source metadata plus the transient walk snapshot it needs during derivation

### 1. `FileStoreRequestContext` was carrying rollout-specific plumbing

The file-store API had grown a request-context object mainly to support the earlier generated-source design.

That object bundled together:

- sanitized files
- path-map/provenance details
- request-scoped distinctions that the final cache model does not actually want to keep

Once the generated-source cache is made project-derived again, that extra request-context layer becomes unnecessary.

### 2. The orchestrator had generated-source-specific behavior that belonged inside the store

The previous design forced `file-stores/index.ts` to know too much about generated-source behavior, including disabled behavior and request-specific handling.

That is backwards. The store itself should own:

- whether it is active
- what invalidates it
- what it keeps in cache
- what is transient walk-time state vs persistent derived state

### 3. The docs had drifted toward the rollout intermediate state

The documentation described a hybrid request-filtered generated-source cache and request-context behavior that we do not want to keep as the long-term design.

This PR is not only code cleanup. It also brings the docs back to the actual intended architecture.

### 1. Removed `FileStoreRequestContext`

The file-store API now takes plain sanitized `inputFiles?: AnalyzableFiles`.

Concretely:

- removed `FileStoreRequestContext` from `projectAnalysis.ts`
- updated `FileStore` to accept `inputFiles` directly
- updated all file stores to the simpler signature
- updated sanitization helpers to return only `files` + `pathMap`
- updated the HTTP/gRPC entrypoints to pass sanitized `files` directly to `initFileStores(...)`

This removes request-specific plumbing that was only there to serve the intermediate generated-source design.

### 2. Simplified `sourceFileStore`

`sourceFileStore` still preserves the important behavior that explicit `inputFiles` are authoritative, but it now reseeds directly from the sanitized file map without wrapping that map in an extra request-context object.

### 3. Moved generated-source internals under `file-stores/generated-sources/`

Generated-source derivation, detectors, observability helpers, snapshot helpers, and telemetry helpers are now grouped under the generated-source file-store area.

This makes the ownership boundary clearer:

- file-store concerns live with the file store
- rule helpers consume the result instead of owning the detection implementation

### 4. Turned `generatedSourceStore` into a project-derived cache only

Removed the request-sensitive state from `generatedSourceStore`, including the filtered-match/request-key machinery.

The store now keeps only project-derived detector state:

- `derivedFamilyByFile`
- `configPaths`
- `watchedOutputPaths`

`getFamily()` is now project-derived as well.

That means a generated file can still have a family even if it is currently excluded or outside the current analyzable subset. This is intentional: the store caches detector output, not the current request view.

### 5. Moved generated-source observability to analysis time

Added:

- `generatedSourceStore.observeGeneratedSources(configuration, filesToAnalyze)`

`analyzeProject()` now uses this to compute telemetry and emit generated-source logs from the current analyzable set.

That keeps the layering clean:

- the store owns project-derived detector data
- the analysis pipeline owns the current tagged subset
- telemetry/logging are derived from the current analysis snapshot instead of being cached as store state

### 6. Made disabled and no-FS behavior store-owned

`generatedSourceStore` now handles its own disabled behavior:

- `detectGeneratedCode=false` => no-op
- `canAccessFileSystem=false` => no-op

This removes generated-source special casing from the file-store orchestrator.

### 7. Reused existing stores instead of duplicating their responsibilities

A key part of this cleanup is that generated-source derivation now reuses other stores rather than owning duplicate copies of their responsibilities.

Specifically:

- overlapping JS/TS config contents are reused from `sourceFileStore.getFileContent(...)`
- raw `package.json` contents are reused from `dependencyManifestStore.getPackageJsons()`

Store order now matters intentionally in `file-stores/index.ts`:

- `sourceFileStore` must run before `generatedSourceStore`
- `dependencyManifestStore` must run before `generatedSourceStore`

That order is now documented in code and in the docs.

### 8. Reworked generated-source derivation around an explicit transient snapshot

During the shared walk, `generatedSourceStore` captures the transient inputs it needs to derive detector output later:

- walked directories
- walked JS/TS source-file paths
- preloaded detector inputs whose contents may be needed later

Then in `postProcess()` it:

- merges in raw `package.json` files from `dependencyManifestStore`
- builds the project snapshot view
- derives generated-source metadata entirely from memory
- clears the transient walk snapshot immediately afterwards

So the persistent cache stays small, but the derivation still benefits from the single shared walk.

### 9. Kept the single-walk / no-disk tradeoff explicit

One detail worth calling out: the transient preload is still broader than just a tiny hardcoded list of files.

That is intentional.

Some detector inputs are only discoverable after parsing `package.json` scripts in `postProcess()`. For example, a project can declare an explicit config path such as:

- `graphql-codegen --config ./custom-codegen.ts`

By the time we know that path, the filesystem walk is already over.

Because this PR explicitly keeps the constraints above:

- no second walk
- no detector-time disk access

any file whose content may later be needed as a config has to be captured while the walk is happening.

So this PR does **not** try to eliminate all preload of config-like JS/TS/JSON/YAML files. That preload is the tradeoff that lets us keep the one-walk / in-memory derivation model.

What this cleanup *does* do is remove unnecessary duplication and waste within that model:

- `package.json` is no longer read twice by generated-source detection
- raw manifests stay owned by `dependencyManifestStore`
- generated-source snapshot creation reuses that manifest data directly
- the transient snapshot is cleared after derivation
- obviously irrelevant cases such as plain `.tsx` source files are no longer preloaded as config content

### Real SQS / filesystem-backed path

This PR does **not** really change cache hit/miss semantics in the normal production path.

The important properties stay the same:

- `sourceFileStore` is still rebuilt from the explicit `inputFiles`
- `dependencyManifestStore` is still a request-insensitive helper-file cache
- `generatedSourceStore` is still a request-insensitive project/config cache

So Java-side cache hits still mainly affect the current tagged subset indirectly, because those files are simply absent from `sourceFileStore`.

What *does* improve is the generated-source **miss path**.

Before this cleanup, a miss could reuse the project walk and then still go back to disk again for detector config/output discovery. Now the derivation reuses the same walk-time snapshot and existing store-owned inputs.

So the main effect in the real path is:

- same cache-keying semantics
- cheaper miss path

### BaseDir-only / Node-owned discovery path

Same conclusion.

The invalidation split is still:

- `sourceFileStore`: analyzable-file configuration
- `generatedSourceStore`: project-discovery/generated-source inputs

So scope-only changes can still refresh `sourceFileStore` while reusing `generatedSourceStore`, which is exactly the intended `resolved` vs `tagged` split.

### `inputFiles` + `canAccessFileSystem=false`

This PR keeps that path simple by skipping generated-source derivation entirely.

That is deliberate. We do not want a parallel request-only generated-source cache model in the no-FS path.

## Result

This cleanup leaves us with:

- a simpler file-store API
- no request-sensitive generated-source cache state
- no generated-source special casing in `file-stores/index.ts`
- clear ownership between `sourceFileStore`, `dependencyManifestStore`, and `generatedSourceStore`
- telemetry/logging aligned with the only distinction we really want to publish (`resolved` vs `tagged`)
- generated-source derivation that stays within the intended constraints: one walk, no detector disk access, in-memory post-process derivation
- docs that match the implementation again

## Validation

Ran targeted validation for the touched areas:

```sh
npx tsx --test packages/analysis/tests/file-stores.test.ts packages/analysis/tests/jsts/project-metadata/generated-sources.test.ts packages/analysis/tests/dependency-manifests.test.ts packages/analysis/tests/telemetry.test.ts packages/grpc/tests/analyze-project-normalize.test.ts
./node_modules/.bin/prettier --check packages/analysis/src/file-stores/generated-sources/store.ts packages/analysis/src/file-stores/generated-sources/snapshot-files.ts packages/analysis/src/file-stores/index.ts packages/analysis/tests/jsts/project-metadata/generated-sources.test.ts docs/generated-sources.md docs/file-stores.md docs/node-analysis-caches.md
```

I intentionally kept validation focused on the cleanup surface instead of running the full build in this draft.


vdiez:
Follow-up on the earlier telemetry discussion in #7215, now that this cleanup changed how generated-source inputs are collected.

I re-checked the cache behavior across the two real analysis paths plus the explicit-input/no-FS edge case.

For the real SQS path (`request.files` present), this does **not** really change cache hit/miss semantics:
- `sourceFileStore` is still rebuilt for every request with explicit `inputFiles`
- `dependencyManifestStore` and `generatedSourceStore` are still request-insensitive caches keyed from project/config state, not from `request.files`
- so Java-side cache hits still only affect the `tagged` subset indirectly, because those files are simply absent from `sourceFileStore`

What *does* change is the generated-source **miss path**. On a miss, we already pay for the project walk for the pending stores. Before this cleanup, generated-source detection would then go back to disk again for GraphQL/OpenAPI/etc. config/output discovery. Now it reuses that same walk to preload the small set of detector-specific inputs and derives mostly from the in-memory snapshot. So the main effect is: same hit/miss behavior, cheaper miss path.

For the baseDir-only path (Node owns discovery), same conclusion:
- `sourceFileStore` still invalidates on the full analyzable-file config key
- `generatedSourceStore` still invalidates on project-discovery/generated-source inputs
- so scope-only changes can still miss `sourceFileStore` while hitting `generatedSourceStore`, which is exactly the intended `resolved` vs `tagged` split

The only scenario where behavior really changes is the non-standard `inputFiles` + `canAccessFileSystem=false` case. Previously, generated-source derivation was effectively empty there. Now, if the request also includes the relevant helper files/configs, the generated-source store can derive from that in-memory snapshot. I don’t think this affects the real SQS path, but it is the one case that is more than just a miss-cost optimization.

So I still think the telemetry contract should stay coarse:
- `resolved`: detector derived this generated path from project metadata
- `tagged`: this path ended up in `sourceFileStore`

For the real SQS path, anything finer than that is still an inferred explanation for why a resolved file was not tagged, and Node still cannot reliably distinguish “filtered in Java”, “omitted because of Java cache hit”, and “filtered later in Node”.

francois-mora-sonarsource:
Approved.

I checked the cleanup against `#7215`, especially the `resolved` semantics. `resolved` now means “resolved from the shared project snapshot used by analysis,” not “detector-derived before analyzer-side filtering.” That matches the design here: a single shared walk, project-derived store state, and analysis-time tagging/telemetry.

The layering is cleaner overall. The main follow-up is doc alignment, since the generated-source docs still read closer to the older `resolved`/`tagged` contract.

A few simplifications also look possible in follow-up cleanup:
- several snapshot-based detector helpers are still `async` even though the new path is in-memory only
- `postProcess()` still carries some defensive setup/lifecycle logic that `initFileStores()` now centralizes
- `package.json` parsing / task-invocation collection is done once to collect declared preload paths and again during derivation, so there may be room to share that intermediate later

vdiez:
@francois-mora-sonarsource I pushed a follow-up pass on top of this cleanup to address the points you raised.

What changed:
- docs: tightened the generated-source wording so `resolved` now explicitly means "derived from the shared project snapshot used by analysis", rather than the older looser phrasing
- snapshot helpers: removed several leftover `async` wrappers in the snapshot-only detector path (`detector-api`, GraphQL/OpenAPI/proto detector internals) now that those helpers are purely in-memory
- lifecycle: removed the defensive `setup(...)` fallback from `generatedSourceStore.postProcess()`. The store lifecycle now fully follows what `initFileStores()` centralizes, and the extra `postProcess()` initialization assertion was dropped as redundant once that flow was cleaned up
- shared intermediate: introduced a small shared package-context helper so `package.json` parsing and task-invocation collection are now done once per generated-source post-process and reused both for declared-preload collection and detector derivation

Validation after the follow-up:

```sh
npx tsx --test packages/analysis/tests/file-stores.test.ts packages/analysis/tests/jsts/project-metadata/generated-sources.test.ts packages/analysis/tests/dependency-manifests.test.ts packages/analysis/tests/telemetry.test.ts packages/grpc/tests/analyze-project-normalize.test.ts
./node_modules/.bin/prettier --check packages/analysis/src/file-stores/generated-sources/package-contexts.ts packages/analysis/src/file-stores/generated-sources/contracts.ts packages/analysis/src/file-stores/generated-sources/derive.ts packages/analysis/src/file-stores/generated-sources/preload-paths.ts packages/analysis/src/file-stores/generated-sources/store.ts packages/analysis/src/file-stores/generated-sources/detector-api.ts packages/analysis/src/file-stores/generated-sources/detectors/openapi-generator.ts packages/analysis/src/file-stores/generated-sources/detectors/graphql-codegen.ts packages/analysis/src/file-stores/generated-sources/detectors/proto-loader-gen-types.ts docs/generated-sources.md
```

Both were green locally, and after the final lifecycle cleanup I reran the focused test suite and it stayed green.

francois-mora-sonarsource (approved):

PR #7283, opened 2026-06-15
JS-1801 Narrow S7766 typed union suppression

francois-mora-sonarsource:
## Summary

This PR is a follow-up to the Peach impact analysis of [JS-1801](https://sonarsource.atlassian.net/browse/JS-1801), which revealed suppressed true positives after the initial false-positive escape.

The false-positive escape was introduced with the assumption that a union type meant the value should not be treated as numeric. Peach showed that this assumption was too broad: in practice, a union can still mean the value is numeric, but the type annotations are not narrow enough to tie it to only `number`.

The consequence is that unions containing `number` must not go through the false-positive escape. This change narrows the suppression logic so any union with a numeric constituent is reported instead of suppressed.

## Validation

- `npx tsx --tsconfig packages/tsconfig.test.json --test --test-concurrency=1 --test-reporter=spec --test-reporter-destination=stdout packages/analysis/src/jsts/rules/S7766/unit.test.ts`
- GitHub CI on PR #7283

[JS-1801]: https://sonarsource.atlassian.net/browse/JS-1801?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/analysis/src/jsts/rules/S7766/decorator.ts
+++ b/packages/analysis/src/jsts/rules/S7766/decorator.ts
@@ -113,8 +113,8 @@ function getTypedMinMaxEvidence(
  * Converts a TypeScript type into a reporting decision.
  *
  * Decision rules:
- * 1. Union: return `unknown` if any branch is unknown, `suppress` if any branch
- *    is suppressible, otherwise `report`.
+ * 1. Union: return `unknown` if any branch is unknown, `report` if any branch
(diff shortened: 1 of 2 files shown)


guillemsarda (approved):
Image

PR #7285, opened 2026-06-15
JS-1898 Add check for multer sub-function

alessandro-mizzaro-sonarsource:
Part of 

----
## Summary by Gitar

- **Rule Analysis Improvements:**
  - Added logic in `checkMulter` to ignore destructured imports of `multer` sub-functions like `diskStorage`.
  - Added support for binary expressions in `getSizeValue` to correctly evaluate arithmetic operations when calculating file size limits.
- **Test Infrastructure:**
  - Added a new fixture file `cb.fixture.js` to validate compliant and noncompliant multer configurations.
  - Added `cb.test.ts` test suite using the comment-based checker for rule `S5693` validation.

### DIFF
--- a/packages/analysis/src/jsts/rules/S5693/cb.fixture.js
+++ b/packages/analysis/src/jsts/rules/S5693/cb.fixture.js
@@ -23,7 +23,7 @@ const config = {
     },
   }),
   limits: {
-    fileSize: 10 * 1024 * 1024,
+    fileSize: 5 * 1024 * 1024,
     fieldSize: 1024 * 1024,
     fieldNameSize: 5,
     files: 1,
--- a/packages/analysis/src/jsts/rules/S5693/cb.fixture.js
+++ b/packages/analysis/src/jsts/rules/S5693/cb.fixture.js
@@ -1,7 +1,6 @@
-const multer = require('multer');
-const { diskStorage } = require('multer');
-const { randomBytes } = require('crypto');
-const { extname } = require('path');
+import multer, { diskStorage } from 'multer';
+import { randomBytes } from 'node:crypto';
+import { extname } from 'node:path';
 
 const tmpFolder = '/tmp';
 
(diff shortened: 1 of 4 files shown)


packages/analysis/src/jsts/rules/S5693/cb.fixture.js
```diff
@@ -0,0 +1,36 @@
...
+    fieldSize: 1024 * 1024,
+    fieldNameSize: 5,
+    files: 1,
+  },
+};
+
+// multer without limits is not compliant
+const upload = multer({ dest: '/uploads' }); // Noncompliant {{Make sure the content length limit is safe here.}}
+
+const upload_safe = multer(config); // Compliant
```
nathsou, line 36: Why is this example considered safe when `config.fileSize` is greater than the default threshold of 8MB? I guess this is voluntary but I would like to understand the context :)

packages/analysis/src/jsts/rules/S5693/rule.ts
nathsou, line 266: Minor: 

To support patterns like:

```js
const MB = 1024 ** 2; // bytes
multer({ limits: { fileSize: 8 * MB } });
```

where constants used in the fileSize binary expression are themselves expressions (not number literals).

We can add a constant resolution step:


```suggestion
  const resolved = getUniqueWriteUsageOrNode(context, node, true);
  if (resolved !== node) {
    return getSizeValue(context, resolved);
  }

  const literal = getValueOfExpression(context, node, 'Literal');
```

packages/analysis/src/jsts/rules/S5693/rule.ts
```diff
@@ -215,6 +223,30 @@ function getSizeValue(context: Rule.RuleContext, node: estree.Node): number | nu
...
+          break;
+        case '+':
+          result = left + right;
+          break;
+        case '-':
+          result = left - right;
+          break;
+        case '/':
+          result = left / right;
+          break;
```
nathsou: Minor: supporting `**` can also be useful:

```suggestion
        case '/':
          result = left / right;
          break;
        case '**':
          result = left ** right;
          break;
```

nathsou (approved):
LGTM, I just have two minor comments but it's not blocking

PR #7322, opened 2026-06-19
JS-1914 Add ensure-rule-data lifecycle for RSPEC outputs

francois-mora-sonarsource:
## Summary

Adds `npm run ensure-rule-data` as the SonarJS-owned entrypoint for preparing generated RSPEC rule data. The command reuses prepared outputs only when the rule data, per-language `rspec.sha` files, current checkout, root `rspec.sha` pin, and ignored `resources/rule-data-state.json` stamp line up.

`npm run generate-meta` now runs this freshness check before `generate-meta:raw`, and CI prepares RSPEC rule data through the same command. The prepared artifact now includes the state stamp and is downloaded at the repository root so the generated paths are restored consistently.

The lifecycle also guards RSPEC pins: stale stamped, unpinned data clears generated per-language pins before regeneration to avoid reusing pins from another checkout, while prepared data with no state stamp preserves those pins for legacy or direct Maven pinned outputs.

## Why

External automation and local metadata generation need an idempotent SonarJS command that can prepare rule data without keeping separate hidden state. Making the stamp visible in the checkout also lets downstream jobs and later `generate-meta` calls tell when prepared RSPEC data belongs to the current revision and root pin.

## Validation

- `node --test tools/ensure-rule-data.test.mjs tools/generate-meta.test.mjs`
- `npm exec --yes --package=prettier@3.8.4 -- prettier --no-config --print-width 100 --trailing-comma all --single-quote --arrow-parens avoid --end-of-line lf --check tools/ensure-rule-data.mjs tools/ensure-rule-data.test.mjs docs/DEV.md`
- `git diff --check`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -173,7 +173,7 @@ jobs:
         uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
         with:
           name: rspec-rule-data-${{ github.sha }}
-          path: sonar-plugin
+          path: .
       - &maven_cache
         uses: ./.github/actions/maven-cache
         with:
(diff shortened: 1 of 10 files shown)


vdiez (approved):
Approving the change. This closes a real gap for reused workspaces and moves Nigel-specific lifecycle logic into SonarJS where it belongs.

Please add or plan dedicated RSPEC sync / rule-data lifecycle documentation; the distinction between resources/rule-data-state.json, root rspec.sha, and generated per-language rspec.sha is not obvious from the PR alone.

PR #7324, opened 2026-06-19
JS-1916 S6268: reduce false positives via local constant propagation

Luqmansonar:
## Why

S6268 was generating noise on projects using `bypassSecurityTrust*` with provably compile-time constant values. During a noise-reduction investigation, sample files were downloaded and found to contain false positives. The rule's existing check only suppressed direct inline literals — any variable or object property access, even when assigned a hardcoded string in the same file, was incorrectly flagged.

## What

Extend hardcoded-literal detection to two additional patterns via local constant propagation:

1. An identifier with a single write site that resolves to a string literal
2. A static member expression (`obj.prop`) where the object resolves to a locally-defined object literal with a literal property value

## How

Uses existing AST helpers (`getUniqueWriteUsageOrNode`, `getValueOfExpression`, `getProperty`) — no new dependencies. Change is isolated to `isHardcodedLiteral` in `rule.ts`.

**Before (both flagged as FP):**
```typescript
const trustedHtml = '<b>Hello</b>';
sanitizer.bypassSecurityTrustHtml(trustedHtml); // FP

const config = { content: '<svg>...</svg>' };
sanitizer.bypassSecurityTrustHtml(config.content); // FP
```

**After (neither flagged):**
```typescript
const trustedHtml = '<b>Hello</b>';
sanitizer.bypassSecurityTrustHtml(trustedHtml); // OK

const config = { content: '<svg>...</svg>' };
sanitizer.bypassSecurityTrustHtml(config.content); // OK
```

**Out of scope:** Callback parameters (e.g. `array.map(item => sanitizer.bypassSecurityTrustHtml(item.content))`) and cross-file tracking are not covered by this change.

### DIFF
--- a/packages/analysis/src/jsts/rules/S6268/rule.ts
+++ b/packages/analysis/src/jsts/rules/S6268/rule.ts
@@ -19,7 +19,13 @@
 import type { Rule } from 'eslint';
 import type estree from 'estree';
 import { generateMeta } from '../helpers/generate-meta.js';
-import { isLiteral, isMemberWithProperty } from '../helpers/ast.js';
+import {
+  getProperty,
+  getUniqueWriteUsageOrNode,
+  getValueOfExpression,
+  isLiteral,
+  isMemberWithProperty,
+} from '../helpers/ast.js';
 import * as meta from './generated-meta.js';
 
 const bypassMethods = [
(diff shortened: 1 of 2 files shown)


daniel-teuchert-sonarsource (commented):
The approach sounds good to me. An engineer should review this though to make sure it works as intended.
Thanks for making this improvement!

francois-mora-sonarsource (approved):

PR #7328, opened 2026-06-19
JS-1910: S8785 new rule test callbacks synchronous

guillemsarda:
Part of JS-1910

----
## Summary by Gitar

  - Added `S8785` rule to detect asynchronous callbacks in test suites for Jest, Mocha, and Cypress.
  - Implemented reporting for `async` callbacks, including a quick fix for removing the `async` keyword and flagging tests dropped due to top-level awaits.
- **Test infrastructure:**
  - Created unit tests in `unit.test.ts` covering various framework scenarios, async patterns, and edge cases.
  - Added rule metadata and registered the new rule in the `Sonar way` quality profile.


guillemsarda:
> CI failed: The integration tests failed due to a JSON parsing error triggered by the new rule [S8785](https://sonarsource.github.io/rspec/#/rspec/S8785/javascript), which is generating invalid issue messages containing double-quoted strings within the JSON payload.
> ### Overview
> The integration tests failed due to an error in the implementation of the new rule [S8785](https://sonarsource.github.io/rspec/#/rspec/S8785/javascript). The scanner encountered a `java.lang.IllegalStateException` when processing analysis results because the rule is producing malformed JSON issue messages.
> 
> ### Failures
> #### JSON Parsing Error in Rule [S8785](https://sonarsource.github.io/rspec/#/rspec/S8785/javascript) (confidence: high)
> * **Type**: build
> * **Affected jobs**: 82680684379
> * **Related to change**: yes
> * **Root cause**: The rule logic for [S8785](https://sonarsource.github.io/rspec/#/rspec/S8785/javascript) generates an issue message that includes double quotes inside a double-quoted string (e.g., `"Remove the "async" keyword..."`). This invalidates the JSON structure when the analysis result is serialized.
> * **Suggested fix**: Update the rule implementation for [S8785](https://sonarsource.github.io/rspec/#/rspec/S8785/javascript) to correctly escape internal double quotes in the issue message (e.g., `\"async\"`) or adjust the message formatting to avoid conflicts with JSON syntax.
> 
> ### Summary
> * **Change-related failures**: 1, specifically a malformed issue message payload produced by the new rule implementation causing a [build-time](https://sonarsource.atlassian.net/browse/BUILD-time) crash during integration tests.
> * **Infrastructure/flaky failures**: 0
> * **Recommended action**: Review the code responsible for generating the issue message for rule [S8785](https://sonarsource.github.io/rspec/#/rspec/S8785/javascript). Ensure that all generated text is properly escaped for JSON compliance before returning it as part of the analysis results.
> 
> Code Review 👍 Approved with suggestions 0 resolved / 1 findings
> Implements rule [S8785](https://sonarsource.github.io/rspec/#/rspec/S8785/javascript) to flag asynchronous test callbacks with quick-fix support. Secondary location reporting for dropped tests requires updates to resolve nested declaration detection.
> 
> 💡 Edge Case: Dropped-test secondary locations miss nested declarations
> 📄 [packages/analysis/src/jsts/rules/S8785/rule.ts:223-237](https://github.com/SonarSource/SonarJS/pull/7328/files#diff-5f74c910306341c78d6c2d95eb7205cbf08848f24fdaf5290a1232f860f89257R223-R237)
> 
> `findTestsRegisteredAfter` only iterates the suite callback's direct child statements (`callback.body.body`) when collecting the tests that are silently dropped after a top-level await. Tests/suites declared after the await but nested inside a block — e.g. inside an `if`, `try`, or loop body — are not collected and therefore get no secondary location highlight, even though they are equally dropped at runtime.
> 
> Example not covered:
> 
> describe('svc', async () => { await loadConfig(); if (cond) { it('a', () => {}); // dropped, but not highlighted } });
> 
> The primary `moveAsyncSetup` issue still reports correctly, so impact is limited to under-reported secondary locations. If full coverage is desired, traverse descendant statements (stopping at nested function boundaries, like `findFirstTopLevelAwait` does) rather than only the top-level statement list.
> 
> Walk descendants (skipping nested functions) instead of only direct body statements so dropped tests inside blocks are also collected.
> ```
> const awaitEnd = sourceCode.getRange(awaitNode)[1];
> const tests: estree.Node[] = [];
> const stack: estree.Node[] = [...childrenOf(callback.body, sourceCode.visitorKeys)];
> while (stack.length > 0) {
>   const current = stack.pop();
>   if (current === undefined || isFunctionNode(current)) continue;
>   if (
>     current.type === 'ExpressionStatement' &&
>     current.expression.type === 'CallExpression' &&
>     sourceCode.getRange(current)[0] >= awaitEnd
>   ) {
>     const parts = getMochaCalleeParts(current.expression.callee);
>     if (parts && TEST_AND_SUITE_NAMES.has(parts.base.name)) {
>       tests.push(current.expression.callee);
>       continue;
>     }
>   }
>   stack.push(...childrenOf(current, sourceCode.visitorKeys));
> }
> return tests;
> ```
> 
> 🤖 Prompt for agents
> ```
> Code Review: Implements rule S8785 to flag asynchronous test callbacks with quick-fix support. Secondary location reporting for dropped tests requires updates to resolve nested declaration detection.
> 
> 1. 💡 Edge Case: Dropped-test secondary locations miss nested declarations
>    Files: packages/analysis/src/jsts/rules/S8785/rule.ts:223-237
> 
>    `findTestsRegisteredAfter` only iterates the suite callback's direct child statements (`callback.body.body`) when collecting the tests that are silently dropped after a top-level await. Tests/suites declared after the await but nested inside a block — e.g. inside an `if`, `try`, or loop body — are not collected and therefore get no secondary location highlight, even though they are equally dropped at runtime.
>    
>    Example not covered:
>    
>      describe('svc', async () => {
>        await loadConfig();
>        if (cond) {
>          it('a', () => {}); // dropped, but not highlighted
>        }
>      });
>    
>    The primary `moveAsyncSetup` issue still reports correctly, so impact is limited to under-reported secondary locations. If full coverage is desired, traverse descendant statements (stopping at nested function boundaries, like `findFirstTopLevelAwait` does) rather than only the top-level statement list.
> 
>    Fix (Walk descendants (skipping nested functions) instead of only direct body statements so dropped tests inside blocks are also collected.):
>    const awaitEnd = sourceCode.getRange(awaitNode)[1];
>    const tests: estree.Node[] = [];
>    const stack: estree.Node[] = [...childrenOf(callback.body, sourceCode.visitorKeys)];
>    while (stack.length > 0) {
>      const current = stack.pop();
>      if (current === undefined || isFunctionNode(current)) continue;
>      if (
>        current.type === 'ExpressionStatement' &&
>        current.expression.type === 'CallExpression' &&
>        sourceCode.getRange(current)[0] >= awaitEnd
>      ) {
>        const parts = getMochaCalleeParts(current.expression.callee);
>        if (parts && TEST_AND_SUITE_NAMES.has(parts.base.name)) {
>          tests.push(current.expression.callee);
>          continue;
>        }
>      }
>      stack.push(...childrenOf(current, sourceCode.visitorKeys));
>    }
>    return tests;
> ```
> 
> Tip
> 
> Comment `Gitar fix CI` or enable auto-apply: `gitar auto-apply:on`
> 
> Options
> Auto-apply is off → Gitar will not commit updates to this branch.Display: compact → Showing less information.
> 
> Comment with these commands to change:
> 
> Auto-apply	Compact
> ```
> gitar auto-apply:on         
> ```
> 
> ```
> gitar display:verbose         
> ```
> 
> Was this helpful? React with 👍 / 👎 | [Gitar](https://gitar.ai)                                                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                               
  The AI failure summary was wrong about the mechanism. It claimed the message's double quotes break JSON serialization and that escaping them (\"async\") would fix it. That's a red herring — the message on the failing path is never serialized with JSON.stringify at all, and JSON.parse('Remove the async...') would 
  still fail on the leading R even with no quotes.                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                            
  Actual root cause (its/sources/projects/desktop/app/test/unit/path-test.ts, with failFast=true aborting the whole project):                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                            
  - meta.ts:20 declares hasSecondaries = true. Because of that, the linter's decodeSecondaryLocations runs JSON.parse on every issue message S8785 produces (decode.ts:34-42).                                                                                                                                              
  - The moveAsyncSetup branch reports via the report helper, which JSON-encodes the message — fine.                                                                                                                                                                                                                         
  - The removeAsync branch used raw context.report with a plain string. The moment a real file triggered it, the decoder tried to JSON.parse that plain string → IllegalStateException: Failed to analyze file → analysis crash.                                                                                            
                                                                                                                                                                                                                                                                                                                            
  Fix (rule.ts): route removeAsync through the same report helper (with no secondary locations) so all of the rule's messages are encoded consistently. The rule is invariant: once a rule declares hasSecondaries, every one of its report paths must use the encoding helper.                                             
                                                                                                                                                                                                                                                                                                                            
  I also added a regression test covering removeAsync under settings: { sonarRuntime: true } — the combination that was previously untested and let this slip through.                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                            
  Files changed:                                                                                                                                                                                                                                                                                                            
  - packages/analysis/src/jsts/rules/S8785/rule.ts                                                                                                                                                                                                                                                                          
  - packages/analysis/src/jsts/rules/S8785/unit.test.ts

packages/analysis/src/jsts/rules/S8785/rule.ts
```diff
@@ -0,0 +1,274 @@
...
+  SUITE_FUNCTION_NAMES,
+  TEST_FUNCTION_NAMES,
+} from '../helpers/mocha-style-test-frameworks.js';
+import * as meta from './generated-meta.js';
+
+/**
+ * Frameworks that run a suite's callback synchronously during the discovery phase. Vitest is
+ * intentionally excluded: it awaits the suite callback, so an async callback is safe there.
+ */
+const SUPPORTED_FRAMEWORKS = ['jest', 'mocha', 'cypress'];
```
nathsou: `@jest/globals` is a supported Jest entry point, but this gate only whitelists `jest`, `mocha`, and `cypress`. When the file imports `describe` from `@jest/globals`, `getFullyQualifiedName()` resolves it under `@jest`, so `isSupportedFrameworkConstruct()` below returns false and we miss the issue entirely. `S8780` already handles this by treating Jest as `["jest", "@jest/globals"]`.

nathsou: Oh also vitest?

nathsou (approved):
Image

PR #7343, opened 2026-06-22
JS-1011: S8907 - Native APIs should be preferred over Lodash and Underscore.js methods

nathsou:
Part of [JS-1011](https://sonarsource.atlassian.net/browse/JS-1011)
RSpec PR: https://github.com/SonarSource/rspec/pull/7140

----
## Summary by Gitar

- **New feature:**
  - Implemented rule `S8907` to suggest native JavaScript alternatives for Lodash and Underscore methods.
- **Logic:**
  - Added comprehensive support for identifying `lodash`, `lodash-es`, and `underscore` imports and method calls, including module name mapping.
  - Implemented type-aware logic to handle shape-specific alternatives (e.g., `Array.prototype` vs `String.prototype` vs `Object` methods) and EcmaScript version gating.
- **Testing:**
  - Added extensive test suite and baseline expectations for various projects (e.g., `Ghost`, `sonar-web`, `tailwindcss`) to verify rule accuracy.
- **Configuration:**
  - Registered the rule in the `Sonar agentic AI` and `Sonar way` quality profiles and updated `deploy-rule-data.ts`.

[JS-1011]: https://sonarsource.atlassian.net/browse/JS-1011?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/analysis/src/jsts/rules/S8907/rule.ts
```diff
@@ -0,0 +1,524 @@
...
+    'Array.prototype.includes()',
+    'the library also accepts strings, objects, and nullish values',
+    ES2016,
+  ),
+  drop: cautious('Array.prototype.slice()', ARRAY_NULLISH_REASON, ES5, '`array.slice(n)`'),
+  dropRight: cautious(
+    'Array.prototype.slice()',
+    ARRAY_NULLISH_REASON,
+    ES5,
+    '`n === 0 ? array.slice() : array.slice(0, -n)`',
```
guillemsarda: This `replacements` block looks like a good candidate to move into a helper inside the same rule folder.

Right now `rule.ts` mixes three different concerns:
1. the visitor/reporting flow,
2. the call-detection logic,
3. the static replacement catalog and its derived lookup structures.

Moving the replacement catalog into something like `catalog.ts` or `replacements.ts` would make the rule easier to scan and maintain without changing the design.

For example, `rule.ts` could stay focused on the analysis flow:

```ts
import {
  replacements,
  methodNames,
  methodNamesByLowerCase,
  shapeDependentMethods,
  callbackCollectionMethods,
} from './catalog.js';

File: packages/analysis/src/jsts/rules/S8907/rule.ts [hunk 2]
```diff
@@ -0,0 +1,524 @@
...
+  return (
+    searchValue?.type === 'SpreadElement' ||
+    (searchValue !== undefined && isRegExpExpression(getConstantExpression(searchValue, context)))
+  );
+}
+
+function getConstantExpression(
+  node: estree.Expression,
+  context: Rule.RuleContext,
+  visited = new Set<string>(),
```
guillemsarda: **Very small consideration:** `getConstantExpression()` looks fragile because cycle detection is currently based on the identifier name instead of the resolved binding.

In JavaScript, two different variables can legally share the same name in different scopes, so using `Set<string>` here can stop resolution too early and break the conservative exemptions of the rule.

Example:

```js
import _ from 'lodash';

const pattern = /a/;

function demo() {
  const search = pattern;
  {
    const pattern = search;
    _.includes('abc', pattern);
  }
}

guillemsarda (approved):
Image

PR #7351, opened 2026-06-23
JS-1929: S8927: Default imports from modular utility libraries should not be used

nathsou:
Part of [JS-1929](https://sonarsource.atlassian.net/browse/JS-1929)

[JS-1929]: https://sonarsource.atlassian.net/browse/JS-1929?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

----
## Summary by Gitar

  - Added `S8927` to identify and discourage default imports from modular utility libraries like `lodash` and `rxjs`.
  - Implemented logic in `rule.ts` to check project dependencies and recommend subpath imports based on minimum supported versions.
- **Metadata and Configuration:**
  - Created `S8927.json` definition and enabled the rule in `Sonar way` and `Sonar agentic AI` profiles.
- **Testing:**
  - Added comprehensive test suite `unit.test.ts` with fixtures covering supported, unsupported, and non-semver dependency scenarios.
- **Rule Maintenance:**
  - Updated `S2925` metadata and marked `S6766`, `S7735`, and `S6774` as deprecated or removed from default quality profiles.
  - Updated `rspec.sha` to synchronize documentation with current rule implementations.
- **Codebase Improvements:**
  - Added `isTypeOnlyImport` helper and refactored `getProjectMinVersion` to improve reusability across rule logic.


packages/analysis/src/jsts/rules/S8927/rule.ts
```diff
@@ -0,0 +1,129 @@
...
+  if (!versionRange) {
+    return false;
+  }
+  try {
+    const minimumSupportedVersion = minVersion(versionRange);
+    return minimumSupportedVersion !== null && gte(minimumSupportedVersion, minimumVersion);
+  } catch {
+    return false;
+  }
+}
```
guillemsarda, line 142: ### `*` and `latest` version specs are silently skipped, though they always resolve to a supported version

**File:** `packages/analysis/src/jsts/rules/S8927/rule.ts:119-128`

`supportsRecommendation` derives the version to compare with `minVersion(versionRange)` — the *lowest* version the declared range admits — then checks `gte(min, minimumVersion)`. For unbounded specs this floor is meaningless:

```ts
function supportsRecommendation(versionRange: string | undefined, minimumVersion: string): boolean {
  if (!versionRange) {
    return false;
  }
  try {
    const minimumSupportedVersion = minVersion(versionRange);
    return minimumSupportedVersion !== null && gte(minimumSupportedVersion, minimumVersion);
  } catch {
    return false;
  }
}
```

- `minVersion("*")` → `0.0.0` → `gte(0.0.0, threshold)` is `false` → skipped.
- `minVersion("latest")` → throws → caught → returns `false` → skipped.

**Scenario:**

```jsonc
// package.json
{ "dependencies": { "lodash-es": "*" } }
```

```js
import _ from 'lodash-es'; // should be Noncompliant — "*" resolves to the newest published version, which supports named imports
```

A `*` or `latest` dependency resolves to the newest published version (always at or above the threshold), yet the rule never reports it. The sibling helper `isValidDependencySignal` already special-cases `*`/`latest` for this exact reason; this rule has no comparable handling.

**Suggested fix:** treat `*` / `latest` (and other unbounded specs that imply the newest version) as supported, rather than letting `minVersion` collapse them to `0.0.0` / throw.

packages/analysis/src/jsts/rules/S8927/rule.ts
```diff
@@ -0,0 +1,129 @@
...
+
+function hasNamespaceImport(node: estree.ImportDeclaration): boolean {
+  return node.specifiers.some(specifier => specifier.type === 'ImportNamespaceSpecifier');
+}
+
+function isTypeOnlyImport(node: estree.ImportDeclaration): boolean {
+  return (node as estree.ImportDeclaration & { importKind?: string | null }).importKind === 'type';
+}
+
+function supportsRecommendation(versionRange: string | undefined, minimumVersion: string): boolean {
```
guillemsarda, line 132: ### (Potential reuse) Version-coercion logic is duplicated from existing helper

**File:** `packages/analysis/src/jsts/rules/S8927/rule.ts:119-128`

`supportsRecommendation` reimplements "coerce a declared range to a comparable minimum version, swallowing non-semver garbage" — logic that already exists twice in the codebase, including in a module S8927 already imports from.

S8927's copy:

```ts
function supportsRecommendation(versionRange: string | undefined, minimumVersion: string): boolean {
  if (!versionRange) {
    return false;
  }
  try {
    const minimumSupportedVersion = minVersion(versionRange);
    return minimumSupportedVersion !== null && gte(minimumSupportedVersion, minimumVersion);
  } catch {
    return false;
  }
}
```

`helpers/validate-version.ts:77` already does this and is **more robust** — it adds a `{ loose: true }` fallback for non-canonical ranges:

```ts
function getProjectMinVersion(versionRange: string) {
  try {
    return semver.minVersion(versionRange);
  } catch {
    // Some projects publish non-canonical ranges such as ">=10.00.0".
    try {
      return semver.minVersion(versionRange, { loose: true });
    } catch {
      return null;
    }
  }
}
```

**Suggestion**: consider if it's worth to reuse the helper to get the min version.

packages/analysis/src/jsts/rules/S8927/rule.ts
```diff
@@ -0,0 +1,129 @@
...
+import { getDependenciesSanitizePaths } from '../helpers/dependency-manifests/dependencies.js';
+import * as meta from './generated-meta.js';
+
+const messages = {
+  useNamedImports: 'Use named imports from "{{library}}" instead of a default import.',
+  useMethodImports:
+    'Import used {{library}} methods from method-level subpaths such as "{{example}}".',
+};
+
+type LibraryRecommendation = {
```
guillemsarda: ### The `LIBRARIES` table can express illegal states

**File:** `packages/analysis/src/jsts/rules/S8927/rule.ts:32-61, 95, 103-104`

`messageId` and `example` are perfectly correlated — `useMethodImports` always needs an `example` (lodash, validator), `useNamedImports` never uses one (lodash-es, rxjs, rambda) — but nothing in the types enforces that pairing:

```ts
type LibraryRecommendation = {
  minimumVersion: string;
  messageId: keyof typeof messages;
  example?: string;   // optional for everyone, required-in-practice for only one messageId
};
```

The single `report` call then injects an empty placeholder for the named-import libraries:

```ts
data: {
  library,
  example: recommendation.example ?? '',   // dead data for useNamedImports entries
},
```

**Cost:**
- A `useMethodImports` entry that forgets `example` compiles fine and renders `…subpaths such as ""` at runtime — the invariant lives only in the literal, with no compile-time guard.
- `example: … ?? ''` is dead data that exists only to satisfy the shared interpolation call.
- `getRecommendation` widens the lookup with `LIBRARIES[library as keyof typeof LIBRARIES]`, so a mismatched/typo'd key isn't caught — the `string` param accepts anything and silently returns `undefined`.

**Suggested direction:** model the recommendation as a discriminated union so illegal combinations are unrepresentable, and report per-kind:

```ts
type LibraryRecommendation =
  | { minimumVersion: string; kind: 'named' }
  | { minimumVersion: string; kind: 'method'; example: string };
```

Now `example` is required exactly when it's used, and the empty-string fallback disappears.

packages/analysis/src/jsts/rules/S8927/rule.ts
```diff
@@ -0,0 +1,129 @@
...
+
+function hasDefaultImport(node: estree.ImportDeclaration): boolean {
+  return node.specifiers.some(specifier => specifier.type === 'ImportDefaultSpecifier');
+}
+
+function hasNamespaceImport(node: estree.ImportDeclaration): boolean {
+  return node.specifiers.some(specifier => specifier.type === 'ImportNamespaceSpecifier');
+}
+
+function isTypeOnlyImport(node: estree.ImportDeclaration): boolean {
```
guillemsarda: ### The `importKind` type-only cast is duplicated across three rules

**File:** `packages/analysis/src/jsts/rules/S8927/rule.ts:115-117`

`isTypeOnlyImport` works around a real estree/TSESTree typing gap (estree's `ImportDeclaration` has no `importKind`) with an ad-hoc cast:

```ts
function isTypeOnlyImport(node: estree.ImportDeclaration): boolean {
  return (node as estree.ImportDeclaration & { importKind?: string | null }).importKind === 'type';
}
```

The same cast is reinvented in two other rules:

```ts
// S6747/rule.ts:110-113
const importKind =
  'importKind' in i ? (i as { importKind?: string | null }).importKind : undefined;
```

```ts
// S7763/decorator.ts also accesses importKind via its own cast
```

**Cost:** three independent ad-hoc casts of the same missing-type gap. When the estree/TSESTree typing situation shifts on a dependency bump, every cast site must be found and fixed individually rather than in one place; and each site can drift in how it reads/normalizes the field.

**Suggested direction:** add a single properly-typed helper in `helpers/ast.ts` — e.g. `getImportKind(node)` or `isTypeOnlyImport(node)` — that centralizes the cast, and have S8927, S6747, and S7763 consume it.

guillemsarda (changes requested):

guillemsarda (commented):
Image

guillemsarda (approved):

PR #7359, opened 2026-06-24
JS-1934: S5976 - Similar tests should be grouped in a single Parameterized test

nathsou:
## Summary
- Implement S5976 for Jest/Vitest-compatible `test` and `it` cases
- Detect groups of 3+ structurally equivalent tests that differ only in up to 3 literal parameter positions
- Add focused unit tests for globals, imports, aliases, require destructuring, scoped Jest globals, secondaries, and conservative non-reporting cases

RSPEC PR: https://github.com/SonarSource/rspec/pull/7178

## Validation
- `node node_modules/tsx/dist/cli.mjs --tsconfig packages/tsconfig.test.json --test packages/analysis/src/jsts/rules/S5976/unit.test.ts`
- `node_modules/.bin/tsgo -p packages/tsconfig.app.json`
- `node_modules/.bin/tsgo -p packages/tsconfig.test.json`
- `node node_modules/tsx/dist/cli.mjs tools/generate-meta.ts`
- `git diff --check`
- `git diff --cached --check`

Note: this PR includes `rspec.sha` pinned to `65b108317e` from the matching RSPEC branch.

----
## Summary by Gitar

- **Rule Updates:**
  - Deprecated `S6766` and `S7735` by removing them from default quality profiles.
  - Updated `S2925` metadata, including an increased remediation cost and updated attributes.
- **Configuration:**
  - Added `S5976` to the `Sonar way` quality profile.


packages/analysis/src/jsts/rules/S5976/rule.ts
```diff
@@ -0,0 +1,476 @@
...
+  const handled = new Set<TestCandidate>();
+  for (const [index, test] of tests.entries()) {
+    if (handled.has(test)) {
+      continue;
+    }
+
+    const literalCollector = new LiteralDifferenceCollector(context, test.body);
+    const equivalentTests: TestCandidate[] = [];
+
+    for (const otherTest of tests.slice(index + 1)) {
```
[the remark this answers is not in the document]
nathsou: Fixed in e5af26a92. I added a valid regression case with three identical tests, verified it failed before the production change, and then skip reporting when `parameterNodes.size === 0` because there is no value to parameterize. The focused S5976 unit test passes after the change.

packages/analysis/src/jsts/rules/S5976/rule.ts
```diff
@@ -0,0 +1,476 @@
...
+  handled: Set<TestCandidate>,
+) {
+  if (equivalentTests.length + 1 < MIN_SIMILAR_TESTS) {
+    return;
+  }
+
+  handled.add(test);
+  for (const equivalentTest of equivalentTests) {
+    handled.add(equivalentTest);
+  }
```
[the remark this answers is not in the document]
nathsou: Leaving this unchanged intentionally. The current behavior consumes the maximal neighboring equivalent group before the max-parameter guard, so a group that would need more than 3 parameters is suppressed as a whole instead of reporting a partial subgroup. That keeps the rule conservative and matches the Java-rule-inspired behavior we planned. The valid fixture with `builds case 1` through `builds case 5` covers this: the full neighboring group has 4 parameter positions, while a later subgroup could fit under the limit, and we intentionally report nothing.

packages/analysis/src/jsts/rules/S5976/rule.ts
```diff
@@ -0,0 +1,506 @@
...
+        if (candidate !== undefined && frame !== undefined && !frame.disabled) {
+          frame.tests.push(candidate);
+        }
+      },
+    };
+  },
+};
+
+function importsUnsupportedTestFramework(context: Rule.RuleContext): boolean {
+  return getImportDeclarations(context).some(
```
guillemsarda: ### `require()`-based unsupported frameworks don't disable global `it`/`test` → false positives

**File:** `packages/analysis/src/jsts/rules/S5976/rule.ts:166-171`

`importsUnsupportedTestFramework` only scans **ESM `import`** declarations via `getImportDeclarations`, but framework *activation* (`importsOrDependsOnModule`) also honors `require()` and `package.json` dependencies. The two detection paths are asymmetric, so an unsupported framework loaded through `require()` is invisible to the guard that disables global `it`/`test`.

**Scenario** — `package.json` lists `jest` as a dependency, and a spec uses Cypress via `require()`:

```js
// package.json → { "dependencies": { "jest": "*" } }
const cy = require('cypress');

test('renders 200', () => {
  cy.visit('/status/200');
  cy.contains('200');
});

test('renders 201', () => {
  cy.visit('/status/201');
  cy.contains('201');
});

test('renders 202', () => {
  cy.visit('/status/202');
  cy.contains('202');
});
```

Jest is "active" (via the dependency), and the Cypress `require` is never seen by `importsUnsupportedTestFramework`, so `allowGlobalTestFunctions` stays `true`. The three global `test(...)` blocks get flagged — exactly the Cypress case the `import 'cypress'` valid fixture was meant to suppress, but only the ESM `import` form is suppressed.

**Suggested fix:** detect unsupported frameworks through the same `require()`/dependency-aware path used for activation (e.g. `importsOrDependsOnModule`) instead of `getImportDeclarations`, so both checks see the same modules.
~

packages/analysis/src/jsts/rules/S5976/rule.ts
```diff
@@ -0,0 +1,506 @@
...
+  constructor(
+    private readonly context: Rule.RuleContext,
+    private readonly baseBody: estree.Statement[],
+  ) {
+    this.literalNodesByTokenRange = collectLiteralNodes(context, baseBody);
+  }
+
+  matches(otherBody: estree.Statement[]): boolean {
+    return areEquivalent(this.baseBody, otherBody, this.context.sourceCode, (left, right) => {
+      if (left.value === right.value) {
```
guillemsarda, line 384: ### Literal equality is compared on raw token text, not value → quote/format-only differences become "parameters"

**File:** `packages/analysis/src/jsts/rules/S5976/rule.ts:390-406`

The comparator passed to `areEquivalent` receives **tokens**, and the early `left.value === right.value` check compares *source text*. A `String` token's value includes its surrounding quotes, and a `Numeric` token's value is the literal text. So two literals that are semantically equal but written differently (quote style, numeric format) are treated as a difference, added to `nodesToParameterize`, and reported as a value to parameterize.

**Scenario** — three otherwise-identical tests differing only in quote style and numeric format:

```js
test('normalizes x', () => {
  const result = normalize(f('x'));
  expect(result).toBe(1);
});

test('normalizes x again', () => {
  const result = normalize(f("x"));   // same value 'x', different quotes
  expect(result).toBe(1.0);           // same value 1, different format
});

test('normalizes x once more', () => {
  const result = normalize(f(`x`));
  expect(result).toBe(1);
});
```

Although `'x'`, `"x"` and `` `x` `` carry the same value (and `1` === `1.0`), the raw token texts differ, both sides share the same literal category, so the base literal is added as a parameter. The rule reports "Replace these 3 tests with a single Parameterized one" and highlights the quote-/format-only differences as values to parameterize — a noisy false positive on what are effectively duplicate tests.

**Suggested fix:** compare the AST literals' actual values (the `node.value` already resolved via `literalNodesByTokenRange`) rather than relying on the raw `token.value` text in the early-equality short-circuit, so quote-style and numeric-format variations are not mistaken for parameterizable differences.

packages/analysis/src/jsts/rules/S5976/rule.ts
```diff
@@ -0,0 +1,506 @@
...
+  const moduleName = SUPPORTED_MODULE_FQNS.find(moduleName => fqn.startsWith(`${moduleName}.`));
+  const testName = moduleName === undefined ? undefined : fqn.slice(moduleName.length + 1);
+  return testName !== undefined && TEST_FUNCTION_NAMES.has(testName) ? testName : undefined;
+}
+
+function isSupportedModule(fqn: string | null): boolean {
+  return fqn !== null && SUPPORTED_MODULE_FQNS.includes(fqn);
+}
+
+function isConcreteModifier(modifier: string, activeFrameworks: ActiveFrameworks): boolean {
```
guillemsarda, line 271: ### Unrecognized per-framework modifiers silently drop the whole test → false negatives

**File:** `packages/analysis/src/jsts/rules/S5976/rule.ts:220-228` (with `isConcreteModifier`, `275-282`)

`classifyTestCall` returns `'concrete'` only when *every* modifier is concrete for the active framework, `'ignored'` when any modifier is non-concrete (`each`/`skip`/`todo`/`fixme`), and otherwise `undefined`. A modifier that is neither — because it isn't in the active framework's concrete set nor in `NON_CONCRETE_TEST_MODIFIERS` — falls through to `undefined`, so the call is treated as *not a test at all* rather than as a concrete test. There is no "unknown modifier ⇒ ignore" fallback.

**Scenario** — a Vitest file using `test.fails(...)`. Vitest's real modifier is `fails`, but `VITEST_TEST_MODIFIERS` only lists `concurrent`/`sequential` (`failing` is Jest-only):

```ts
import { test, expect } from 'vitest';

test.fails('rejects 200', () => {
  const result = normalize(200);
  expect(result).toBe(-1);
});

test.fails('rejects 201', () => {
  const result = normalize(201);
  expect(result).toBe(-1);
});

test.fails('rejects 202', () => {
  const result = normalize(202);
  expect(result).toBe(-1);
});
```

`fails` is in no modifier set, so each call classifies as `undefined` and the trio is never grouped — a false negative on a perfectly parameterizable group. The same happens for any future/typo'd modifier name, which silently disables grouping for that test.

**Suggested fix:** add Vitest's `fails` (and audit the other frameworks' modifier lists for completeness), and consider treating an otherwise-unrecognized modifier as `'ignored'` rather than `undefined`, so unknown modifiers fail safe instead of silently dropping the test.

packages/analysis/src/jsts/rules/S5976/rule.ts
```diff
@@ -0,0 +1,506 @@
...
+  for (const child of childrenOf(node, context.sourceCode.visitorKeys)) {
+    collectLiteralNode(context, child, result);
+  }
+}
+
+function tokenRangeKey(token: { range?: [number, number] }): string {
+  return token.range?.join(':') ?? '';
+}
+
+function literalCategory(node: estree.Literal): LiteralCategory | undefined {
```
guillemsarda, line 464: ### Literal matching is more indirect than it needs to be

**File:** `packages/analysis/src/jsts/rules/S5976/rule.ts:390-460` (and `466-502`)

The comparator passed to `areEquivalent` receives **tokens**, but to report and parameterize we need the AST **`Literal` node**. Bridging token → node currently takes three pieces that all solve that one problem:

1. A second walk over the base body to build a token-range → node map:

```ts
this.literalNodesByTokenRange = collectLiteralNodes(context, baseBody); // walks baseBody again
```

2. Two parallel "what kind of literal is this" functions — one for tokens, one for nodes:

```ts
function literalCategory(node: estree.Literal) { /* uses typeof node.value */ }
function tokenLiteralCategory(token) { /* uses token.type: 'Numeric', 'String', ... */ }
```

3. An assertion that the two agree, just to recover the node:

```ts
const leftCategory = tokenLiteralCategory(left);
if (leftCategory === undefined || leftCategory !== tokenLiteralCategory(right)) return false;
const leftLiteral = this.literalNodesByTokenRange.get(tokenRangeKey(left));
if (leftLiteral === undefined || literalCategory(leftLiteral) !== leftCategory) return false; // glue
```

**Why this is worth a look:** the category is computed twice for the same literal, and the two functions must stay in lockstep. If someone later adds a new literal kind (e.g. templates or regex) and updates only one of them, the mismatch path just `return false`s — the literal silently stops being parameterized, with no error or failing test to catch it. The base body is also traversed twice.

**Suggested direction:** compare literals and collect parameter nodes directly during a single equivalence walk over **AST nodes**, where `node` and `node.value` are both already in hand. That removes the second traversal (`collectLiteralNodes`), the `literalNodesByTokenRange` map (and the `tokenRangeKey` `''` fallback), and one of the two category functions.

Not a bug — it works today — but it's indirection that's easy to break silently later. Lower priority than the correctness findings.

packages/analysis/src/jsts/rules/S5976/rule.ts
```diff
@@ -0,0 +1,506 @@
...
+    case 'Numeric':
+      return token.value.endsWith('n') ? 'bigint' : 'number';
+    case 'String':
+      return 'string';
+    default:
+      return undefined;
+  }
+}
+
+function isFunctionNode(node: estree.Node | estree.SpreadElement): node is FunctionNode {
```
guillemsarda: ### Local `isFunctionNode` shadows the shared helper with a different meaning

**File:** `packages/analysis/src/jsts/rules/S5976/rule.ts:504-506`

The rule defines its own `isFunctionNode`:

```ts
function isFunctionNode(node: estree.Node | estree.SpreadElement): node is FunctionNode {
  return node.type === 'FunctionExpression' || node.type === 'ArrowFunctionExpression';
}
```

But `helpers/ast.ts` already exports one with the same name:

```ts
// packages/analysis/src/jsts/rules/helpers/ast.ts:220
export function isFunctionNode(node: estree.Node): node is FunctionNodeType {
  return FUNCTION_NODES.includes(node.type); // FunctionDeclaration | FunctionExpression | ArrowFunctionExpression
}
```

The two have the **same name but different membership** — the local one deliberately excludes `FunctionDeclaration` (the rule only cares about callbacks). That's a legitimate narrowing, but two predicates sharing a name invites a maintainer to import or edit the wrong one, and the intent of the difference isn't obvious.

**Suggested direction:** give the local predicate a name that states the narrowing (e.g. `isFunctionExpressionNode` / `isCallbackFunction`), or build on the shared helper:

```ts
function isFunctionExpressionNode(node: estree.Node): node is FunctionNode {
  return isFunctionNode(node) && node.type !== 'FunctionDeclaration';
}
```

Not a bug — lower priority than the correctness findings.

guillemsarda (changes requested):

packages/analysis/src/jsts/rules/S5976/rule.ts
nathsou: I addressed the semantic bug this thread could make harder to see by comparing the resolved AST literal values before marking a literal as parameterizable. That removes the quote/numeric-format false positive without changing the shared equivalence helper.

I am intentionally leaving the broader AST-node equivalence refactor out of this PR. It would change a shared helper/design point rather than just S5976 behavior, and the current token-to-literal bridge is now covered by focused regressions for the cases that were actually failing. I would rather keep this PR scoped to the rule and revisit a cleaner AST-equivalence API separately if we see more rules needing it.

guillemsarda (approved):
Image

PR #7360, opened 2026-06-24
JS-1927: Create new rule S8784 - Assertions should be placed inside test cases or hooks

guillemsarda:
Part of JS-1927

----
## Summary by Gitar

- **New rule S8784:**
  - Implemented rule `S8784` to flag assertions located outside of test cases or hooks, with support for `mocha`, `jest`, `vitest`, `cypress`, `playwright`, `chai`, `sinon`, `supertest`, and `node:assert`.
  - Integrated robust detection for "runner-bound" vs "script-capable" assertions to handle standalone scripts correctly.
- **Refactoring:**
  - Centralized assertion detection logic into `assertion-detection.ts` to share logic between `S2699` and `S8784`.
  - Optimized `S2699` by utilizing the new shared assertion identification and updated `vitest` logic to treat setup helpers as compile-time assertions.
- **Configuration:**
  - Updated `Sonar way` profile to include `S8784` and removed deprecated rules (`S6766`, `S7735`, `S6774`) from various profiles.


packages/analysis/src/jsts/rules/helpers/assertion-detection.ts
```diff
@@ -0,0 +1,213 @@
...
+  while (current.type === 'MemberExpression') {
+    current = current.object;
+  }
+
+  if (current.type !== 'CallExpression') {
+    return false;
+  }
+
+  const innerCall = current;
+  return innerCall.callee.type === 'Identifier' && innerCall.callee.name.startsWith('expect');
```
nathsou: **False positive: any `expect*`-prefixed call is treated as an assertion.**

`isGlobalExpectExpressionJS` matches `IDENT(...).method()` whenever the inner callee name merely `.startsWith('expect')`, with no import/FQN resolution. In S2699 this only *suppressed* warnings, but S8784 *reports*, so the over-match now surfaces as a false positive.

Because `hasSupportedAssertionLibrary` is satisfied project-wide whenever `jest`/`cypress`/etc. is a devDependency, a non-test production file containing e.g. `expectation(config).validate();` or `expected(x).toThrow();` at top level gets flagged *"Move this assertion into a test case or a lifecycle hook."* (reproduced against the fixtures' `package.json`).

The prefix breadth exists to cover `expectObservable`/`expectSubscriptions`/`expectTypeOf`, but `startsWith('expect')` is far wider than that intent — consider matching a known set of `expect*` entry points (or resolving the binding) instead.

packages/analysis/src/jsts/rules/S8784/rule.ts
```diff
@@ -0,0 +1,225 @@
...
+ * Playwright check mirrors `isPlaywrightDescribe`'s FQN-then-name resolution so the
+ * two agree (notably for the `test` extended-fixtures pattern, where the FQN does
+ * not resolve to `@playwright/test`).
+ */
+function isTestStructureConstruct(context: Rule.RuleContext, call: estree.CallExpression): boolean {
+  return (
+    isMochaTestConstruct(context, call, SUITE_FUNCTION_NAMES) ||
+    isMochaTestConstruct(context, call, TEST_FUNCTION_NAMES) ||
+    (getPlaywrightTestQualifiers(context, call.callee) ??
+      getPlaywrightDescribeQualifiers(call.callee)) !== undefined
```
nathsou: **False positive: a locally-defined `test` is treated as test structure.**

`getPlaywrightDescribeQualifiers` matches `isIdentifier(base, 'test')` by name only (no binding check) and returns `[]` for any bare `test(...)`. Since `[] !== undefined`, this branch flips `hasTestStructure = true` for an unrelated local `function test(){}; test();`.

A genuinely standalone top-level assertion (e.g. `node:assert`) in such a file is then wrongly reported. This contradicts the invariant `local-describe.fixture.js` pins — locally-bound `describe`/`it` must *not* count — because `test` is the one name that slips through the Playwright name-only fallback. The same name-only resolution also affects `isPlaywrightDescribe` (suite-body classification) below.

packages/analysis/src/jsts/rules/S8784/rule.ts
```diff
@@ -0,0 +1,225 @@
...
+ * (`it`/`test`/`specify`), or any Playwright `test`/`test.describe` call. It reuses
+ * `isMochaTestConstruct`, so a locally-defined `describe`/`it` does not count. The
+ * Playwright check mirrors `isPlaywrightDescribe`'s FQN-then-name resolution so the
+ * two agree (notably for the `test` extended-fixtures pattern, where the FQN does
+ * not resolve to `@playwright/test`).
+ */
+function isTestStructureConstruct(context: Rule.RuleContext, call: estree.CallExpression): boolean {
+  return (
+    isMochaTestConstruct(context, call, SUITE_FUNCTION_NAMES) ||
+    isMochaTestConstruct(context, call, TEST_FUNCTION_NAMES) ||
```
nathsou: **False negative: `describe.each` / `test.each` / `it.each` are not recognized.**

`isMochaTestConstruct` rejects the `each` modifier (it is in none of the modifier sets in `mocha-style-test-frameworks.ts`), so every `.each(...)(...)` form fails both `isTestStructureConstruct` (here) and `isSuiteCallback` (below). Two misses follow:

1. An assertion directly in a `describe.each([...])('n', cb)` body is classified `inside-function` and silently skipped — the exact suite-body misplacement the rule targets.
2. A file whose only markers are `.each` forms never sets `hasTestStructure`, so a top-level script-capable assertion is mistaken for a standalone script.

`.each` is a first-class, common Jest/Vitest API and no fixture covers it. The fix likely belongs in the shared helper (gate `each` like the other framework modifiers), which would also benefit sibling rules.

nathsou (commented):
Automated review (xhigh) — three potential issues on the new S8784 rule, all verified against the code. Findings 1 and 2 are false positives stemming from name-based matching without binding/FQN resolution; finding 3 is a false-negative coverage gap for parameterized `.each` suites. None block the design.

packages/analysis/src/jsts/rules/helpers/assertion-detection.ts
nathsou: **False positive: any `expect*`-prefixed call is treated as an assertion.**

`isGlobalExpectExpressionJS` matches `IDENT(...).method()` whenever the inner callee name merely `.startsWith('expect')`, with no import/FQN resolution. In S2699 this only *suppressed* warnings, but S8784 *reports*, so the over-match now surfaces as a false positive.

Because `hasSupportedAssertionLibrary` is satisfied project-wide whenever `jest`/`cypress`/etc. is a devDependency, a non-test production file containing e.g. `expectation(config).validate();` or `expected(x).toThrow();` at top level gets flagged *"Move this assertion into a test case or a lifecycle hook."* (reproduced against the fixtures' `package.json`).

The prefix breadth exists to cover `expectObservable`/`expectSubscriptions`/`expectTypeOf`, but `startsWith('expect')` is far wider than that intent — consider matching a known set of `expect*` entry points (or resolving the binding) instead.

packages/analysis/src/jsts/rules/S8784/rule.ts
nathsou: **False positive: a locally-defined `test` is treated as test structure.**

`getPlaywrightDescribeQualifiers` matches `isIdentifier(base, 'test')` by name only (no binding check) and returns `[]` for any bare `test(...)`. Since `[] !== undefined`, this branch flips `hasTestStructure = true` for an unrelated local `function test(){}; test();`.

A genuinely standalone top-level assertion (e.g. `node:assert`) in such a file is then wrongly reported. This contradicts the invariant `local-describe.fixture.js` pins — locally-bound `describe`/`it` must *not* count — because `test` is the one name that slips through the Playwright name-only fallback. The same name-only resolution also affects `isPlaywrightDescribe` (suite-body classification) below.

packages/analysis/src/jsts/rules/S8784/rule.ts
nathsou: **False negative: `describe.each` / `test.each` / `it.each` are not recognized.**

`isMochaTestConstruct` rejects the `each` modifier (it is in none of the modifier sets in `mocha-style-test-frameworks.ts`), so every `.each(...)(...)` form fails both `isTestStructureConstruct` (here) and `isSuiteCallback` (below). Two misses follow:

1. An assertion directly in a `describe.each([...])('n', cb)` body is classified `inside-function` and silently skipped — the exact suite-body misplacement the rule targets.
2. A file whose only markers are `.each` forms never sets `hasTestStructure`, so a top-level script-capable assertion is mistaken for a standalone script.

`.each` is a first-class, common Jest/Vitest API and no fixture covers it. The fix likely belongs in the shared helper (gate `each` like the other framework modifiers), which would also benefit sibling rules.

nathsou (commented):
Automated review (xhigh) — three potential issues on the new S8784 rule, all verified against the code. Findings 1 and 2 are false positives stemming from name-based matching without binding/FQN resolution; finding 3 is a false-negative coverage gap for parameterized `.each` suites. None block the design.

packages/analysis/src/jsts/rules/helpers/assertion-detection.ts
nathsou: **False positive: any `expect*`-prefixed call is treated as an assertion.**

`isGlobalExpectExpressionJS` matches `IDENT(...).method()` whenever the inner callee name merely `.startsWith('expect')`, with no import/FQN resolution. In S2699 this only *suppressed* warnings, but S8784 *reports*, so the over-match now surfaces as a false positive.

Because `hasSupportedAssertionLibrary` is satisfied project-wide whenever `jest`/`cypress`/etc. is a devDependency, a non-test production file containing e.g. `expectation(config).validate();` or `expected(x).toThrow();` at top level gets flagged *"Move this assertion into a test case or a lifecycle hook."* (reproduced against the fixtures' `package.json`).

The prefix breadth exists to cover `expectObservable`/`expectSubscriptions`/`expectTypeOf`, but `startsWith('expect')` is far wider than that intent — consider matching a known set of `expect*` entry points (or resolving the binding) instead.

packages/analysis/src/jsts/rules/S8784/rule.ts
nathsou: **False positive: a locally-defined `test` is treated as test structure.**

`getPlaywrightDescribeQualifiers` matches `isIdentifier(base, 'test')` by name only (no binding check) and returns `[]` for any bare `test(...)`. Since `[] !== undefined`, this branch flips `hasTestStructure = true` for an unrelated local `function test(){}; test();`.

A genuinely standalone top-level assertion (e.g. `node:assert`) in such a file is then wrongly reported. This contradicts the invariant `local-describe.fixture.js` pins — locally-bound `describe`/`it` must *not* count — because `test` is the one name that slips through the Playwright name-only fallback. The same name-only resolution also affects `isPlaywrightDescribe` (suite-body classification) below.

packages/analysis/src/jsts/rules/S8784/rule.ts
nathsou: **False negative: `describe.each` / `test.each` / `it.each` are not recognized.**

`isMochaTestConstruct` rejects the `each` modifier (it is in none of the modifier sets in `mocha-style-test-frameworks.ts`), so every `.each(...)(...)` form fails both `isTestStructureConstruct` (here) and `isSuiteCallback` (below). Two misses follow:

1. An assertion directly in a `describe.each([...])('n', cb)` body is classified `inside-function` and silently skipped — the exact suite-body misplacement the rule targets.
2. A file whose only markers are `.each` forms never sets `hasTestStructure`, so a top-level script-capable assertion is mistaken for a standalone script.

`.each` is a first-class, common Jest/Vitest API and no fixture covers it. The fix likely belongs in the shared helper (gate `each` like the other framework modifiers), which would also benefit sibling rules.

nathsou (changes requested):
Automated review (xhigh) — three potential issues on the new S8784 rule, all verified against the code. Findings 1 and 2 are false positives stemming from name-based matching without binding/FQN resolution; finding 3 is a false-negative coverage gap for parameterized `.each` suites. None block the design.

nathsou (approved):
Image

PR #7361, opened 2026-06-24
Simplify the RSPEC refresh workflow

vdiez:
## Summary

- make `npm run bbf` a purely local build again: clear `lib`, regenerate TS metadata from the tracked JS rule JSON, and compile
- keep one explicit RSPEC refresh entrypoint, `npm run rspec:refresh`, implemented as a direct root Maven command instead of SonarJS JS wrapper logic
- configure `rspec-maven-plugin` once with both JavaScript and CSS targets, so one refresh pulls RSPEC once and generates both languages in one execution
- support one logical SonarJS RSPEC pin via root `rspec.sha`, with pinned refresh now selecting the SHA path correctly
- keep the tracked plugin JSON in Git for RSPEC site coverage/history and keep derived per-language `rspec.sha` files as release-traceability markers in published artifacts
- remove the old `ensure-rule-data` / `generate-meta:raw` layering and the former SonarJS-specific RSPEC wrapper scripts
- keep CI on a single prepare-once / reuse-many path for RSPEC data, including the nightly `Generated Files Freshness` workflow

## Why

The old `master` flow mixed two different use cases behind `npm run bbf`:

- normal local development, which should reuse the tracked JSON already in the repo and stay offline
- explicit RSPEC refresh, which is intentionally slower and needs auth because it pulls from `SonarSource/rspec`

That hidden layering is what made clean-checkout `bbf` surprisingly slow and hard to reason about.

This PR separates those use cases again while keeping the tracked JSON in Git for the RSPEC site coverage logic, the nightly freshness PR flow, and local builds.

The later follow-ups on top of the first draft were:

- move the multi-target JS/CSS refresh and optional root `rspec.sha` support into `rspec-maven-plugin`
- improve shared `sonar-rule-api` refresh performance
- fix CI consumers so every downstream job unpacks the prepared RSPEC artifact under the correct `sonar-plugin/` root
- make `JS/TS Ruling` generate proto files explicitly instead of depending on incidental prior state
- keep the nightly `Generated Files Freshness` path working against the same prepared RSPEC artifact model

## Requirements Assessment

After rechecking the implementation against the documented requirements, I think the branch now satisfies the intended RSPEC-sync contract well.

Validated:

- `npm run bbf` works immediately after `npm ci` from the tracked metadata already in the checkout
- default `bbf` is fast and offline; it does not check RSPEC freshness and it does not require GitHub auth
- SonarJS has one explicit refresh entrypoint (`npm run rspec:refresh`) plus an explicit refresh+build convenience command (`npm run bbf:latest`)
- the tracked plugin JSON stays in Git, which preserves the RSPEC GitHub Pages coverage / "Covered since" behavior
- SonarJS speaks in terms of one RSPEC pin (`rspec.sha`) instead of separate JS/CSS pins
- JavaScript and CSS stay on the same RSPEC refresh execution and the same logical pin
- both unpinned refresh and pinned refresh via root `rspec.sha` work from a clean `SONAR_USER_HOME`
- local explicit refresh works with ordinary developer auth (`gh auth login`) and with `GITHUB_TOKEN`
- `mvn clean` preserves committed JSON/profile files and removes only derived HTML / per-language `rspec.sha`
- normal module `generate-resources` no longer performs hidden RSPEC refresh; it regenerates metadata and Java classes only
- CI prepares RSPEC explicitly once and reuses the resulting artifact downstream instead of relying on hidden `bbf` behavior
- the nightly `Generated Files Freshness` job still works with the prepared-artifact model and remains the mechanism that keeps tracked generated files on `master` from drifting behind RSPEC
- `resources/rule-data-state.json` is no longer part of the normal lifecycle contract
- the per-language `rspec.sha` files remain derived outputs, but are now documented as release-traceability markers for historical rebuilds
- the shared-stack direction is preserved more strongly than before: the multi-language refresh and root pin file handling now live in `rspec-maven-plugin`, not in SonarJS JS glue

Still imperfect:

- repeated unpinned `npm run rspec:refresh` runs are still not a true no-op for long-lived automation workspaces because the shared branch-refresh path still does work on each refresh
- `deploy-rule-data.ts` is still a SonarJS-specific deployment layer for tracked JSON/profiles/derived `rspec.sha` files; this PR removes the refresh wrapper around Maven, but not that deployment layer

## Performance

Benchmark setup:

- clean worktree from current PR head for each clean scenario
- `npm ci` before the first timed run
- dedicated empty `SONAR_USER_HOME` for clean explicit-refresh scenarios; warm refresh reuses the same worktree and `SONAR_USER_HOME`
- master numbers below are carried forward from the earlier benchmark already recorded on this PR
- current-branch numbers below were remeasured after the later `rspec-maven-plugin` and `sonar-rule-api` changes on top of the first draft

| Scenario | master | current PR | delta |
| --- | ---: | ---: | ---: |
| Clean `npm run bbf` after `npm ci` | 75.77s | 11.66s | -84.6% |
| Warm `npm run bbf` | 14.84s | 11.19s | -24.6% |
| Clean explicit refresh | 65.87s | 38.52s | -41.5% |
| Warm explicit refresh | 43.20s | 20.58s | -52.4% |

Additional validation on the current branch:

- pinned refresh via root `rspec.sha` from a clean `SONAR_USER_HOME`: 37.19s
- `mvn -q -ntp -pl sonar-plugin/javascript-checks -am clean`: 1.37s
- `mvn -q -ntp -pl sonar-plugin/javascript-checks -am -DskipTests generate-resources`: 6.13s

A few important observations from those numbers:

- the main developer-path win is still the same: clean `npm run bbf` now only does `generate-meta -> bridge:compile`
- the later `rspec-maven-plugin` and `sonar-rule-api` work materially improved the explicit-refresh path as well, especially warm refresh
- JavaScript and CSS are refreshed in one plugin execution from one resolved RSPEC checkout
- the remaining explicit-refresh cost is now much smaller, but still dominated by the shared refresh path rather than SonarJS wrapper logic

## CI / Nightly Notes

- downstream CI jobs now consume the same prepared RSPEC artifact under `sonar-plugin/`, which keeps Linux, Windows, unit-test, ruling, knip, ESLint-plugin, and nightly freshness jobs on the same generated rule data for a given workflow run
- ordinary PR / branch CI still uses ephemeral refreshed rule data that is not committed back to the branch
- the nightly `Generated Files Freshness` workflow remains the separate automated path that opens or updates a PR when tracked generated files on `master` lag behind latest RSPEC

## Merge Blocker

This PR must **not** be merged yet.

Current dependency chain:

- SonarJS currently points at unreleased `rspec-maven-plugin` `1.5.0.145`
- the latest public `rspec-maven-plugin` release is still `1.4.0.122`
- `rspec-maven-plugin#85` currently depends on unreleased `sonar-rule-api` `2.25.0.5965`
- the latest public `sonar-rule-api` release is still `2.24.0.5949`

Required sequence:

1. release `sonar-rule-api` with the changes from `sonar-rule-api#308`
2. update `rspec-maven-plugin#85` to the final released `sonar-rule-api` version and release `rspec-maven-plugin`
3. update this SonarJS PR to the final released `rspec-maven-plugin` version
4. only then merge SonarJS

Reason:

- PR artifacts / temporary builds are not a stable merge target
- GitHub PR artifacts are cleaned up after 30 days
- SonarJS should merge only after it depends on final released artifacts in the chain above

## Possible Follow-Ups

- optimize the shared branch-refresh path further before adding any “quick latest head” precheck; that is still where most of the remaining explicit-refresh cost lives
- add a pinned early-exit for the FP-fixes-bot-style workflow when the requested root `rspec.sha` already matches the prepared outputs
- move more of the SonarJS-specific deployment logic into shared infrastructure if we ever want to simplify `deploy-rule-data.ts` further

I do not think checking the latest RSPEC head during every normal `npm run bbf` is the right next step: the default dev path should stay offline and deterministic.


francois-mora-sonarsource (approved):
image

One non-blocking follow-up: the new explicit RSPEC refresh contract is accepted, but a few documentation/command surfaces still point users at bare Maven builds and could be updated to avoid confusion:

- `README.md` still suggests `npm install` + `mvn clean install` as the build path. For a fresh/clean build, this should mention `npm run rspec:refresh` before the build.
- `docs/BUILD.md` still lists `mvn install` / `mvn clean install` as common full-build commands and later says to run a regular `mvn install`; these should clarify whether RSPEC artifacts must already be prepared, or point to `npm run rspec:refresh` / `npm run bbf:latest` for fresh refreshed builds.
- `docs/DEV.md` still tells contributors to use `mvn clean install` and `mvn clean install -DskipTests`; that section should include the refresh step after clean/fresh checkout.
- `.claude/skills/build/SKILL.md` still lists `mvn clean install` as a full clean build without the refresh prerequisite.
- `package.json` keeps `plugin:build`, `plugin:build:fast`, and `pbf` as bare Maven aliases; if these are meant to be fresh/clean-safe entrypoints, they may need a refreshed companion alias or updated documentation.

vdiez:
@francois-mora-sonarsource I pushed a follow-up for the remaining doc/command surfaces.

The intended contract stays the same: the main plugin build entry point is still Maven (`mvn install` / `mvn clean install`). `npm run bbf` is only the fast Node-focused path for bridge / metadata work; it is not meant to replace the normal Maven build.

I also removed the `plugin:*` / `pbf` npm wrappers so `package.json` stays Node-centric. When a refreshed Maven build is needed, the flow is now documented as `npm run rspec:refresh` followed by invoking Maven directly.

PR #7365, opened 2026-06-25
JS-1937 Implement rule S8932: Lodash memoize calls should provide an explicit cache key function

nathsou:
## Summary
- Implement S8932 to report Lodash/Lodash-ES/Underscore `memoize` calls when a resolvable memoized function has multiple runtime parameters and no explicit cache key function.
- Cover inline callbacks, local function references, import/require forms, missing `undefined`/`null` cache key function arguments, rest parameters, and TypeScript `this` parameters.
- Pin RSPEC data to SonarSource/rspec#7189 commit `f1d4ece05ecec5c851a000b3270122f71d07da96`.

## Test Plan
- `PATH=/Users/nathan.soufflet/.cache/codex-runtimes/codex-primary-runtime/dependencies/node/bin:$PATH ./node_modules/.bin/tsx --test packages/analysis/src/jsts/rules/S8932/**/*.test.ts`
- `PATH=/Users/nathan.soufflet/.cache/codex-runtimes/codex-primary-runtime/dependencies/node/bin:$PATH ./node_modules/.bin/tsc -p tsconfig-plugin.json --noEmit`
- `git diff --check`


packages/analysis/src/jsts/rules/S8932/rule.ts
```diff
@@ -0,0 +1,166 @@
...
+  }
+  return node.type === 'CallExpression';
+}
+
+function isMemoizeCall(
+  fullyQualifiedName: string,
+  syntacticMethod: typeof methodName | null,
+): boolean {
+  const parts = fullyQualifiedName.replaceAll('/', '.').split('.');
+  if (parts.length !== 2) {
```
guillemsarda, line 162: ### 1. Deep submodule imports are silently skipped → false negative

**File:** `packages/analysis/src/jsts/rules/S8932/rule.ts:147`

`isMemoizeCall` requires the fully-qualified name to split into **exactly two parts**:

```ts
const parts = fullyQualifiedName.replaceAll('/', '.').split('.');
if (parts.length !== 2) {
  return false;
}
```

But `getFullyQualifiedName` unshifts every `/`-segment of the import source (`module.ts:317`), so a deep submodule path produces more than two parts. `lodash/fp/memoize` is a real, shipped path and resolves to `lodash.fp.memoize` (3 parts) — which this check rejects.

**Scenario:**

```js
import memoize from 'lodash/fp/memoize';

memoize((amount, locale) => amount.toLocaleString(locale)); // should be Noncompliant, but is NOT reported
```

The FQN is `lodash.fp.memoize` → `parts.length === 3` → `isMemoizeCall` returns `false` → no report, even though this is genuine lodash `memoize` with multiple parameters and no resolver.

**Suggested fix:** match on the module and method segments instead of requiring exactly two parts:

```ts
const parts = fullyQualifiedName.split('.');
const moduleName = parts[0];
const qualifier = parts.at(-1);
return (
  supportedModules.has(moduleName) &&
  qualifier === methodName &&
  (syntacticMethod === null || qualifier === syntacticMethod)
);
```

(The `replaceAll('/', '.')` is also a no-op — `getFullyQualifiedName` never returns a string containing `/`.)

packages/analysis/src/jsts/rules/S8932/rule.ts
```diff
@@ -0,0 +1,166 @@
...
+}
+
+function getSyntacticMethodName(
+  callee: estree.Expression | estree.Super,
+): typeof methodName | null | undefined {
+  if (callee.type === 'ChainExpression') {
+    return getSyntacticMethodName(callee.expression);
+  }
+  if (callee.type === 'MemberExpression') {
+    if (callee.computed || !isIdentifier(callee.property)) {
```
guillemsarda: ### 2. Computed member access is skipped → false negative

**File:** `packages/analysis/src/jsts/rules/S8932/rule.ts:121`

`getSyntacticMethodName` returns `undefined` for any computed `MemberExpression` and bails *before* the fully-qualified name is computed:

```ts
if (callee.computed || !isIdentifier(callee.property)) {
  return undefined;
}
```

But `getFullyQualifiedName` resolves string-literal computed properties fine (`reduceTo` in `module.ts:434` handles `property.type === 'Literal'`), so the bail-out is the only reason this case is missed.

**Scenario:**

```js
import _ from 'lodash';

_['memoize']((amount, locale) => amount.toLocaleString(locale)); // should be Noncompliant, but is NOT reported
```

This is semantically identical to the Noncompliant fixture on line 21, yet produces no report. Low real-world frequency, but a true false negative.

**Suggested fix:** this is fixed for free if detection moves to the FQN-only idiom (see the redundant-machinery finding) — `getFullyQualifiedName` already handles computed string-literal access.

packages/analysis/src/jsts/rules/S8932/rule.ts
```diff
@@ -0,0 +1,166 @@
...
+): FunctionNodeType | null {
+  if (argument === undefined || argument.type === 'SpreadElement') {
+    return null;
+  }
+  return (
+    resolveFunction(context, argument) ??
+    getValueOfExpression(context, argument, 'ArrowFunctionExpression', true) ??
+    getValueOfExpression(context, argument, 'FunctionExpression', true) ??
+    null
+  );
```
guillemsarda, line 99: ### Coverage gap (not a bug): variable-bound arrow callbacks are untested

**File:** `packages/analysis/src/jsts/rules/S8932/rule.ts:94-99`

The two `getValueOfExpression` fallbacks in `resolveMemoizedFunction` are **not** dead code — they are the only path that resolves a callback passed by reference when it's a variable bound to an arrow/function expression. `resolveFunction` only resolves `function` *declarations*, so this case relies entirely on the fallback:

```ts
return (
  resolveFunction(context, argument) ??
  getValueOfExpression(context, argument, 'ArrowFunctionExpression', true) ??
  getValueOfExpression(context, argument, 'FunctionExpression', true) ??
  null
);
```

**Untested scenario:**

```js
const formatArrow = (amount, locale) => amount.toLocaleString(locale);
_.memoize(formatArrow); // should be Noncompliant — depends on the getValueOfExpression fallback
```

The fixtures only cover inline arrows/expressions and `function`-declaration references (`formatAmount`, `formatDate`). If the fallback chain regressed, the test suite would still pass green.

**Suggested fix:** add a fixture line for a variable-bound arrow/function expression passed by reference, so the fallback path is actually exercised.

packages/analysis/src/jsts/rules/S8932/rule.ts
```diff
@@ -0,0 +1,166 @@
...
+  meta: generateMeta(meta, {
+    messages: {
+      provideResolver: 'Provide an explicit function to compute the cache key.',
+    },
+  }),
+  create(context: Rule.RuleContext) {
+    return {
+      CallExpression(node: estree.Node) {
+        const call = node as estree.CallExpression;
+        const syntacticMethod = getSyntacticMethodName(call.callee);
```
guillemsarda, line 46: ### Detection machinery is redundant with the standard idiom (and is the root of findings 1 & 2)

**File:** `packages/analysis/src/jsts/rules/S8932/rule.ts:46-54, 114-156`

Sibling rules that match a specific library method — S7790, S6442, S2598 — call `getFullyQualifiedName(context, node)` once and test the result against a known set. S8932 instead computes a *syntactic* method name (`getSyntacticMethodName`), a *fully-qualified* name, and cross-checks them inside `isMemoizeCall`, plus a hand-rolled string parse:

```ts
const parts = fullyQualifiedName.replaceAll('/', '.').split('.');
```

The `replaceAll('/', '.')` is a no-op — `getFullyQualifiedName` already splits submodule paths on `/`, so the returned string never contains `/`.

**Why this is worth a look:** the bespoke syntactic pre-check is what causes the two false negatives:
- the `parts.length !== 2` restriction drops deep submodule paths like `lodash/fp/memoize` (finding 1);
- the `callee.computed` bail-out drops `_['memoize'](...)` (finding 2).

Both disappear if detection collapses to the standard idiom.

**Suggested direction:** reduce the whole detection to a single FQN check with membership on the module and method segments:

```ts
const parts = getFullyQualifiedName(context, call)?.split('.') ?? [];
if (parts.length < 2 || !supportedModules.has(parts[0]) || parts.at(-1) !== methodName) {
  return;
}
```

This removes `getSyntacticMethodName`, `isCallResult`, and `isMemoizeCall`, and fixes findings 1 and 2 in one move.

---

### Lesser cleanup

- **ChainExpression unwrapping is hand-rolled three times** — in `getSyntacticMethodName` (117-119), `isCallResult` (136-138), and `getReportNode` (159-161). A single unwrap helper, or reusing `reduceTo` from `module.ts`, would centralize it; today a change to chain handling must be applied in three places or they drift.
- **`node as estree.CallExpression` cast** (44-45) — type the visitor parameter directly (`CallExpression(node: estree.CallExpression)`), which is the more common idiom in the codebase, to drop the cast and keep type-checker help.

guillemsarda (changes requested):

packages/analysis/src/jsts/rules/S8932/rule.ts
nathsou: `formatAmount` already covered the variable-bound arrow-function case. The missing one was the variable-bound function expression case, which depends on the same fallback chain. I added that coverage in 8dd5f3f5d with `formatAmountExpression` in `cb.fixture.js`.

packages/analysis/src/jsts/rules/S8932/rule.ts
nathsou: This one is out of scope for this rule. `lodash/fp/memoize` has a different method signature: locally it is typed as `<T extends (...args: any) => any>(func: T) => T & lodash.MemoizedFunction`, so it does not expose the optional cache-key function that this rule is about. I’m keeping the detection limited to the Lodash and Underscore.js `memoize` APIs that support that second argument.

packages/analysis/src/jsts/rules/S8932/rule.ts
nathsou: The concrete false negative from finding 2 is fixed in 8dd5f3f5d. I’m not taking the broader refactor in this PR: the remaining part of finding 1 is out of scope here because `lodash/fp/memoize` is a different API surface and does not support the optional cache-key function this rule recommends.

packages/analysis/src/jsts/rules/S8932/rule.ts
nathsou: Fixed in 8dd5f3f5d. `getSyntacticMethodName` now accepts computed string-literal member access, `getReportNode` reports on the literal node for that form, and `cb.fixture.js` now covers `_['memoize'](...)`.

guillemsarda (approved):
Remove the rspec.sha
Image

PR #7370, opened 2026-06-26
JS-1942 Suppress S6582 when both comparison sides are nullable

erwan-leforestier-sonarsource:
## Summary
Suppresses S6582 when a comparison has multiple distinct nullable subjects (e.g. `!a || !b || a.length !== b.length`). The optional-chaining rewrite silently changes semantics in this pattern: the original treats "either side null" as an incompatibility condition, while `a?.length !== b?.length` returns `false` when both are null, inverting the guard.

RSPEC PR: https://github.com/SonarSource/rspec/pull/7200

## Changes
- Added `collectGuardedSubjects` helper that walks the guards side of a logical chain and collects identifiers null-guarded as `!x` (in `||` chains) or bare `x` (in `&&` chains)
- Added `matchesMultipleNullableSubjectsFalsePositive` which suppresses the report when both operands of the binary comparison are member expressions on distinct, guarded identifiers
- Wired the new check into `isFalsePositiveSuppression`
- Added unit tests covering both `null` and `undefined` nullable types, plus a positive case ensuring the rule still reports when only one comparison side is nullable

## Functional Validation
Artifact: [JS-1942-fv.zip](https://github.com/user-attachments/files/29508200/JS-1942-fv.zip)

    ./run.sh

Expected output:
```text
Analyzing "sample.ts"...
Results:
    - Rule "S6582" -> sample.ts:L.<line>

****** Branch "fix/js-1942-gate-s6582-es2020" ******
Analyzing "sample.ts"...
Results:
    (none)
```

### DIFF
--- a/packages/analysis/src/jsts/rules/S6582/rule.ts
+++ b/packages/analysis/src/jsts/rules/S6582/rule.ts
@@ -78,23 +78,6 @@ function findReportNode(
  * let result: string | undefined;
  * result = value?.property;
  */
-// Collects identifiers that are null-guarded in a logical chain:
-// `!x` in || chains and bare `x` in && chains.
-function collectGuardedSubjects(node: Rule.Node, subjects: Set<string>): void {
-  if (
-    node.type === 'UnaryExpression' &&
-    (node as Rule.Node & { operator: string }).operator === '!' &&
-    (node as Rule.Node & { argument: Rule.Node }).argument?.type === 'Identifier'
-  ) {
-    subjects.add((node as Rule.Node & { argument: Rule.Node & { name: string } }).argument.name);
-  } else if (node.type === 'Identifier') {
-    subjects.add((node as Rule.Node & { name: string }).name);
-  } else if (node.type === 'LogicalExpression') {
-    collectGuardedSubjects(node.left, subjects);
-    collectGuardedSubjects(node.right, subjects);
-  }
-}
-
 function allowsUndefined(type: ts.Type): boolean {
   const constituents: ts.Type[] = type.isUnion() ? type.types : [type];
   return constituents.some(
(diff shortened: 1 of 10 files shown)


packages/analysis/src/jsts/rules/S6582/rule.ts
```diff
@@ -89,6 +89,19 @@ function allowsUndefined(type: ts.Type): boolean {
   );
 }
 
+function allowsNullOrUndefined(type: ts.Type): boolean {
+  const constituents: ts.Type[] = type.isUnion() ? type.types : [type];
+  return constituents.some(
+    constituent =>
+      (constituent.flags & ts.TypeFlags.Null) !== 0 ||
+      (constituent.flags & ts.TypeFlags.Undefined) !== 0,
```
francois-mora-sonarsource, line 97: We already treat `void` like `undefined` in `allowsUndefined`, but not here.

So for `T | void`, S6582 is considered safe in the existing undefined-safety check, while this new nullable-subject check still ignores it. That leaves the same multi-subject false positive unsuppressed for `void`-typed subjects.

Could we include `TypeFlags.Void` here as well and add a small regression test?

erwan-leforestier-sonarsource: Adding `void` to `allowsNullOrUndefined`: it's unreachable. Upstream's `isValidFalseBooleanCheckType` only allows Null | Undefined | Object, so a T | void subject is filtered out before it hits our suppressor. No FP to fix, no regression test possible. Left the flags as-is, by I agree that names of these 2 functions are really confusing because they have different targets.

I renamed both to make the intent obvious:

- allowsUndefined → targetAcceptsUndefined: "is undefined assignable to the target type?" (includes any/unknown/void).
- allowsNullOrUndefined → subjectIsNullable: "does the subject's type explicitly include null/undefined?" (excludes them).

francois-mora-sonarsource (approved):

PR #7371, opened 2026-06-26
JS-1928 Fix S2871 FP on sort normalization comparisons

francois-mora-sonarsource:
## Summary
- narrow S2871 so sort-based normalization comparisons are ignored only for supported safe patterns
- cover the new S2871 cases with additional unit tests, including helper-based comparisons, string join serialization, and escaped normalizer references
- move the string-join helper to module scope to satisfy Sonar

## Validation
- `npx tsx --tsconfig packages/tsconfig.test.json --test packages/analysis/src/jsts/rules/S2871/unit.test.ts`
- `npm run bridge:build:fast`


erwan-leforestier-sonarsource (commented):
`packages/analysis/src/jsts/rules/S2871/false-positives/compared-normalizer.ts` lines 122–137

`isUsedOnlyInJsonStringifyComparisons` only checks where the helper is **called** (are all reads inside `JSON.stringify()` comparisons?), not what the helper body actually does. A helper like:

```js
function normalize(arr) {
  if (arr.length > 2) return arr.sort();
  return arr;
}
JSON.stringify(normalize(a)) === JSON.stringify(normalize(b))
```

passes the call-site check — both reads are inside a `JSON.stringify()` equality — so the `.sort()` is silenced. But this helper is not a symmetric canonicalizer: when `a.length <= 2` and `b.length > 2`, one side is sorted and the other isn't. That is exactly the asymmetric-sort case S2871 is meant to catch.

Should the exemption also verify that the helper body always, unconditionally returns the sort result (e.g. single-expression body, or a single `return arr.sort()` with no other return path)?

erwan-leforestier-sonarsource (commented):
`packages/analysis/src/jsts/rules/S2871/false-positives/compared-normalizer.ts` lines 106–117

`getLocalFunctionVariable` correctly rejects `export function normalize()` because a `FunctionDeclaration`'s direct parent is `ExportNamedDeclaration`. But for `export const normalize = (arr) => arr.sort()` the `ArrowFunctionExpression`'s direct parent is `VariableDeclarator`, not the export node — so the guard on line 107 never fires and the function is treated as local.

If that exported arrow has at least one read inside a `JSON.stringify()` equality, `isUsedOnlyInJsonStringifyComparisons` returns `true` and its sort is silently suppressed — even though the function is exported and could be imported by another file for actual ordering.

The fix is to walk up one more level: when the parent is `VariableDeclarator`, also check whether its `VariableDeclaration` sits inside an `ExportNamedDeclaration`.

erwan-leforestier-sonarsource (commented):
`packages/analysis/src/jsts/rules/S2871/false-positives/compared-normalizer.ts` lines 106–117

`getLocalFunctionVariable` correctly rejects `export function normalize()` — a `FunctionDeclaration`'s direct parent is `ExportNamedDeclaration`, so the guard on line 107 fires. But for an exported arrow function the AST is different:

```
ExportNamedDeclaration
  └── VariableDeclaration
        └── VariableDeclarator   ← direct parent of the arrow
              └── ArrowFunctionExpression
```

The guard only looks one level up (to `VariableDeclarator`), misses the export, and treats the function as file-private. So this case is incorrectly suppressed:

```js
// file-a.ts
export const normalize = (arr: number[]) => arr.sort();
JSON.stringify(normalize(a)) === JSON.stringify(normalize(b))  // only local read → suppressed

// file-b.ts
import { normalize } from './file-a';
const sorted = normalize([3, 1, 2]);  // actual ordering use — S2871 should have fired
```

The fix is to also walk up from `VariableDeclarator` → `VariableDeclaration` and return `null` if that sits inside an `ExportNamedDeclaration`.

erwan-leforestier-sonarsource (commented):
`packages/analysis/src/jsts/rules/S2871/unit.test.ts` lines 541–583

The `invalid` section covers exported `function` declarations:

```js
// already tested — correctly reported
export function normalize(arr) { return arr.sort(); }
JSON.stringify(normalize(a)) === JSON.stringify(normalize(b))
```

But there is no equivalent test for the exported arrow form:

```js
// missing — should be reported, but currently suppressed
export const normalize = (arr: number[]) => arr.sort();
JSON.stringify(normalize(a)) === JSON.stringify(normalize(b))
```

Adding this to the `invalid` section would document the expected behavior and guard against regressions in the export-detection logic.

francois-mora-sonarsource:
@erwan-leforestier-sonarsource thanks, I tightened the suppression boundaries along the lines you pointed out:

- For the conditional-return normalizer case, the compared-normalizer escape is now intentionally limited to trivial one-parameter wrappers: either an expression-bodied arrow returning the reported sort call directly, or a block body with exactly one `return <reported sort call>;`. Helpers with branches such as `if (...) return arr.sort(); return arr;` now keep the S2871 issue. This is covered by invalid tests for multi-return normalizers and derived sort results in `7b1728f33`.

- For `export const normalize = (...) => ...`, the local-function lookup now checks the `VariableDeclarator -> VariableDeclaration -> ExportNamedDeclaration` parent chain, so exported arrow/function-expression normalizers are not treated as local. The missing exported-arrow cases are covered for both `sort()` and `toSorted()` in the invalid tests; the export detection fix is in `a36cf48d5`.

- I also made the join-based normalization escape more conservative in `893c31d46`: it now only suppresses for `number[]` / `bigint[]` receivers with a safe non-empty punctuation-style separator. `string[]`, `join('')`, and numeric-looking separators remain reported.

francois-mora-sonarsource:
Replying to https://github.com/SonarSource/SonarJS/pull/7371#pullrequestreview-4601328043

Good point. I narrowed this escape so we no longer suppress helpers with conditional or multiple return paths.

The compared-normalizer is now accepted only when the helper is a trivial one-parameter wrapper around the reported sort call: either an expression-bodied arrow, or a block body with exactly one `return <reported sort call>;`. So a helper like:

```js
function normalize(arr) {
  if (arr.length > 2) return arr.sort();
  return arr;
}
```

continues to raise S2871. This is covered by the new invalid tests for multi-return normalizers and derived sort results in `7b1728f33`.

francois-mora-sonarsource:
Replying to https://github.com/SonarSource/SonarJS/pull/7371#pullrequestreview-4601469182

Fixed. `getLocalFunctionVariable` now handles the `export const normalize = (...) => ...` shape by walking from the `VariableDeclarator` to its `VariableDeclaration`, then rejecting the helper when that declaration is inside an `ExportNamedDeclaration` or `ExportDefaultDeclaration`.

That means exported arrow/function-expression normalizers are no longer treated as local, so their comparator-less `sort()` / `toSorted()` calls keep reporting even if there is also a local `JSON.stringify(normalize(a)) === JSON.stringify(normalize(b))` comparison.

Implemented in `a36cf48d5` and covered by exported-arrow invalid tests.

francois-mora-sonarsource:
Replying to https://github.com/SonarSource/SonarJS/pull/7371#pullrequestreview-4601476178

Yes, this AST shape was the missing case. The implementation now explicitly checks:

```text
ArrowFunctionExpression -> VariableDeclarator -> VariableDeclaration -> ExportNamedDeclaration
```

and returns `null` for the helper variable when the variable declaration is exported. So the example with `export const normalize = (arr: number[]) => arr.sort()` is no longer suppressed; S2871 is raised because the helper can be consumed from another file where ordering matters.

This is in `a36cf48d5`.

francois-mora-sonarsource:
Replying to https://github.com/SonarSource/SonarJS/pull/7371#pullrequestreview-4601731972

Added the missing exported-arrow regression coverage.

The invalid tests now include exported arrow normalizers for both families:

```ts
export const normalize = (value: string[]): string[] => value.sort();
```

and

```ts
export const normalize = (value: string[]): string[] => value.toSorted();
```

Both continue to report S2871 even when used locally in the `JSON.stringify(normalize(before)) !== JSON.stringify(normalize(after))` comparison. The tests landed with the export-detection fix in `a36cf48d5`, with the later stricter compared-normalizer boundary in `7b1728f33` still preserving those cases.

erwan-leforestier-sonarsource (approved):

PR #7389, opened 2026-06-29
JS-1950 Fix S5906 ambiguous Jasmine length assertions

nathsou:
## Summary
- Prefer explicit assertion-framework imports over project-wide dependency signals.
- Suppress S5906 length-equality suggestions for global expect() when Jasmine and Jest dependency signals are both present.
- Add mixed Jasmine/Jest fixture coverage for ambiguous globals and explicit Jasmine/Vitest imports.

## Community report
- https://community.sonarsource.com/t/eslint-prefer-specific-assertions-wants-me-to-use-tohavelength-but-i-cant/184082

## Verification
- ./node_modules/.bin/tsx --test packages/analysis/src/jsts/rules/S5906/**/*.test.ts
- ./node_modules/.bin/tsc -b packages

Jira: JS-1950

### DIFF
--- a/packages/analysis/src/jsts/rules/S5906/fixtures/mixed-jasmine-jest/package.json
+++ b/packages/analysis/src/jsts/rules/S5906/fixtures/mixed-jasmine-jest/package.json
@@ -1,0 +1,6 @@
+{
+  "devDependencies": {
+    "jasmine-core": "5.7.1",
+    "jest": "30.0.0"
+  }
+}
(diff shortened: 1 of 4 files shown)


francois-mora-sonarsource (approved):
Approved. The type-import/global-injection concern does not block this PR: a type-only Jasmine import can be potential evidence that Jasmine globals are available, so it is reasonable to keep treating it as an ambiguity signal rather than requiring runtime binding evidence for this fix.

PR #7395, opened 2026-06-30
JS-1957 JS-1958 Split S2925 and add S8959

nathsou:
## Summary
- narrow S2925 to fixed waits in Cypress and Playwright tests
- add S8959 for Cypress and Playwright UI test debug commands
- pin `rspec.sha` to `3abaa64c0f0b6fa7efa72c5c04b8b546a176d8aa` so CI resolves the matching RSPEC change

## RSPEC
- SonarSource/rspec#7258

## Test Plan
- npm run generate-meta
- npm run generate-java-rule-classes
- npm run count-rules
- ./node_modules/.bin/tsx --tsconfig packages/tsconfig.test.json --test packages/analysis/src/jsts/rules/S2925/unit.test.ts packages/analysis/src/jsts/rules/S8959/unit.test.ts


packages/analysis/src/jsts/rules/S8959/rule.ts
```diff
@@ -0,0 +1,71 @@
...
+
+import type { Rule } from 'eslint';
+import type estree from 'estree';
+import { isIdentifier, isMethodCall } from '../helpers/ast.js';
+import { chainStartsWithCy } from '../helpers/cypress.js';
+import { generateMeta } from '../helpers/generate-meta.js';
+import { isTestRelatedFile } from '../helpers/test-file-pattern.js';
+import * as meta from './generated-meta.js';
+
+export const rule: Rule.RuleModule = {
```
guillemsarda, line 28: I think it's worth adding the quickfix.

guillemsarda: https://github.com/SonarSource/rspec/pull/7258/changes#r3506341939

packages/analysis/src/jsts/rules/S8959/unit.test.ts
```diff
@@ -0,0 +1,138 @@
...
+import { rule } from './rule.js';
+
+const ruleTester = new NoTypeCheckingRuleTester();
+
+describe('S8959', () => {
+  it('S8959', () => {
+    ruleTester.run('UI test debug commands should not be committed to version control', rule, {
+      valid: [
+        {
+          code: `
```
guillemsarda, line 28: This test case is missing: `cy?.pause()` and `cy?.get(...).debug()` (optional chaining) are not covered. The
`chainStartsWithCy` helper handles `ChainExpression` nodes, so these would be correctly flagged — but without a test, a future refactor removing that branch would go undetected:

  ```js
  cy?.pause();
  cy?.get('button.save').debug();

packages/analysis/src/jsts/rules/S8959/rule.ts
```diff
@@ -0,0 +1,71 @@
...
+      },
+    };
+  },
+};
+
+function isUiTestDebugCommand(callee: estree.MemberExpression & { property: estree.Identifier }) {
+  const { object, property } = callee;
+  switch (property.name) {
+    case 'pause':
+      return chainStartsWithCy(object) || isIdentifier(object, 'page');
```
guillemsarda, line 59: **Playwright `page.pause()` detection is name-based** — the rule uses `isIdentifier(object, 'page')`, which only matches when the receiver is literally named `page`. Common Playwright patterns would be silently missed:

```js
// renamed variable — not flagged
const p = await browser.newPage();
p.pause();

// BrowserContext also has pause() — not flagged
await page.context().pause();
```

Is this scope intentional, or should we consider a dependency-based detection similar to how Cypress is handled via chainStartsWithCy?

guillemsarda (changes requested):

packages/analysis/src/jsts/rules/S8959/unit.test.ts
nathsou: Added an invalid case covering `cy?.pause()` and `cy?.get('button.save').debug()`.

One correction for the record: I traced the AST and neither snippet actually exercises the `ChainExpression` branch in `chainStartsWithCy` — `ChainExpression` only wraps the outermost node of an optional chain, so both calls resolve through the plain `Identifier`/`MemberExpression`/`CallExpression` branches instead. The `ChainExpression` branch would only be hit by something like `(cy?.get('x')).pause()` (a chain broken by parens). Keeping the added test as useful real-world coverage regardless — let me know if you'd like the parenthesized case added too to actually cover that branch.

packages/analysis/src/jsts/rules/S8959/rule.ts
nathsou: This is intentional and actually mirrors the Cypress side: `chainStartsWithCy` is also purely name-based (`isIdentifier(node, 'cy')`), not dependency- or type-based, so `const c = cy; c.pause()` is equally unflagged today. Both frameworks rely on matching the conventional identifier (global `cy`, destructured `page` fixture), consistent with the RSPEC examples. A real fix for renamed variables / `BrowserContext.pause()` would need type information, and this rule deliberately uses `NoTypeCheckingRuleTester`. Not planning to expand scope here, but open to revisiting if you feel strongly.

packages/analysis/src/jsts/rules/S8959/rule.ts
nathsou: Update: reclassified RSPEC S8959 as `quickfix: covered` (rspec#7258, SonarSource/rspec@d51bc4d) and bumped the pinned `rspec.sha` here so `generate-meta` picks it up. Implemented the ESLint fix in `rule.ts`:
- standalone calls (`cy.pause();`, `await page.pause();`) remove the whole statement
- mid-chain calls (`cy.get('button.save').debug().click();`) splice out just the `.debug()`/`.pause()` link, matching the RSPEC compliant example

Added regression tests with `output` assertions covering both cases. All green locally (`node:test` + `tsc -b packages`).

packages/analysis/src/jsts/rules/S8959/rule.ts
```diff
@@ -0,0 +1,111 @@
...
+): Rule.Fix {
+  const statement = unwrapToStatementExpression(call);
+  if (hasParent(statement)) {
+    const { parent } = statement;
+    if (parent.type === 'ExpressionStatement' && parent.expression === statement) {
+      return removeNodeWithLeadingWhitespaces(context, parent, fixer);
+    }
+  }
+  return fixer.removeRange([call.callee.object.range![1], call.range![1]]);
+}
```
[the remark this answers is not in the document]
nathsou: Fixed in `fade66e5c`. `removeDebugCommand` now deletes the whole statement only when the receiver is the root `cy`/`page` call, so tail-of-chain Cypress cases keep the preceding command (`cy.get(...).pause()` -> `cy.get(...);`). Added regression coverage for both tail-of-chain `pause()` and `debug()` forms, including optional chaining, and re-ran the focused `S8959` test plus `validate-quickfix` and `tsc -b packages` locally.

guillemsarda (approved):
```
          .     .
         / \___/ \
        /  o   o  \
       |     ^     |
       |   \___/   |
        \         /
         \_______/
        .-'/   \'-.
       /  /|   |\  \
      /__/ |___| \__\
         _/   \_
        /_\   /_\
```

PR #7396, opened 2026-06-30
JS-1960: S8960 - Test and hook callbacks should use a single completion style

nathsou:
----
## Summary by Gitar

- **New Rule:**
  - Implemented `S8960` to detect mixed completion styles (e.g., `async` combined with a `done` callback) in test and hook functions.
- **Framework Support:**
  - Added support for `Jest`, `Mocha`, and `Jasmine` test frameworks.
- **Configuration:**
  - Registered `S8960` in the `Sonar way` quality profile.


packages/analysis/src/jsts/rules/S8960/rule.ts
```diff
@@ -0,0 +1,217 @@
...
+
+function getCallback(node: estree.CallExpression): FunctionNode | undefined {
+  const callback = node.arguments.find(
+    argument =>
+      argument.type === 'FunctionExpression' || argument.type === 'ArrowFunctionExpression',
+  );
+  return callback?.type === 'FunctionExpression' || callback?.type === 'ArrowFunctionExpression'
+    ? callback
+    : undefined;
+}
```
guillemsarda: `getCallback` re-checks `callback?.type === 'FunctionExpression' || 'ArrowFunctionExpression'` after `.find` has already narrowed to exactly those two types — the second check can never fail.

packages/analysis/src/jsts/rules/S8960/rule.ts
```diff
@@ -0,0 +1,217 @@
...
+          allowGlobalConstructs,
+        );
+        if (constructKind === undefined) {
+          return;
+        }
+
+        const asyncToken = context.sourceCode.getFirstToken(callback);
+        if (asyncToken?.value !== 'async') {
+          return;
+        }
```
guillemsarda: This check is redundant because you have already checked in line 81 that the callback is async. This will never be false here.

packages/analysis/src/jsts/rules/S8960/rule.ts
```diff
@@ -0,0 +1,217 @@
...
+  'before',
+  'after',
+  'beforeEach',
+  'afterEach',
+  'beforeAll',
+  'afterAll',
+]);
+const SUPPORTED_CONSTRUCT_NAMES = new Set([...TEST_FUNCTION_NAMES, ...HOOK_FUNCTION_NAMES]);
+const SUPPORTED_MODULE_FQNS = ['jest', '@jest.globals', 'mocha', 'jasmine'];
+const COMMON_TEST_MODIFIERS = new Set(['only']);
```
guillemsarda, line 53: Two test constructs that accept a `done` callback are not currently covered:

- **`xit` / `xtest` / `fit` aliases** — Jest exposes these as standalone globals (`xit`, `xtest`
  as skip aliases; `fit` as a focus/only alias — see [Jest API docs](https://jestjs.io/docs/api#xittestname-fn-timeout)).
  Jasmine does the same ([Jasmine Global API](https://jasmine.github.io/api/edge/global)).
  They accept the exact same callback signature as `it`/`test` but are not in
  `TEST_FUNCTION_NAMES`, so `async (done) => {}` inside them goes undetected.

- **`it.each` / `test.each`** — Both Jest and Jasmine allow `done` as the *last* parameter
  in parameterized tests (see [jest-each on npm](https://www.npmjs.com/package/jest-each)):

  ```js
  it.each([1, 2])('test %i', async (val, done) => {
    await doWork(val);
    done();
  });```

`each` is not a recognised modifier so these calls are silently skipped. This one is
  trickier since `getCompletionCallbackParameter` checks the first param, but here `done ` is last. Worth a conscious decision on whether to support it.

guillemsarda (changes requested):
Just raised some questions

packages/analysis/src/jsts/rules/S8960/rule.ts
nathsou: Added alias coverage for `fit`, `xit`, and `xtest` in `02a66f1f4`.

I left parameterized `.each(...)` out of scope for this rule on purpose: the current implementation only targets simple outer callbacks with a first-parameter completion callback, and supporting `.each` would mean broadening that detection to trailing `done` parameters.

packages/analysis/src/jsts/rules/S8960/rule.ts
nathsou: Fixed in `02a66f1f4`: `getCallback` now delegates the narrowing to `isCallback`, so the redundant post-`find` type check is gone.

packages/analysis/src/jsts/rules/S8960/rule.ts
nathsou: Fixed in `02a66f1f4`: the extra async guard is gone; after the early `!callback?.async` return we now just locate the `async` token and report.

packages/analysis/src/jsts/rules/S8960/rule.ts
nathsou: Keeping `.each(...)` out of scope in this PR. The alias gap is fixed in `02a66f1f4`, but parameterized callbacks are not uniformly `(...data, done)`: in Jest the completion callback only exists for some table shapes, while tuple/array rows are pure data parameters, and Jasmine does not expose `it.each` / `test.each` in its global API. Since this rule is purely syntactic, treating the trailing identifier in `.each(...)` callbacks as a completion callback would create false positives on multi-column tables. If we want `.each(...)` support, it should come in a dedicated follow-up with table-shape analysis.

guillemsarda (approved):
Image

PR #7401, opened 2026-06-30
JS-1947 Group S2933 issues by class

francois-mora-sonarsource:
## Summary

Recreates the closed work from #7366 on top of the latest `origin/master`, using the new JS Jira key instead of the old `AT-23` branch.

This groups `S2933` (`typescript-eslint/prefer-readonly`) reports by class, preserving each readonly-eligible member as a secondary location and keeping the aggregate quick fix for adding `readonly`.

The branch also carries the nyc ESM loader source-map fallback from the original PR, adapted to the newer `SourceMaps` cache/content-hash implementation now present on `master`.

## Notes

The old generated rule JSON changes already match current `origin/master`. The profile files from the old PR were not replayed because current `master` contains newer profile entries and regenerating rule data produced no profile diff.

## Validation

- `npm install`
- `npm run generate-rule-data:maven`
- `npm run generate-meta`
- `npx tsx --tsconfig packages/tsconfig.test.json --test packages/analysis/src/jsts/rules/S2933/unit.test.ts packages/analysis/tests/jsts/linter/index.test.ts packages/analysis/tests/tools/nyc-esm-loader.test.ts`
- `npx tsc -b packages`
- `npm run bbf`


francois-mora-sonarsource:
Follow-up note on inline ESLint suppressions: this grouping decorator intercepts raw upstream `context.report()` calls and re-reports one grouped issue on the class. ESLint applies inline disable comments only after rule execution, so a member-level disable such as `eslint-disable-next-line @typescript-eslint/prefer-readonly` would not match the grouped class-level report.

I do not think this should block JS-1947: inline suppressions for this rule are expected to be uncommon, and the PR improves the default reporting experience substantially. Handling this properly also deserves a generic solution for decorators that relocate or aggregate reports, instead of special-casing S2933 or duplicating ESLint directive semantics.

Tracked as JS-2001: add a reusable metadata-driven guard so decorators can opt out of custom grouping/relocation when relevant inline ESLint disables are present, forwarding original reports and letting ESLint apply its native suppression logic.

packages/analysis/src/jsts/rules/S2933/decorator.ts
```diff
@@ -0,0 +1,334 @@
...
+
+function findInnermostClassContaining(classes: ClassNode[], index: number): ClassNode | null {
+  let owner: ClassNode | null = null;
+
+  for (const classNode of classes) {
+    if (
+      classNode.range[0] <= index &&
+      index <= classNode.range[1] &&
+      (!owner || classNode.range[0] >= owner.range[0])
+    ) {
```
erwan-leforestier-sonarsource: Worth adding a comment about what this condition is checking because it's not clear to me

packages/analysis/src/jsts/rules/S2933/decorator.ts
```diff
@@ -0,0 +1,334 @@
...
+  report(
+    context,
+    {
+      loc: primaryLocation.loc,
+      message: 'Mark these members as `readonly`.',
+      ...(fix && { fix }),
+    },
+    secondaryLocations,
+  );
+}
```
erwan-leforestier-sonarsource: I saw that you tried to extract this `reportGroupedIssue` in a reusable helper in this commit: http://github.com/SonarSource/SonarJS/pull/7401/changes/bf16d3e62c8ba32c120dcbc04416bf004fa21efc
But it was then reverted. Why not having this in a separate helper ?

erwan-leforestier-sonarsource (approved):

PR #7404, opened 2026-06-30
JS-1731 Fix S6551 typeof guard false positives

francois-mora-sonarsource:
## Summary

Fixes JS-1731 by adding a narrow S6551 decorator suppression for identifiers used in implicit string contexts when an enclosing `if` consequent is guarded by `typeof` checks proving the same bound variable is primitive.

The suppression handles primitive equality checks, reversed comparisons, `typeof value !== "object"`, conjunctions, and disjunctions. It resolves identifiers through ESLint scope so same-name shadowing does not suppress incorrectly, and it rejects suppression after reassignment before the report site.

## Why

`@typescript-eslint/no-base-to-string` can still report after guards such as `typeof value !== "object"` because TypeScript may expose a broad narrowed type like `{}` / `{} | undefined`, even though runtime object stringification cannot occur in the guarded branch.

## Tests

- `npx tsx --tsconfig packages/tsconfig.test.json --test --test-reporter=spec --test-reporter-destination=stdout packages/analysis/src/jsts/rules/S6551/cb.test.ts`
- `npm run bridge:compile`
- `git diff --check`
- `npx prettier --check packages/analysis/src/jsts/rules/S6551/decorator.ts packages/analysis/src/jsts/rules/S6551/cb.fixture.ts`

----
## Summary by Gitar

- **CI/CD pipeline updates:**
  - Modified `build.yml` to export and validate JavaScript coverage reports via `actions/upload-artifact` and `actions/download-artifact`.
  - Removed redundant coverage caching logic in favor of direct artifact propagation between CI jobs.
- **Refactored S6551 rule logic:**
  - Extracted false positive suppression logic into a dedicated `false-positives/` module to improve maintainability and support complex conditional checks.


francois-mora-sonarsource:
## Review feedback

Reviewed the diff — logic checks out overall (scope resolution, shadowing, reassignment/destructuring/loop invalidation, and the conjunction/disjunction semantics in `conditionMakesBranchSafeToStringify` all behave correctly against the fixture cases). Two minor, non-blocking points:

1. **Likely dead code** — `isImplicitlyStringified` in [`guarded-by-typeof-check.ts`](packages/analysis/src/jsts/rules/S6551/false-positives/guarded-by-typeof-check.ts#L104-L127) climbs through `TSAsExpression`/`TSTypeAssertion`/`TSNonNullExpression` parents to "transparently walk through TypeScript-only wrappers." But the caller (`isGuardedByTypeofCheckFalsePositive`) already requires `node.type === 'Identifier'` before calling in. Per the upstream `no-base-to-string` rule, when a template-literal/`+` operand is wrapped (e.g. `` `${value as any}` `` or `` `${value!}` ``), the reported node is the wrapper itself, not the inner identifier — so by the time this loop runs, `node.parent` being a TS wrapper shouldn't occur from a real report. Worth a quick `RuleTester` check with a wrapped-identifier fixture case; if confirmed dead, the branch (and its comment) could be dropped for clarity.

2. **Completeness gap (safe direction)** — the same function only recognizes `TemplateLiteral` and `BinaryExpression` (`+`) as implicit-stringification contexts, not `AssignmentExpression` with `+=`. So a guarded case like `if (typeof value !== 'object') { rendered += value; }` (no template literal) stays unsuppressed even though it's just as safe as the tested `` rendered += `${value}` `` case. This only means a missed suppression, never an incorrect one, so it's safe but incomplete relative to the PR's goal — could be a good follow-up if worth the extra branch.

Both are minor/non-blocking — no correctness bugs found, and test coverage (fixture Compliant/Noncompliant cases + upstream sentinel tests) is solid.

asya-vorobeva (approved):
LGTM 🥇

PR #7407, opened 2026-06-30
Implement S8967: inline snapshots should not contain interpolations

nathsou:
## Summary
- implement S8967 for inline snapshot template interpolations in Jest and Vitest
- keep the detection narrow to built-in inline snapshot matchers rooted in `expect(...)`, including `not`, `resolves`, and `rejects`
- add focused unit tests and sync the generated rule metadata/profile data

## Links
- Jira: JS-1964
- RSPEC: https://github.com/SonarSource/rspec/pull/7269

----
## Summary by Gitar

  - Implemented `S8967` to detect template interpolations in Jest and Vitest inline snapshots.
- **Shared logic extraction:**
  - Extracted chain-walking logic into a new helper, `expect-call-chain.ts`, to support `S8967` and refactor `S8780`.
- **Refactoring:**
  - Updated `S8780` to utilize the new `collectCallChain` and `getRootCall` helpers, improving handling of computed member access in expectation chains.
- **Testing and metadata:**
  - Added comprehensive unit tests for `S8967` and updated `Sonar_way_profile.json` to include the new rule.


packages/analysis/src/jsts/rules/S8960/rule.ts
```diff
@@ -0,0 +1,217 @@
+/*
```
guillemsarda, line 1: This folder shouldn't be in this PR

packages/analysis/src/jsts/rules/S8967/rule.ts
```diff
@@ -0,0 +1,133 @@
...
+): boolean {
+  const fqn = getFullyQualifiedName(context, call.callee);
+  if (fqn) {
+    return fqn === '@jest.globals.expect' || fqn === 'vitest.expect';
+  }
+
+  return (
+    (frameworks.jest || frameworks.vitest) &&
+    isIdentifier(call.callee, 'expect') &&
+    isUnresolved(call.callee, context)
```
guillemsarda: `isUnresolved()` feels too broad for the fallback bare-`expect(...)` check. This code is really trying to answer “is this the framework global expect?”, so a more specific helper based on scope/defs would make the intent clearer and reduce the risk of future false positives from shadowed locals (although very unlikely). For instance:

```ts
function isGlobalExpect(context: Rule.RuleContext, node: estree.Identifier): boolean {
  const variable = getVariableFromName(context, node.name, node);
  return node.name === 'expect' && (!variable || variable.defs.length === 0);
}
```

and then:

```ts
return (
  (frameworks.jest || frameworks.vitest) &&
  isIdentifier(call.callee, 'expect') &&
  isGlobalExpect(context, call.callee)
);
```

guillemsarda (changes requested):
No big changes requested

packages/analysis/src/jsts/rules/S8960/rule.ts
nathsou: Fixed by rebasing

guillemsarda (approved):
Image

PR #7418, opened 2026-07-01
Implement S8968: Tests should be skipped explicitly

nathsou:
## Summary

- Implement S8968: report a conditional `return` (no `else`, no value) used as the first statement of an `it`/`test` callback to bail out of a test, when it's followed by more code — instead of calling the test framework's explicit skip mechanism. Every target framework reports such a test as **passed**, not skipped, hiding the fact its assertions never ran.
- Scoped to Mocha, Vitest, Playwright, node:test, and Bun. Jest and Jasmine are intentionally out of scope; AVA/QUnit are excluded because both already fail a test by default when it completes without an assertion, so the false-pass premise doesn't apply there.
- Detection deliberately does not reuse `helpers/mocha.ts`'s `extractTestCase` as-is: that helper assumes the callback is always argument index 1, which misses the 3-argument `test(name, options, fn)` form used by node:test and Playwright. Uses `isTestCase` plus a local "last function-typed argument" extraction instead, and excludes already-`.skip()`-marked calls (a permanently skipped test can't produce the misleading "passed" outcome this rule targets).
- No type information required — purely syntactic, identical behavior for JS and TS.

## Testing

- `cb.fixture.js`/`cb.fixture.ts` cover all 5 frameworks' noncompliant/compliant shapes from the RSPEC, plus: if/else present, unconditional return with nothing after it, return with a value, return inside a nested `forEach` callback, `beforeEach` hook (out of scope), a second guard clause in a two-guard sequence (only the first is flagged), the `specify` alias, `it.skip(...)`/`it.only(...)` interaction, 3-argument `test(name, { ... }, fn)` forms, and the chained `.skipIf(cond)(...)` shape (not recognized as a test case at all — no crash, no false positive).
- `npx tsx --test packages/analysis/src/jsts/rules/S8968/*.test.ts` passes.

## Links

- Jira: https://sonarsource.atlassian.net/browse/JS-1974
- RSPEC PR: https://github.com/SonarSource/rspec/pull/7272

----
## Summary by Gitar

- **New rule added:**
  - Implemented `S8968` to identify tests using implicit guard returns instead of explicit skip mechanisms.
- **Rule cleanup:**
  - Removed `S8960` rule files and configuration, as these were superseded in this PR.
- **Infrastructure improvements:**
  - Added comprehensive fixtures for `S8968` covering `bun`, `mocha`, `node-test`, `playwright`, and `vitest`.
  - Updated `Sonar_way_profile.json` to include `S8968` and `S8960` (the latter was added to the profile in the same commit batch).


packages/analysis/src/jsts/rules/S8968/rule.ts
```diff
@@ -0,0 +1,181 @@
...
+  // dependency manifest for Playwright/Vitest, which sometimes run test files that
+  // don't literally import the framework (e.g. via a shared fixture re-export).
+  if (importsOrDependsOnModule(context, PLAYWRIGHT_MODULES, PLAYWRIGHT_MODULES)) {
+    return 'playwright';
+  }
+  if (importsOrDependsOnModule(context, VITEST_MODULES, VITEST_MODULES)) {
+    return 'vitest';
+  }
+  return 'mocha';
+}
```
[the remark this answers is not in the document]
nathsou: This is a deliberate, precedented tradeoff rather than an oversight: sibling rules S3415, S8782, and S6079 all use the same `isTestCase`/`extractTestCase` helpers with zero import-based gating, relying on the syntactic shape alone. Requiring a weak Mocha signal (e.g. a `mocha`/`describe` import) before falling back to the Mocha message would regress the dominant real-world case, since most Mocha test files rely purely on globals injected by the test runner and have no import to gate on at all. Keeping this as-is, consistent with existing convention. No change.

sonar-plugin/javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/Sonar_way_profile.json
```diff
@@ -422,8 +422,10 @@
     "S8950",
     "S8951",
     "S8957",
```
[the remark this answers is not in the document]
nathsou: Good catch. Verified against origin/master's actual content: S8957/S8961/S8962 are legitimately already there from unrelated, already-merged PRs, so those weren't the issue. But S8960 had indeed leaked in — from an earlier rebase conflict resolution mistake on my end (S8960's rule implementation is a separate, still-open, unmerged PR #7396, not part of this branch at all). Removed S8960 in ceddf1491.

packages/analysis/src/jsts/rules/S8968/rule.ts
```diff
@@ -0,0 +1,192 @@
...
+  // Neither imported directly in this file nor excluded: fall back to the project's
+  // dependency manifest for Playwright/Vitest, which sometimes run test files that
+  // don't literally import the framework (e.g. via a shared fixture re-export).
+  if (importsOrDependsOnModule(context, PLAYWRIGHT_MODULES, PLAYWRIGHT_MODULES)) {
+    return 'playwright';
+  }
+  if (importsOrDependsOnModule(context, VITEST_MODULES, VITEST_MODULES)) {
+    return 'vitest';
+  }
+  if (importsOrDependsOnModule(context, MOCHA_MODULES, MOCHA_MODULES)) {
```
guillemsarda, line 128: `detectFramework()` now gives precedence to explicit file-level imports over project-wide exclusion signals for Playwright/Vitest, but Mocha still misses that treatment.

In particular, this branch:

```ts
if (importsOrDependsOnModule(context, JEST_IMPORTS, JEST_DEPENDENCIES)) {
  return 'excluded';
}
```
runs before Mocha is ever considered, while Mocha is only checked later via:
```ts
if (importsOrDependsOnModule(context, MOCHA_MODULES, MOCHA_MODULES)) {
  return 'mocha';
}
```

That means an explicit Mocha test file is silently excluded in a mixed-runner repo that also depends on Jest:
```
import { it } from 'mocha';

it('reorders columns', function () {
  if (readOnlyMode) {
    return;
  }
  db.reorderColumns();
  expect(db.columns).toEqual(['a', 'b']);
});
```
In that file, the rule should clearly apply as mocha, but the current precedence returns excluded as soon as the project-level Jest dependency is seen.
So I think Mocha needs the same “explicit import wins over dependency-based exclusion” treatment that was already added for Playwright/Vitest.

packages/analysis/src/jsts/rules/S8968/rule.ts
```diff
@@ -0,0 +1,192 @@
...
+/**
+ * A recognized it/test call's callback, or null. Does not reuse Mocha.extractTestCase,
+ * which only looks at argument index 1 and therefore misses the 3-argument
+ * `test(name, options, fn)` form used by node:test and Playwright: the callback is
+ * always the last argument regardless of how many arguments precede it.
+ * Calls already marked `.skip` are ignored: such a test can never be misreported as
+ * "passed", since it never runs at all.
+ */
+function extractCallback(call: estree.CallExpression): estree.Function | null {
+  if (!isTestCase(call)) {
```
guillemsarda: Is this scope narrowed on purpose?

We have more cases that we could cover:
- aliased imports
- namespace imports
- `require(...)` namespace bindings
- Vitest modifiers already recognized elsewhere (f.i., `.concurrent` or `.sequential`
- Playwright alias / modifier forms

packages/analysis/src/jsts/rules/S8968/rule.ts
```diff
@@ -0,0 +1,192 @@
...
+ * Mocha works off globals and is not necessarily imported in the file, so it's
+ * checked last via `importsOrDependsOnModule` against the project's package.json
+ * dependencies. Unlike the other frameworks, Mocha has no other positive signal
+ * to rely on: without this dependency check, every file with no other framework
+ * signal would be assumed to be Mocha, which produces false positives on
+ * frameworks this rule doesn't otherwise recognize (e.g. Jasmine loaded from a
+ * vendored copy rather than an npm dependency). If nothing indicates Mocha
+ * either, the file is excluded rather than guessed at.
+ */
+function detectFramework(context: Rule.RuleContext): Framework | 'excluded' {
```
guillemsarda, line 91: **Refactor suggestion:** `detectFramework()` is basically encoding an ordered decision table as a long `if` ladder:
- positive file-level signals for covered frameworks
- exclusion signals
- fallback dependency-based signals
- final default

That makes the precedence harder to scan and easier to break when a new framework is added.

A cleaner shape would be to encode it as ordered rules, for example:
```ts
type DetectionStep = {
  match: (context: Rule.RuleContext) => boolean;
  result: Framework | 'excluded';
};

const DETECTION_STEPS: DetectionStep[] = [
  { match: context => importsModule(context, PLAYWRIGHT_MODULES), result: 'playwright' },
  { match: context => importsModule(context, NODE_TEST_MODULES), result: 'nodeTest' },
  { match: context => importsModule(context, BUN_MODULES), result: 'bun' },
  { match: context => importsModule(context, VITEST_MODULES), result: 'vitest' },

  { match: context => importsOrDependsOnModule(context, JEST_IMPORTS, JEST_DEPENDENCIES), result: 'excluded' },
  { match: context => importsOrDependsOnModule(context, JASMINE_MODULES, JASMINE_MODULES), result: 'excluded' },
  { match: context => importsOrDependsOnModule(context, AVA_MODULES, AVA_MODULES), result: 'excluded' },
  { match: context => importsOrDependsOnModule(context, QUNIT_MODULES, QUNIT_MODULES), result: 'excluded' },

  { match: context => importsOrDependsOnModule(context, PLAYWRIGHT_MODULES, PLAYWRIGHT_MODULES), result: 'playwright' },
  { match: context => importsOrDependsOnModule(context, VITEST_MODULES, VITEST_MODULES), result: 'vitest' },
  { match: context => importsOrDependsOnModule(context, MOCHA_MODULES, MOCHA_MODULES), result: 'mocha' },
];

function detectFramework(context: Rule.RuleContext): Framework | 'excluded' {
  return DETECTION_STEPS.find(step => step.match(context))?.result ?? 'excluded';
}
```

guillemsarda (changes requested):

packages/analysis/src/jsts/rules/S8968/rule.ts
```diff
@@ -0,0 +1,192 @@
+/*
```
guillemsarda, line 1: Also, there is the option for quickfixes in Node and Playwright.

packages/analysis/src/jsts/rules/S8968/rule.ts
nathsou: Good catch, thanks — fixed in 8efeae3b5. `detectFramework()` now checks `importsModule(context, MOCHA_MODULES)` right after the other three explicit-import checks (Playwright/node:test/Bun/Vitest) and before the Jest/Jasmine/AVA/QUnit exclusion block, so an explicit `import { it } from 'mocha'` wins over a project-wide Jest dependency just like the other frameworks. Added a regression test in `unit.test.ts` mirroring the existing jest-and-playwright one (`fixtures/jest-and-mocha/`).

packages/analysis/src/jsts/rules/S8968/rule.ts
nathsou: Addressed in 8efeae3b5, with one honest caveat.

Mocha/Vitest now go through `isMochaTestConstruct` and Playwright through `getPlaywrightTestQualifiers`, both from `helpers/mocha-style-test-frameworks.ts` (the same helper S8960 already relies on). That gets us: aliased named imports (`import { it as t } from 'vitest'`), destructured `require()` bindings (`const { it } = require('mocha')`), and modifier chains — Vitest's `.concurrent`/`.sequential`, Playwright's `.only`/`.fail` — for free, since that's exactly what the shared helper resolves. Added fixtures covering the aliased+modifier combos in `vitest.fixture.js`, `playwright.fixture.js`, and `unit.test.ts`.

The one thing that does *not* work, and I want to flag rather than silently miss: pure namespace-style access — `import * as ns from 'mocha'; ns.it(...)` or `const ns = require('mocha'); ns.it(...)`. I traced this through `getFullyQualifiedName`/`checkFqnFromImport` in `helpers/module.ts`: the property access (`.it`) gets consumed as a "modifier" by `getMochaCalleeParts` before the base identifier's FQN is resolved, and that resolution doesn't know to re-append it, so it comes back as just `'mocha'` instead of `'mocha.it'` and fails the match. This is a pre-existing gap in the shared helper (I confirmed it also affects other rules that already use `isMochaTestConstruct`/`getPlaywrightTestQualifiers`), not something introduced here, and fixing it means touching `module.ts`'s FQN resolution shared across the codebase — that felt like the wrong scope for this PR. Happy to file it separately if you'd like it tracked.

packages/analysis/src/jsts/rules/S8968/rule.ts
nathsou: Thanks for the suggestion — I'd like to hold off on this one for now. The current `if` ladder is short (4 explicit imports → 4 exclusions → 3 dependency fallbacks → default) and each branch is directly documented by the precedence comment above it, so I don't think the `DETECTION_STEPS` table meaningfully reduces the complexity here — it's the same number of ordered checks, just behind a layer of indirection (an array of closures + a `.find()`) instead of plain control flow. Per this repo's general preference for avoiding abstractions beyond what's needed, I'd rather keep it as a straightforward function for now. Happy to revisit if `detectFramework()` grows enough (e.g. more frameworks) that the ladder becomes hard to scan.

packages/analysis/src/jsts/rules/S8968/rule.ts
nathsou: Right, and in Vitest

guillemsarda (approved):
Image

PR #7419, opened 2026-07-01
JS-1976 Cache sonar-rule-api RSPEC clone in CI

vdiez:
## Summary

- add an explicit cache for `~/.sonar/rule-api`
- restore it on non-default branches without saving, following the repository cache policy
- save it on the default branch with a rolling key so the cached RSPEC clone can evolve independently of SonarJS sources
- apply it only to `prepare_rspec_rule_data`
- keep the cache key simple: `rule-api-${github.run_id}` with `rule-api-` restore fallback

## Why

We already reuse the prepared RSPEC rule-data artifact across jobs inside the same workflow, but the `sonar-rule-api` cache directory is currently recreated from scratch on every workflow run. That means cold-start `rspec:refresh` runs still pay the full RSPEC clone/update cost.

Caching `~/.sonar/rule-api` makes those cold starts cheaper across workflow runs while keeping the current job topology unchanged. Only `prepare_rspec_rule_data` needs this cache because downstream jobs already consume the uploaded RSPEC artifact.

The cache key uses `github.run_id` on the default branch so the cache can keep moving forward even when the SonarJS checkout itself does not change. We intentionally do not rotate this cache monthly because the payload is a Git clone that updates in place rather than an ever-growing dependency store. We also do not split the key by OS because the cache is only used by the Linux `prepare_rspec_rule_data` job.

## Validation

- `git diff --check`
- `ruby -e 'require "yaml"; YAML.load_file(".github/actions/rule-api-cache/action.yml"); YAML.load_file(".github/workflows/build.yml")'`

### DIFF
--- a/.github/actions/rule-api-cache/action.yml
+++ b/.github/actions/rule-api-cache/action.yml
@@ -25,8 +25,8 @@ runs:
       uses: actions/cache/restore@v5
       with:
         path: ${{ env.RULE_API_CACHE_DIR }}
-        key: ${{ inputs.key-prefix }}-${{ runner.os }}-${{ github.run_id }}
-        restore-keys: ${{ inputs.key-prefix }}-${{ runner.os }}-
+        key: ${{ inputs.key-prefix }}-${{ github.run_id }}
+        restore-keys: ${{ inputs.key-prefix }}-
 
     # Full cache on default branch (save enabled) - uses run_id for rolling cache because the RSPEC
     # clone contents evolve independently of the SonarJS source tree.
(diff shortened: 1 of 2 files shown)


.github/actions/rule-api-cache/action.yml
```diff
@@ -0,0 +1,43 @@
...
+      with:
+        path: ${{ env.RULE_API_CACHE_DIR }}
+        key: ${{ inputs.key-prefix }}-${{ github.run_id }}
+        restore-keys: ${{ inputs.key-prefix }}-
+
+    # Full cache on default branch (save enabled) - uses run_id for rolling cache because the RSPEC
+    # clone contents evolve independently of the SonarJS source tree.
+    - name: Cache Rule API data (default branch)
+      if: github.ref_name == github.event.repository.default_branch && inputs.save == 'true'
+      uses: actions/cache@v5
```
nathsou: Any reason to use v5 instead of v6 here (and on line 25)?

The other workflows seem to be using v6: https://github.com/SonarSource/SonarJS/blob/1b9b2fe637925bf79b3d89c61abed0be26f8cc50/.github/actions/orchestrator-cache/action.yml#L28C9-L28C37

nathsou (approved):

.github/actions/rule-api-cache/action.yml
vdiez: Good catch. No specific reason to stay on v5 here; I aligned the new action with the rest of the repository and switched both cache steps to v6 in bb17b77.

PR #7423, opened 2026-07-01
JS-1981 Expose propNamePattern for S6478

francois-mora-sonarsource:
## Summary
- expose `propNamePattern` alongside `allowAsProps` for S6478
- keep the upstream default glob semantics (`render*`) and describe the option as a prop-name glob
- add focused tests covering `render*`, `*Enhancer`, and `*Render` suppression while keeping non-matching props noncompliant

## Validation
- `npm run generate-meta`
- `npx tsx --test packages/analysis/src/jsts/rules/S6478/**/*.test.ts`

## Notes
- `npm run generate-java-rule-classes` completes with an unrelated `RSPEC metadata not found for rule S8957.` warning in this checkout

### DIFF
--- a/packages/analysis/src/jsts/rules/S6478/config.ts
+++ b/packages/analysis/src/jsts/rules/S6478/config.ts
@@ -28,7 +28,7 @@ export const fields = [
     },
     {
       field: 'propNamePattern',
-      default: '{render*,*Enhancer,*Render}',
+      default: '{render*,*Enhancer,*Renderer}',
       description:
         'Glob pattern matching prop names whose inline functions should be treated as render props rather than nested components.',
       displayName: 'propNamePattern',
--- a/packages/analysis/src/jsts/rules/S6478/config.ts
+++ b/packages/analysis/src/jsts/rules/S6478/config.ts
@@ -23,7 +23,15 @@ export const fields = [
     {
       field: 'allowAsProps',
       default: false,
-      description: 'Allow React components defined inline when they are passed as props.',
+      description:
+        'Allow React components defined inline when they are passed as props. Use this broad escape hatch when propNamePattern is not precise enough.',
+    },
+    {
+      field: 'propNamePattern',
+      default: 'render*',
+      description:
+        'Glob pattern matching prop names whose inline functions should be treated as render props rather than nested components.',
+      displayName: 'propNamePattern',
     },
   ],
 ] as const satisfies ESLintConfiguration;
--- a/packages/analysis/src/jsts/rules/S6478/config.ts
+++ b/packages/analysis/src/jsts/rules/S6478/config.ts
@@ -28,7 +28,7 @@ export const fields = [
     },
     {
       field: 'propNamePattern',
-      default: 'render*',
+      default: '{render*,*Enhancer,*Render}',
       description:
         'Glob pattern matching prop names whose inline functions should be treated as render props rather than nested components.',
       displayName: 'propNamePattern',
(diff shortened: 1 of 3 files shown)


packages/analysis/src/jsts/rules/S6478/unit.test.ts
```diff
@@ -36,13 +36,52 @@ describe('S6478', () => {
...
+    assert.ok(
+      fields.some(
+        field =>
+          Array.isArray(field) &&
+          field.some(
+            option =>
+              option.field === 'propNamePattern' &&
+              option.default === '{render*,*Enhancer,*Render}' &&
+              option.displayName === 'propNamePattern' &&
+              typeof option.description === 'string',
```
erwan-leforestier-sonarsource: I don't really see the need for such test, like others on top. It's just testing that indeed what was defined in `S6478/config.ts` matches something we want but it's not really testing anything real. I think unit tests should focus only on analysis result and not on rule configuration. Wdyt ?

packages/analysis/src/jsts/rules/S6478/unit.test.ts
```diff
@@ -132,6 +171,70 @@ describe('S6478', () => {
     });
   });
 
+  it('should suppress configured render-prop names with propNamePattern', () => {
+    const ruleTester = new DefaultParserRuleTester();
+
+    ruleTester.run('S6478', rule, {
+      valid: [
+        {
+          options: [{ propNamePattern: '{render*,*Enhancer,*Render}' }],
```
erwan-leforestier-sonarsource: Worth testing with a propNamePattern that is different from the default `propNamePattern` value, otherwise we don't know if it's the custom one or the default one that is used.

packages/analysis/src/jsts/rules/S6478/config.ts
```diff
@@ -23,7 +23,15 @@ export const fields = [
     {
       field: 'allowAsProps',
       default: false,
-      description: 'Allow React components defined inline when they are passed as props.',
+      description:
+        'Allow React components defined inline when they are passed as props. Use this broad escape hatch when propNamePattern is not precise enough.',
+    },
+    {
+      field: 'propNamePattern',
+      default: '{render*,*Enhancer,*Render}',
```
erwan-leforestier-sonarsource: According to ticket JS-1981: 

> This leaves a gap for render-prop APIs whose prop names do not start with render, for example BaseWeb startEnhancer/endEnhancer and other *Enhancer or *Render**er** props.

Shouldn't the default value be `'{render*,*Enhancer,*Renderer}'` instead ?

packages/analysis/src/jsts/rules/S6478/unit.test.ts
francois-mora-sonarsource: Removed that config-shape assertion in the follow-up push. The unit test now stays focused on analysis behavior only.

packages/analysis/src/jsts/rules/S6478/unit.test.ts
francois-mora-sonarsource: Adjusted in the follow-up push. The test now uses a non-default pattern (`{*Render,*Slot}`) so it exercises the custom option rather than the built-in default.

packages/analysis/src/jsts/rules/S6478/config.ts
francois-mora-sonarsource: Updated in the follow-up push. The exposed default is now `{render*,*Enhancer,*Renderer}` to match the ticket wording.

erwan-leforestier-sonarsource (approved):

PR #7426, opened 2026-07-01
JS-1077 Handle Chai-like assertions in S2699

francois-mora-sonarsource:
## Summary
Improve `assertions-in-tests` (S2699) recognition of Chai-like assertions with a rule-local detector, without changing the shared assertion helper.

## Changes
- ignore a standalone `.should` access so it no longer satisfies S2699 by itself
- recognize Chai `should` assertions structurally: any `.should` access that is *extended* by a further property access or call is an assertion, regardless of the terminal shape (`.be.true`, `.property(...)`, `.lengthOf(...)`, `.calledWith(...)`, `.equal(...)`, …) — replacing the initial hardcoded terminal allow-lists
- recognize the additional Chai assert methods `assert.instance(...)` and `assert.is(...)`
- add focused JS and TS regression coverage, including negative cases for bare `.should`, unrelated method names, and shadowed local `assert` objects

## Behavioral note
The type-aware path previously only recognized `should` chains whose fully-qualified name resolved through Chai's type augmentation, so should chains rooted in locals (e.g. `localVar.should.equal(x)`) or ending in a plain property (`.should.be.empty`) were false-positived. This PR closes that gap. As a consequence, `chai.fixture.js` line for `throwsTypeError.should.to.not.throw(...)` flips from `Noncompliant` to `Compliant` — it is a genuine Chai assertion that was previously (incorrectly) flagged. The rule now stops flagging local-rooted `.should` chains in type-aware analysis.

## Validation
- `./node_modules/.bin/tsx --tsconfig packages/tsconfig.test.json --test packages/analysis/src/jsts/rules/S2699/unit.test.ts`
- `./node_modules/.bin/tsx --tsconfig packages/tsconfig.test.json --test packages/analysis/src/jsts/rules/S2699/cb.test.ts`
- `./node_modules/.bin/tsx --tsconfig packages/tsconfig.test.json --test packages/analysis/src/jsts/rules/S2699/typescript-detection/cb.test.ts`


packages/analysis/src/jsts/rules/helpers/assertion-detection.ts
```diff
@@ -144,6 +159,73 @@ export function isTSAssertion(services: ParserServicesWithTypeInformation, node:
...
+    return (
+      ts.isVariableStatement(statement) &&
+      statement.declarationList.declarations.some(declaration =>
+        isChaiRequireInitializer(declaration.initializer),
+      )
+    );
+  });
+  chaiImportCache.set(sourceFile, result);
+  return result;
+}
```
erwan-leforestier-sonarsource: Don't we already have some helpers to detect if a module is imported ?

packages/analysis/src/jsts/rules/helpers/assertion-detection.ts
```diff
@@ -144,6 +159,73 @@ export function isTSAssertion(services: ParserServicesWithTypeInformation, node:
...
+    !ts.isPropertyAccessExpression(parent) ||
+    parent.expression !== node
+  ) {
+    return false;
+  }
+  const grandparent = parent.parent;
+  return (
+    (ts.isPropertyAccessExpression(grandparent) && grandparent.expression === parent) ||
+    (ts.isCallExpression(grandparent) && grandparent.expression === parent)
+  );
```
erwan-leforestier-sonarsource: [Reported by both claude and codex]
`isTSExtendingShouldChainParent` requires the grandparent to be a `PropertyAccessExpression` or `CallExpression`, so it only covers chains with at least two nodes after `.should`:
```js
value.should.equal(2) // ok, grandparent is a call
value.should.have.property('x') // ok, grandparent is a member access
value.should.exist // still a FP, grandparent is ExpressionStatement
```

The non-TS path (isExtendingShouldChainParent, line 274) only looks one level up and handles the third case correctly. Can isTSExtendingShouldChainParent be aligned to do the same?

packages/analysis/src/jsts/rules/helpers/assertion-detection.ts
francois-mora-sonarsource: The helper was missing for TypeScript, it has been added.

packages/analysis/src/jsts/rules/helpers/assertion-detection.ts
francois-mora-sonarsource: This was  a simple heuristic to avoid hardcoding all the Chai keywords. But in the end, I decided to add that list and have a more accurate handling of Chai `should` expressions and the false positive suppression will be more accurate.

erwan-leforestier-sonarsource (approved):

PR #7430, opened 2026-07-02
JS-1983 Refine S6819 role=group exception

erwan-leforestier-sonarsource:
## Summary
Refine the S6819 `role="group"` exception so the rule no longer relies on `<form>` or `<fieldset>` ancestry.
The rule now keeps `group` for composite-widget subgroups and non-fieldset-like content, and only reports when the container looks like a real `<fieldset>` replacement.
RSPEC PR: https://github.com/SonarSource/rspec/pull/7277

## Changes
- Replace the form-ancestor heuristic with a content-based check: shared accessible name plus at least two native form controls
- Suppress composite-widget group substructures, including `aria-owns` ownership
- Refresh focused S6819 tests to show that form ancestry is irrelevant and that fieldset-like groups still report

### DIFF
--- a/its/ruling/src/test/expected/desktop/typescript-S6819.json
+++ b/its/ruling/src/test/expected/desktop/typescript-S6819.json
@@ -1,10 +1,4 @@
 {
-"desktop:app/src/ui/changes/commit-message.tsx": [
-753
-],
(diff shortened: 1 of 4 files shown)


asya-vorobeva (approved):
🔢

PR #7431, opened 2026-07-02
JS-1982 Guard S6819 menuitem suggestions

erwan-leforestier-sonarsource:
## Summary
This adds a local S6819 guard that suppresses suggestions pointing to obsolete `<menuitem>` tags if an upstream prefer-tag-over-role report reintroduces them.
It also adds regression coverage for the real rule behavior and for the decorator-level guard that keeps valid replacements intact.

## Changes
- Suppress any S6819 report whose suggested replacement starts with `<menuitem`
- Add regression tests for `menuitem`, `menuitemcheckbox`, and `menuitemradio`
- Add a focused decorator test that still reports valid replacements such as `<button>`

## Functional Validation
Artifact: [JS-1982-fv.zip](https://github.com/user-attachments/files/29588505/JS-1982-fv.zip)

    ./run.sh

Expected output:
```text
Analyzing "suggestions.jsx"...
Results:
    - Rule "S6819" -> L.1: Use <menuitem> instead of the "menuitem" role to ensure accessibility across all devices.
    - Rule "S6819" -> L.2: Use <menuitem type="checkbox"> instead of the "menuitemcheckbox" role to ensure accessibility across all devices.
    - Rule "S6819" -> L.3: Use <menuitem type="radio"> instead of the "menuitemradio" role to ensure accessibility across all devices.
    - Rule "S6819" -> L.4: Use <button> instead of the "button" role to ensure accessibility across all devices.

****** Branch "fix/js-1982-s6819-menuitem-suggestion" ******
Analyzing "suggestions.jsx"...
Results:
    - Rule "S6819" -> L.4: Use <button> instead of the "button" role to ensure accessibility across all devices.
```

### DIFF
--- a/packages/analysis/src/jsts/rules/S6819/decorator.ts
+++ b/packages/analysis/src/jsts/rules/S6819/decorator.ts
@@ -83,6 +83,13 @@ export function decorate(rule: Rule.RuleModule): Rule.RuleModule {
         return;
       }
 
+      const suggestedTag = (
+        reportDescriptor as Rule.ReportDescriptor & { data?: { tag?: unknown } }
+      ).data?.tag;
+      if (typeof suggestedTag === 'string' && suggestedTag.startsWith('<menuitem')) {
+        return;
+      }
+
       if (isValidAriaPattern(node)) {
         // Suppress for valid ARIA patterns
         return;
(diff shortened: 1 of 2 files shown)


erwan-leforestier-sonarsource:
Closing. This doesn't reproduce on master: aria-query 5.3.2 no longer maps `menuitem`, `menuitemcheckbox`, or `menuitemradio` to any HTML element, so the upstream `prefer-tag-over-role` rule never suggests `<menuitem>` for them. The guard added here defends against a case the pinned upstream cannot produce.

Verified by running the real upstream rule (zero reports on all three roles, while `role="button"` still reports) and by deleting the guard: the real-rule test stayed green and only the synthetic mock test exercised it.

PR #7434, opened 2026-07-02
JS-1993 S5906: clarify that length matchers apply beyond collections

nathsou:
## Summary

Reviewers have reported the S5906 rewrite `expect(x.length).toBe(n)` → `expect(x).toHaveLength(n)` (and the Jasmine/Chai equivalents `toHaveSize`/`lengthOf`) as a false positive when `x` isn't literally a collection, e.g. `expect(diagram.length).toBe(1400)` where `diagram` is a rendered string.

These matchers are actually generic: Jest/Vitest's `toHaveLength` and Jasmine's `toHaveSize` pass whenever the tested value exposes a numeric `length`/`size` property with the expected value, whatever that value represents (Vitest's own test suite even asserts `expect({ length: 3 }).toHaveLength(3)`). The rewrite is behaviorally identical regardless of what `x` is, so this isn't a detection bug — it's a communication gap.

This PR:
- Tags length-based suggestions with `kind: 'length'` on the `Suggestion` type (`assertion-utils.ts`), set at every call site that builds a `.length`/`.size` rewrite across `rule.ts` and `assertion-suggestions.ts`.
- Adds a dedicated `preferSpecificLengthAssertion` message that explicitly states the matcher works on any value with a numeric length property, not only arrays or collections, used instead of the generic `preferSpecificAssertion` message for these cases.
- Updates existing length-related test expectations across `unit.test.ts`, `chai-bdd.unit.test.ts`, and `integrations.unit.test.ts` to assert the new message id.

RSPEC companion PR (adds the same clarification plus a string-based example to "Why is this an issue?"): https://github.com/SonarSource/rspec/pull/7279

## Test plan
- [x] `npx tsx --test packages/analysis/src/jsts/rules/S5906/*.test.ts` — all 3 suites pass
- [x] `npx tsc --noEmit` on the analysis package — no new errors
- [x] `npx prettier --check` on changed files

### DIFF
--- a/packages/analysis/src/jsts/rules/S5906/assertion-utils.ts
+++ b/packages/analysis/src/jsts/rules/S5906/assertion-utils.ts
@@ -22,7 +22,7 @@ import { getVariableFromName, isIdentifier, isMethodCall } from '../helpers/ast.
 export type Suggestion = {
   assertion: string;
   replacement?: string;
-  kind?: 'length';
+  messageId?: 'preferSpecificAssertion' | 'preferSpecificLengthAssertion';
 };
 
 const PLAYWRIGHT_LOCATOR_FACTORIES = new Set([
@@ -125,11 +125,11 @@ export function replacement(
   replacementText: string,
   _node: estree.CallExpression,
   _sourceCode?: SourceCode,
-  kind?: Suggestion['kind'],
+  messageId?: Suggestion['messageId'],
 ): Suggestion {
   return {
     assertion: replacementText,
     replacement: replacementText,
-    ...(kind ? { kind } : {}),
+    ...(messageId ? { messageId } : {}),
   };
 }
--- a/packages/analysis/src/jsts/rules/S5906/assertion-utils.ts
+++ b/packages/analysis/src/jsts/rules/S5906/assertion-utils.ts
@@ -22,6 +22,7 @@ import { getVariableFromName, isIdentifier, isMethodCall } from '../helpers/ast.
 export type Suggestion = {
   assertion: string;
   replacement?: string;
+  kind?: 'length';
 };
 
 const PLAYWRIGHT_LOCATOR_FACTORIES = new Set([
@@ -124,9 +125,11 @@ export function replacement(
   replacementText: string,
   _node: estree.CallExpression,
   _sourceCode?: SourceCode,
+  kind?: Suggestion['kind'],
 ): Suggestion {
   return {
     assertion: replacementText,
     replacement: replacementText,
+    ...(kind ? { kind } : {}),
   };
 }
--- a/packages/analysis/src/jsts/rules/S5906/rule.ts
+++ b/packages/analysis/src/jsts/rules/S5906/rule.ts
@@ -48,7 +48,7 @@ const messages = {
   preferSpecificAssertion:
     'Prefer "{{assertion}}" over this generic assertion; dedicated matchers read better and report clearer failures.',
   preferSpecificLengthAssertion:
-    'Prefer "{{assertion}}" over this generic assertion. It also works for any object with a numeric length property.',
+    'Prefer "{{assertion}}" over this generic assertion for better reporting; it works on any object with a numeric length property.',
   quickfix: 'Replace with {{assertion}}.',
 };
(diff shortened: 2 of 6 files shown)


packages/analysis/src/jsts/rules/S5906/assertion-utils.ts
```diff
@@ -22,6 +22,7 @@ import { getVariableFromName, isIdentifier, isMethodCall } from '../helpers/ast.
 export type Suggestion = {
   assertion: string;
   replacement?: string;
+  kind?: 'length';
```
guillemsarda: The change looks correct functionally, but I think the new `kind?: 'length'` field on `Suggestion` adds avoidable coupling between suggestion construction and reporting.

Right now, `report()` needs to know that `kind === 'length'` means it should switch from `preferSpecificAssertion` to `preferSpecificLengthAssertion`. That works, but it pushes message-selection logic away from the place where the suggestion is actually built.

I think it would be cleaner to let `Suggestion` carry the `messageId` directly, for example:

```ts
type Suggestion = {
  assertion: string;
  replacement?: string;
  messageId: 'preferSpecificAssertion' | 'preferSpecificLengthAssertion';
};
```

Then report() becomes generic and does not need to interpret kind:
```ts
context.report({
  node: node.callee,
  messageId: suggestion.messageId,
  data: { assertion: suggestion.assertion },
  ...(suggest ? { suggest } : {}),
});
```
This keeps the wording decision at the suggestion-construction site, avoids growing an ad hoc taxonomy in Suggestion, and should scale better if we introduce more message variants later.

packages/analysis/src/jsts/rules/S5906/rule.ts
```diff
@@ -46,7 +46,9 @@ import { getPlaywrightLocatorSuggestion } from './playwright-suggestions.js';
 
 const messages = {
   preferSpecificAssertion:
-    'Prefer a more specific assertion instead of this generic one, e.g. "{{assertion}}".',
+    'Prefer "{{assertion}}" over this generic assertion; dedicated matchers read better and report clearer failures.',
+  preferSpecificLengthAssertion:
+    'Prefer "{{assertion}}" over this generic assertion; it reads better, reports clearer failures, and applies to any object with a numeric length property.',
```
guillemsarda: One minor comment on the new `preferSpecificLengthAssertion` wording: I think the message is a bit long for usual Sonar style.

Current version:
```text
Prefer "{{assertion}}" over this generic assertion; it reads better, reports clearer failures, and applies to any object with a numeric length property.
```
It is accurate, but it mixes the action, the rationale, and the clarification in a single sentence. In SonarJS, issue messages are usually easier to scan when they stay compact and leave most of the explanation to RSPEC.
I think this would read more cleanly with a shorter variant, for example:
```text
Prefer "{{assertion}}" over this generic assertion. It also works for any object with a numeric length property.
```
The clarification is useful, but I would keep the issue message a bit tighter.

guillemsarda (commented):
I left two suggestions. Lmkwyt

packages/analysis/src/jsts/rules/S5906/assertion-utils.ts
nathsou: Good catch — done. `Suggestion.kind?: 'length'` is now `Suggestion.messageId?: 'preferSpecificAssertion' | 'preferSpecificLengthAssertion'`, so `report()` just forwards `suggestion.messageId` instead of re-deriving it from `kind`. All 8 call sites building length suggestions now pass `'preferSpecificLengthAssertion'` directly. See 906efd414.

packages/analysis/src/jsts/rules/S5906/rule.ts
nathsou: Tightened the wording: `Prefer "{{assertion}}" over this generic assertion for better reporting; it works on any object with a numeric length property.` — shorter than the original, keeps a compact nod to the rationale shared with `preferSpecificAssertion`, and keeps "numeric" to match the RSPEC description. See 9f3698bc6.

guillemsarda (approved):
Image

PR #7435, opened 2026-07-02
JS-1991 Fix S5914: narrow to trivially-true assertions

nathsou:
## Summary

- Narrow S5914 (`no-trivial-assertions`) to only report assertions that are **guaranteed to always succeed**, instead of both always-succeed and always-fail directions.
- Rationale: an always-failing assertion is self-correcting (CI turns red immediately), while an always-succeeding assertion silently stops testing anything. A reported false positive showed the common `expect(true).toBe(false)` "unreachable code" sentinel idiom being flagged for the wrong reason.
- `resolveConstantPrimitiveValue` replaces the boolean-only `isConstantPrimitiveValue`: it resolves the actual constant value (literals, `const` bindings, unary/binary/logical expressions using real JS operator semantics) so predicate and constant-vs-constant comparison assertions can take `.not`/negation into account and only report when the assertion is guaranteed to succeed.
- The identity-comparison-against-a-freshly-created-value check (`freshIdentity`) keeps being flagged in **both** directions as a documented exception, since it usually signals a `toBe`/`toEqual` mixup regardless of pass/fail.
- Updated `messages.issue` text and unit tests (moved guaranteed-fail fixtures to compliant, added coverage for the reported FP pattern and for negated guaranteed-pass/guaranteed-fail cases).
- Manually synced the rule title in `sonar-plugin/javascript-checks/.../S5914.json` ahead of the RSPEC PR merging.

## Links

- Jira: https://sonarsource.atlassian.net/browse/JS-1990
- RSPEC PR: https://github.com/SonarSource/rspec/pull/7280


packages/analysis/src/jsts/rules/S5914/rule.ts
```diff
@@ -184,6 +612,12 @@ function resolveConstBinding(
   if (def.parent.kind !== 'const' || def.node.id.type !== 'Identifier') {
     return undefined;
   }
+  // a read that textually precedes its `const` declaration is a temporal-dead-zone violation:
+  // it throws at runtime rather than resolving to the initializer, so treat it as unresolvable
+  // rather than as a value (the assertion is then guaranteed to crash, not to succeed).
+  if (!node.range || !def.node.range || node.range[0] < def.node.range[0]) {
```
guillemsarda: `resolveConstBinding()` now treats any read whose identifier appears textually before the `const` declaration as a TDZ access:

```ts
if (!node.range || !def.node.range || node.range[0] < def.node.range[0]) {
  return undefined;
}
```

That is not a sound proxy for TDZ. TDZ is about runtime execution order, not source order, so this introduces false negatives for deferred reads such as callbacks/function bodies:
```js
test('x', () => {
  expect(answer).toBe(42);
});

const answer = 42;
```
This code is valid at runtime and the assertion is still trivially true, but the new guard stops S5914 from resolving answer.
So this check fixes the direct top-level case, but it also makes the analysis regress on any reference that is syntactically earlier and executed later. I think we should either model execution boundaries more precisely here, or avoid the source-order bailout and keep the previous behavior.

packages/analysis/src/jsts/rules/S5914/rule.ts
```diff
@@ -58,107 +75,518 @@ export const rule: Rule.RuleModule = {
...
+      return false;
+    default: {
+      const exhaustiveCheck: never = predicate;
+      return exhaustiveCheck;
+    }
+  }
+}
+
+type ConstantPrimitive = boolean | string | number | bigint | null | undefined;
+
```
guillemsarda: This file is starting to mix rule orchestration with a fairly large amount of S5914-specific evaluation logic. I think it would read better if part of that logic moved into a helper module under the same rule folder.

In particular, functions like `predicateHolds`, `freshReferencePredicateHolds`, `resolveConstantPrimitiveValue`, and the unary/binary evaluation helpers seem cohesive enough to live together in something like `S5914/constant-evaluation.ts`, leaving `rule.ts` focused on classifying assertions and reporting issues/suggestions.

guillemsarda (changes requested):

guillemsarda (approved):
Image

PR #7436, opened 2026-07-02
JS-1994 Fix S8780 false positive on done-callback tests

nathsou:
## Summary

- Fix a false positive in S8780 ("Async test assertions should be awaited or returned") for test callbacks that use a `done`-style completion parameter (e.g. `it('...', (done) => {...})`).
- In that pattern, Jest/Jasmine/Vitest determine test completion by the callback's arity, not by the returned/awaited value — mixing `done` with an awaited or returned assertion makes the runtime throw `Test functions cannot both take a 'done' callback and return something`, so the rule's own suggested fix doesn't apply to these tests.
- Locally verified with jest 30.4.1 that a failing assertion in this pattern is not silently swallowed either (unlike the scenario the rule targets) — it surfaces as an unhandled rejection that crashes the worker and exits non-zero.
- `usesDoneCallback` skips analysis when the test callback's first parameter is a plain identifier, while leaving Playwright's destructured fixtures parameter (`({ page })`) unaffected, since it's never a plain identifier.
- Added regression tests covering the reported pattern (Jest) and a Vitest done-callback variant.

## Links

- Jira: https://sonarsource.atlassian.net/browse/JS-1994

### DIFF
--- a/packages/analysis/src/jsts/rules/S8780/rule.ts
+++ b/packages/analysis/src/jsts/rules/S8780/rule.ts
@@ -123,6 +123,10 @@ export const rule: Rule.RuleModule = {
           return;
         }
 
+        if (usesDoneCallback(callback)) {
+          return;
+        }
+
         forEachExpressionStatement(callback.body, statement => {
           const assertionNode = getAsyncAssertionNode(context, statement.expression, frameworks);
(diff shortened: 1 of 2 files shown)


guillemsarda (approved):
Image

PR #7441, opened 2026-07-02
JS-1997 Fix S1848 RegExp validation FP

francois-mora-sonarsource:
Part of JS-1997

## Summary
- Suppress S1848 for discarded built-in RegExp constructors used as a validation step before returning parsed regex data.
- Keep reporting arbitrary discarded RegExp constructors, unrelated returns, and locally shadowed RegExp/globalThis constructors.
- Add JS/TS-focused unit coverage for the Jira reproducer and guardrails.

## Validation
- npm run generate-meta
- npx tsx --test packages/analysis/src/jsts/rules/S1848/**/*.test.ts
- npm run eslint-plugin:check
- npx prettier --check packages/analysis/src/jsts/rules/S1848/rule.ts packages/analysis/src/jsts/rules/S1848/unit.test.ts

### DIFF
--- a/packages/analysis/src/jsts/rules/S1848/rule.ts
+++ b/packages/analysis/src/jsts/rules/S1848/rule.ts
@@ -35,6 +35,8 @@ const DOM_SELECTION_METHODS = [
 ];
 /** jQuery/$ function names */
 const JQUERY_IDENTIFIERS = ['$', 'jQuery'];
+const STATEMENT_ANCESTOR_OFFSET = 1;
+const CONTAINER_ANCESTOR_OFFSET = 2;
 
 export const rule: Rule.RuleModule = {
   meta: generateMeta(meta, {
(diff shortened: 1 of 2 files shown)


packages/analysis/src/jsts/rules/S1848/rule.ts
```diff
@@ -88,6 +93,133 @@ function isTryable(node: estree.Node, context: Rule.RuleContext) {
...
+ */
+function isRegExpValidation(node: estree.NewExpression, context: Rule.RuleContext): boolean {
+  if (!isBuiltInRegExpConstructor(node, context)) {
+    return false;
+  }
+
+  const nextStatement = getNextSiblingStatement(node, context);
+  if (nextStatement?.type !== 'ReturnStatement') {
+    return false;
+  }
```
erwan-leforestier-sonarsource, line 117: I would challenge the need for verifying that the next statement is a "return statement". Most likely a `new RegExp(...)` statement is always about validating a string, regardless of what follows this line. Wdyt ?

francois-mora-sonarsource: I kept the `return` check intentionally to make the exception prove a very specific intent: the discarded `new RegExp(...)` is used to validate a value that is immediately returned.

A later use of the same identifier is weaker evidence. For example, this should still be reported:

```js
new RegExp(pattern); // forgot: this.regexp = new RegExp(pattern);
this.pattern = pattern;
```

Here `pattern` is reused afterwards, but that does not show the constructor was only for validation. It is also compatible with a real S1848 bug where the regexp result was meant to be stored and got discarded by mistake.

So the current heuristic is deliberately narrow:
- built-in `RegExp`
- immediate following `return`
- returned value directly contains one of the validated identifier arguments

That felt like the safest boundary for `JS-1997`. If we want to broaden it later, I think we would need more concrete cases than just `identifier` being used afterwards.

erwan-leforestier-sonarsource: Thanks for the explanation, i'm ok with that

erwan-leforestier-sonarsource (approved):

PR #7445, opened 2026-07-02
JS-1995 Suppress S2486 on single-statement try blocks

erwan-leforestier-sonarsource:
## Summary
Suppress S2486 when a `try` block contains a single top-level statement, which reduces noise on intentional try/fallback patterns. Keep reporting unused catch parameters for broader `try` blocks.

## Changes
- Skip the S2486 report when the enclosing `try` body has exactly one top-level statement
- Extend the rule fixture with compliant single-statement cases and a noncompliant multi-statement case
- Add a standalone FV that compares `master` and the review branch on the same sample

----
## Summary by Gitar

- **Rule logic:**
  - Modified `S2486` to exempt `try` blocks containing exactly one top-level statement, accounting for `LabeledStatement` wrappers.
- **Testing & Tooling:**
  - Updated `build.yml` to improve JS coverage artifacts handling.
  - Updated several `ruling` test expectations following the logic change in `S2486`.


francois-mora-sonarsource:
`isSingleSimpleStatement()` is still a bit too broad here. It now rejects labeled loops/switches, but it still treats a single nested block as “simple”, so this becomes compliant:

```js
try {
  { foo(); bar(); }
} catch (err) {}
```

I verified that locally with a RuleTester repro: the current rule reports 0 issues for that shape. That feels outside the intended single-operation try/fallback pattern and turns a multi-step catch body into a false negative. Could we keep reporting for block wrappers as well, and add an invalid regression for that boundary next to the existing loop/switch cases?

erwan-leforestier-sonarsource:
@francois-mora-sonarsource 
Good catch, you were right. I tightened the exemption so wrapped blocks are still reported and added a regression test for that case.

francois-mora-sonarsource (approved):
Looks good to me now.

The wrapped-block and labeled-loop boundaries are covered, the ruling delta stays aligned with the intended suppression scope, and the focused local regression check passes.

One non-blocking follow-up thought: depending on how we want to define “single statement”, we may later want to decide whether top-level `IfStatement` and nested `TryStatement` should also count as non-simple. I do not see that as a blocker for this PR as written.

PR #7452, opened 2026-07-03
JS-1971: Extend S8785 coverage

guillemsarda:
----
## Summary by Gitar

- **Rule S8785 enhancements:**
  - Extended detection to `async` helper functions that silently drop tests when called within test suites.
  - Added TypeScript-based resolution for callbacks passed by reference to identify async declarations.
- **Refactored analysis logic:**
  - Extracted analysis logic into `async-analysis.ts` and `call-to-declaration.ts` for improved maintainability.
  - Updated `rule.ts` to orchestrate detection of both top-level `await` and unawaited `async` helper calls.
- **Rule metadata updates:**
  - Renamed rule title to "Tests should not be registered asynchronously in suite callbacks" and updated quick-fix status.
  - Adjusted `S5914` and `S6582` rule metadata (titles, severity, and quick-fix indicators).

### DIFF
--- a/packages/analysis/src/jsts/rules/S8785/call-to-declaration.ts
+++ b/packages/analysis/src/jsts/rules/S8785/call-to-declaration.ts
@@ -18,6 +18,7 @@ import type estree from 'estree';
 import type { TSESTree } from '@typescript-eslint/utils';
 import ts from 'typescript';
 import type { RequiredParserServices } from '../helpers/parser-services.js';
+import { getSignatureFromCallee } from '../helpers/type.js';
 
 function normalizeToFunctionLikeDeclaration(
   declaration: ts.Declaration | undefined,
@@ -50,11 +51,7 @@ export function followCallToDeclaration(
   call: estree.CallExpression,
   services: RequiredParserServices,
 ): ts.SignatureDeclaration | undefined {
-  const tsCall = services.esTreeNodeToTSNodeMap.get(call as unknown as TSESTree.Node);
-  const signature = services.program
-    .getTypeChecker()
-    .getResolvedSignature(tsCall as ts.CallLikeExpression);
-  return normalizeToFunctionLikeDeclaration(signature?.declaration);
+  return normalizeToFunctionLikeDeclaration(getSignatureFromCallee(call, services)?.declaration);
 }
 
 /**
--- a/packages/analysis/src/jsts/rules/S8785/call-to-declaration.ts
+++ b/packages/analysis/src/jsts/rules/S8785/call-to-declaration.ts
@@ -1,0 +1,84 @@
+/*
+ * SonarQube JavaScript Plugin
+ * Copyright (C) SonarSource Sàrl
+ * mailto:info AT sonarsource DOT com
+ *
+ * You can redistribute and/or modify this program under the terms of
+ * the Sonar Source-Available License Version 1, as published by SonarSource Sàrl.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the Sonar Source-Available License for more details.
+ *
+ * You should have received a copy of the Sonar Source-Available License
+ * along with this program; if not, see https://sonarsource.com/license/ssal/
+ */
+import type estree from 'estree';
+import type { TSESTree } from '@typescript-eslint/utils';
+import ts from 'typescript';
+import type { RequiredParserServices } from '../helpers/parser-services.js';
+
+function normalizeToFunctionLikeDeclaration(
+  declaration: ts.Declaration | undefined,
+): ts.SignatureDeclaration | undefined {
+  if (declaration === undefined) {
+    return undefined;
+  }
+  if (ts.isFunctionLike(declaration)) {
+    return declaration;
+  }
+  if (
+    ts.isVariableDeclaration(declaration) &&
+    declaration.initializer !== undefined &&
+    ts.isFunctionLike(declaration.initializer)
+  ) {
+    return declaration.initializer;
+  }
+  return undefined;
+}
+
+/**
+ * Resolves a call expression to the function-like declaration it targets, using the TypeScript
+ * type checker's `getResolvedSignature`. Returns `undefined` when the callee does not resolve to
+ * a known function declaration (e.g. built-ins, overloads with no single declaration, or
+ * cross-package symbols without source).
+ *
+ * Only one hop is followed: the directly called function. Deeper call chains are not resolved.
+ */
+export function followCallToDeclaration(
+  call: estree.CallExpression,
+  services: RequiredParserServices,
+): ts.SignatureDeclaration | undefined {
+  const tsCall = services.esTreeNodeToTSNodeMap.get(call as unknown as TSESTree.Node);
+  const signature = services.program
+    .getTypeChecker()
+    .getResolvedSignature(tsCall as ts.CallLikeExpression);
+  return normalizeToFunctionLikeDeclaration(signature?.declaration);
+}
+
+/**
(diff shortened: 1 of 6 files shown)


packages/analysis/src/jsts/rules/S8785/rule.ts
```diff
@@ -140,144 +297,174 @@ function isSupportedFrameworkConstruct(
...
-function findFirstTopLevelAwait(
+function reportUnawaitedAsyncHelperCalls(
   context: Rule.RuleContext,
-  callback: estree.FunctionExpression | estree.ArrowFunctionExpression,
-): estree.Node | undefined {
+  services: RequiredParserServices,
+  callback: FunctionLike,
+  skipAwait?: estree.Node,
+): boolean {
+  if (callback.body?.type !== 'BlockStatement') {
```
nathsou: `reportUnawaitedAsyncHelperCalls` bails out here whenever `callback.body` isn't a `BlockStatement`. That means a concise-body arrow describe callback never gets checked for unawaited async helper calls:

```js
describe('suite', () => testing()); // never analyzed
describe('suite', () => { testing(); }); // correctly analyzed (per this PR's own tests)
```

Both forms are semantically identical and `testing()` can drop tests the same way in either, but only the block-bodied form is covered. Worth handling the expression-body case too (wrap it as if it were the statement, or special-case a top-level `CallExpression`/`AwaitExpression` body).

packages/analysis/src/jsts/rules/S8785/rule.ts
```diff
@@ -140,144 +297,174 @@ function isSupportedFrameworkConstruct(
...
+      // Awaited call: ExpressionStatement → AwaitExpression → CallExpression
+      // Skip if this is the top-level-await already handled by the outer detection branch.
+      if (
+        expr.type === 'AwaitExpression' &&
+        expr !== skipAwait &&
+        expr.argument.type === 'CallExpression' &&
+        !shouldSkipHelperResolution(context, expr.argument)
+      ) {
+        reported =
+          reportAsyncHelperCall(context, services, expr.argument, callback, expr) || reported;
```
nathsou: For a describe callback with two sequential awaited helper calls (`await helperA(); it(...); await helperB(); it(...);`), only the first top-level await is special-cased by `analyzeAsyncSuiteCallback`/`tryReportAwaitedAsyncHelperCall`. This branch then reports the second (and any further) awaited helper call independently, but `findTestsRegisteredAfter` has no upper bound at the *next* await — it only checks `offset >= suspendOffset`. So `helperA`'s report ends up listing tests that actually belong to `helperB` as dropped secondaries too, and those same tests get listed again in `helperB`'s own report. The two reports end up with overlapping/misleading secondary locations. Consider bounding `findTestsRegisteredAfter` by the next top-level await (if any) when computing `droppedInDescribe`.

packages/analysis/src/jsts/rules/S8785/rule.ts
```diff
@@ -140,144 +297,174 @@ function isSupportedFrameworkConstruct(
...
-    node.type === 'AwaitExpression' ||
-    (node.type === 'ForOfStatement' && (node as estree.ForOfStatement & { await: boolean }).await)
-  );
+function shouldSkipHelperResolution(
+  context: Rule.RuleContext,
+  call: estree.CallExpression,
+): boolean {
+  const calleeParts = getMochaCalleeParts(call.callee);
+  const isKnownFrameworkRegistration =
+    calleeParts !== undefined && TEST_FRAMEWORK_STRUCTURE_FUNCTIONS.has(calleeParts.base.name);
```
nathsou: `shouldSkipHelperResolution` skips helper analysis purely on a bare-name match against `TEST_FRAMEWORK_STRUCTURE_FUNCTIONS` (via `getMochaCalleeParts`), with no verification that the identifier actually resolves to an imported test-framework binding — unlike `isSupportedFrameworkConstruct` above, which does check import provenance for the same kind of name. A user's own async helper named `before`, `after`, or `test` (all in `TEST_FRAMEWORK_STRUCTURE_FUNCTIONS`) will be silently exempted from detection even though it's not actually the framework hook, so a genuine dropped-test bug in that helper won't be reported.

packages/analysis/src/jsts/rules/S8785/call-to-declaration.ts
```diff
@@ -0,0 +1,84 @@
...
+ * Only one hop is followed: the directly called function. Deeper call chains are not resolved.
+ */
+export function followCallToDeclaration(
+  call: estree.CallExpression,
+  services: RequiredParserServices,
+): ts.SignatureDeclaration | undefined {
+  const tsCall = services.esTreeNodeToTSNodeMap.get(call as unknown as TSESTree.Node);
+  const signature = services.program
+    .getTypeChecker()
+    .getResolvedSignature(tsCall as ts.CallLikeExpression);
```
nathsou: This duplicates `getSignatureFromCallee` in `helpers/type.ts` (already used by S3801/S2999/S2234), which does the same `esTreeNodeToTSNodeMap.get(...)` + `getResolvedSignature(...)` steps. Could call it directly (`getSignatureFromCallee(call, services)?.declaration`) instead of re-deriving the checker lookup here.

packages/analysis/src/jsts/rules/S8785/call-to-declaration.ts
```diff
@@ -0,0 +1,84 @@
...
+ * is a function expression / arrow function. Aliases such as `const cb = setup; describe(..., cb)`
+ * intentionally return `undefined`.
+ */
+export function followReferenceToDeclaration(
+  node: estree.Identifier,
+  services: RequiredParserServices,
+): ts.SignatureDeclaration | undefined {
+  const tsNode = services.esTreeNodeToTSNodeMap.get(node as unknown as TSESTree.Node);
+  const checker = services.program.getTypeChecker();
+  let symbol = checker.getSymbolAtLocation(tsNode);
```
nathsou: Same duplication here: `helpers/type.ts` already exports `getSymbolAtLocation` (used by S2817/S4043) which does exactly this `checker.getSymbolAtLocation(tsNode)` call. Reusing it would avoid a second TS-checker resolution path in the rules codebase for the same operation.

packages/analysis/src/jsts/rules/S8785/rule.ts
```diff
@@ -71,17 +103,121 @@ export const rule: Rule.RuleModule = {
...
+    ? analyzeAsyncSuiteCallback(context, parserServices, callback)
+    : undefined;
+
+  if (asyncSuiteOutcome?.kind === 'noTopLevelAwait') {
+    // Async callback, no top-level await: remove the misleading async keyword.
+    reportRemoveAsync(context, callback, !byReference);
+  }
+
+  if (isRequiredParserServices(parserServices)) {
+    reportUnawaitedAsyncHelperCalls(
```
nathsou, line 136: Minor efficiency note: for an async describe callback, this call runs unconditionally after `analyzeAsyncSuiteCallback` has already walked the same callback body via `findFirstTopLevelAwait` (and possibly `findTestsRegisteredAfter`). That's up to 3 independent full stack-walks of the same body per suite call site. Probably not significant given typical describe-body sizes, but a single combined pass could collect the top-level-await, dropped-tests, and helper-call information in one walk instead of three.

nathsou (changes requested):
Reviewed the S8785 extension logic (async-analysis.ts, call-to-declaration.ts, rule.ts). Found a real false-negative gap, an overlapping-secondaries bug, a false-negative from a bare-name check, a reuse opportunity, and a minor efficiency note. Details inline.

packages/analysis/src/jsts/rules/S8785/rule.ts
guillemsarda: I'll keep it as it is because it is more readable than doing one walk

nathsou (approved):
Image

PR #7470, opened 2026-07-06
JS-2008 Automate SQAA release handoff

vdiez:
## Summary
- automate the SonarJS SQAA handoff from the standard release flow while reusing the same release build number as the Java artifacts
- publish official SQAA release images to the durable Repox release registry so they do not expire after 90 days
- keep `.github/workflows/docker-sqaa.yml` as the manual SQAA-only entry point, with inputs that allow rerunning an official release with the original build number

### `.github/workflows/automated-release.yml`
- add a SonarJS-level `sqaa-integration` input
- keep the generic `release-github-actions` SQAA integration disabled because SonarJS does not use `gradle/sonar-plugins.versions.toml`
- run a local `sqaa_release` job after the standard release succeeds
- derive the SQAA Docker tag explicitly from the released `X.Y.Z.BuildNumber` before calling the local SQAA workflow
- build SQAA from `refs/tags/<release-version>` so the Docker image comes from the exact released commit
- let SQAA depend only on `sqaa-integration`, not on `sqc-integration`
- fall back to `release-ticket-url` when no SQC ticket is created
- pass `publish-to-release-repo: true` so official release images are also pushed to `repox-sonarsource-docker-releases.jfrog.io`
- make `bump_versions` wait for a successful SQAA handoff whenever `sqaa-integration` is enabled, so derive or handoff failures block the next-dev PR

### `.github/workflows/sqaa-release.yml`
- add a reusable workflow for SQAA builds and handoff
- resolve the Docker tag from an explicit `build-number` input, or the `.BuildNumber` suffix of `release-version`, or a fresh `get-build-number` call for manual non-release runs
- build and push the image to `repox-sonarsource-docker-builds.jfrog.io/a3s/analysis/javascript:<build_number>`
- optionally also push the same tag to `repox-sonarsource-docker-releases.jfrog.io/a3s/analysis/javascript:<build_number>` when `publish-to-release-repo=true`
- align npm registry setup with the current `build.yml` flow by configuring Repox manually with the Artifactory private-reader token instead of `config-npm@v1`
- create the `sonar-analysis-as-a-service` PR automatically by updating `analysis/js_ts_image_tag`

### `.github/workflows/docker-sqaa.yml`
- turn the manual workflow into a thin wrapper over the reusable SQAA workflow
- allow building from a branch or a tag
- add an optional `build-number` input so an official release can be rerun with the original tag
- add `publish-to-release-repo` so manual reruns of official releases can republish the durable release image

### `docs/RELEASE.md`
- document the shared build-number behavior between the standard release and SQAA
- document the durable Repox release-registry publication
- document the manual rerun procedure for official releases
- document that `sonar-analysis-as-a-service` must switch its pull source to the release registry

### Standard release path
- the standard SonarJS release produces `X.Y.Z.BuildNumber`
- the SQAA image reuses that same `BuildNumber`
- the SQAA image is built from the release tag, not the moving branch head
- the official release tag is published to both `repox-sonarsource-docker-builds.jfrog.io` and `repox-sonarsource-docker-releases.jfrog.io`
- the `sonar-analysis-as-a-service` PR is created automatically by updating `analysis/js_ts_image_tag`
- the next-dev bump PR is blocked if the requested SQAA handoff fails

### Manual rerun path for an official release
Run `.github/workflows/docker-sqaa.yml` with:
- `branch`: the release tag
- `build-number`: the original released build number
- `publish-to-release-repo`: `true`

This avoids minting a new Docker tag and republishes the official image to the durable release registry.

## Design Assessment

SonarJS intentionally does not use the generic SQAA integration from `release-github-actions`:

- most JVM analyzers publish plugin JARs to Repox, and the generic integration updates `gradle/sonar-plugins.versions.toml` so `sonar-analysis-as-a-service` can rebuild its shared analyzer image
- .NET publishes analyzer and service packages that are assembled into a Docker image by `sonar-analysis-as-a-service`
- SonarJS, like the custom Jasmin/Armor path, owns a self-contained gRPC Docker image and is integrated through an image-tag file rather than a plugin version catalog

For SonarJS, keeping the Docker build in this repository and updating `analysis/js_ts_image_tag` is therefore simpler and more accurate than introducing AAAS-side Gradle or packaging scaffolding. Reusing the standard release build number also gives the Java artifacts and Docker image one release identity, while publishing to the Docker release registry removes the 90-day retention problem of the builds registry.

### Known tradeoff: release-time rebuild

The SQAA image is rebuilt during the release from the exact release tag. This guarantees the same source commit and build number, but it is not strict "build once, use everywhere": the workflow reruns `npm ci`, `rspec:refresh`, and `grpc:build` instead of promoting an image produced by the original `build.yml` run. In particular, refreshed RSPEC data could change between the validated build and the release-time Docker build.

The stronger long-term model is to build and publish the SQAA image to the Docker builds registry during `build.yml`, using the same build number and prepared RSPEC artifact as the standard analyzer build. The release workflow would then promote or copy that exact image digest to the Docker releases registry before opening the AAAS PR.

This PR deliberately keeps that larger build-pipeline refactor out of scope. It automates the current custom-image release model while improving traceability, failure gating, and artifact durability.
## Follow-up Outside This PR

1. `SonarSource/sonar-analysis-as-a-service` must pull the JS/TS analyzer image from:

   ```text
   repox-sonarsource-docker-releases.jfrog.io/a3s/analysis/javascript:${source_tag}
   ```

   instead of the expiring `repox-sonarsource-docker-builds.jfrog.io` source.

2. For strict build reproducibility, move SQAA image creation into `build.yml` and make the release workflow promote the already validated image digest rather than rebuilding it.
## Verification
- `git diff --check`
- `python -c "import yaml, pathlib; yaml.safe_load(pathlib.Path('.github/workflows/sqaa-release.yml').read_text())"`
- `python -c "import yaml, pathlib; yaml.safe_load(pathlib.Path('.github/workflows/automated-release.yml').read_text())"`
- `python -c "import yaml, pathlib; yaml.safe_load(pathlib.Path('.github/workflows/docker-sqaa.yml').read_text())"`

## Notes
- I did not run the workflows themselves from this environment.
- I did not verify the Vault secret contents for `development/artifactory/token/{REPO_OWNER_NAME_DASH}-docker-release`; the workflow assumes that token exists with permission to push to the Repox release registry.

### DIFF
--- a/.github/workflows/sqaa-release.yml
+++ b/.github/workflows/sqaa-release.yml
@@ -156,16 +156,20 @@ jobs:
           artifactory-reader-role: private-reader
           disable-caching: 'true'
 
-      - uses: SonarSource/ci-github-actions/config-npm@v1
-
       - name: Access vault secrets
         id: secrets
         uses: SonarSource/vault-action-wrapper@v3
         with:
           secrets: |
+            development/artifactory/token/{REPO_OWNER_NAME_DASH}-private-reader access_token | ARTIFACTORY_ACCESS_TOKEN;
             development/artifactory/token/{REPO_OWNER_NAME_DASH}-qa-deployer access_token | ARTIFACTORY_DEPLOY_PASSWORD;
             development/artifactory/token/{REPO_OWNER_NAME_DASH}-qa-deployer username | ARTIFACTORY_DEPLOY_USERNAME;
             development/github/token/{REPO_OWNER_NAME_DASH}-rspec token | RSPEC_GITHUB_TOKEN;
+
+      - name: Configure npm registry
+        run: |
+          npm config set //repox.jfrog.io/artifactory/api/npm/:_authToken=${{ fromJSON(steps.secrets.outputs.vault).ARTIFACTORY_ACCESS_TOKEN }}
+          npm config set registry https://repox.jfrog.io/artifactory/api/npm/npm/
 
       - name: Install NPM dependencies
         run: npm ci
--- a/.github/workflows/sqaa-release.yml
+++ b/.github/workflows/sqaa-release.yml
@@ -1,0 +1,281 @@
+name: SQAA Release
+
+on:
+  workflow_call:
+    inputs:
+      ref:
+        description: Git ref to build from
+        required: true
+        type: string
+      release-version:
+        description: Released version in major.minor.patch.build format
+        required: false
+        type: string
+      build-number:
+        description: Build number to reuse for the Docker image tag
+        required: false
+        type: string
+      ticket-url:
+        description: Jira ticket URL used for the sonar-analysis-as-a-service PR
+        required: false
+        type: string
+      create-aaas-pr:
+        description: Create the sonar-analysis-as-a-service PR after pushing the image
+        required: false
+        type: boolean
+        default: false
+      aaas-base-branch:
+        description: Base branch for the sonar-analysis-as-a-service PR
+        required: false
+        type: string
+        default: master
+      reviewers:
+        description: Comma-separated list of GitHub reviewers for the sonar-analysis-as-a-service PR
+        required: false
+        type: string
+      draft:
+        description: Create the sonar-analysis-as-a-service PR as draft
+        required: false
+        type: boolean
+        default: false
+      release-automation-secret-name:
+        description: Vault secret name for a GitHub token with sonar-analysis-as-a-service access
+        required: false
+        type: string
+        default: SonarJS-release-automation
+      runner:
+        description: GitHub runner to use
+        required: false
+        type: string
+        default: warp-custom-ubuntu-24-04
+    outputs:
+      build-number:
+        description: Docker image build number
+        value: ${{ jobs.resolve_build_number.outputs.build-number }}
+      pull-request-url:
+        description: URL of the created sonar-analysis-as-a-service pull request
+        value: ${{ jobs.handoff_to_analysis_as_a_service.outputs.pull-request-url }}
+
+env:
+  FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: 'true'
+  DOCKER_REPOX_BUILDS_REGISTRY: repox-sonarsource-docker-builds.jfrog.io
+  # The registry repository still uses the historical a3s prefix.
+  DOCKER_IMAGE: "a3s/analysis/javascript"
+
+jobs:
+  resolve_build_number:
+    name: Resolve build number
+    runs-on: ${{ inputs.runner }}
+    permissions:
+      id-token: write
+      contents: read
+    outputs:
+      build-number: ${{ steps.finalize.outputs.build-number }}
+      ticket-key: ${{ steps.finalize.outputs.ticket-key }}
+    steps:
+      - name: Derive build metadata from inputs
+        id: derive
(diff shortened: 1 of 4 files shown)


.github/workflows/sqaa-release.yml
```diff
@@ -0,0 +1,316 @@
...
+        uses: SonarSource/ci-github-actions/get-build-number@master
+
+      - name: Finalize build metadata
+        id: finalize
+        shell: bash
+        env:
+          DERIVED_BUILD_NUMBER: ${{ steps.derive.outputs.build-number }}
+          GENERATED_BUILD_NUMBER: ${{ steps.get-build-number.outputs.BUILD_NUMBER }}
+          DERIVED_TICKET_KEY: ${{ steps.derive.outputs.ticket-key }}
+          CREATE_AAAS_PR: ${{ inputs.create-aaas-pr == true && 'true' || 'false' }}
```
aurelien-coet-sonarsource, line 114: Why is the `&& 'true' || 'false'` needed here ?

.github/workflows/sqaa-release.yml
```diff
@@ -0,0 +1,316 @@
...
+        run: |
+          npm config set //repox.jfrog.io/artifactory/api/npm/:_authToken=${{ fromJSON(steps.secrets.outputs.vault).ARTIFACTORY_ACCESS_TOKEN }}
+          npm config set registry https://repox.jfrog.io/artifactory/api/npm/npm/
+
+      - name: Install NPM dependencies
+        run: npm ci
+
+      - uses: ./.github/actions/rule-api-cache
+
+      - name: Refresh RSPEC rule data
```
aurelien-coet-sonarsource, line 179: Why is the RSPEC data refreshed between the "standard" build and the SQAA build ?

aurelien-coet-sonarsource (approved):
LGTM, just a couple of questions.

.github/workflows/sqaa-release.yml
vdiez: It is not strictly required. `workflow_call` preserves the input as a boolean and GitHub would stringify it when exporting the environment variable. The explicit `&& 'true' || 'false'` just normalizes the value at the workflow boundary because the bash code below compares it as a string. The same convention is used for `DRAFT` and `PUBLISH_TO_RELEASE_REPO` in this workflow.

.github/workflows/sqaa-release.yml
vdiez: Good point. The SQAA Docker image is already built separately from the standard analyzer today, so this PR does not introduce that second build. It is still a substantial improvement: the separate build is now automated from the exact release tag, reuses the analyzer release build number, publishes to the durable release registry, and gates the release handoff.

We will address strict reproducibility in a subsequent PR by building the SQAA image during `build.yml` with the prepared RSPEC artifact and promoting that exact image digest during release. I have documented this limitation and follow-up explicitly in the PR description.

vdiez: I created [JS-2076](https://sonarsource.atlassian.net/browse/JS-2076) to track this follow-up. It covers moving the SQAA image build into `build.yml`, reusing the prepared RSPEC artifact, and promoting the exact validated image digest during release.

PR #7478, opened 2026-07-07
JS-1972: Remove type assertions from rule S8784

guillemsarda:
----
## Summary by Gitar

- **Rule update:**
  - Modified rule `S8784` to ignore Vitest's compile-time type assertions `expectTypeOf` and `assertType`.
  - Implemented `isVitestTypeLevelAssertion` to distinguish these static type checks from runtime test assertions.

### DIFF
--- a/packages/analysis/src/jsts/rules/S8784/rule.ts
+++ b/packages/analysis/src/jsts/rules/S8784/rule.ts
@@ -30,8 +30,8 @@ import {
   hasSupportedAssertionLibrary,
   isAssertion,
   isScriptCapableAssertion,
+  isTypeLevelAssertion,
 } from '../helpers/assertion-detection.js';
-import { isTypeLevelAssertion } from '../helpers/vitest.js';
 import * as meta from './generated-meta.js';
 
 const messages = {
--- a/packages/analysis/src/jsts/rules/S8784/rule.ts
+++ b/packages/analysis/src/jsts/rules/S8784/rule.ts
@@ -31,7 +31,7 @@ import {
   isAssertion,
   isScriptCapableAssertion,
 } from '../helpers/assertion-detection.js';
-import { getFullyQualifiedName } from '../helpers/module.js';
+import { isTypeLevelAssertion } from '../helpers/vitest.js';
 import * as meta from './generated-meta.js';
 
 const messages = {
@@ -70,7 +70,7 @@ export const rule: Rule.RuleModule = {
         if (!isAssertion(context, node)) {
           return;
         }
-        if (isVitestTypeLevelAssertion(context, node)) {
+        if (isTypeLevelAssertion(context, node)) {
           return;
         }
         const statement = findEnclosingExpressionStatement(context, node);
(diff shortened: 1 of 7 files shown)


packages/analysis/src/jsts/rules/S8784/rule.ts
```diff
@@ -219,3 +223,38 @@ function findEnclosingFunctionIndex(ancestors: estree.Node[]): number {
...
+  }
+  let current: estree.Expression | estree.Super = node.callee.object;
+  while (current.type === 'MemberExpression') {
+    current = current.object;
+  }
+  return (
+    current.type === 'CallExpression' &&
+    current.callee.type === 'Identifier' &&
+    VITEST_TYPE_LEVEL_CALLEE_NAMES.has(current.callee.name)
+  );
```
nathsou: **Unscoped name-based fallback causes false negatives.**

This branch matches any chained call whose innermost identifier is literally named `expectTypeOf`/`assertType`, with no import/FQN check at all. That's broader than the stated rationale ("Vitest's compile-time type checks").

This is demonstrably wrong in this PR's own diff: `fixtures/no-test-structure.fixture.js` changes
```js
expectTypeOf(value).toEqualTypeOf(); // Noncompliant
```
to `// Compliant`, in a file that has **no vitest import anywhere** — right next to `expectObservable(source$).toBe('a-b'); // Noncompliant`, an equally-unimported global name that's still correctly flagged. Any project with an unrelated top-level `expectTypeOf`/`assertType`-shaped call (a local helper, or the real `expect-type` npm package that Vitest's own `expectTypeOf` is built on) will now silently stop being flagged.

Suggest dropping this fallback and matching by fully-qualified name only (reusing `ASSERTION_ROOTS` from `helpers/vitest.ts`, see next comment).

packages/analysis/src/jsts/rules/S8784/rule.ts
```diff
@@ -219,3 +223,38 @@ function findEnclosingFunctionIndex(ancestors: estree.Node[]): number {
   }
   return -1;
 }
+
+// Vitest's `expectTypeOf` and `assertType` are compile-time type checks, not runtime
+// assertions — Vitest's type-checking pipeline passes these files to `tsc` for static
+// analysis and never executes them. Placing them outside `test()` is idiomatic in
+// `.test-d.ts` files, so S8784 must not flag them.
+const VITEST_TYPE_LEVEL_ROOTS = ['vitest.expectTypeOf', 'vitest.assertType'];
```
nathsou: `VITEST_TYPE_LEVEL_ROOTS` re-literals `'vitest.expectTypeOf'`/`'vitest.assertType'`, which already exist in `helpers/vitest.ts`'s `ASSERTION_ROOTS` (line 40). A second, unsynchronized source of truth — a future addition to Vitest's type-level API surface only needs to be added in one place to compile and pass tests, silently leaving the other stale.

Consider exporting a predicate or root-list from `helpers/vitest.ts` instead (mirroring the existing `isTSSetupCall`/`NON_ASSERTION_EXPECT_MEMBERS` carve-out pattern already in that file) and importing it here.

packages/analysis/src/jsts/rules/S8784/rule.ts
```diff
@@ -219,3 +223,38 @@ function findEnclosingFunctionIndex(ancestors: estree.Node[]): number {
...
+const VITEST_TYPE_LEVEL_CALLEE_NAMES = new Set(['expectTypeOf', 'assertType']);
```
nathsou: The `'assertType'` entry in this set is unreachable. The name-based branch below only fires for chained calls (`node.callee.type === 'MemberExpression'`), but `assertType(value)` is never chained in Vitest's real API — any genuine `assertType` usage is already caught by the FQN branch above, which returns first. This entry can never be the one that matches.

nathsou (changes requested):
Reviewed the S8784 Vitest type-level exception. The exclusion is currently too broad and can silently suppress real S8784 violations for code that has nothing to do with Vitest — see the inline comment for a concrete repro using this PR's own fixture. Two smaller cleanup notes alongside it.

packages/analysis/src/jsts/rules/S8784/rule.ts
guillemsarda: Good catch! I've added as well a fqn check for `expec-type`'s `expectTypeOf`

nathsou (approved):
Image

packages/analysis/src/jsts/rules/helpers/assertion-detection.ts
```diff
@@ -215,6 +215,26 @@ export function isScriptCapableAssertion(context: Rule.RuleContext, node: estree
...
+ */
+export function isTypeLevelAssertion(context: Rule.RuleContext, node: estree.Node): boolean {
+  if (node.type !== 'CallExpression') {
+    return false;
+  }
+  const fqn = getFullyQualifiedName(context, node);
+  return (
+    fqn !== null &&
+    TYPE_LEVEL_ASSERTION_ROOTS.some(root => fqn === root || fqn.startsWith(`${root}.`))
+  );
```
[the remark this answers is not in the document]
guillemsarda: The report is fair but we have already handled this case in previous PR comments, please see them. For next time, gather the context of the previous comments (if any) and see if what you'll state has already been considered.

PR #7480, opened 2026-07-07
JS-2014: Implement S8980: react testing library calls should not be wrapped in act()

nathsou:
## Summary

Implements S8980: flags `act(...)` calls (from `@testing-library/react` or `react-dom/test-utils`) whose callback consists entirely of React Testing Library / `user-event` calls, since those utilities already wrap themselves in `act()` internally.

- Decorates `eslint-plugin-testing-library`'s `no-unnecessary-act` rule (`implementation: 'decorated'`) rather than reimplementing Testing Library import/call resolution from scratch.
- Forces `isStrict: false` (upstream default is `true`) so only callbacks made **entirely** of Testing Library calls are flagged; mixed content stays compliant.
- Overrides `context.settings` locally in the rule's own `decorator.ts` (not the shared linter settings) to disable the upstream plugin's "aggressive" name-based heuristic detection, forcing precise import-resolution-only matching.
- Drops the upstream rule's `noUnnecessaryActEmptyFunction` report (`act(() => {})`) via `interceptReport`, since that case is already covered by S1186.
- Gated by `requiredDependency: ['@testing-library/react']`.
- Added `eslint-plugin-testing-library@7.16.2` as a devDependency (consistent with the recent `eslint-plugin-vue` precedent for S8950/JS-1952).

## Known upstream limitation

Found and worked around (not patched) a real ordering bug in `eslint-plugin-testing-library@7.16.2`: importing `@testing-library/user-event` before `react-dom/test-utils` in the same file causes the plugin's own `react-dom/test-utils`-imported `act` detection to be silently skipped, due to a stray `!importedUserEventLibraryNode` guard instead of `!importedReactDomTestUtilsNode`. This is an inherited false-negative risk in that narrow scenario, tracked as a follow-up rather than a blocker.

## Testing

Comment-based fixtures cover: `render`/`fireEvent`/implicit-return `act(() => screen.getByRole(...))`/`userEvent`/`waitFor` noncompliant cases, `act` imported from `react-dom/test-utils` (isolated fixture, see limitation above), plain RTL calls with no `act` wrapper (compliant), `act` wrapping only non-Testing-Library code (compliant — proves the rule won't mask genuine act warnings), mixed TL/non-TL content (compliant, verifies `isStrict: false`), and empty `act(() => {})` (not reported by this rule, verifies no overlap with S1186).

## Links

- Jira: https://sonarsource.atlassian.net/browse/JS-2014
- RSPEC PR: https://github.com/SonarSource/rspec/pull/7322


guillemsarda (approved):
Image

PR #7483, opened 2026-07-07
JS-2015: Implement S8981: regular expressions used in Testing Library queries should not have the global flag

nathsou:
## Summary

- Implement S8981: flag a regular expression with the global (`g`) flag used as a matcher in a Testing Library query (`getByText`, `getByRole`'s `name` option, etc.). The `g` flag makes the regex stateful (`lastIndex`), and since Testing Library reuses the same regex instance to test several candidate DOM nodes, a query can silently skip an element it should have matched.
- Wire up `eslint-plugin-testing-library` as a new external rule source (`external/testing-library.ts` + `external/registry.ts`, mirroring the existing `react`/`vue`/`unicorn` wrappers) and re-export its `no-global-regexp-flag-in-query` rule as a bare `external` rule, reusing its detection and autofix as-is instead of reimplementing the query-matching heuristics.
- Deliberately keeps the upstream "Aggressive Reporting" default on (no file-scope gate, no restricting `testing-library/*` settings): it flags calls by name pattern (`getBy*`/`queryBy*`/`findBy*` family) even without a `@testing-library/*` import in the file, which is what's needed to catch real occurrences such as Nx's generated Storybook `play` functions (`canvas.getByText(/name/gi)`), which don't import Testing Library directly.
- The generated Java check class targets `TestFileCheck` (RSPEC `scope: Tests`) to bound this to the canonical `*.test.*`/`*.spec.*` usage on the SonarQube/SonarLint side, since that distribution has no combined Main+Test targeting option. The standalone `eslint-plugin-sonarjs` package has no such restriction and will still flag non-test files like Storybook stories.
- No dedicated rule-level tests added, matching the established convention for every other bare `external` rule in this repo (verified against S1656, S4204, S6750, S6756, S6807, S7650, S7726, S7744 — none carry rule-specific test files); correctness relies on upstream's own test suite plus the repo's "Plugin public API" consistency test. Verified locally with a manual smoke test (literal `g`+`i` direct arg, `getByRole` `name` option, and a variable-held regex all fire with the correct autofix; the non-`g` case is silent).

## Links

- Jira: https://sonarsource.atlassian.net/browse/JS-2015
- RSPEC PR: https://github.com/SonarSource/rspec/pull/7324

---

> [!NOTE]
> **Low Risk**
> Test-scoped lint rule and build-time bundling fix; no auth, security, or production runtime logic changes beyond enabling a new ESLint plugin in the analyzer bundle.
> 
> **Overview**
> Adds **S8981** as an external rule that re-exports `testing-library/no-global-regexp-flag-in-query`, flagging regex matchers with the global (`g`) flag in Testing Library `getBy*` / `queryBy*` / `findBy*` calls (with autofix to drop `g`).
> 
> **eslint-plugin-testing-library** `7.16.2` is added as a dependency, with a new `external/testing-library` wrapper and registry entry, SonarQube metadata (`scope: Tests`, Sonar way), docs links, and **unit tests** for the re-exported rule.
> 
> The bridge bundle gains an **esbuild text-replace** (plus a build-time guard) so the plugin’s `createRequire(import.meta.url)` package.json read does not break bundled CJS startup.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d93705da24fc98880704e8a8df00e9b139a8da49. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/analysis/src/jsts/rules/S8981/meta.ts
```diff
@@ -0,0 +1,20 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the Sonar Source-Available License for more details.
+ *
+ * You should have received a copy of the Sonar Source-Available License
+ * along with this program; if not, see https://sonarsource.com/license/ssal/
+ */
+export const implementation = 'external';
+export const eslintId = 'no-global-regexp-flag-in-query';
+export const externalPlugin = 'testing-library';
```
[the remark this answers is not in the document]
nathsou: Fixed in 6ae9a7a07: added `case 'testing-library'` to the `externalURL` switch in `tools/generate-external-rules-docs.ts` and regenerated `packages/analysis/src/jsts/rules/README.md` to include the S8981 row. Thanks for catching this.

packages/analysis/src/jsts/rules/S8981/meta.ts
```diff
@@ -0,0 +1,28 @@
...
+export const requiredDependency = [
+  '@testing-library/dom',
+  '@testing-library/react',
+  '@testing-library/vue',
+  '@testing-library/angular',
+  '@testing-library/svelte',
+  'storybook',
```
guillemsarda, line 26: `storybook` does not appear on the list of available libraries: https://github.com/testing-library/eslint-plugin-testing-library/blob/main/src/utils/index.ts#L23

packages/analysis/src/jsts/rules/S8981/unit.test.ts
```diff
@@ -0,0 +1,68 @@
...
+            code: `
+              const re = /foo/g;
+              screen.getByText(re);
+            `,
+            output: `
+              const re = /foo/;
+              screen.getByText(re);
+            `,
+            errors: [{ messageId: 'noGlobalRegExpFlagInQuery' }],
+          },
```
guillemsarda, line 67: The test suite only exercises getBy* queries. I'd suggest adding at least one case for the queryBy* and findBy* families, and a valid case confirming non-query calls are not flagged:

```ts
valid: [
  // ...existing cases...
  {
    // global flag on a regex not passed to a query — should not be flagged
    code: `const re = /foo/g; someNonQueryFn(re);`,
  },
],
invalid: [
  // ...existing cases...
  {
    // queryBy* family
    code: `screen.queryByText(/hello/g);`,
    output: `screen.queryByText(/hello/);`,
    errors: [{ messageId: 'noGlobalRegExpFlagInQuery' }],
  },
  {
    // findBy* (async) family
    code: `await screen.findByText(/hello/g);`,
    output: `await screen.findByText(/hello/);`,
    errors: [{ messageId: 'noGlobalRegExpFlagInQuery' }],
  },
],
```

guillemsarda (changes requested):

packages/analysis/src/jsts/rules/S8981/meta.ts
guillemsarda: If you confirm it, we should update the rspec too

esbuild-common.mjs
```diff
@@ -161,6 +164,21 @@ export async function buildBundle({ entryPoint, outfile, additionalAssets = [] }
...
+        pattern: [
+          [
+            'createRequire(import.meta.url)(resolve(dirname(fileURLToPath(import.meta.url)), "../package.json"))',
+            JSON.stringify({
+              name: testingLibraryPkgJson.name,
+              version: testingLibraryPkgJson.version,
+            }),
+          ],
+        ],
+      }),
```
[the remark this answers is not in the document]
nathsou: Good catch — verified esbuild-plugin-text-replace's replaceAll-based implementation directly, it has no match-count check and would silently no-op. Fixed in f06df01b5: added a build-time assertion that the exact createRequire(import.meta.url)(...) expression still exists in the bundled eslint-plugin-testing-library source before relying on the replace, throwing a clear error otherwise.

packages/analysis/src/jsts/rules/S8981/unit.test.ts
nathsou: Fixed in d93705da2 — added the queryByText/findByText invalid cases with output assertions, plus the non-query negative case, matching your suggested snippet exactly.

packages/analysis/src/jsts/rules/S8981/meta.ts
nathsou: Confirmed - storybook isn't in LIBRARY_MODULES, and that list is real (it drives the plugin's own import-detection). But it's not actually load-bearing here: with aggressive query reporting on (the default we keep), isQuery() in detect-testing-library-utils.ts is a pure name-pattern check (/^(get|query|find)(All)?By.+$/) with zero dependency on imports or LIBRARY_MODULES — confirmed by reading that function directly.

storybook is in a different list: SonarJS's own requiredDependency/filterDependency, which gates whether this rule runs at all for a given project (separate mechanism from anything upstream does). It's there because Storybook's test utilities (@storybook/test, storybook/test) only pull in @testing-library/dom transitively — a Storybook-only project (the strongest real-world case motivating this rule, from the RSPEC history) would never satisfy the gate without it. Happy to add a code comment on the requiredDependency array clarifying this distinction if useful — let me know.

guillemsarda (approved):
Image

PR #7484, opened 2026-07-07
JS-2013 Align S6842 with the ARIA in HTML conformance table

erwan-leforestier-sonarsource:
## Summary
Align `S6842` with the full [ARIA in HTML conformance table](https://w3c.github.io/html-aria/#docconformance), not only the composite-widget overrides. jsx-a11y's element→roles allowlist cannot express context, so the rule is now decorated: `config.ts` keeps the flat, spec-derived allowlists and `decorator.ts` handles everything the allowlist cannot. This brings the JS/TS analyzer to parity with the sonar-html implementation (SONARHTML-403) — both now agree with the spec on every element/role pair.

## Changes
- `config.ts`: enumerated allowlists per the conformance table — `ul`/`ol`/`menu` (drop `treegrid`), `nav`, `h1`–`h6`, `fieldset`, `td`, `progress`; `li` is left to the decorator.
- `decorator.ts` (new): suppresses the combinations the table permits but the allowlist cannot describe:
  - elements that accept **any role** (`table`/`tbody`/`thead`/`tfoot`, text-level elements such as `p`, `blockquote`, `time`, `output`, …);
  - context-sensitive elements — `li` (only restricted when a parent exposes a `list` role), `img` (interactive roles allowed only with an accessible name), `figure` (restricted only with a `figcaption`), `label` (restricted only when associated with a control);
  - `toolbar` treated as a structure role (non-interactive), consistent with sonar-html and the WAI-ARIA role taxonomy.

## Fixes vs. the previous revision
- Removes false positives on `nav`, headings, named `img`, and any-role elements.
- Restores the missing `treegrid`-on-`ul`/`ol` report (spec forbids it).
- Drops the incorrect `table[role="button"]` / `tbody[role="button"]` true positives — the table family accepts any role.

## Tests
`unit.test.ts` rewritten as a spec matrix (valid + invalid) covering every case above.

### DIFF
--- a/its/ruling/src/test/expected/desktop/typescript-S6842.json
+++ b/its/ruling/src/test/expected/desktop/typescript-S6842.json
@@ -1,5 +1,0 @@
-{
-"desktop:app/src/ui/lib/vertical-segmented-control/segmented-item.tsx": [
-71
-]
-}
--- a/packages/analysis/src/jsts/rules/S6842/config.ts
+++ b/packages/analysis/src/jsts/rules/S6842/config.ts
@@ -17,35 +17,30 @@
 // https://sonarsource.github.io/rspec/#/rspec/S6842/javascript
 
 import type { ESLintConfiguration } from '../helpers/configs.js';
-import { getUpstreamRecommendedConfiguration } from '../external/a11y.js';
 
-type Allowlist = Record<string, string[]>;
+// Interactive roles that the ARIA in HTML conformance table
+// (https://w3c.github.io/html-aria/#docconformance) permits per element.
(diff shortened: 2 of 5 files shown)


francois-mora-sonarsource (approved):
LGTM.

Non-blocking follow-up if you decide to cover one more JSX edge case: template-literal accessible names on `img` still look worth adding to the valid matrix, for example:

```jsx
<img aria-label={`${x} logo`} role="button" />
<img alt={`Logo`} role="button" />
```

The current dynamic-name coverage handles identifiers and calls; this would close the same accessible-name path for template literals too.

PR #7497, opened 2026-07-08
JS-2018 Fix S2699 typed assertion helper traversal

jdkandersson:
## Summary
- extract S8785 call/reference declaration resolution into a shared helper
- add implementation fallback for typed function values whose call resolves to a type signature
- use the shared implementation resolver in S2699's TypeScript assertion traversal
- add S2699 regression for node:test + node:assert/strict helper-wrapped assertions, matching the knip sample pattern

## Validation
- npm ci
- npm run generate-meta
- ./node_modules/.bin/tsx --tsconfig packages/tsconfig.test.json --test --test-reporter=spec --test-reporter-destination stdout packages/analysis/tests/jsts/rules/helpers/assertion-detection.test.ts packages/analysis/tests/jsts/rules/helpers/call-to-declaration.test.ts packages/analysis/src/jsts/rules/S2699/unit.test.ts packages/analysis/src/jsts/rules/S2699/cb.test.ts packages/analysis/src/jsts/rules/S2699/jest/cb.test.ts packages/analysis/src/jsts/rules/S2699/jasmine/cb.test.ts packages/analysis/src/jsts/rules/S2699/cypress/cb.test.ts packages/analysis/src/jsts/rules/S2699/playwright/cb.test.ts packages/analysis/src/jsts/rules/S2699/typescript-detection/cb.test.ts packages/analysis/src/jsts/rules/S8785/unit.test.ts
- npm run bridge:compile
- git diff --check


packages/analysis/src/jsts/rules/S2699/rule.ts
```diff
@@ -241,11 +242,14 @@ class TestCaseAssertionVisitor {
...
-      const callNode = services.program
-        .getTypeChecker()
-        .getResolvedSignature(node as ts.CallLikeExpression);
-      if (callNode?.declaration) {
-        nodeHasAssertions ||= this.visitTSNode(services, callNode.declaration, visitedTSNodes);
+      const estreeNode = services.tsNodeToESTreeNodeMap.get(node);
+      const declaration =
+        estreeNode?.type === 'CallExpression'
+          ? followCallToImplementation(estreeNode, services)
+          : services.program.getTypeChecker().getResolvedSignature(node as ts.CallLikeExpression)
```
nathsou: `tsNodeToESTreeNodeMap` is built per-file during parsing, so it only has entries for nodes in the currently-linted source file. `followCallToDeclaration` (via `getResolvedSignature`) can resolve a call to a declaration in a *different* file (e.g. an imported typed helper), and `visitTSNode` recurses into it via `forEachChild`. Any nested `CallExpression` inside that cross-file declaration has no entry in this file's `tsNodeToESTreeNodeMap`, so `estreeNode` is `undefined` and the code falls back to the old `getResolvedSignature(...).declaration` path here — meaning the new fix doesn't apply, and the original false-positive bug still reproduces for typed helpers declared in another file.

Worth a regression test with the assertion helper imported from a separate module, if that's a realistic case for this rule.

packages/analysis/src/jsts/rules/helpers/call-to-declaration.ts
```diff
@@ -76,3 +102,22 @@ export function followReferenceToDeclaration(
...
+ */
+export function followCallToImplementation(
+  call: SupportedCallExpression,
+  services: RequiredParserServices,
+): ts.SignatureDeclaration | undefined {
+  const declaration = followCallToDeclaration(call, services);
+  if (hasBody(declaration)) {
+    return declaration;
+  }
+  if (call.callee.type === 'Identifier') {
```
nathsou, line 278: This fallback only triggers when `call.callee.type === 'Identifier'`. A typed assertion helper invoked via member access (e.g. `const helpers = { check: (a, b) => assert.deepEqual(a, b) } satisfies { check: Helper }; test('x', () => helpers.check(1, 1))`) still resolves to a bodyless call-signature declaration — `hasBody` is false, but the callee is a `MemberExpression`, so no fallback runs here and the bodyless declaration is returned as-is. That reproduces the exact false positive this PR fixes, just one property access away, and isn't covered by either new test file.

Might be worth handling `MemberExpression`/property-access callees too (e.g. resolve the property's value declaration the same way), or at least noting it as a known limitation in the doc comment above `followCallToImplementation`.

packages/analysis/src/jsts/rules/helpers/call-to-declaration.ts
```diff
@@ -16,8 +16,12 @@
  */
 import type estree from 'estree';
 import ts from 'typescript';
-import type { RequiredParserServices } from '../helpers/parser-services.js';
-import { getSignatureFromCallee, getSymbolAtLocation } from '../helpers/type.js';
+import type { TSESTree } from '@typescript-eslint/utils';
+import type { RequiredParserServices } from './parser-services.js';
+import { getSignatureFromCallee, getSymbolAtLocation } from './type.js';
+
+type SupportedCallExpression = estree.CallExpression | TSESTree.CallExpression;
```
nathsou: The `SupportedCallExpression`/`SupportedIdentifier` unions plus the internal `as estree.CallExpression`/`as estree.Identifier` casts (lines 74, 78, 92-93) add type surface across all three exported functions to support the one new S2699 call site that hands in a TSESTree node. Since `getSignatureFromCallee`/`getSymbolAtLocation` in `type.ts` already accept the wider `estree.Node`, and these casts don't add real type safety (they're blind assertions), this could likely be simplified to a single cast at the S2699 call site instead of threading the estree/TSESTree distinction through the whole shared module.

nathsou (commented):
Automated review of the S2699 typed-assertion-helper traversal fix. The new `followCallToImplementation` fallback is a real improvement for the identifier-callee case, but a few gaps and a cleanup item are worth a look before merge (see inline comments).

One more note that doesn't map to a changed line: `isFunctionLikeDeclaration` in `helpers/call-to-declaration.ts` (lines 51-63) is a byte-for-byte duplicate of the private helper already in `packages/analysis/src/jsts/rules/S4623/rule.ts` (lines 104-115). Since this PR already created a shared helpers module, it may be worth exporting one copy and having S4623 import it, to avoid two hand-maintained SyntaxKind allowlists drifting apart.

packages/analysis/src/jsts/rules/S2699/rule.ts
jdkandersson: Addressed in 1046f59f40. S2699 now uses `followTypeScriptCallToImplementation` when a recursively visited TS call has no ESTree mapping, so imported helper declarations can still resolve typed helper implementations. I also added a multi-file regression in `call-to-declaration.test.ts` covering an imported declaration body with a typed helper call.

packages/analysis/src/jsts/rules/helpers/call-to-declaration.ts
jdkandersson: Addressed in 3cbdfa2828. `followCallToImplementation` now handles non-computed member expressions by resolving `const` object-literal member implementations, with S2699 and helper-level regressions for `helpers.check(...)`. Follow-up 872c917bc2 also avoids trusting that initializer when the object member is reassigned.

packages/analysis/src/jsts/rules/helpers/call-to-declaration.ts
jdkandersson: Addressed in 3cbdfa2828. The shared helper API is back to ESTree-only inputs; the TSESTree support surface and internal blind casts were removed, and S2699 now performs the one explicit cast at the guarded `estreeNode?.type === "CallExpression"` call site.

packages/analysis/src/jsts/rules/helpers/call-to-declaration.ts
```diff
@@ -0,0 +1,298 @@
...
+  if (symbol === undefined) {
+    return undefined;
+  }
+  if ((symbol.flags & ts.SymbolFlags.Alias) !== 0) {
+    symbol = services.program.getTypeChecker().getAliasedSymbol(symbol);
+  }
+  if (symbol.declarations?.length !== 1) {
+    return undefined;
+  }
+  return normalizeToFunctionLikeDeclaration(symbol.declarations[0]);
```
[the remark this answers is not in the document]
jdkandersson: Addressed in 23c6e790cb. The ESTree and TypeScript entry points now delegate to shared symbol-based helpers (`followSymbolToDeclaration` and `followObjectLiteralSymbolMemberToDeclaration`), so alias handling, const object-member resolution, and reassignment guards are maintained in one place.

nathsou (approved):
All prior findings are addressed:

- Cross-file fallback gap: `followTypeScriptCallToImplementation`/`followTypeScriptCalleeToDeclaration` now handle the ESTree-less path, with a dedicated cross-file test.
- `MemberExpression` callee gap: `followMemberPropertyToDeclaration`/`followObjectLiteralMemberToDeclaration` now resolve `obj.check(...)`-style calls, with tests for both a plain property and a method-shorthand property.
- `SupportedCallExpression`/`SupportedIdentifier` union + casts: removed; the estree-facing functions kept their original signatures and the ESTree-less path got its own `ts.Node`-native functions instead.
- `isFunctionLikeDeclaration` duplication with S4623: resolved by exporting it from the shared helper and having S4623 import it.
- Bonus: the reassignment edge case (`let`-bound and reassigned object-member helpers) is now guarded via `isConstVariableDeclaration`/`hasPropertyWrite`, with tests for both directions.

One non-blocking note I raised after the previous round (duplicated symbol-resolution logic between the estree-facing and ts-native functions) has also since been factored out into shared `followSymbolToDeclaration`/`followObjectLiteralSymbolMemberToDeclaration` helpers in the latest commit — nothing outstanding from my side. LGTM.

PR #7504, opened 2026-07-08
JS-2025 Use official cache and artifact actions in CI

vdiez:
## Summary

- replace `SonarSource/gh-action_cache` usage in `build.yml` with official GitHub cache actions where the workflow depends on standard `lookup-only` semantics
- use workflow artifacts for explicit job-to-job handoff of JS coverage reports and the ESLint plugin tarball
- keep the existing local GitHub cache wrappers for Maven, orchestrator, and rule-api data

## Why

This follows up on #7481 and #7494.

The recent `gh-action_cache` behavior change exposed that parts of our workflow were relying on cache side effects rather than explicit handoff. For the SonarJS build workflow, we do not need the SonarSource S3 cache wrapper for these paths, and using the official GitHub cache/artifact actions makes the behavior align with what the workflow is actually trying to do.

## Testing

- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/build.yml"); puts "build.yml parsed"'`
- CI not run locally

----
## Summary by Gitar

- **CI documentation:**
  - Added `docs/CI.md` as a canonical reference for the CI pipeline architecture and data flow.
  - Updated `docs/BUILD.md` to link to the new documentation.
- **Cleanup workflow:**
  - Added `pr-cleanup.yml` to automatically prune PR-scoped GitHub caches and artifacts upon closure.
- **Dependency management:**
  - Aligned `populate_npm_cache_win` with `get_build_number` dependency and added shared environment variables.
  - Integrated `validate_js_coverage_reports` as a shared step in `test_js` and `analyze_primary` to ensure data integrity during artifact handoff.
- **Action pinning:**
  - Pinned all `actions/cache` and `actions/download-artifact` references to specific SHAs and versions for stability.

### DIFF
--- a/.github/actions/maven-cache/action.yml
+++ b/.github/actions/maven-cache/action.yml
@@ -15,7 +15,7 @@ runs:
     # Restore only on branches (no save) - allows branches to read from default branch's cache
     - name: Restore Maven cache (branches)
       if: github.ref_name != github.event.repository.default_branch
-      uses: actions/cache/restore@v6
+      uses: actions/cache/restore@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
       with:
         path: ~/.m2/repository
         key: maven-${{ runner.os }}-${{ inputs.cache-month }}-${{ inputs.maven-hash }}
@@ -24,7 +24,7 @@ runs:
     # Full cache on default branch (save enabled)
     - name: Cache Maven dependencies (default branch)
       if: github.ref_name == github.event.repository.default_branch
-      uses: actions/cache@v6
+      uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
       with:
         path: ~/.m2/repository
         key: maven-${{ runner.os }}-${{ inputs.cache-month }}-${{ inputs.maven-hash }}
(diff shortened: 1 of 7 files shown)


nathsou (approved):
Approving, with two non-blocking cleanup comments:

- `.github/workflows/pr-cleanup.yml:11-18` — The cleanup workflow is not fork-safe. It runs for every closed PR and requests `actions: write`, but fork-PR runs normally receive a read-only `GITHUB_TOKEN` unless the repo/org setting explicitly sends write tokens to fork workflows. If a fork PR has PR-scoped caches, `SonarSource/ci-github-actions/pr_cleanup@v1` will reach `gh cache delete` without error suppression and fail the cleanup job. Consider guarding this job to internal PRs, moving cleanup to a safe privileged trigger, or making cache deletion explicitly best-effort for fork PRs.

- `.github/workflows/build.yml:401-416` — The tarball relocation fallback is dead under the current artifact contract. `build_eslint_plugin` uploads the fixed file `lib/eslint-plugin-sonarjs-0.0.0-SNAPSHOT.tgz`, and `test_eslint_plugin` downloads that artifact into `lib`, so the first `test -f "$TARBALL"` is the expected and sufficient check. If the package name/version contract drifts, the upload step already fails with `if-no-files-found: error`; the later `find`/`mv` path either never runs or papers over a layout drift that should stay visible.

vdiez:
Thanks. The fork-safety note is accepted; we do not run CI for forks here. I also checked the ESLint tarball point against the current artifact contract, and you're right: with a single named artifact downloaded by name into `lib`, the relocation fallback is dead code and can hide a contract drift. I removed that fallback and kept only the explicit existence check.

PR #7508, opened 2026-07-08
JS-2025 Improve CI reproducibility and ruling maintainability

vdiez:
## Summary
- restore the shared `node_modules` cache in `build_eslint_plugin` and stop `eslint-plugin:build` from wiping it with `npm ci`
- strengthen JS skip-cache keys so workflow and dependency changes do not silently reuse old success
- move the `js_ts_ruling` PR comment and fix-PR automation out of workflow YAML into the repo-local `./.github/actions/ruling_bot` action
- make `ruling_bot` reusable for sonar-lits analyzers by passing explicit result-root inputs and source/link inputs instead of hard-coding the SonarJS layout
- preserve the existing SonarJS PR ruling report format in `ruling_bot`, including rule-grouped sections, source links, inline snippets, and the collapsible full report
- make `npm run ruling-sync` reuse the same sync implementation as `ruling_bot`
- pin internal `SonarSource/ci-github-actions` refs from `@master` to the stable `@v1` major tag
- update the CI and build docs to match the workflow behavior

## Out of scope
- collapsing the plugin QA family into a reusable workflow or stronger parameterization still deserves a dedicated follow-up PR
- RSPEC refresh is still live in CI; this PR only tightens action-ref reproducibility and cache-key safety

## Validation
- `ruby -e "require 'yaml'; YAML.load_file('.github/workflows/build.yml'); puts 'build.yml ok'"`
- `ruby -e "require 'yaml'; YAML.load_file('.github/actions/ruling_bot/action.yml'); puts 'ruling_bot action ok'"`
- `node --check .github/actions/ruling_bot/sync-results.mjs`
- `node --check .github/actions/ruling_bot/generate-report.mjs`
- `git diff --check`
- synthetic parity check: compared `ruling_bot` output against the previous SonarJS-specific ruling reporter on a controlled ruling diff and matched the generated PR report body
- `npm ci` could not be completed locally because the configured npm registry auth in this environment returns `E401`; full runtime validation is expected from CI


nathsou (approved):
Image

PR #7524, opened 2026-07-10
JS-2050: S8980: explain why act() is unnecessary in the issue message

nathsou:
## Summary

The S8980 issue message inherited from upstream (`eslint-plugin-testing-library`'s `no-unnecessary-act`) only states what's wrong ("Avoid wrapping Testing Library util calls in `act`"), not why it matters or what to do about it.

- Overrides the message at report time (same pattern already used by S8950 for its own decorated rule) to: **"Remove this redundant `act()` call; the wrapped call already flushes its own updates."**
- Updated the three comment-based fixtures to assert the new message text.
- No detection-logic change — verified first via a fresh `uc peach diff` review that the current detection is sound (4/4 new real-world issues since merge confirmed true positives by inspecting the checked-out source imports directly, 0 false positives, 0 unclear) before touching only the wording.

## Links

- Jira: https://sonarsource.atlassian.net/browse/JS-2050
- Original rule PR: https://github.com/SonarSource/SonarJS/pull/7480

### DIFF
--- a/packages/analysis/src/jsts/rules/S8980/cb.fixture.tsx
+++ b/packages/analysis/src/jsts/rules/S8980/cb.fixture.tsx
@@ -6,23 +6,23 @@ function Form() {
 }
 
 function submitsForm() {
-    act(() => { // Noncompliant {{Avoid wrapping Testing Library util calls in `act`}}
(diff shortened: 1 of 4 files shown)


guillemsarda (approved):
Image

PR #7526, opened 2026-07-10
JS-2043 Decorate new rule S8988: Side effects should not be performed inside computed properties

guillemsarda:
Part of 

----
## Summary by Gitar

- **New Rule S8988:**
  - Implemented `no-side-effects-in-computed-properties` rule, wrapping the existing `vue/no-side-effects-in-computed-properties` rule.
  - Added comprehensive unit tests covering Options API and Composition API for both JS and TS.
- **Configuration:**
  - Registered the new rule in the SonarJS plugin and included it in the `Sonar way` profile.
  - Updated documentation in `README.md` and added JSON rule definition.
- **Refactor:**
  - Cleaned up `S8985.json` tags and dependency list.


nathsou (approved):
Image

PR #7527, opened 2026-07-10
JS-2042 Decorate rule S8987: "v-if" and "v-for" should not be used on the same element

guillemsarda:
Part of 

----
## Summary by Gitar

  - Added rule `S8987` to detect and report `v-if` and `v-for` used on the same Vue element.
  - Implemented `decorator.ts` to intercept and customize error messages based on `v-if` dependency on loop variables.
- **Testing:**
  - Added comprehensive unit tests in `unit.test.ts` covering both standard Vue and TypeScript configurations.
- **Rule Metadata:**
  - Created `S8987.json` and registered the new rule in `Sonar_way_profile.json`.
- **Minor fixes:**
  - Updated the title of `S8988.json` to include "Vue" context.


packages/analysis/src/jsts/rules/S8987/unit.test.ts
```diff
@@ -0,0 +1,168 @@
...
+</template>
+`,
+          errors: [
+            {
+              message:
+                'Replace `getUsers()` with a computed property that returns a filtered array.',
+            },
+          ],
+        },
+        {
```
nathsou, line 90: Missing test coverage: this `movedToWrapper` invalid case (v-if unrelated to the loop variable) is only tested under default options. There's no test combining `options: [{ allowUsingIterationVar: true }]` with this unrelated-`v-if` scenario to confirm it's still reported as invalid. The only place `allowUsingIterationVar: true` is exercised is the `valid` case above (line 68), which only proves the loop-var-referencing case gets suppressed — it doesn't prove the option is scoped correctly and doesn't also suppress this unrelated case. A future change that widens the option's suppression scope would go undetected. Suggest adding an `invalid` case with `v-if="showList"` + `options: [{ allowUsingIterationVar: true }]` asserting `movedToWrapper` is still reported.

packages/analysis/src/jsts/rules/S8987/decorator.ts
```diff
@@ -0,0 +1,47 @@
...
+  return interceptReport(
+    {
+      ...rule,
+      meta: generateMeta(meta, rule.meta),
+    },
+    (context, reportDescriptor) => {
+      if ('messageId' in reportDescriptor) {
+        const { messageId, data, ...rest } = reportDescriptor;
+        if (messageId === 'shouldUseComputed') {
+          const { iteratorName } = data as { iteratorName: string; kind: string };
```
nathsou: `data as { iteratorName: string; kind: string }` is an unsound cast — ESLint's own types leave `data` as `MessagePlaceholderData | undefined` (`Record<string, string | number | boolean | bigint | null | undefined>`), and `messageId` is typed as plain `string` by default, so narrowing on `messageId === 'shouldUseComputed'` gives no compiler-enforced link to `data`'s actual shape. It works today only because the upstream `eslint-plugin-vue` rule happens to always supply this exact shape for `shouldUseComputed`. Low practical risk right now, but if a future upstream version changes or drops `data` for this messageId, this would silently produce `undefined` in the message rather than failing a type check or an explicit runtime guard.

nathsou (changes requested):
Two findings from review, both verified against the actual code/types (not just skimmed).

packages/analysis/src/jsts/rules/S8987/unit.test.ts
guillemsarda: Good catch. Investigating this revealed that `allowUsingIterationVar` isn't exposed by S8987 at all (`defaultOptions: []`, no `config.ts`). The valid case at line 68 was already testing upstream rule behaviour that users can't trigger from SonarJS. Rather than adding the suggested invalid counterpart (which would have the same problem), we removed both option-based cases. The remaining tests cover all the behaviour SonarJS actually exposes.

nathsou (approved):
Image

PR #7535, opened 2026-07-13
JS-2062: Implement S9012: Snapshots should not be generated inside Testing Library async utilities

nathsou:
## Summary

Implements S9012 by decorating Testing Library's `no-wait-for-snapshot` rule with strict import resolution and the Sonar diagnostic wording.

## Links

- Jira: https://sonarsource.atlassian.net/browse/JS-2062
- RSPEC: https://github.com/SonarSource/rspec/pull/7382

## Verification

- `uc check`


packages/analysis/src/jsts/rules/S9012/unit.test.ts
```diff
@@ -0,0 +1,110 @@
...
+          `,
+            errors: [
+              {
+                message:
+                  'Snapshots should not be generated inside `waitFor`; retries can create unstable snapshot artifacts.',
+              },
+            ],
+          },
+          {
+            code: `
```
guillemsarda, line 78: **Optional tests coverage**

**Missing coverage for strict import-resolution override**
This rule’s main added behavior is withStrictImportResolution(...), but the tests only cover the default no-settings case. It would be worth adding a case where the file explicitly sets Testing Library aggressive settings and proving the decorator still ignores a same-named local helper.
```js
{
  code: `
    import { waitFor } from './test-utils';

    await waitFor(() => expect(document.body).toMatchSnapshot());
  `,
  settings: {
    'testing-library/utils-module': './test-utils',
  },
}
```
That would lock in the actual decorator contract, not just the upstream rule behavior.

**Missing import-shape coverage for a decorated rule**
The invalid cases only exercise named ESM imports. Since this decorator depends on import resolution and rewrites the reported utility name, it would be safer to cover at least one namespace import and one CommonJS import shape too.

```js
{
  code: `
    import * as rtl from '@testing-library/react';

    await rtl.waitFor(() => expect(document.body).toMatchSnapshot());
  `,
  errors: [
    {
      message:
        'Snapshots should not be generated inside `waitFor`; retries can create unstable snapshot artifacts.',
    },
  ],
}
{
  code: `
    const { waitFor } = require('@testing-library/react');

    await waitFor(() => expect(document.body).toMatchSnapshot());
  `,
  errors: [
    {
      message:
        'Snapshots should not be generated inside `waitFor`; retries can create unstable snapshot artifacts.',
    },
  ],
}
```
That would make the decorated-rule coverage less tied to a single import style.

guillemsarda (approved):
I left 2 optional suggestions for tests.

Image

packages/analysis/src/jsts/rules/README.md
```diff
@@ -684,5 +684,6 @@ The following rules are used in SonarJS but not available in this ESLint plugin.
...
 | S8988           | [vue/no-side-effects-in-computed-properties](https://eslint.vuejs.org/rules/no-side-effects-in-computed-properties.html)                                                                                                                                                                                                                                                                |
 | S8990           | [testing-library/no-promise-in-fire-event](https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/no-promise-in-fire-event.md)                                                                                                                                                                                                                           |
+| S9012           | [testing-library/no-wait-for-snapshot](https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/no-wait-for-snapshot.md)                                                                                                                                                                                                                                   |
```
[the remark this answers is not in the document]
nathsou: Confirmed: S8987 is already enabled in Sonar Way on current master, but its decorated-rules README row is missing there as well. This PR did not introduce the documentation drift; the master merge only surfaced it. I will leave this unrelated fix out of S9012.

packages/analysis/src/jsts/rules/S9012/unit.test.ts
nathsou: Addressed in commit 0f9bc4394. The S9012 tests now cover the strict settings override with a local helper, a namespace import, and a CommonJS import. The focused S9012 test and uc check pass.

PR #7537, opened 2026-07-13
JS-2066: Implement S9020: Use "find*" to query Testing Library elements that may not be available yet

nathsou:
## Summary

- Implement S9020 as a decorated `testing-library/prefer-find-by` rule.
- Resolve Testing Library utilities strictly by import and retain only safe upstream quick fixes.
- Cover supported rewrites, no-fix safeguards, aliases, valid `waitFor` patterns, and TypeScript syntax.

## Links

- Jira: https://sonarsource.atlassian.net/browse/JS-2066
- RSPEC: https://github.com/SonarSource/rspec/pull/7391

----
## Summary by Gitar

- **Rule Metadata:**
  - Integrated `S9020` into `Sonar way` and `Sonar agentic AI` quality profiles.
  - Defined metadata including `RSPEC-9020` mapping and `testing-library` dependency requirements.
- **Decorator Logic:**
  - Implemented `decorator.ts` to selectively disable risky autofixes for `waitFor` patterns involving variable options or synchronous wrappers.
  - Added `isKnownNonTestingLibraryQuery` to prevent false positives when using third-party libraries like `playwright`.


packages/analysis/src/jsts/rules/S9020/decorator.ts
```diff
@@ -0,0 +1,123 @@
...
+import type { Rule } from 'eslint';
+import type estree from 'estree';
+import { interceptReport } from '../helpers/decorators/interceptor.js';
+import { generateMeta } from '../helpers/generate-meta.js';
+import { withStrictImportResolution } from '../helpers/testing-library.js';
+import * as meta from './generated-meta.js';
+
+export function decorate(rule: Rule.RuleModule): Rule.RuleModule {
+  return interceptReport(
+    { ...withStrictImportResolution(rule), meta: generateMeta(meta, rule.meta) },
```
guillemsarda, line 29: Same issue as before for potential future issues:

This rule still inherits one upstream limitation around query detection.
`waitFor` is now resolved strictly by import, which is good, but the query inside the callback is still recognized mostly by its name. That means a non-Testing-Library API such as `Playwright`’s `page.getByRole(...)` can still be treated as if it were a Testing Library query.
Example:
```js
import { waitFor } from '@testing-library/react';
import { page } from '@playwright/test';

const status = await waitFor(() => page.getByRole('status', { name: 'Saved' }));
```
This gets reported, and can even be autofixed to page.findByRole(...), even though this is not a Testing Library call.
So the rule can still raise false positives on same-named helpers from other libraries.

guillemsarda (approved):
Minor comment to be aware of.
Image

packages/analysis/src/jsts/rules/S9020/decorator.ts
nathsou: Fixed in 4a15d7c91. S9020 now resolves member-query fully qualified names and suppresses reports for known non-Testing Library imports, including Playwright page.getByRole, while preserving Testing Library rewrites. Added a regression test for the Playwright example. The project check passes: 4 passed, 2 expected skips.

packages/analysis/src/jsts/rules/S9020/decorator.ts
```diff
@@ -0,0 +1,165 @@
...
+import { withStrictImportResolution } from '../helpers/testing-library.js';
+import * as meta from './generated-meta.js';
+
+const RECOGNIZED_MODULE_PREFIXES = [
+  '@testing-library.dom',
+  '@testing-library.react',
+  '@testing-library.vue',
+  '@testing-library.angular',
+  '@testing-library.svelte',
+];
```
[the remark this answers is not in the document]
nathsou: Fixed in bff1dc9f1. The guard now recognizes any @testing-library/* fully qualified query instead of a hard-coded framework list, so packages such as preact (and react-native when supported upstream) are not suppressed. Added an @testing-library/preact regression test; the project check passes: 4 passed, 2 expected skips.

PR #7538, opened 2026-07-13
JS-2054 New rule S9011: "<button>" elements should have an explicit "type" attribute

guillemsarda:
----
## Summary by Gitar

  - Added rule `S9011` to enforce explicit `type` attributes on HTML `<button>` elements.
  - Implemented logic by decorating and merging existing `react/button-has-type` and `vue/html-button-has-type` rules.
- **Configuration and metadata:**
  - Registered `S9011` in the `Sonar_way_profile.json` and added it to the rules `README.md`.
  - Included rule definition in `S9011.json` with appropriate severity and tags.
- **Testing:**
  - Added comprehensive unit tests in `unit.test.ts` covering both React and Vue scenarios.
  - Included expected ruling results for various test projects to ensure rule compliance across the codebase.


sonar-plugin/javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S9011.json
```diff
@@ -0,0 +1,32 @@
+{
+  "title": "\"<button>\" elements should have an explicit \"type\" attribute",
+  "type": "CODE_SMELL",
+  "status": "ready",
+  "remediation": {
+    "func": "Constant/Issue",
+    "constantCost": "2min"
+  },
+  "tags": [
+    "angular",
```
erwan-leforestier-sonarsource: No angular

packages/analysis/src/jsts/rules/S9011/rule.ts
```diff
@@ -0,0 +1,81 @@
...
+export const rule: Rule.RuleModule = {
+  meta: generateMeta(meta, {
+    messages: {
+      ...reactBaseRule.meta?.messages,
+      ...vueBaseRule.meta?.messages,
+    },
+  }),
+  create(context: Rule.RuleContext) {
+    return mergeRules(reactRule.create(context), vueRule.create(context));
+  },
```
erwan-leforestier-sonarsource, line 155: The ruling bot on this PR flagged 93 new issues on jsts-test-sources, most of them unrelated to form submission. Two examples: `courselit/packages/rich-text/src/EditorWithToolbar.js:122` is a toolbar "Add" button inside a rich-text editor, and `console/src/views/account/ResetPasswordView/ResetPasswordView.tsx:85` fires its action via `onClick`, not native form submission. The rule has no form-context check, so it reports every button missing a `type` attribute regardless of whether it is inside or associated with a form. Should we restrict to buttons that have a `<form>` ancestor or carry a `form=` attribute?

packages/analysis/src/jsts/rules/S9011/rule.ts
```diff
@@ -0,0 +1,81 @@
...
+    context.report({
+      ...rest,
+      message: `Replace this invalid "type" value "${value}" with one of "button", "submit", or "reset".`,
+    });
+  }
+}
+
+const reactRule = interceptReportForReact(reactBaseRule, onReport);
+const vueRule = interceptReport(vueBaseRule, onReport);
+
```
erwan-leforestier-sonarsource, line 135: Vue's `:type` binding is accepted as soon as the expression is non-empty, so `:type="'action'"` and `:type="isSubmit ? 'submit' : 'action'"` both pass. The same literals in JSX (`type={'action'}`) are correctly flagged. Simple bound values like string literals and ternaries with literal branches can be evaluated statically before we skip the check.

packages/analysis/src/jsts/rules/S9011/rule.ts
```diff
@@ -0,0 +1,81 @@
...
+function onReport(context: Rule.RuleContext, reportDescriptor: Rule.ReportDescriptor) {
+  if (!('messageId' in reportDescriptor)) {
+    context.report(reportDescriptor);
+    return;
+  }
+  const { messageId, data, ...rest } = reportDescriptor;
+  if (MISSING_TYPE_MESSAGE_IDS.has(messageId)) {
+    context.report({ ...rest, message: MISSING_TYPE_MESSAGE });
+  } else if (INVALID_TYPE_MESSAGE_IDS.has(messageId)) {
+    const { value } = data as { value?: string };
```
erwan-leforestier-sonarsource: `<button type="">` in React goes through the `invalidValue` branch and gets the message "Replace this invalid type value '' with one of 'button', 'submit', or 'reset'", while Vue routes it to `emptyTypeAttribute` and says "Add an explicit type attribute". The React message is confusing because it refers to replacing an invalid value, but the real intent is to add a meaningful one. Both should say the same thing.

packages/analysis/src/jsts/rules/S9011/rule.ts
```diff
@@ -0,0 +1,81 @@
...
+    context.report({
+      ...rest,
+      message: `Replace this invalid "type" value "${value}" with one of "button", "submit", or "reset".`,
+    });
```
erwan-leforestier-sonarsource, line 77: HTML's `type` attribute is case-insensitive, so `<button type="Submit">` is valid and browsers treat it as a submit button. The rule flags it as invalid. Lowercasing the value before the membership check would fix it.

packages/analysis/src/jsts/rules/S9011/rule.ts
guillemsarda: In my opinion, I think we should keep the current behaviour of keeping the upstream plugins logic and raise the issue always when the button type is not defined. It is true that there is the risk of raising a lot of issues but it is a [good practice](https://www.w3schools.com/Tags/att_button_type.asp) to explicitly define the type. So I think we should enforce it.

Moreover, if we limit the scope to just fire when we know that a button is inside a form, we would be missing many true positives. 

For instance in React, we create a button component that we do not know if it's going to be used in a form.
```jsx
function Button() {
  return <button>Submit</button> // We do not raise because it is not explicitly inside a form
}
```
Then in another file we use it:
```jsx
function Form() {
  return ( 
    <form>
      <Button /> // We should raise it because it's inside a form but we can't resolve the jsx
    </form>
  )
}
```

What do you think about it?

packages/analysis/src/jsts/rules/S9011/rule.ts
guillemsarda: Good catch! I'll update the code

packages/analysis/src/jsts/rules/S9011/rule.ts
guillemsarda: Nice! I'll apply the changes

packages/analysis/src/jsts/rules/S9011/rule.ts
guillemsarda: Done!

packages/analysis/src/jsts/rules/S9011/rule.ts
erwan-leforestier-sonarsource: Yeah I know but better having less FP and less TP than more FP and more TP. It's up to you to decide, in any case we will see if this rule is widely adopted or rejected and correct later if need be.

erwan-leforestier-sonarsource (approved):

packages/analysis/src/jsts/rules/S9011/rule.ts
guillemsarda: I'll take the risk and go forward with it. I'll add a comment in the task to have it in mind during the evaluation of the rule.

PR #7547, opened 2026-07-14
JS-2061 Create rule S9025: Asynchronous actions should not be used in Vue computed properties

guillemsarda:
----
## Summary by Gitar

- **New rule:**
  - Implemented rule `S9025` to detect asynchronous actions within Vue computed properties.
  - Added rule decorator in `decorator.ts` to provide specific feedback for Options and Composition API.
- **Rule integration:**
  - Configured metadata in `meta.ts` and registered the rule in `index.ts` using `vue/no-async-in-computed-properties`.
- **Testing:**
  - Added `unit.test.ts` covering various scenarios including `Promise`, `async/await`, `setTimeout`, and `nextTick` for both JavaScript and TypeScript.


packages/analysis/src/jsts/rules/S9025/decorator.ts
```diff
@@ -0,0 +1,152 @@
...
+): NodeWithParent<FunctionNodeType> | undefined {
+  let current = node.parent as NodeWithParent<estree.Node> | undefined;
+  while (current) {
+    if (isFunctionNode(current)) {
+      return current as NodeWithParent<FunctionNodeType>;
+    }
+    current = current.parent as NodeWithParent<estree.Node> | undefined;
+  }
+  return undefined;
+}
```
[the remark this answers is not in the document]
guillemsarda: We decided to not cover these cases.

packages/analysis/src/jsts/rules/S9025/decorator.ts
```diff
@@ -0,0 +1,152 @@
...
+    call.arguments.includes(fn) &&
+    getFullyQualifiedName(context, call) === 'vue.computed'
+  );
+}
+
+function createNextTickImportListener(context: Rule.RuleContext): Rule.RuleListener {
+  return {
+    CallExpression(node: estree.CallExpression) {
+      const call = node as NodeWithParent<estree.CallExpression>;
+      if (getFullyQualifiedName(context, call) !== 'vue.nextTick') {
```
nathsou: **Confirmed double report.** Any call reachable via an identifier imported from `'vue'` resolves to the FQN `vue.nextTick` here, regardless of the local binding name — `getFullyQualifiedName` prepends the module path unconditionally, it doesn't care whether the import is named, default, or namespace. So `import Vue from 'vue'; ...; Vue.nextTick(() => {})` inside a computed function matches this listener's `vue.nextTick` check *and* the upstream rule's own naive `callee.object.name === 'Vue'` check, producing two identical `"Move this asynchronous action out of the computed function."` reports on the same node.

Verified by running the rule through `RuleTester` on:
```js
import Vue from 'vue';
import { computed, ref } from 'vue';
const value = ref(0);
const data = computed(() => {
  Vue.nextTick(() => {});
  return value.value;
});
```
→ 2 errors reported instead of 1.

The existing unit tests don't catch this because the `Vue.nextTick` test case never actually imports `Vue` (relies on the bare undeclared global), which sidesteps this listener's FQN resolution entirely.

Separately: per JS-2061, this whole `nextTick`-import-detection addition (this function, `findEnclosingFunction`, `getOptionsApiComputedPropertyName`, `isCompositionApiComputedCallback`) is out of the ticket's scope — the ticket only asks for message rewording via `interceptReport`, following the `S8988` pattern with no extra detection logic. Worth confirming whether this feature should be here at all, or dropped in favor of matching the ticket.

packages/analysis/src/jsts/rules/S9025/decorator.ts
```diff
@@ -0,0 +1,152 @@
...
+// Composition API: computed(() => { ... }), where `computed` is imported from 'vue'
+function isCompositionApiComputedCallback(
+  context: Rule.RuleContext,
+  fn: NodeWithParent<FunctionNodeType>,
+): boolean {
+  // a FunctionDeclaration can never be a call argument, only an expression can
+  if (fn.type === 'FunctionDeclaration') {
+    return false;
+  }
+  const call = fn.parent;
```
nathsou: **Confirmed false negative for writable computed properties.** Vue supports a `get`/`set` object for writable computed refs in both APIs — `computed({ get() {...}, set(v) {...} })` (Composition API) and `computed: { foo: { get() {...}, set(v) {...} } }` (Options API). The upstream rule already unwraps the `get` function for all of its own checks (`Promise`, `await`, `setTimeout`, `this.$nextTick`, `Vue.nextTick`) via `getGetterBodyFromComputedFunction`/`getComputedProperties` — but this function only recognizes `fn` when it's the *direct* call argument (`call.arguments.includes(fn)`), so a `get` method nested inside an `ObjectExpression` argument never matches, and `getOptionsApiComputedPropertyName` above has the same direct-parent-chain assumption for the Options API equivalent.

Verified with `RuleTester`:
```js
const data = computed({
  get() {
    nextTick(() => {});
    return value.value;
  },
  set(v) {
    value.value = v;
  }
});
```
→ 0 errors reported (expected 1), even though `await`/`new Promise`/`setTimeout`/`this.$nextTick` inside the same `get()` would correctly be flagged via the upstream rule, unaffected by this gap.

No test in `unit.test.ts` covers the writable-computed form, which is why this wasn't caught. Given this is also not part of JS-2061's scope (see the other comment on this file), it may be simpler to drop this bare-`nextTick`-import detection entirely rather than extend it to handle `get`/`set`.

packages/analysis/src/jsts/rules/S9025/decorator.ts
guillemsarda: Thank you for the catch! Fixed it

packages/analysis/src/jsts/rules/S9025/decorator.ts
guillemsarda: Yes, I'll drop the feature at the end because we were covering more than expected. I'll leave a comment in the task to be aware that this is a FN.

packages/analysis/src/jsts/rules/S9025/decorator.ts
guillemsarda: Rolling back.

nathsou (approved):
Image

PR #7558, opened 2026-07-15
JS-393 Add regression for destructured hook props

erwan-leforestier-sonarsource:
## Summary
Adds regression coverage for the original JS-393 S6767 false-positive scenario where a component forwards its props to a helper that destructures the parameter. The current rule behavior is already correct on `master`, and this PR locks that behavior in.

## Changes
- add the Jira reproducer as a TypeScript S6767 regression test
- cover the combination of whole-props forwarding and helper-parameter destructuring
- keep the change test-only, with no analyzer behavior change


erwan-leforestier-sonarsource:
Was already fixed by PR #7458

PR #7559, opened 2026-07-15
JS-2086: Implement S9026: Promises returned by Testing Library async queries should be handled

nathsou:
## Summary

- Decorate `testing-library/await-async-queries` as S9026.
- Use the selected direct-query and wrapper issue messages.
- Preserve upstream handling while suppressing unsafe direct-query fixes outside async functions.
- Pin refreshed rule data to the linked RSPEC revision.

## Links

- Jira: https://sonarsource.atlassian.net/browse/JS-2086
- RSPEC PR: https://github.com/SonarSource/rspec/pull/7404


packages/analysis/src/jsts/rules/S9026/decorator.ts
```diff
@@ -0,0 +1,74 @@
+/*
```
guillemsarda, line 1: The upstream rule has a known FP which could potentially affect the performance of the rule: https://github.com/testing-library/eslint-plugin-testing-library/issues/673.
I think it's fine if we go forward with this implementation for now, but we should add this issue in the Jira ticket to be aware of it in the future.

guillemsarda (approved):
Image

packages/analysis/src/jsts/rules/S9026/decorator.ts
nathsou: Documented in https://sonarsource.atlassian.net/browse/JS-2086?focusedCommentId=964449

PR #7560, opened 2026-07-15
JS-2087: Implement S9027: Testing Library queries should match presence assertions

nathsou:
## Summary

- add original rule S9027 for Testing Library `screen` queries used with incompatible presence or absence assertions
- provide precise fixes between `getBy*` and `queryBy*` query families
- pin RSPEC metadata to the matching specification commit until it merges

## Validation

- focused S9027 comment-based test
- `npm run validate-quickfix`
- `uc check` (type check, unit test, merge, and Knip checks)

## Links

- Jira: https://sonarsource.atlassian.net/browse/JS-2087
- RSPEC PR: https://github.com/SonarSource/rspec/pull/7405


packages/analysis/src/jsts/rules/S9027/rule.ts
```diff
@@ -0,0 +1,221 @@
...
+
+const PRESENCE_MATCHERS = new Set(['toBeInTheDocument', 'toBeTruthy', 'toBeDefined']);
+const ABSENCE_MATCHERS = new Set(['toBeNull', 'toBeFalsy']);
+
+export const rule: Rule.RuleModule = {
+  meta: generateMeta(meta, { messages, fixable: 'code' }),
+  create(context: Rule.RuleContext) {
+    return {
+      CallExpression(node: estree.Node) {
+        if (node.type !== 'CallExpression') {
```
guillemsarda, line 49: This check is redundant, the common pattern is doing type casting: `const call = node as estree.CallExpression;`

packages/analysis/src/jsts/rules/S9027/rule.ts
```diff
@@ -0,0 +1,221 @@
...
+          return;
+        }
+
+        const assertion = getAssertion(node);
+        if (!assertion) {
+          return;
+        }
+
+        const query = getQueryCall(assertion.actual);
```
guillemsarda, line 58: **on getQueryCall (line 137)**

**FN: TS non-null/type assertions on the query result bypass the rule**

`getQueryCall` only recognizes a bare `CallExpression` or a `MemberExpression` (`.length`/index) directly on the query call. It doesn't unwrap `TSNonNullExpression` (`!`) or `TSAsExpression` (`as X`), so these common TS idioms are silently missed:

```ts
expect(screen.queryByRole('dialog')!).toBeInTheDocument();
expect(screen.queryByRole('dialog') as HTMLElement).toBeInTheDocument();
```

Since `queryBy*` returns `T | null`, asserting non-null or casting is a very natural thing to reach for in a TS test — which is exactly the audience this rule targets — so this isn't a corner case, it's a realistic gap in recall.

Suggest unwrapping `TSNonNullExpression`/`TSAsExpression` (and `TSSatisfiesExpression` for completeness) before the existing `CallExpression`/`MemberExpression` checks, e.g.:

```ts
function unwrapAssertions(expr: estree.Expression): estree.Expression {
  const e = expr as any;
  if (e.type === 'TSNonNullExpression' || e.type === 'TSAsExpression' || e.type === 'TSSatisfiesExpression') {
    return unwrapAssertions(e.expression);
  }
  return expr;
}
```

Not a blocker, but worth a fast-follow since it's likely to matter in real codebases.

packages/analysis/src/jsts/rules/S9027/rule.ts
```diff
@@ -0,0 +1,221 @@
...
+const TESTING_LIBRARY_MODULES = new Set([
+  '@testing-library/dom',
+  '@testing-library/react',
+  '@testing-library/vue',
+  '@testing-library/angular',
+  '@testing-library/svelte',
+]);
+
+const PRESENCE_MATCHERS = new Set(['toBeInTheDocument', 'toBeTruthy', 'toBeDefined']);
+const ABSENCE_MATCHERS = new Set(['toBeNull', 'toBeFalsy']);
```
guillemsarda, line 42: **Minor: two narrower-than-necessary matches**

1. `ABSENCE_MATCHERS` covers `toBeNull`/`toBeFalsy` but not `toBeUndefined`. `queryBy*` resolves to `null` rather than `undefined`, so this is a minor gap, but `.toBeUndefined()` shows up in the wild as a loose absence check.
2. The `AllBy*` + index branch (`isArrayIndex`) only recognizes a literal, non-negative integer property (`arr[0]`), not a variable index (`arr[i]`). `screen.getAllByText('item')[i]).not.toBeDefined()` won't be flagged even though it's the same pattern as the `arr[0]` case already covered by the fixture.

Neither is a correctness bug — both are just missed detections, and expanding either is optional scope for a v1. Flagging in case they're easy wins alongside the non-null/`as` fix.

guillemsarda (approved):
Non-blocking comments
Image

packages/analysis/src/jsts/rules/S9027/rule.ts
nathsou: Thanks! Keeping this guard is intentional. ESLint types this listener parameter as estree.Node, so the runtime check preserves type safety and follows the same narrowing pattern used by existing SonarJS rules (for example, S5906 and S8967). Replacing it with an unchecked type assertion would weaken that safety, so I am leaving it unchanged.

PR #7562, opened 2026-07-15
JS-2089 Support Playwright test.expect assertions

erwan-leforestier-sonarsource:
## Summary
Fix S2699 false positives on Playwright tests that use `test.expect(...)`.
The rule now treats the Playwright `test.expect` alias as an assertion in both the AST-only and type-aware detection paths.

## Changes
- extend shared assertion detection to recognize Playwright `test.expect(...)`
- add Playwright fixture coverage for the comment-based rule tests
- add a type-aware regression fixture for the alias

## Functional Validation
Artifact: [JS-2089-fv.zip](https://github.com/user-attachments/files/30052877/JS-2089-fv.zip)

    ./run.sh

Expected output:
```text
Analyzing "sample/input.ts"...
Results:
    - Rule "S2699" -> L.3

****** Branch "fix/js-2089-support-playwright-test-expect" ******
Analyzing "sample/input.ts"...
Results:
    (none)
```

### DIFF
--- a/packages/analysis/src/jsts/rules/helpers/assertion-detection.ts
+++ b/packages/analysis/src/jsts/rules/helpers/assertion-detection.ts
@@ -98,6 +98,10 @@ const GLOBAL_EXPECT_NAMES = new Set([
 
 const PLAYWRIGHT_TEST_EXPECT_FQN = '@playwright.test.test.expect';
 
+function isPlaywrightTestExpectFqn(fqn: string | null | undefined): boolean {
+  return fqn?.replaceAll('/', '.') === PLAYWRIGHT_TEST_EXPECT_FQN;
+}
+
 const CHAI_NON_TERMINAL_PROPERTY_NAMES = new Set([
   'all',
   'also',
@@ -308,8 +312,7 @@ function isGlobalExpectExpressionJS(
   const innerCall = current;
   return (
     (innerCall.callee.type === 'Identifier' && GLOBAL_EXPECT_NAMES.has(innerCall.callee.name)) ||
-    getFullyQualifiedName(context, innerCall.callee)?.replaceAll('/', '.') ===
-      PLAYWRIGHT_TEST_EXPECT_FQN
+    isPlaywrightTestExpectFqn(getFullyQualifiedName(context, innerCall.callee))
   );
 }
 
@@ -461,8 +464,7 @@ function isGlobalExpectExpression(
   return (
     (innerCallExpression.expression.kind === ts.SyntaxKind.Identifier &&
       GLOBAL_EXPECT_NAMES.has((innerCallExpression.expression as ts.Identifier).text)) ||
-    getFullyQualifiedNameTS(services, innerCallExpression.expression)?.replaceAll('/', '.') ===
-      PLAYWRIGHT_TEST_EXPECT_FQN
+    isPlaywrightTestExpectFqn(getFullyQualifiedNameTS(services, innerCallExpression.expression))
   );
 }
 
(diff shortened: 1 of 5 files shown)


packages/analysis/src/jsts/rules/helpers/assertion-detection.ts
```diff
@@ -96,6 +96,25 @@ const GLOBAL_EXPECT_NAMES = new Set([
...
+function isPlaywrightExpectFqn(fqn: string | null | undefined): boolean {
+  if (!fqn) {
+    return false;
+  }
+  const normalized = fqn.replaceAll('/', '.');
+  return PLAYWRIGHT_EXPECT_FQN_ROOTS.some(
+    root => normalized === root || normalized.startsWith(root + '.'),
+  );
+}
+
```
victor-peixoto-sonarsource: Suggestion: move this to a separate module, similar to the other test frameworks.

packages/analysis/src/jsts/rules/helpers/assertion-detection.ts
```diff
@@ -300,7 +323,10 @@ function isGlobalExpectExpressionJS(node: estree.CallExpression): boolean {
   }
 
   const innerCall = current;
-  return innerCall.callee.type === 'Identifier' && GLOBAL_EXPECT_NAMES.has(innerCall.callee.name);
+  return (
+    (innerCall.callee.type === 'Identifier' && GLOBAL_EXPECT_NAMES.has(innerCall.callee.name)) ||
+    isPlaywrightExpectFqn(getFullyQualifiedName(context, innerCall.callee))
```
victor-peixoto-sonarsource: Even with the suggestion above to move the playwright logic to a separate module, we might still need to keep this here instead of replicating the global expression walk there and having this in two places.

packages/analysis/src/jsts/rules/helpers/assertion-detection.ts
```diff
@@ -446,8 +475,9 @@ function isGlobalExpectExpression(node: ts.CallExpression) {
 
   const innerCallExpression = current as ts.CallExpression;
   return (
-    innerCallExpression.expression.kind === ts.SyntaxKind.Identifier &&
-    GLOBAL_EXPECT_NAMES.has((innerCallExpression.expression as ts.Identifier).text)
+    (innerCallExpression.expression.kind === ts.SyntaxKind.Identifier &&
+      GLOBAL_EXPECT_NAMES.has((innerCallExpression.expression as ts.Identifier).text)) ||
+    isPlaywrightExpectFqn(getFullyQualifiedNameTS(services, innerCallExpression.expression))
```
victor-peixoto-sonarsource: Same as above. Even if you move the playwright logic to a separate module, I believe it is best to keep this check here.

victor-peixoto-sonarsource (approved):
Looks good! Approved with suggestions.

packages/analysis/src/jsts/rules/helpers/assertion-detection.ts
erwan-leforestier-sonarsource: Good call. I moved the Playwright-specific `expect` matching into `playwright.ts`.

packages/analysis/src/jsts/rules/helpers/assertion-detection.ts
erwan-leforestier-sonarsource: Yep, agreed. I kept the shared expression walk here and only delegated the Playwright-specific check.

packages/analysis/src/jsts/rules/helpers/assertion-detection.ts
erwan-leforestier-sonarsource: Agreed. I kept this check here as well and only extracted the Playwright helper.

PR #7566, opened 2026-07-16
IST-5039: publish CycloneDX SBOM with npm bridge payload

tomverin:
## Summary
- Generate a Maven CycloneDX BOM for `sonar-javascript-plugin`
- Generate an npm CycloneDX BOM for bundled bridge production dependencies (`cyclonedx-npm`)
- Hierarchically merge both BOMs and publish as `sonar-javascript-plugin-*-cyclonedx.json` to Repox

Addresses the npm/`pkg:npm` completeness gap for the JavaScript analyzer (IST-5039 Phase 2).

## Context
[IST-5039](https://sonarsource.atlassian.net/browse/IST-5039)

## Test plan
- [ ] `mvn -pl sonar-plugin/sonar-javascript-plugin -am package` produces `bom-merged.json`
- [ ] Merged BOM contains `pkg:npm` components
- [ ] CI deploy publishes `sonar-javascript-plugin:json:cyclonedx` to Repox
- [ ] Downstream SQS edition BOM merge can consume the sidecar when on matching plugin version

[IST-5039]: https://sonarsource.atlassian.net/browse/IST-5039?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sonar-plugin/sonar-javascript-plugin/pom.xml
```diff
@@ -558,6 +558,87 @@
...
+                <argument>--hierarchical</argument>
+                <argument>--input-files</argument>
+                <argument>${project.build.directory}/bom.json</argument>
+                <argument>${maven.multiModuleProjectDirectory}/target/bom-npm.json</argument>
+                <argument>--output-file</argument>
+                <argument>${project.build.directory}/bom-merged.json</argument>
+                <argument>--group</argument>
+                <argument>${project.groupId}</argument>
+                <argument>--name</argument>
+                <argument>${project.artifactId}</argument>
```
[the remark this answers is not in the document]
tomverin: Applied — merge now uses tools/merge-cyclonedx-bom.sh which downloads cyclonedx-cli v0.25.0 when not present on the runner.

sonar-plugin/sonar-javascript-plugin/pom.xml
```diff
@@ -558,6 +558,87 @@
...
+            </goals>
+            <configuration>
+              <executable>cyclonedx</executable>
+              <workingDirectory>${project.basedir}</workingDirectory>
+              <arguments>
+                <argument>merge</argument>
+                <argument>--hierarchical</argument>
+                <argument>--input-files</argument>
+                <argument>${project.build.directory}/bom.json</argument>
+                <argument>${maven.multiModuleProjectDirectory}/target/bom-npm.json</argument>
```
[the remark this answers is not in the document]
tomverin: Applied — plugin package phase now runs npm cyclonedx:npm from the repo root before merge, so bom-npm.json exists even without -am.

sonar-plugin/sonar-javascript-plugin/pom.xml
```diff
@@ -558,6 +558,97 @@
...
+          </execution>
+        </executions>
+        <configuration>
+          <projectType>library</projectType>
+          <schemaVersion>1.6</schemaVersion>
+          <includeTestScope>false</includeTestScope>
+          <outputFormat>json</outputFormat>
+          <outputName>bom</outputName>
+        </configuration>
+      </plugin>
```
julien-carsique-sonarsource: Plugin version and configuration (but eventually not the execution) should come from a parent pom, in the `pluginManagement` section.

sonar-plugin/sonar-javascript-plugin/pom.xml
```diff
@@ -558,6 +558,97 @@
...
+
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>exec-maven-plugin</artifactId>
+        <version>3.6.3</version>
```
julien-carsique-sonarsource: Set the plugin version in a parent pom, in the `pluginManagement` section.

sonar-plugin/sonar-javascript-plugin/pom.xml
```diff
@@ -558,6 +558,97 @@
...
+              </arguments>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>build-helper-maven-plugin</artifactId>
+        <version>3.6.1</version>
```
julien-carsique-sonarsource: Set the plugin version in a parent pom, in the pluginManagement section.

sonar-plugin/sonar-javascript-plugin/pom.xml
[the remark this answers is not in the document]
tomverin: Fixed in c2f6bf1 (merge script + bash invocation) and 1dbf205 (prepare-package/verify phase ordering, OS-aware cyclonedx-cli download including Windows .exe). The stale diff hunk still showed `<executable>cyclonedx</executable>`; current pom uses `bash tools/merge-cyclonedx-bom.sh`.

sonar-plugin/sonar-javascript-plugin/pom.xml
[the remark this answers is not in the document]
tomverin: Fixed in c2f6bf1 — `generate-npm-cyclonedx-bom` exec runs `npm run cyclonedx:npm` from the repo root during package before merge (now in verify). 1dbf205 also ensures Maven bom.json is generated in prepare-package first.

sonar-plugin/sonar-javascript-plugin/pom.xml
tomverin: Applied — cyclonedx-maven-plugin version and shared configuration moved to root `pluginManagement`; this module keeps only the `make-bom` execution.

sonar-plugin/sonar-javascript-plugin/pom.xml
tomverin: Applied — exec-maven-plugin version is inherited from root `pluginManagement` (inline version removed here).

sonar-plugin/sonar-javascript-plugin/pom.xml
tomverin: Applied — build-helper-maven-plugin version moved to root `pluginManagement`.

tools/merge-cyclonedx-bom.sh
```diff
@@ -0,0 +1,56 @@
...
+version="$7"
+
+if command -v cyclonedx >/dev/null 2>&1; then
+  cli="cyclonedx"
+else
+  arch="$(uname -m)"
+  case "$arch" in
+    x86_64|amd64) arch="x64" ;;
+    aarch64|arm64) arch="arm64" ;;
+    i686|i386) arch="x86" ;;
```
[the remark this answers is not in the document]
tomverin: Applied — i686/i386 is now accepted only on Windows (where cyclonedx-win-x86.exe exists); 32-bit Linux/macOS fail fast.

vdiez:
## Full assessment

### Verdict

I would not treat this PR as ready to close IST-5039 / SONAR-30945. It is a useful producer-side building block, and its publication mechanism works, but it currently misses bundled npm packages and has no downstream consumer.

| Area | Assessment |
|---|---|
| Maven/Repox workflow fit | Good |
| npm payload completeness | Insufficient |
| End-to-end Jira resolution | Not solved |
| Supply-chain security | Acceptable after checksum hardening |
| Maintainability | Medium-low without follow-up changes |

### What works well

The publication path follows our established conventions:

- The merged BOM is attached as `json:cyclonedx` and listed in [`artifactsToPublish`](https://github.com/SonarSource/SonarJS/blob/7350dc8a2981646fe837f96b1f37e07981021482/pom.xml#L84).
- CI build `43600` signed and deployed `sonar-javascript-plugin-13.3.0.43600-cyclonedx.json` and its signature.
- Promotion moved the complete build from `sonarsource-public-qa` to `sonarsource-public-dev`.
- Linux deploy and Windows verification both pass.
- The generated CycloneDX 1.6 file validates successfully and contains Maven and `pkg:npm` components.
- This meets most of the internal SBOM guidance: generated during the build, signed, attached, pushed to Repox, and promoted.

The direct executable download is also substantially safer after the latest fixes: the version and platform-specific checksums are pinned, cached binaries are reverified, and the hashes match GitHub's official release digests.

### Blocking correctness gaps

#### 1. The npm BOM still omits shipped dependencies

The npm command in [`package.json`](https://github.com/SonarSource/SonarJS/blob/7350dc8a2981646fe837f96b1f37e07981021482/package.json#L47) uses `--omit dev`. However, two dev dependencies are imported into the production bridge bundle:

- `eslint-plugin-vue`
- `eslint-plugin-testing-library`

I compared the esbuild metafile inputs with the generated BOM:

- 347 unique npm packages were inputs to the shipped bundle.
- Five were absent from the BOM:
  - `eslint-plugin-vue`
  - `eslint-plugin-testing-library`
  - `boolbase`
  - `nth-check`
  - `xml-name-validator`

The last three are transitive dependencies of the omitted plugins. Therefore, the PR does not yet satisfy "all JS/npm dependencies represented."

At minimum, the two runtime plugins should move to `dependencies`, or the BOM generation should use a bundle-aware input rather than the root dependency classification.

#### 2. Nothing consumes this sidecar downstream

The merged [sonar-enterprise PR #17746](https://github.com/SonarSource/sonar-enterprise/pull/17746) combines the application and edition BOMs. It does not resolve or merge analyzer `json:cyclonedx` classifiers.

Sonar Enterprise still declares only the JavaScript `multi` JAR. Gradle transitivity can discover Maven dependencies, but it cannot automatically ingest the attached CycloneDX JSON. Consequently:

- Phase 1 restores the missing JavaScript analyzer and its Maven-visible dependencies.
- This PR publishes npm information.
- The final SQS BOM still receives none of that npm information.

A further `sonar-enterprise` change must resolve the matching SonarJS sidecar and include it in the hierarchical merge.

### Maintainability concerns

- The npm BOM is generated twice: once from [`bridge/pom.xml`](https://github.com/SonarSource/SonarJS/blob/7350dc8a2981646fe837f96b1f37e07981021482/sonar-plugin/bridge/pom.xml#L286), then again from [`sonar-javascript-plugin/pom.xml`](https://github.com/SonarSource/SonarJS/blob/7350dc8a2981646fe837f96b1f37e07981021482/sonar-plugin/sonar-javascript-plugin/pom.xml#L210). CI logs confirm both executions. The bridge execution is redundant.
- The unconditional `package` executions bypass `-Dskip-nodejs`. A package/install/deploy build with that flag still runs npm, Bash, curl, and the merger, contradicting the documented [`-Dskip-nodejs` workflow](https://github.com/SonarSource/SonarJS/blob/7350dc8a2981646fe837f96b1f37e07981021482/docs/BUILD.md#L221).
- Internal SBOM guidance says generation should be a separate task so it does not disrupt normal fail-fast builds. Here it is embedded directly in `package`.
- Each fresh Linux and Windows CI build downloads a roughly 72 MB executable. It is cached only under the workspace `target`, not the cross-run cache. In the latest CI run this added roughly 21 seconds on Linux and 28 seconds on Windows.
- It introduces an undocumented Bash/curl/SHA-tool prerequisite for local Windows builds.
- The CycloneDX npm tool adds 137 lockfile package entries.
- CI verifies that generation succeeds, but there is no assertion that required packages are present or that the downstream artifact contains `pkg:npm`.

The BOM is also manifest-conservative rather than artifact-exact. It includes clear build-only packages such as `patch-package` and `@typescript/native-preview`, while missing the two bundled dev dependencies above. That may be acceptable if explicitly documented, but it is not currently accurate to the shipped esbuild payload.

### Recommended path

Before merging:

1. Move the two bundled ESLint plugins into production dependencies, and review whether `patch-package` and `@typescript/native-preview` should instead be dev dependencies.
2. Remove the bridge-level duplicate BOM execution.
3. Put SBOM generation behind an explicit profile/property that CI and releases enable, while preserving `-Dskip-nodejs`.
4. Cache or centrally provision CycloneDX CLI rather than downloading 70+ MB in every build.
5. Add checks that:
   - the result validates;
   - it contains `pkg:npm`;
   - every npm package in the esbuild metafile appears in the BOM.
6. Add the downstream `sonar-enterprise` consumption and merge.
7. Verify the promoted release artifact—not only the PR/dev build—against the final SQS `lib/extensions` inventory.

Longer term, generating the npm component set from esbuild's metafile would give the best payload fidelity. Syft against the final ZIP is useful as an independent completeness check, but it cannot reliably reconstruct package identities from a minified JavaScript bundle, so it should not replace producer metadata.

Bottom line: merge only as an explicitly incomplete prerequisite after fixing the missing bundled packages and lifecycle issues. Do not close [IST-5039](https://sonarsource.atlassian.net/browse/IST-5039) or [SONAR-30945](https://sonarsource.atlassian.net/browse/SONAR-30945) until the sidecar is consumed by SQS and the final promoted Enterprise BOM is verified.


[IST-5039]: https://sonarsource.atlassian.net/browse/IST-5039?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

vdiez:
@tomverin @julien-carsique-sonarsource I'm taking it from here, should be merged by tomorrow

vdiez (approved):

PR #7569, opened 2026-07-16
JS-2003 Use shared secret exclusion patterns in hardcoded secrets rules

mstachniuk:
Part of 

----
## Summary by Gitar

- **Secret exclusion mechanism:**
    - Added `isExcludedSecretValue` helper to ignore common placeholder or fake secret patterns.
    - Introduced `tools/generate-secret-patterns.ts` to sync patterns from `@sonarsource/analyzer-commons-configurations`.
- **Rule enhancements:**
    - Integrated shared secret patterns into `S2068`, `S6418`, and `S6437` to reduce false positives.
    - Extended `S6437` to support variable resolution and validation of hardcoded strings.
- **Build and config updates:**
    - Added `generate-secret-patterns` script to build pipeline.
    - Updated `package.json` with new dependency `@sonarsource/analyzer-commons-configurations`.


mstachniuk:
gitar auto-apply:on

its/sources/custom/jsts/S6437.js
daniel-teuchert-sonarsource, line 1: As these tests are mainly about detecting the secrets in different contexts I suggest to use a random string and reuse it for all instances. Something that should not be detected by our fake secret filters.

packages/analysis/src/jsts/rules/S6418/cb.fixture.ts
```diff
@@ -69,3 +69,8 @@ function memberExpressionAssignment(req, res, next) {
   req.session['token'] = '1IfHMPanImzX8ZxC-Ud6+YhXiLwlXq$f_-3v~.='; // Noncompliant {{"token" detected here, make sure this is not a hard-coded secret.}}
   req.session.safe = '1IfHMPanImzX8ZxC-Ud6+YhXiLwlXq$f_-3v~.='; // Compliant - property name doesn't match secret words
 }
+
+function excludedButHighEntropyValues() {
+  const apiSecret = 'vK4n8QzR2mXtLpJ6wYbGdCafeBabeFh9rSk3aWq'; // Compliant - matches known-fake CAFEBABE pattern despite high entropy
+  const authToken = 'xK9process.env.wR3mQ8vL5nT2yH6dF1cS0uJ7bA4gEoIiPmZ'; // Compliant - references an environment variable despite high entropy
```
daniel-teuchert-sonarsource: This looks weird. I suggest to either remove it or change it to something else.

packages/analysis/src/jsts/rules/S6437/cb.fixture.ts
daniel-teuchert-sonarsource, line 1: Same as above. IMO there is no need to come up with many different secrets for testing here. One pattern could be enough to check that it is detected in different contexts.

vdiez (changes requested):
Thanks for working on this, Marcin. The overall approach of consuming the analyzer-commons configuration package at build time looks good to me, but I found one analysis crash that needs to be fixed before merging.

S6437 resolves identifiers through hardcodedVariables, which is a file-global map keyed only by identifier name. With aliasing and lexical shadowing, the map becomes cyclic and getHardcodedStringValue recurses until the analysis crashes. I ran the following reproducer against the current PR head (287caffb):

```js
import crypto from "node:crypto";

const outer = "zQ9mPfXtRk3wDbn";
const alias = outer;

{
  const outer = alias;
  crypto.createHmac("sha256", outer);
}
```

It throws:

```text
RangeError: Maximum call stack size exceeded
```

The map ends up resolving outer to alias and alias back to outer. Please use scope-aware identifier resolution (there are existing helpers such as getUniqueWriteUsageOrNode) and add this as a regression test. The same name-based map can also resolve a use to the wrong value after ordinary shadowing, so cycle detection alone would not be sufficient.

There is also a packaging blocker for external contributors: package-lock.json pins @sonarsource/analyzer-commons-configurations to a tarball under repox.jfrog.io. Anonymous access to that URL returns 401, and the package is not available on registry.npmjs.org yet. Once the analyzer-commons package is published publicly, please regenerate package-lock.json against npmjs so its resolved URL points to registry.npmjs.org. Otherwise npm ci on a public network will continue trying the private RepoX URL even after the package is published.

mstachniuk:
I addressed the call stack recursion problem and published finally the dependency in npm: https://www.npmjs.com/package/@sonarsource/analyzer-commons-configurations So it should be safe to merge

vdiez (changes requested):
package-lock.json still points to repox, can you change it? 

Also, merge the [ruling changes PR](https://github.com/SonarSource/SonarJS/pull/7623) if you agree with the changes.

Also, there is an issue in the QG, seems an easy fix

vdiez:
I double-checked the detection delta against `master` and the contents/semantics of the analyzer-commons configuration.

Overall, the observed ruling changes are coherent with the shared policy:

- **S2068:** 15 findings disappear and none are added. The removed findings are test placeholders (`areallygoodpassword`, `newpassword`, `wrongpassword`, `rightpassword`, `validpassword`, `mypassword`) or the handler name `onChangePasswordClick`. These suppressions make sense. Values not covered by the shared classifier, such as `Sl1m3rson`, still report.
- **S6418:** there is no ruling delta. It still reports high-entropy values unless the shared classifier excludes them.
- **S6437:** 23 of the 27 original custom ruling findings are suppressed because their values match analyzer-commons patterns; the four original `privateKeyPEM` cases remain. The 23 random-hex findings added to the fixture replace the suppressed cases, so these are new source lines rather than newly detected findings in unchanged code. There are no new real-project S6437 ruling findings.

The scope-aware S6437 resolver also looks correct independently of the filtering: it can resolve a uniquely assigned literal, avoids false positives from shadowed variables and multiple/dynamic writes, and the cycle guard prevents the previous `RangeError`.

The main caveat is a product-policy trade-off in analyzer-commons rather than a SonarJS implementation problem. Patterns such as `pass`, `secret`, `test`, `change`, and `admin` are unanchored, and every value of five characters or fewer is excluded. This can hide genuine secrets. I verified, for example, that S6418 reports a high-entropy token but suppresses the same token when `secret` is embedded in it. For S6437 this is particularly worth noting because the rule already has high-confidence sink context: literals such as `hardcodedSecret123`, `hardcodedPassword123`, `root`, and `key` no longer report.

I would keep SonarJS aligned with analyzer-commons rather than fork the policy locally, but this false-negative trade-off should be understood/validated by AppSec. I would also add a direct S6418 rule test with one excluded high-entropy value and one nearby value that must still report; currently only the helper and the other two rules directly exercise the integration.

I ran the targeted S2068, S6418, S6437, secret-exclusion, and AST helper tests: all 19 tests pass.

vdiez (approved):

PR #7575, opened 2026-07-17
JS-2101: Extend JavaScript test rules to bun:test and node:test

nathsou:
## Summary

- Recognize `bun:test` and `node:test` bindings in shared test-rule detection.
- Add Bun support to S8754, S8781, S8784, S8780, S5976, and S2699.
- Add Node `assert.rejects()` and `assert.doesNotReject()` support to S8780.

## Validation

- `npm run bridge:compile`
- Focused S2699, S5976, S8754, S8780, S8781, and S8784 suites

## Links

- Jira: https://sonarsource.atlassian.net/browse/JS-2101
- RSPEC PR: https://github.com/SonarSource/rspec/pull/7463

### DIFF
--- a/packages/analysis/src/jsts/rules/S8780/rule.ts
+++ b/packages/analysis/src/jsts/rules/S8780/rule.ts
@@ -38,6 +38,8 @@ const VITEST_MODULES = ['vitest'];
 const JASMINE_MODULES = ['jasmine'];
 const JASMINE_DEPENDENCIES = ['jasmine', 'jasmine-core', 'jasmine-node', 'karma-jasmine'];
 const PLAYWRIGHT_MODULES = ['@playwright/test'];
+const BUN_MODULES = ['bun:test'];
+const NODE_TEST_MODULES = ['node:test'];
 
 const TEST_FUNCTION_NAMES = new Set([
   'it',
@@ -97,6 +99,8 @@ type Frameworks = {
   vitest: boolean;
   jasmine: boolean;
   playwright: boolean;
+  bun: boolean;
+  nodeTest: boolean;
 };
 
 type FunctionArgument = estree.FunctionExpression | estree.ArrowFunctionExpression;
@@ -110,6 +114,8 @@ export const rule: Rule.RuleModule = {
       vitest: importsOrDependsOnModule(context, VITEST_MODULES, VITEST_MODULES),
       jasmine: importsOrDependsOnModule(context, JASMINE_MODULES, JASMINE_DEPENDENCIES),
       playwright: importsOrDependsOnModule(context, PLAYWRIGHT_MODULES, PLAYWRIGHT_MODULES),
+      bun: importsOrDependsOnModule(context, BUN_MODULES, BUN_MODULES),
+      nodeTest: importsOrDependsOnModule(context, NODE_TEST_MODULES, NODE_TEST_MODULES),
     };
 
     if (!Object.values(frameworks).some(Boolean)) {
@@ -243,6 +249,7 @@ function getAsyncAssertionNode(
   return (
     getJestOrVitestAsyncNode(context, unwrapped, frameworks) ??
     getJasmineAsyncNode(context, unwrapped, frameworks) ??
+    getNodeAssertAsyncNode(context, unwrapped, frameworks) ??
     getPlaywrightAsyncNode(context, unwrapped, frameworks)
   );
 }
@@ -252,7 +259,7 @@ function getJestOrVitestAsyncNode(
   call: estree.CallExpression,
   frameworks: Frameworks,
 ): estree.Node | null {
-  if (!frameworks.jest && !frameworks.vitest) {
+  if (!frameworks.jest && !frameworks.vitest && !frameworks.bun) {
     return null;
   }
 
@@ -270,6 +277,24 @@ function getJestOrVitestAsyncNode(
   }
 
   return asyncSegment.node;
+}
+
+function getNodeAssertAsyncNode(
+  context: Rule.RuleContext,
+  call: estree.CallExpression,
+  frameworks: Frameworks,
+): estree.Node | null {
+  if (!frameworks.nodeTest) {
+    return null;
+  }
+
+  const fqn = getFullyQualifiedName(context, call.callee);
+  return fqn === 'assert.rejects' ||
+    fqn === 'assert.doesNotReject' ||
+    fqn === 'assert.strict.rejects' ||
+    fqn === 'assert.strict.doesNotReject'
+    ? call.callee
+    : null;
 }
 
 function getJasmineAsyncNode(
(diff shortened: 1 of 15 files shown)


packages/analysis/src/jsts/rules/S8780/rule.ts
```diff
@@ -173,17 +191,42 @@ function isFunctionArgument(node: estree.Node | estree.SpreadElement): node is F
...
+ * Jest/Jasmine/Vitest and Bun pass a `done` callback whenever the test function declares a
+ * parameter (arity-based, regardless of its name). Mixing that with an awaited/returned assertion
+ * makes the runtime throw ("cannot both take a 'done' callback and return something"), so the fix
+ * this rule suggests doesn't apply to such tests. Node's first callback parameter is a
+ * `TestContext`, not a `done` callback, so node:test calls are excluded from this heuristic.
+ * Playwright's fixtures parameter is always a destructuring pattern, never a plain identifier, so
+ * it isn't mistaken for a done callback.
  */
-function usesDoneCallback(callback: estree.Function): boolean {
+function usesDoneCallback(
```
guillemsarda, line 203: **Edge Case: node:test's genuine two-parameter done-callback is now flagged as a missing await**

📄 `packages/analysis/src/jsts/rules/S8780/rule.ts:202-213`

`usesDoneCallback` now unconditionally treats *any* node:test call as not using a done callback:

```ts
function usesDoneCallback(
  context: Rule.RuleContext,
  call: estree.CallExpression,
  callback: estree.Function,
  frameworks: Frameworks,
): boolean {
  if (frameworks.nodeTest && isNodeTestCall(context, call)) {
    return false;
  }
  const [firstParam] = callback.params;
  return firstParam?.type === 'Identifier';
}
```

This correctly fixes the single-param case (`(t) => {...}`, where `t` is a `TestContext`, not a done callback). But node:test also supports a genuine **two-parameter** done-callback form — `test('name', (t, done) => { ...; done(); })` — where the *second* param is the real done callback. Blanket-excluding node:test from the heuristic makes this a false positive:

```js
import test from 'node:test';
import assert from 'node:assert/strict';

test('uses explicit done alongside TestContext', (t, done) => {
  assert.rejects(failingJob()); // flagged: "Await or return this async assertion."
  done();
});
```

This is valid, idiomatic node:test code — awaiting `assert.rejects` here would be wrong, since completion is signaled via `done()`, not by returning/awaiting inside the test body.

Reproduced with a local rule-tester case (currently uncovered by the PR's fixtures — all node:test test cases use a single-param callback).

**Suggestion:** make the node:test exclusion arity-aware instead of blanket, e.g.:

```ts
if (frameworks.nodeTest && isNodeTestCall(context, call)) {
  return callback.params.length >= 2;
}
```

so a single `TestContext` param is correctly excluded (the fix this PR intended) while a real `(t, done)` done-callback is still recognized.

packages/analysis/src/jsts/rules/helpers/mocha-style-test-frameworks.ts
guillemsarda: **Design Smell: `getMochaConstructName` and `getMochaConstructAndModifiers` can disagree on the same input**

📄 `packages/analysis/src/jsts/rules/helpers/mocha-style-test-frameworks.ts:130-168`
📄 `packages/analysis/src/jsts/rules/S8754/rule.ts:262-274`

This PR adds default-`node:test`-import handling (`import test from 'node:test'; test.describe(...)` / `test.it(...)`) inside `getMochaConstructAndModifiers`, which shifts the first modifier into the construct name:

```ts
function getMochaConstructAndModifiers(
  context: Rule.RuleContext,
  calleeParts: { base: estree.Identifier; modifiers: string[] },
): { constructName: string | undefined; modifiers: string[] } {
  const constructName = getMochaConstructName(context, calleeParts.base);
  if (
    constructName === 'test' &&
    isNodeTestDefaultBinding(context, calleeParts.base) &&
    calleeParts.modifiers.length > 0
  ) {
    const [nodeTestConstructName, ...modifiers] = calleeParts.modifiers;
    ...
  }
  return { constructName, modifiers: calleeParts.modifiers };
}
```

But the plain `getMochaConstructName` — which is still exported and still called directly elsewhere — doesn't know this trick. `S8754/rule.ts`'s `isNonConcreteMochaSuite` calls it directly, bypassing the shift entirely:

```ts
function isNonConcreteMochaSuite(context: Rule.RuleContext, node: estree.Node): boolean {
  const calleeParts = getMochaCalleeParts(node);
  if (calleeParts === undefined) {
    return false;
  }

  const constructName = getMochaConstructName(context, calleeParts.base);
  if (constructName === undefined || !SUITE_FUNCTION_NAMES.includes(constructName)) {
    return false;
  }
  ...
}
```

For `test.describe.skip(...)` with a default `node:test` import, `getMochaConstructName(context, 'test')` returns `'test'` (not `'describe'`), so this function can never recognize a default-bound node:test suite — the two functions give different answers to "what construct is this?" for the identical input.

I probed this directly (a skipped default-import suite alone, and one nested inside an active outer suite) and in both cases S8754's final verdict was still correct — some other gating in the rule (`isInConcreteCollectionCallback`) happens to absorb the gap, so I can't point to a currently-reproducible false positive/negative from this specific path. But the correctness there looks accidental rather than intentional, and it's fragile: any other caller that reaches for `getMochaConstructName` directly (instead of going through `isMochaTestConstruct` / `getMochaConstructAndModifiers`) will silently misclassify default-import `node:test` namespaced calls (`test.describe`, `test.it`, etc.).

**Suggestion:** fold the default-binding modifier-shift into `getMochaConstructName` itself (or make it the only supported entry point, with `getMochaConstructAndModifiers` just adjusting `modifiers`), so there's a single source of truth for "what construct is this call" instead of a basic version and a fixed version that different callers can pick between.

packages/analysis/src/jsts/rules/S8780/rule.ts
guillemsarda: https://nodejs.org/api/test.html#testname-options-fn
https://nodejs.org/api/test.html#test-runner

guillemsarda (changes requested):

packages/analysis/src/jsts/rules/helpers/mocha-style-test-frameworks.ts
nathsou: Fixed in `1b70b1200`. The base resolver is now private, `getMochaConstructAndModifiers` is the shared exported entry point, and S8754's `isNonConcreteMochaSuite` uses it too. Callers now share one construct-plus-modifier interpretation for default node:test namespace calls.

packages/analysis/src/jsts/rules/S8780/rule.ts
nathsou: Fixed in `1b70b1200`. `usesDoneCallback` now returns `callback.params.length >= 2` for recognized node:test calls, preserving single-parameter TestContext behavior while excluding genuine two-parameter done callbacks. Added a valid S8780 regression case.

guillemsarda (approved):
Image

PR #7590, opened 2026-07-20
JS-2115: Implement S8998: Parameterized tests should not have empty datasets

nathsou:
## Summary
- add the original S8998 Jest, Vitest, and Bun empty parameterized-dataset rule
- recognize Jest globals, Bun test.each and describe.each, Vitest suites, runnable modifiers, and registration-time empty local tables
- pin generated rule data to the matching RSPEC change

## Links
- Jira: https://sonarsource.atlassian.net/browse/JS-2115
- RSPEC: https://github.com/SonarSource/rspec/pull/7465

## Validation
- uc check


packages/analysis/src/jsts/rules/helpers/mocha-style-test-frameworks.ts
```diff
@@ -135,13 +141,14 @@ export function getMochaConstructName(
 }
 
 function getMochaConstructNameFromFqn(fqn: string | null): string | undefined {
-  const parts = fqn?.split('.');
-  if (parts?.length !== 2) {
+  if (fqn === null) {
```
guillemsarda: ## `@jest/globals` recognition silently expands to 4 other shipped rules

Adding `'@jest/globals'` / `'@jest.globals'` to the shared `SUPPORTED_TEST_FRAMEWORKS`/`MOCHA_STYLE_TEST_FRAMEWORKS` in `mocha-style-test-frameworks.ts`, and rewriting `getMochaConstructNameFromFqn` to prefix-match instead of requiring an exact 2-segment FQN split, changes behavior for every rule that consumes `getMochaConstructName`/`isMochaTestConstruct` — not just S8998.

Verified empirically: on `master`, S8754 (`no-duplicate-test-title`) does **not** flag two identically-titled tests imported from `@jest/globals`. At this PR's head commit, the same input **is** flagged. S8754, S8781, S8784, and S8968 all consume these shared helpers and are affected the same way. Existing test suites for those 4 rules still pass (no regressions), but none of them gained new tests covering `@jest/globals`, so this expansion of already-shipped rule behavior is unvalidated.

This is the same category of risk already flagged in this file for `bun:test` scope creep — which was fixed by keeping Bun detection local to S8998. The equivalent fix wasn't applied to the `@jest/globals` addition, which was left in the shared file.

**Suggestion:** either scope the `@jest/globals` FQN-matching change to S8998 only (mirroring the Bun treatment), or, if this is an intentional shared-helper bug fix, call it out explicitly and add regression tests to S8754/S8781/S8784/S8968 to lock in the new behavior.

packages/analysis/src/jsts/rules/S8781/unit.test.ts
```diff
@@ -76,6 +76,13 @@ test.for([1, 2])(' ', () => {});
         },
         {
           code: `
+import { test } from 'bun:test';
```
guillemsarda: Is this regression added on purpose?

packages/analysis/src/jsts/rules/S8998/unit.test.ts
```diff
@@ -0,0 +1,143 @@
...
+
+describe('S8998', () => {
+  it('reports runnable Jest and Vitest parameterized declarations with empty datasets', () => {
+    const ruleTester = new NoTypeCheckingRuleTester();
+    const jestFixture = path.join(import.meta.dirname, 'fixtures', 'jest', 'test.js');
+
+    ruleTester.run('Parameterized tests should not have empty datasets', rule, {
+      valid: [
+        {
+          code: "import { test } from '@jest/globals';\ntest.each([[1]])('case %i', value => expect(value).toBe(1));",
```
guillemsarda: Is there a reason why you are using strings with scape values? Usually we use templates which help readability

packages/analysis/src/jsts/rules/S8998/unit.test.ts
```diff
@@ -0,0 +1,143 @@
...
+        },
+        {
+          code: "import { suite } from 'vitest';\nconst cases: number[] = [[1]];\nsuite.each(cases)('case %i', () => {});",
```
guillemsarda: I think it may be worth adding as an explicitly known valid case this one: 
```js
test.each([[], []])('case %i', () => {}) 
```

guillemsarda (changes requested):
Left some comments, the main one is to add coverage to the affected rules

packages/analysis/src/jsts/rules/S8781/unit.test.ts
nathsou: Yes, this valid case is intentional as a regression guard. Bun detection is scoped to S8998, so S8781 should continue treating bun:test as unsupported and should not report empty titles. The test ensures future shared-helper changes do not expand S8781 behavior; no product behavior change is intended.

packages/analysis/src/jsts/rules/helpers/mocha-style-test-frameworks.ts
nathsou: Addressed in commit 980e4b768. The shared helper is restored to its previous framework and FQN behavior, while S8998 handles the @jest/globals FQN locally. This keeps S8754, S8781, S8784, and S8968 behavior unchanged. Focused tests pass for S8998, S8781, S8754, S8784, S8785, S8960, S8968, and S5976.

packages/analysis/src/jsts/rules/S8998/unit.test.ts
nathsou: Addressed in commit 980e4b768 by converting the multiline S8998 code fixtures from escaped newline strings to template literals. The focused rule tests and formatter checks pass.

packages/analysis/src/jsts/rules/S8998/unit.test.ts
nathsou: Addressed in commit 980e4b768 by adding `test.each([[], []])` as an explicit valid S8998 case. The focused S8998 tests pass.

guillemsarda (approved):
Image

PR #7591, opened 2026-07-20
JS-2117: Implement S9072: synchronous exception assertions should receive synchronous callbacks

nathsou:
## Summary

- Implement S9072 for type-checked Jest, Vitest, and `bun:test` assertions.
- Report concrete non-callable values passed to synchronous `toThrow` and `toThrowError` assertions.
- Report explicit async callbacks passed to synchronous exception assertions.
- Provide conservative suggestions for wrapping direct operations and converting async test assertions to `await expect(...).rejects.toThrow(...)` or `toThrowError(...)`.
- Skip analysis when required type information is unavailable or the type is uncertain.

## Verification

- Focused S9072 tests pass.
- SonarJS plugin TypeScript compilation passes.
- `uc check --base master`: merge conflicts and unit tests pass; the check remains red only for the pre-existing missing `eslint-plugin-testing-library` dependency and CycloneDX Knip complaint. Ruling tests are skipped because S9072 has no baseline.

Jira: https://sonarsource.atlassian.net/browse/JS-2117
RSPEC: https://github.com/SonarSource/rspec/pull/7466


its/ruling/src/test/expected/vitest/typescript-S9072.json
```diff
@@ -0,0 +1,5 @@
+{
+"vitest:test/core/test/jest-expect.test.ts": [
```
guillemsarda, line 2: Just leaving this present in the PR: this example is a FP because it is a meta testing of the `toThrow` from vitest. A very niche case not worth addressing.

packages/analysis/src/jsts/rules/S9072/rule.ts
```diff
@@ -0,0 +1,463 @@
...
+const TEST_CONSTRUCTS = new Set([
+  ...TEST_FUNCTION_NAMES,
+  'before',
+  'after',
+  'beforeEach',
+  'afterEach',
+  'beforeAll',
+  'afterAll',
+]);
+const TEST_MODIFIERS = new Set(['only', 'concurrent', 'failing', 'sequential']);
```
guillemsarda: **Minor: quick-fix suggestion silently dropped for `test.skip`/`test.todo`**

`packages/analysis/src/jsts/rules/S9072/rule.ts:61`

```js
const TEST_MODIFIERS = new Set(['only', 'concurrent', 'failing', 'sequential']);
```

`isRecognizedTestOrHook` (line 445) requires every modifier segment after the base construct name to be in `TEST_MODIFIERS`. Since skip and todo aren't included, an async-callback misuse inside `test.skip(...)`/`test.todo(...)` (or it.skip/it.todo) still gets the asyncCallback diagnostic — that part is unaffected — but getAsyncCallbackSuggestion returns undefined because getEnclosingAsyncTestCallback bails out via `!isRecognizedTestOrHook(...)`, so no quick fix is offered for an otherwise-fixable case.

Suggest adding 'skip' and 'todo' to TEST_MODIFIERS so the suggestion isn't dropped for these common modifiers.

guillemsarda (approved):
Image

PR #7592, opened 2026-07-20
JS-2088 Exclude vue 2 from rule S8961

guillemsarda:
----
## Summary by Gitar

- **Rule logic:**
    - Added a decorator to `S8961` to automatically disable it for Vue 2 projects.
- **Dependency helpers:**
    - Added `getVueVersion` to `dependencies.ts` to retrieve and parse Vue versions from `package.json`.
- **Testing:**
    - Added comprehensive unit tests for both Vue 2 and Vue 3 environments to verify correct reporting behavior.
- **Documentation:**
    - Updated `README.md` to reclassify `S8961` as a decorated rule.


nathsou (approved):
seal-of-approval-2


[P2] Check the dependency range instead of its minimum version

`getVueVersion` currently returns `minVersion(...)`, so a major-version check would classify ranges such as `>=2.7.0` and `^2.7.0 || ^3.0.0` as Vue 2 even when Vue 3 is installed. Please preserve the raw dependency range and use `semver.intersects(range, ">=3.0.0")`; unknown ranges should keep reporting.

PR #7593, opened 2026-07-20
JS-2095 FP in S9019: Enable ref-to-ref reassignment

guillemsarda:
----
## Summary by Gitar

- **Bug fixes:**
    - Suppress false positives in `S9019` when reassigning a ref to another ref-creating function call.
    - Added `isRefToRefReassignment` to detect and ignore legitimate `ref` or `computed` reassignments.

### DIFF
--- a/packages/analysis/src/jsts/rules/S9019/decorator.ts
+++ b/packages/analysis/src/jsts/rules/S9019/decorator.ts
@@ -17,9 +17,38 @@
 // https://sonarsource.github.io/rspec/#/rspec/S9019/javascript
 
 import type { Rule } from 'eslint';
+import type estree from 'estree';
 import { generateMeta } from '../helpers/generate-meta.js';
 import { interceptReport } from '../helpers/decorators/interceptor.js';
 import * as meta from './generated-meta.js';
+
+// eslint-plugin-vue's ref-object-references.js tracks all ref-creating calls for a variable in a
+// shared map keyed by reference node, overwriting entries per call it processes. When a ref-typed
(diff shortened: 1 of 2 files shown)


nathsou (approved):
Image

packages/analysis/src/jsts/rules/S9019/decorator.ts
```diff
@@ -17,10 +17,39 @@
...
+
+function isRefToRefReassignment(node: estree.Node & Rule.NodeParentExtension): boolean {
+  const { parent } = node;
+  return (
+    parent.type === 'AssignmentExpression' &&
+    parent.operator === '=' &&
+    parent.left === node &&
+    parent.right.type === 'CallExpression' &&
+    parent.right.callee.type === 'Identifier' &&
+    REF_CREATING_CALLEES.has(parent.right.callee.name)
```
nathsou: [P2] Recognize aliased Vue imports

eslint-plugin-vue resolves imported aliases, but this predicate only compares the callee’s local spelling with the export names. Therefore `import { ref as makeRef, computed as makeComputed } from "vue"; ... x = makeComputed(...)` still produces the same false positive. Use the upstream report metadata (for example, `data.method`) or resolve the import binding, and add an alias regression test.

guillemsarda: Yes, I left it out on purpose to keep the update simpler. I'll add a comment in the Jira ticket to have it in mind if we want to improve the rule.

UPDATE: I've reconsidered it and will add the feature.

PR #7611, opened 2026-07-21
JS-2127 Demote S6440 from Sonar way

erwan-leforestier-sonarsource:
## Summary
This removes S6440 from the built-in Sonar way profile while keeping the rule available outside that default profile.

RSPEC PR: https://github.com/SonarSource/rspec/pull/7537

## Changes
- refresh S6440 metadata so it is no longer part of Sonar way
- regenerate the Sonar way aggregate profile without S6440
- add targeted tests for Sonar way activation and generated S6440 recommendation metadata


victor-peixoto-sonarsource (approved):
LGTM

PR #7622, opened 2026-07-23
JS-629: Extend S2970 coverage to modern test frameworks

nathsou:
## Summary
Extends rule S2970 (`no-incomplete-assertions`) beyond Chai to cover modern JavaScript/TypeScript test runners and assertion libraries:
- **Jest**, **Vitest**, **Bun** (`bun:test`), **@testing-library/jest-dom**, **@playwright/test**, and **Node.js** (`node:assert`).
- Support for matcher aliases (`toBeCalled`, `toThrowError`, `toReturn`, etc.) and dangling modifiers (`resolves`, `soft`, `poll`, `not`).
- Unwraps `AwaitExpression` so async modifier assertions (e.g. `await expect(promise).resolves;`) are properly detected.

## Jira
https://sonarsource.atlassian.net/browse/JS-629

## Related RSPEC PR
https://github.com/SonarSource/rspec/pull/7601


packages/analysis/src/jsts/rules/S2970/rule.ts
```diff
@@ -23,6 +23,7 @@ import { isIdentifier, isNumberLiteral } from '../helpers/ast.js';
 import * as meta from './generated-meta.js';
 
 const assertionFunctions = [
```
guillemsarda: `assertionFunctions` and `gettersOrModifiers` are now sizeable lookup tables (~130 lines) mixed into `rule.ts` alongside the traversal/reporting logic. This codebase already has a convention for splitting that out into a same-folder file when a rule's lookup data grows — e.g. `S8907/catalog.ts` (a 147-line table of replacements, imported by `rule.ts` via `from './catalog.js'`) and `S8785/async-analysis.ts`.

Suggest extracting `assertionFunctions` and `gettersOrModifiers` into a sibling file (e.g. `matchers.ts`) in `S2970/`, imported the same way:

```ts
// S2970/matchers.ts
export const assertionFunctions = [ /* ... */ ];
export const gettersOrModifiers = [ /* ... */ ];
```

```ts
// S2970/rule.ts
import { assertionFunctions, gettersOrModifiers } from './matchers.js';
```

Keeps `rule.ts` focused on the AST logic and makes the per-library vocabulary (Chai / Jest / Playwright / jest-dom / Node assert) easier to scan and extend on its own, without implying any cross-rule reuse.

packages/analysis/src/jsts/rules/S2970/cb.fixture.js
```diff
@@ -144,6 +144,58 @@ describe("incomplete assertions", function() {
...
+    assert.rejects;  // Noncompliant {{Call this 'rejects' assertion.}}
+  });
+
+  it("dangling modifiers", async function() {
+    expect(value).not;  // Noncompliant {{Complete this assertion; 'not' doesn't assert anything by itself.}}
+    expect(promise).resolves;  // Noncompliant {{Complete this assertion; 'resolves' doesn't assert anything by itself.}}
+    await expect(promise).resolves;  // Noncompliant {{Complete this assertion; 'resolves' doesn't assert anything by itself.}}
+    expect(promise).rejects;  // Noncompliant {{Complete this assertion; 'rejects' doesn't assert anything by itself.}}
+    await expect(promise).rejects;  // Noncompliant {{Complete this assertion; 'rejects' doesn't assert anything by itself.}}
+    expect(locator).soft;  // Noncompliant {{Complete this assertion; 'soft' doesn't assert anything by itself.}}
```
guillemsarda: `expect.soft`/`expect.poll` are static methods on `expect` itself, not properties chained off `expect(value)`'s result ([Playwright docs](https://playwright.dev/docs/test-assertions#soft-assertions)):

```js
await expect.soft(locator).toHaveText('Success');   // real syntax
await expect.poll(async () => ..., {}).toBe(200);   // real syntax
```

The fixture test added for this (`cb.fixture.js:42`) checks a shape that doesn't occur in real code:

```js
expect(locator).soft;  // Noncompliant {{Complete this assertion; 'soft' doesn't assert anything by itself.}}
```

Nobody writes `expect(locator).soft` — `soft`/`poll` are accessed on `expect` before it's called, not after. This test passes, but it isn't exercising real Playwright usage.

The actual realistic mistake — calling `expect.soft(...)`/`expect.poll(...)` correctly and forgetting the trailing matcher — isn't caught at all:

```js
expect.soft(locator);   // called, but no .toBeVisible() etc. chained — genuinely incomplete, not flagged
```

Tracing why: `expr` here is a `CallExpression` (`expect.soft(locator)`), not a `MemberExpression`, so the dangling-property branch never triggers. `isExpectCall(expr)` also requires `node.callee` to be a bare `Identifier` named `expect`, but here `node.callee` is `expect.soft` (a `MemberExpression`), so that check fails too. Nothing reports.

Suggest either:
- adding detection for `expect.soft(...)`/`expect.poll(...)` calls missing a chained matcher, or
- if that's out of scope for this PR, at least fixing the fixture to test the real syntax shape (`expect.soft(locator)`/`expect.soft;`) so the test reflects actual Playwright usage instead of a non-existent pattern.

guillemsarda (changes requested):
I left a comment regarding valid Playwright syntax worth addressing before merging

packages/analysis/src/jsts/rules/S2970/rule.ts
nathsou: Done in 6bae305c2 — extracted `assertionFunctions`, `gettersOrModifiers`, and `expectStaticMethods` into `S2970/matchers.ts`, imported from `rule.ts` the same way as `S8907/catalog.ts`.

packages/analysis/src/jsts/rules/S2970/cb.fixture.js
nathsou: Good catch — confirmed against the Playwright docs that `soft`/`poll` are static methods on `expect`.

Addressed in 6bae305c2:
- Added detection for incomplete `expect.soft(...)`, `expect.poll(...)`, and nested `expect.soft.poll(...)` (including the bare `expect.soft` / `expect.poll` forms).
- Replaced the fictitious `expect(locator).soft` fixture with real Playwright shapes and compliant counterparts.

guillemsarda (approved):
Found on internet
Image

PR #7625, opened 2026-07-23
JS-2140: Extend S8959 to Testing Library debug utils

nathsou:
## Summary
Extends rule `S8959` (`UI test debug commands should not be committed to version control`) to detect Testing Library debug utilities:
- `screen.debug()`
- `screen.logTestingPlaygroundURL()`
- `prettyDOM()` / `logRoles()`

## Related Links
- Jira Ticket: https://sonarsource.atlassian.net/browse/JS-2140
- RSPEC PR: https://github.com/SonarSource/rspec/pull/7602

----
## Summary by Gitar

- **New rules:**
    - Added rule `S9078` to disallow duplicate test cases in parameterized tests with auto-fixing support

### DIFF
--- a/packages/analysis/src/jsts/rules/S8959/rule.ts
+++ b/packages/analysis/src/jsts/rules/S8959/rule.ts
@@ -21,9 +21,19 @@ import type estree from 'estree';
 import { hasParent, isIdentifier, isMethodCall } from '../helpers/ast.js';
 import { chainStartsWithCy } from '../helpers/cypress.js';
 import { generateMeta } from '../helpers/generate-meta.js';
+import { getFullyQualifiedName } from '../helpers/module.js';
 import { removeNodeWithLeadingWhitespaces } from '../helpers/quickfix.js';
 import { isTestRelatedFile } from '../helpers/test-file-pattern.js';
 import * as meta from './generated-meta.js';
+
+const TESTING_LIBRARY_MODULES = [
+  '@testing-library/dom',
+  '@testing-library/react',
+  '@testing-library/vue',
+  '@testing-library/angular',
+  '@testing-library/svelte',
+  '@testing-library/preact',
+];
 
 export const rule: Rule.RuleModule = {
   meta: generateMeta(meta, {
@@ -41,10 +51,10 @@ export const rule: Rule.RuleModule = {
       CallExpression(node: estree.Node) {
         const call = node as estree.CallExpression;
         if (isMethodCall(call)) {
-          if (isUiTestDebugCommand(call.callee)) {
+          if (isUiTestDebugCommand(context, call)) {
             reportDebugCommand(context, call, call.callee.property);
           }
-        } else if (isIdentifier(call.callee, 'prettyDOM', 'logRoles')) {
+        } else if (isTestingLibraryDebugCommand(context, call)) {
           reportDebugCommand(context, call, call.callee);
         }
       },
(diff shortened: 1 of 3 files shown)


packages/analysis/src/jsts/rules/S8959/unit.test.ts
```diff
@@ -83,6 +84,39 @@ describe('S8959', () => {
           `,
           filename: 'tests/debugger.spec.ts',
         },
+        {
```
guillemsarda, line 87: **`render()`'s own return value also exposes a documented `debug` method, separate from `screen.debug()`, and it isn't detected by this rule:**

```js
const { debug } = render(<Component />);
debug();                          // not flagged

render(<Component />).debug();    // not flagged
```

This is an officially documented API ([testing-library.com](https://testing-library.com/docs/react-testing-library/api/#debug) and predates `screen.debug()` as the idiomatic way to debug, so it's likely still common in real test suites.

Root cause: `isTestingLibraryDebugCommand` only matches `screen.debug`, `screen.logTestingPlaygroundURL`, `prettyDOM`, and `logRoles` (`TESTING_LIBRARY_MODULES` check). For `const { debug } = render(...)`, `debug`'s fqn resolution fails because its binding traces back to a call expression (`render(...)`), not an import/require — `getFullyQualifiedName` has no type information to know `render`'s return shape. For `render(...).debug()`, the resolved fqn is `<module>.render.debug`, which isn't in the matched list.

Worth deciding whether to extend detection to cover destructured/chained `render().debug()`, or explicitly scope this rule to `screen`-based usage only and document that as a known limitation.

packages/analysis/src/jsts/rules/S8959/unit.test.ts
```diff
@@ -83,6 +84,39 @@ describe('S8959', () => {
...
+          code: `
+            test('Testing Library queries are valid', () => {
```
guillemsarda, line 89: The fully-qualified-name resolution is single-file and doesn't follow re-exports, so the official "custom test-utils wrapper" pattern from Testing Library's own setup docs defeats this detection entirely:

```js
// test-utils.tsx
export * from '@testing-library/react';
export { customRender as render };
```

```js
// some.test.tsx
import { screen } from '../test-utils';

screen.debug(); // not flagged
```

This is the pattern recommended in Testing Library's [Setup docs](https://testing-library.com/docs/react-testing-library/setup/#custom-render) for injecting providers into `render`, so it's likely common in real projects — probably more common than the direct `@testing-library/react` import case the current tests cover.

Root cause: `getFullyQualifiedName` resolves `screen`'s import to the literal module specifier in the current file (`'../test-utils'`), not to whatever that file re-exports from. Since `'../test-utils'` isn't in `TESTING_LIBRARY_MODULES`, the fqn never matches, regardless of how many re-export layers ultimately lead back to `@testing-library/react`.

Worth flagging as a known limitation (or scoping test coverage/documentation to explicitly call out direct-import-only support), since it's a silent miss — no false positive to notice, it just never fires for these codebases.

guillemsarda (commented):
I leave some comments worth clarifying
Image

packages/analysis/src/jsts/rules/S8959/rule.ts
```diff
@@ -21,10 +21,20 @@ import type estree from 'estree';
...
 import * as meta from './generated-meta.js';
 
+const TESTING_LIBRARY_MODULES = [
+  '@testing-library/dom',
+  '@testing-library/react',
+  '@testing-library/vue',
+  '@testing-library/angular',
+  '@testing-library/svelte',
+  '@testing-library/preact',
+];
```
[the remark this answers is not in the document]
nathsou: Addressed in 353df1cb4. S8959 now matches Testing Library imports from subpath entrypoints such as @testing-library/react/pure, with a regression test covering screen.debug().

packages/analysis/src/jsts/rules/S8959/unit.test.ts
nathsou: Addressed in 353df1cb4. S8959 now includes render.debug in its Testing Library command set, and regression tests cover both render(...).debug() and const { debug } = render(...).

packages/analysis/src/jsts/rules/S8959/unit.test.ts
nathsou: Addressed as an explicit limitation in 353df1cb4. getFullyQualifiedName only resolves imports in the current file, so custom test-utils re-exports are not followed; the rule now documents this and includes a valid regression case to prevent treating wrapper imports as direct Testing Library imports.

guillemsarda (approved):
```
    /\_/\
   ( o.o )
    > ^ <

PR #7633, opened 2026-07-24
JS-2148 Collect package import telemetry

nathsou:
## Summary

- collect literal JS/TS package imports from static imports, exports, `require`, dynamic imports, and TypeScript import-equals declarations
- limit telemetry to a checked-in 655-name public allowlist, direct manifest dependencies, npm specifiers, and public Node/Bun built-ins
- aggregate unique importing-file counts and emit bounded `javascript.telemetry.import.*` properties through the typed gRPC telemetry pipeline

## Privacy

Unknown, private, local, relative, absolute, URL-based, and undeclared path-aliased module names are discarded locally and never enter telemetry.

## Test plan

- `npm run bbf`
- focused package import, telemetry collector, module extraction, and gRPC conversion tests
- `mvn -pl sonar-plugin/sonar-javascript-plugin -am -Dtest=PluginTelemetryTest -Dsurefire.failIfNoSpecifiedTests=false test`
- `uc check`

## Sample SQL queries

The raw telemetry is queryable as `key`/`value` rows in `sonarsource.sonar.fct_sq_analysis_metric`. Import values are locally aggregated counts of unique files importing each package, not raw import-event counts.

```sql
-- Inspect all import telemetry properties with trimmed keys
SELECT 
    REPLACE(key, 'javascript.telemetry.import.', '') AS package_name,
    value, 
    COUNT(*) AS row_count
FROM sonarsource.sonar.fct_sq_analysis_metric
WHERE key LIKE 'javascript.telemetry.import.%'
GROUP BY 1, 2
ORDER BY row_count DESC;
```

```sql
-- Inspect all import telemetry properties
SELECT key, value, COUNT(*) AS row_count
FROM sonarsource.sonar.fct_sq_analysis_metric
WHERE key LIKE 'javascript.telemetry.import.%'
GROUP BY key, value
ORDER BY key, value;
```

```sql
-- Sum reported importing-file counts by package key
SELECT
  key,
  COUNT(*) AS row_count,
  SUM(CAST(value AS BIGINT)) AS importing_file_count_sum
FROM sonarsource.sonar.fct_sq_analysis_metric
WHERE key LIKE 'javascript.telemetry.import.%'
GROUP BY key
ORDER BY importing_file_count_sum DESC;
```

```sql
-- Distribution of per-analysis importing-file counts
SELECT
  key,
  CAST(value AS BIGINT) AS importing_file_count,
  COUNT(*) AS row_count
FROM sonarsource.sonar.fct_sq_analysis_metric
WHERE key LIKE 'javascript.telemetry.import.%'
GROUP BY key, CAST(value AS BIGINT)
ORDER BY key, importing_file_count;
```

## Links

- Jira: https://sonarsource.atlassian.net/browse/JS-2148


packages/analysis/src/jsts/rules/helpers/module.ts
```diff
@@ -66,25 +67,51 @@ function computeCurrentFileImports(sourceCode: SourceCode): void {
...
+    if (moduleName !== undefined) {
+      CURRENT_FILE_IMPORTS.imports.add(moduleName);
+    }
+    for (const visitorKey of sourceCode.visitorKeys[node.type] ?? []) {
+      const child = (node as unknown as Record<string, unknown>)[visitorKey];
+      if (Array.isArray(child)) {
+        pending.push(...child.filter(isNode));
+      } else if (isNode(child)) {
+        pending.push(child);
+      }
```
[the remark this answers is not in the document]
nathsou: Fixed in commit 3da6b828e. AST child arrays are now appended item-by-item instead of being spread into `pending.push`, preventing argument-limit failures on very large generated files. Added a regression test using a 200,000-element array.

packages/analysis/src/jsts/analysis/analyzer.ts
```diff
@@ -86,6 +86,10 @@ export async function analyzeJSTS(input: JsTsAnalysisInput): Promise<JsTsAnalysi
     debug('Clearing dependencies cache');
     clearDependenciesCache();
   }
+  getOptionalProjectAnalysisTelemetryCollector()?.recordPackageImports(
+    filePath,
+    Linter.collectPackageImports(parseResult.sourceCode, filePath),
+  );
```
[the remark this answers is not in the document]
nathsou: No change is needed here. JavaScript optional chaining short-circuits the optional method call, including its arguments, when `getOptionalProjectAnalysisTelemetryCollector()` returns `undefined`. Therefore `Linter.collectPackageImports(...)` is not invoked when telemetry is disabled; a local runtime check confirmed the argument expression is not evaluated.

packages/analysis/src/jsts/rules/helpers/module.ts
```diff
@@ -87,12 +84,117 @@ function computeCurrentFileImports(sourceCode: SourceCode): void {
...
+  if ((node as TSESTree.Node).type === 'TSImportEqualsDeclaration') {
+    const moduleReference = (node as unknown as TSESTree.TSImportEqualsDeclaration).moduleReference;
+    if (
+      moduleReference.type === 'TSExternalModuleReference' &&
+      moduleReference.expression.type === 'Literal' &&
+      typeof moduleReference.expression.value === 'string'
+    ) {
+      return moduleReference.expression.value;
+    }
+  }
```
guillemsarda, line 163: **`TSImportType` isn't recognized as a module reference**

`getReferencedModuleName` (in `module.ts`) checks for `ImportDeclaration`/`Export*Declaration` (via `.source`), `TSImportEqualsDeclaration`, `require(...)`, and dynamic `import(...)`, but doesn't handle inline type-only imports like:

```ts
type Props = import('react').ComponentProps<'div'>;
```
This is a TSImportType node (module specifier in argument, a TSLiteralType wrapping a string Literal), and it's a fairly common idiom in TS codebases for referencing a package's types without a full import statement. Since it's currently skipped, packages referenced only this way won't be counted in the telemetry.

Worth adding a branch for TSImportType alongside the existing cases?

packages/analysis/src/telemetry.ts
```diff
@@ -134,6 +135,7 @@ export class ProjectAnalysisTelemetryCollector {
   private esmFileCount = 0;
   private cjsFileCount = 0;
   private readonly denoImportCountsByProtocol = new Map<string, number>();
+  private readonly packageImportFiles = new Map<string, Set<string>>();
```
guillemsarda: **`packageImportFiles` could be a counter instead of a `Map<string, Set<string>>`**

```ts
private readonly packageImportFiles = new Map<string, Set<string>>();
```
Since each file is analyzed exactly once per project run, and recordPackageImports is always called with an already-deduplicated per-file Set of package names, a given (filePath, packageName) pair can never be recorded twice. That means the Set<string> of file paths per package isn't needed for correctness — only its final .size is ever read (in getTelemetry()).

Swapping this for a Map<string, number> incremented by 1 per occurrence would give identical counts while avoiding retaining every matched file path in memory for the whole analysis — worth doing on large monorepos with many files/matched packages.

guillemsarda (approved):
Two non-blocking comments.
Image

packages/analysis/src/jsts/rules/helpers/module.ts
nathsou: Good catch — fixed in 93014a2a5. `getReferencedModuleName` now handles `TSImportType` via its `source` string literal (the deprecated `argument` path avoided), so inline type-only imports like `type Props = import('react').ComponentProps<'div'>` are collected. Added a test parsing this pattern with `@typescript-eslint/parser`, and updated the supported-patterns doc comment.

packages/analysis/src/telemetry.ts
nathsou: Agreed — done in 93014a2a5. `packageImportFiles` is now a `Map<string, number>` counter incremented per recorded package, and `recordPackageImports` dropped the unused `filePath` parameter (callers pass an already-deduplicated per-file set exactly once per file). The unit test was updated to match the once-per-file recording semantics.

PR #7637, opened 2026-07-24
JS-2149 Suppress S3504 in var-heavy files

erwan-leforestier-sonarsource:
## Summary
S3504 currently raises one issue per `var` declaration even in files that intentionally use `var` throughout. This change suppresses the rule when `var` is both dominant and noisy in a file, while keeping the existing behavior for mixed modern code.

## Changes
- Add a file-level heuristic in the S3504 decorator that suppresses reporting when a file has at least three runtime `var` declarations and `var` is the dominant declaration style.
- Exclude TypeScript `declare var` statements from that heuristic so ambient declarations do not suppress real findings.
- Add rule-level coverage for var-heavy files, mixed modern files, and the ambient-declaration regression case.

----
## Summary by Gitar

- **Rule updates:**
    - Updated quickfix metadata for rule `S8959` from covered to partial
    - Refactored `S8959` rule implementation by removing unused test command checks and helpers


packages/analysis/src/jsts/rules/S3504/unit.test.ts
```diff
@@ -58,4 +58,88 @@ describe('S3504', () => {
       ],
     });
   });
+
+  it('should suppress files that predominantly use var', () => {
```
lijun-chen-sonarsource, line 62: Would it be useful to add the following tests?

* `var` declaration in for loop initializers
* `var` declaration that declare multiple variables. E.g., `var a = 1, b = 2`. I think the current behavior is of counting this as a single `varDeclarations` is good, but maybe worth locking it down with a test.

packages/analysis/tests/analyzeProject-sonarqube.test.ts
```diff
@@ -494,6 +494,11 @@ describe('SonarQube project analysis', () => {
             '/* eslint-disable-next-line sonarjs/S3504 */',
             'var d = 4;',
             'var e = 5;',
+            'let f = e + 1;',
```
lijun-chen-sonarsource, line 498: To a future reader, the extra let/const declarations are non-obvious. A one-line comment would make it clear that they are present only to avoid the file becoming “var-heavy”.
A perhaps better fix is to use a different rule for this test.

packages/analysis/src/jsts/rules/S3504/unit.test.ts
```diff
@@ -58,4 +58,88 @@ describe('S3504', () => {
...
+            let fifth = fourth + 1;
+            let sixth = fifth + 1;
+          `,
+          errors: 3,
+        },
+      ],
+    });
+  });
+
+  it('should ignore ambient declarations when applying the var heuristic', () => {
```
lijun-chen-sonarsource, line 119: May be out of scope for this PR, since already before this rule relied on `declaration.declare` to determine whether a declaration is an ambient declaration. 

Not sure if other kinds of ambient declaration are detected or not. Maybe worth adding some a test.

```typescript
declare module 'foo' {
  var cached: number;
}

declare namespace Legacy {
  var version: string;
}
```

lijun-chen-sonarsource (approved):
LGTM!

I just added a few suggestions for tests.

packages/analysis/src/jsts/rules/S3504/unit.test.ts
erwan-leforestier-sonarsource: I don't feel the need to have tests for these 2 cases.
Loop and multiple variables declarations have nothing special. I would agree with you if we were, by our own, generating the AST but that's not the case. We rely on the AST provided by eslint. Also the code is pretty straightforward:
```js
if (!declaration.declare) {
    totalDeclarations++;
    if (declaration.kind === 'var') {
        varDeclarations++;
    }
}
  ```
  I will skip your comment but if you have a strong opinion about it, let me know and I can add tests.

lijun-chen-sonarsource: No strong opinion. All good with skipping.

packages/analysis/tests/analyzeProject-sonarqube.test.ts
erwan-leforestier-sonarsource: Added a comment to clarify that the extra let/const declarations are only there to keep the fixture below the var-heavy suppression threshold so one S3504 issue remains open. I kept S3504 in this test because it is specifically exercising suppressed-vs-open issue handling for that rule.

packages/analysis/src/jsts/rules/S3504/unit.test.ts
erwan-leforestier-sonarsource: Good catch. I checked the parser output and vars nested inside `declare module` / `declare namespace` are not marked with `declaration.declare`, so this was a real gap rather than only a missing test. I updated the decorator to also treat vars under a `declare` `TSModuleDeclaration` as ambient, and extended the unit test to cover both cases.

PR #7639, opened 2026-07-24
JS-1610 Replace S1481 with @typescript-eslint/no-unused-vars

erwan-leforestier-sonarsource:
## Summary
Switch `S1481` to a decorated `@typescript-eslint/no-unused-vars` implementation so SonarJS honors underscore ignore patterns while keeping Sonar-specific defaults for parameters, catch variables, and rest siblings.

RSPEC PR: https://github.com/SonarSource/rspec/pull/7606

## Changes
- mark `S1481` as a decorated `typescript-eslint/no-unused-vars` rule and expose the upstream mapping in rule metadata
- wrap the upstream rule with Sonar defaults for `args`, `caughtErrors`, and `ignoreRestSiblings`, while suppressing the upstream fixer
- replace the legacy unit tests with focused coverage for the decorated behavior and move `S1481` from the shipped rule list to the improved external rules section in the plugin README

## Functional Validation
Artifact: [JS-1610-fv.zip](https://github.com/user-attachments/files/30408433/JS-1610-fv.zip)

    ./run.sh

Expected output:
```text
Analyzing "sample.ts"...
Results:
    - Rule "sonarjs/no-unused-vars" -> L.2:9 Remove the declaration of the unused '_unused' variable.

****** Branch "feat/js-1610-switch-s1481-to-ts-eslint" ******
Analyzing "sample.ts"...
Results:
    (none)
```


francois-mora-sonarsource (changes requested):
Thanks for preserving S1481's local-only scope. I found two remaining behavior regressions:

**[P2] Preserve JSX identifiers as variable usage**

The previous S1481 implementation treated JSX identifiers as references, so a local binding used only as `<Icon />` was not reported. The delegated `@typescript-eslint/no-unused-vars` rule reports that binding, since JSX usage normally requires a separate React ESLint rule.

```jsx
function render(icon) {
  const Icon = icon;
  return <Icon />;
}
```

This now raises S1481 on `Icon`, unlike the previous implementation. Could we preserve the existing JSX-reference handling and add a regression test?

**[P2] Avoid duplicating S1128 on unused imports**

S1481 now reports unused import bindings, while S1128 already emits the dedicated import-removal issue. With both rules active, one unused import produces two issues at the same location:

```js
import { foo } from './foo';

console.log('used');
```

```text
S1128: Remove this unused import of 'foo'.
S1481: 'foo' is defined but never used.
```

`@typescript-eslint/no-unused-vars` does not provide an option to exclude imports; `enableAutofixRemoval.imports` only controls its autofix. Could the S1481 interceptor suppress reports whose identifier belongs to an import specifier, leaving S1128 as the owner, and add a regression test?

erwan-leforestier-sonarsource:
Addressed the two regressions you pointed out in S1481:
- JSX component identifiers used only in JSX are treated as usage again, preserving the previous behavior.
- Unused import bindings are filtered out of S1481 so S1128 remains the sole owner for that case.

I also added unit regressions for both cases in the S1481 rule tests.

francois-mora-sonarsource (approved):

PR #7653, opened 2026-07-27
JS-2164 Refine TS requiredEcmaVersion gating

erwan-leforestier-sonarsource:
## Summary
This change makes non-downlevelable TypeScript `requiredEcmaVersion` rules respect the raw compiler target while preserving the existing bypass for downlevelable rules. It applies the new gate to `S7781`, so `replaceAll()` is no longer suggested when `target` is below ES2021.

## Changes
- thread the raw TypeScript target ES year through project analysis and rule filtering
- generalize the rule metadata flag to distinguish downlevelable and non-downlevelable TypeScript changes
- mark `S7781` as non-downlevelable, keep `S6582` explicitly downlevelable, and cover the new behavior with focused tests

## Functional Validation
Artifact: [JS-2164-fv.zip](https://github.com/user-attachments/files/30407214/JS-2164-fv.zip)

    ./run.sh

Expected output:
```text
Analyzing "file.ts"...
Results:
    - Rule "S7781" -> L.4
    - Rule "S6582" -> L.6

****** Branch "fix/js-2164-refine-ts-ecma-gating" ******
Analyzing "file.ts"...
Results:
    - Rule "S6582" -> L.6
```


packages/analysis/src/analyzeWithIncrementalProgram.ts
```diff
@@ -98,6 +99,10 @@ export async function analyzeWithIncrementalProgram(
 
     const detectedEsYear =
       program == null ? undefined : (esLibToYear(program.getCompilerOptions().lib) ?? undefined);
+    const targetEsYear =
+      program == null
+        ? undefined
+        : (tsTargetToEsYear(program.getCompilerOptions().target) ?? undefined);
```
lijun-chen-sonarsource, line 105: How about this?

```suggestion
    const targetEsYear = (tsTargetToEsYear(program?.getCompilerOptions().target) ?? undefined);
```

packages/analysis/src/jsts/linter/filters/filter-ecma-version.ts
```diff
@@ -17,17 +17,22 @@
...
   }
-  // On TypeScript the detected ES year is derived from the compiler target/lib, which
-  // describes emitted syntax. TypeScript downlevels newer syntax, so rules promoting
-  // downlevelable syntax must not be gated by it.
-  if (meta.downlevelableSyntax && ctx.fileLanguage === 'ts') {
+  if (ctx.fileLanguage === 'ts') {
+    if (meta.downlevelableSyntax) {
+      return true;
+    }
+    if (meta.downlevelableSyntax === false && ctx.targetEsYear != null) {
```
lijun-chen-sonarsource, line 31: I initially thought this `meta.downlevelableSyntax === false` was redundant, but then I realized `meta.downlevelableSyntax` can be also `undefined` and the behavior here is that if `meta.downlevelableSyntax` is not explicitly set then `targetEsYear` is ignored and we fallback to `detectedYear`.

Is this the expected behavior?

In any case, it may be useful to add a test for it.

lijun-chen-sonarsource (approved):
Overall LGTM!

I left one nitpick suggestion and one comment asking for clarification.

packages/analysis/src/analyzeWithIncrementalProgram.ts
erwan-leforestier-sonarsource: Yep, I applied this locally. Thanks.

packages/analysis/src/jsts/linter/filters/filter-ecma-version.ts
erwan-leforestier-sonarsource: Yes, that is the intended fallback: if downlevelableSyntax is omitted we keep using detectedEsYear. I also added a small test for that case locally.

PR #7655, opened 2026-07-27
JS-2129: Fix S9073: Composite assertions should be split

nathsou:
## Summary

- keep S9073 scoped to test files
- ignore cohesive `&&` guard chains used for definedness, type narrowing, and safe property access
- retain reports for independent predicates, mixed chains, and numeric or temporal ranges
- cover direct and logically equivalent negated assertion forms

## Validation

- `npx tsx --tsconfig packages/tsconfig.test.json --test packages/analysis/src/jsts/rules/S9073/unit.test.ts`
- Prettier check
- S9073 has no type errors; the complete local type/ruling gate is blocked by expired Repox authentication for `@sonarsource/analyzer-commons-configurations`

## Links

- [JS-2129](https://sonarsource.atlassian.net/browse/JS-2129)
- [RSPEC PR #7614](https://github.com/SonarSource/rspec/pull/7614)

[JS-2129]: https://sonarsource.atlassian.net/browse/JS-2129?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/analysis/src/jsts/rules/S9073/rule.ts
```diff
@@ -57,10 +61,281 @@ export const rule: Rule.RuleModule = {
                 actual.argument.type === 'LogicalExpression' &&
                 actual.argument.operator === '&&')));
 
-        if (isComposite) {
+        if (
+          isComposite &&
+          (!truthyConjunction || !isAssertionGuardChain(truthyConjunction, context.sourceCode))
```
rombirli: I prefer single negation for redeability 
```suggestion
          !(truthyConjunction && isAssertionGuardChain(truthyConjunction, context.sourceCode))
```

packages/analysis/src/jsts/rules/S9073/unit.test.ts
```diff
@@ -92,6 +92,68 @@ describe('S9073', () => {
...
+        {
+          code: `
+            import assert from 'node:assert/strict';
+            assert('create' in rule && typeof rule.create === 'function');
+          `,
+        },
+        {
+          code: `
+            import assert from 'node:assert/strict';
+            assert(result.kind === 'success' && result.value);
```
rombirli: Why this one is compliant?

packages/analysis/src/jsts/rules/S9073/unit.test.ts
```diff
@@ -165,6 +227,62 @@ describe('S9073', () => {
...
+          code: `
+            import assert from 'node:assert/strict';
+            assert(typeof value !== 'object' && value !== null);
+          `,
+          errors: [expectedIssue],
+        },
+        {
+          code: `
+            import assert from 'node:assert/strict';
+            assert(typeof value === 'object' && value.property);
```
rombirli: why is this one non-compliant? I understand why the implementation raises there as `isTruthy` is false, but is it the correct behavior? I'm not a JS expert so I might have missed something. IMO if we accept : 
```js
assert(value && typeof value === 'object' && value.property);
```

triple-composite-assertion we could also accept  this double-composite-assertion

```js
assert(typeof value === 'object' && value.property);
```

rombirli (commented):
What is supposed to happen on this specific case? 

```
assert(x && typeof x == 'object' && f(x))
```

here the beginning of the assertion `x && typeof x == 'object'` is also used to validate `x` before passing it to `f`, so it is the same FP as you tried to solve? I couldn't find any such case in the reproducer

packages/analysis/src/jsts/rules/S9073/rule.ts
nathsou: Applied, thanks. The condition now reads `!(truthyConjunction && isGuardChain(truthyConjunction, context.sourceCode))`.

packages/analysis/src/jsts/rules/S9073/unit.test.ts
nathsou: Good question, and the answer changed: it is now **reported**.

It used to be exempt through a dedicated "discriminant" check that treated `result.kind === 'success'` as establishing something about `result`. But comparing a property is a check on that property, not on the object holding it, so the two operands assert two independent facts and a failure still does not say which one broke.

Dropping that exception removed the last special case that could not be explained in one sentence, and RSPEC PR #7614 is updated to match: this line is now listed as noncompliant instead of compliant.

packages/analysis/src/jsts/rules/S9073/unit.test.ts
nathsou: You were right to push on this: it is now **compliant**.

The report came from a guard that required a `typeof x === 'object'` chain to also exclude `null`, because `typeof null === 'object'` means the type check alone does not make `value.property` safe to read. So the issue was really saying "your guard does not exclude null", which is null-safety verification rather than the diagnosability problem this rule is about. The rule also cannot do it correctly: it runs without type information, so when `value`'s type excludes `null` the extra check is redundant and the issue is a false positive.

I removed that requirement, so the two-operand form is now exempt just like the triple-operand one.

One small detail on the mechanism: `isTruthy` is `true` here, since for `assert(...)` the extracted predicate is `truthy` with `negated` false. The report came from the guard-chain check failing, not from `isTruthy`.

The exception is now a single rule: one operand checks the existence, type, or shape of a reference, and every operand only uses that same reference.

nathsou:
Reworked the exception after your review. It is now a single rule instead of a layered set of heuristics: **a conjunction is exempt when one operand checks the existence, type, or shape of a stable reference and every operand only uses that same reference.** `rule.ts` loses 174 lines and gains 85.

What changed in behaviour:

- `typeof value === 'object' && value.property` is no longer reported. The old null-safety guard was a false positive whenever the value's type excludes `null`, and the rule has no types to decide.
- Any single-argument call now counts as a check on its argument, so `isNonEmpty(list) && list[0].id` is exempt. This replaces the `Array.isArray` special case plus the `/^is[A-Z_]/` callee-name regex.
- `result.kind === 'success' && result.value` is now reported: comparing a property is a check on that property, not on the object holding it. RSPEC PR #7614 is updated accordingly.
- `typeof value !== 'object' && value !== null` is now exempt. That is an accepted false negative from the simpler model, pinned by a test with a comment.

I also restored the two ruling baselines this PR had deleted. The new guards do not change what the rule reports on the ruling projects, which is why both ruling jobs were failing: the `ruling-differences` artifact still contained the same four findings the baselines described.

packages/analysis/src/jsts/rules/S9073/rule.ts
```diff
@@ -57,10 +61,167 @@ export const rule: Rule.RuleModule = {
...
+function getCheckedSubjects(operand: estree.Expression): StableReference[] {
+  if (isStableReference(operand)) {
+    return [operand];
+  }
+  if (operand.type === 'CallExpression' && operand.arguments.length === 1) {
+    const [argument] = operand.arguments;
+    return isStableReference(argument) ? [argument] : [];
+  }
+  return operand.type === 'BinaryExpression' ? getComparisonSubjects(operand) : [];
+}
```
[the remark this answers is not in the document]
nathsou: You're right, thanks — this was a real regression, not just a doc gap. `assert(compute(value) && value.ok)` is exactly the pattern the rule exists to catch: two conditions that can independently fail with nothing indicating which one did.

Restricted the call branch back to predicate-named calls (`is[A-Z_]`-style identifiers and `Array.isArray`), which is what the previous implementation already did, and added a regression test pinning `compute(value) && value.ok` as reported. Pushed as `5574875af`.

its/ruling/src/test/expected/ant-design/javascript-S9073.json
```diff
@@ -1,5 +0,0 @@
-{
-"ant-design:components/menu/__tests__/index.test.js": [
-59
```
rombirli, line 3: [This](https://github.com/SonarSource/jsts-test-sources/blob/master/ant-design/components/menu/__tests__/index.test.js#L59) is a TP that is gone, so a FN

its/ruling/src/test/expected/vitest/typescript-S9073.json
```diff
@@ -1,7 +0,0 @@
-{
-"vitest:packages/vitest/src/node/test-run.ts": [
-62,
-196,
-269
```
rombirli, line 5: Those 3 are debatable
- https://github.com/SonarSource/jsts-test-sources/blob/master/vitest/packages/vitest/src/node/test-run.ts#L62
- https://github.com/SonarSource/jsts-test-sources/blob/master/vitest/packages/vitest/src/node/test-run.ts#L196
- https://github.com/SonarSource/jsts-test-sources/blob/master/vitest/packages/vitest/src/node/test-run.ts#L269


here

```ts
    const task = this.vitest.state.idMap.get(testId)
    const entity = task && this.vitest.state.getReportedEntity(task)

    assert(task && entity, `Entity must be found for task ${task?.name || testId}`)
```

entity depends on task, but :
- is already checked at entity definition so `assert(entity)` would be the same
- is variables backtracking expected? It looks overkill to me as we don't have symbolic execution it would be better to keep things simple

rombirli (approved):
The implementation became highly complex, and I can't understand it anymore. This is expected as these heuristics would need symbolic execution to completely eliminate the FPs. For this reason I propose we rely solely on metrics to evaluate how good this fix is. However the ruling results looks weird to me, **could you please check that all 4 changes in ruling are expected before merging**.

its/ruling/src/test/expected/ant-design/javascript-S9073.json
nathsou: Confirmed as a real gap, but it is not caused by the guard-chain simplification: this file is classified as `MAIN` rather than `TEST` by the ruling harness (the `ant-design` project only configures `tests: "tests"`, and `matchesTestPath` ignores the filename/`__tests__` heuristic entirely once a `testDir` is configured — see `filterPathAndGetFileType` in `filter-path.ts`). Since S9073 is `scope: Tests`, `fileTypeTargets` excludes this file regardless of what the rule itself would report on its AST.

I verified this directly: calling `filterPathAndGetFileType` for `ant-design:components/menu/__tests__/index.test.js` with the project's actual config returns `MAIN`. Feeding the file straight into the rule (bypassing scope filtering) still reports line 59, so the guard-chain logic itself is unaffected — the composite `||` assertion is still recognized as composite.

This looks like a pre-existing limitation of the ruling harness's per-project `testDir` heuristic, affecting any `Tests`-scoped rule when a project has test files co-located outside its configured test directory — not something introduced by this PR. I don't think it's in scope here to touch the ruling infra, so I'll leave the baseline removal as is; happy to file a follow-up if you agree it's worth fixing at the harness level.

its/ruling/src/test/expected/vitest/typescript-S9073.json
nathsou: Same root cause as the ant-design thread: all three lines are in `packages/vitest/src/node/test-run.ts`, which is classified as `MAIN` (the `vitest` project only configures `tests: "test"`, a top-level dir this file isn't under). Since S9073 is `scope: Tests`, none of these were ever analyzed — confirmed by running `filterPathAndGetFileType` directly, which returns `MAIN` for this path. So this isn't a guard-chain regression either.

On the backtracking question for `assert(task && entity, ...)`: agreed, we should keep it simple. The rule only looks at what's syntactically checked within the assertion itself, not at how `entity` was defined a few lines above. Since `entity`'s definition (`task && this.vitest.state.getReportedEntity(task)`) isn't a predicate-named call, `getCheckedSubjects` doesn't establish `entity` as depending on `task`, so `mentions` fails both ways and it's reported as composite — which happens to be actionable advice anyway (`assert(entity, ...)` alone would suffice, per your own read). Adding data-flow backtracking to recognize this pattern would meaningfully increase complexity and false-negative risk for a rule that's meant to stay purely syntactic, so I'd rather not go there.

PR #7659, opened 2026-07-27
JS-2064 New rule S9107: vue require prop type constructor

guillemsarda:
Part of 

----
## Summary by Gitar

    - Added rule `S9107` (`vue/require-prop-type-constructor`) ensuring Vue component prop types are constructors.
    - Included rule metadata, decorator, unit tests, JSON configuration, and documentation updates.


packages/analysis/src/jsts/rules/S9107/decorator.ts
```diff
@@ -0,0 +1,68 @@
...
+ * along with this program; if not, see https://sonarsource.com/license/ssal/
+ */
+// https://sonarsource.github.io/rspec/#/rspec/S9107/javascript
+
+import type { Rule } from 'eslint';
+import { generateMeta } from '../helpers/generate-meta.js';
+import { interceptReport } from '../helpers/decorators/interceptor.js';
+import * as meta from './generated-meta.js';
+
+const IDENTIFIER_PATTERN = /^[A-Z][A-Za-z0-9_$]*$/;
```
nathsou, line 24: Preserve the safe `null` quick-fix

`IDENTIFIER_PATTERN` rejects `null`, so `type: 'null'` is now reported without the upstream fix. However, `type: null` is accepted by the upstream Vue rule (and is specified as an exception in the RSPEC: https://github.com/SonarSource/rspec/blob/8fb75c3430f746d53a16e02a49e10d68621850b6/rules/S9107/javascript/rule.adoc#exceptions) and therefore by this decorator, the decorator only changes reports it receives. This turns a valid, issue-resolving fix into no fix. Please allow this case (and add a regression test for `'null'` → `null`).

nathsou (approved):
(Just 1 comment really)

Image

PR #7669, opened 2026-07-28
JS-2176 Fix S6747 false positives for simple JSX props

francois-mora-sonarsource:
Part of JS-2176

## Summary

This replaces the canceled React Three Fiber-specific S6747 proposal with a smaller set of framework-specific JSX prop exceptions.

The reusable refactoring keeps S6747's false-positive handling in a report-first escape registry. The behavior change is limited to simple prop-level exceptions:

- `css` for Compiled, Emotion (`@emotion/react` and legacy `@emotion/core`), and styled-components projects
- `sx` for Theme UI and StyleX projects
- `tw` for `next/og`, `@vercel/og`, `satori`, and `twin.macro` when their runtime APIs are imported
- existing simple `jsx`/`global` exceptions for Next.js and styled-jsx remain covered

Runtime import detection ignores type-only imports so type references do not activate a JSX prop exception.

## Library candidate survey

The PR includes `docs/rule-research/S6747-jsx-property-exception-survey.md`, which records the candidate search and the decision criteria used for S6747 exceptions.

The survey covered styling libraries, compile-time styling systems, custom renderers, mobile/platform abstractions, custom JSX runtimes, UI systems, chart libraries, animation/gesture libraries, and editor stacks.

Several real candidates were intentionally left out because they need config-aware detection, are too broad for a simple prop ignore list, or apply only to library components rather than raw intrinsic JSX elements. This includes Preact, SolidJS, Pigment CSS, Astroturf, Hono, Qwik, React Three Fiber, Pixi React, react-babylonjs, react-blessed, UI systems, chart libraries, editor stacks, Radix `Slot` / `asChild`, Lit React wrappers, and prop-getter libraries.

React Three Fiber support remains intentionally excluded because the safe exception surface requires a large element/property matrix that is not feasible with the current SonarJS implementation.

## Validation

- `npx tsx --tsconfig packages/tsconfig.test.json --test packages/analysis/src/jsts/rules/S6747/*/unit.test.ts`
- `npm run bbf`
- `git diff --check origin/master...HEAD`


asya-vorobeva (approved):
LGTM

PR #7670, opened 2026-07-28
JS-2174: Implement rule S9114: Lodash and Underscore.js debounced or throttled functions should not be recreated on every React render

nathsou:
## Why

New rule S9114 ([JS-2174](https://sonarsource.atlassian.net/browse/JS-2174), RSPEC: https://github.com/SonarSource/rspec/pull/7546): `debounce`/`throttle` wrappers from lodash, lodash-es, and underscore keep their pending state inside the returned function. Creating one directly in a React function component, custom hook, or class component `render` method discards that state on every render, resetting the debounce timer or throttle window.

## What

- `packages/analysis/src/jsts/rules/S9114/`: original rule with no parser services
  - FQN check covers named, aliased, namespace, per-method path imports and `require` forms for lodash, lodash-es, and underscore
  - Uses shared React helpers for function components, custom hooks, and React class components
  - Excludes nested callbacks and memo-hook-wrapped calls; recognizes stable `useRef` lazy initialization
  - Uses the shared `isCallResult` AST helper with S8907 and S8932
- Comment-based fixtures cover imports, function/class/custom-hook components, `React.memo`, `React.forwardRef`, ref guards, and compliant exclusions
- Profiles: Sonar way; `requiredDependency: ["react"]`
- RSPEC PR is unmerged → `rspec.sha` is pinned to its current head

## Test plan

- [x] `npx tsx --test packages/analysis/src/jsts/rules/S9114/**/*.test.ts`
- [x] `uc check` (merge conflict, Knip, type check, unit, ruling, Sonar lint, Sonar issues)

[JS-2174]: https://sonarsource.atlassian.net/browse/JS-2174?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/analysis/src/jsts/rules/S9114/rule.ts
```diff
@@ -0,0 +1,197 @@
...
+  const ancestors = context.sourceCode.getAncestors(call);
+  for (let i = ancestors.length - 1; i >= 0; i--) {
+    const ancestor = ancestors[i];
+    if (ancestor === boundaryFunction) {
+      return false;
+    }
+    if (
+      ancestor.type === 'CallExpression' &&
+      ancestor.callee.type === 'Identifier' &&
+      memoHooks.has(ancestor.callee.name)
```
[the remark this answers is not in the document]
nathsou: Valid catch — fixed in bd3810c64: the memo-hook exemption now resolves the callee via getFullyQualifiedName and strips the react. prefix, so React.useCallback(debounce(fn), []) is exempt. Added a fixture case for the namespace-import form.

packages/analysis/src/jsts/rules/S9114/rule.ts
```diff
@@ -0,0 +1,244 @@
...
+
+  const enclosingClass = findFirstMatchingAncestor(
+    methodDefinition as TSESTree.Node,
+    ancestor => ancestor.type === 'ClassDeclaration' || ancestor.type === 'ClassExpression',
+  );
+  return enclosingClass !== undefined && isReactClassComponent(enclosingClass as estree.Node);
+}
+
+function isRenderMethod(node: estree.Node): boolean {
+  return (
```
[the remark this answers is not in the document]
nathsou: Addressed in 4cb392083. The class-render gate now accepts non-static, non-computed `PropertyDefinition` members named `render`, and the JSX fixture covers `render = () => { ... }`.

nathsou:
Gitar fix CI

packages/analysis/src/jsts/rules/S9114/rule.ts
```diff
@@ -0,0 +1,253 @@
...
+function isMethodName(name: string): name is MethodName {
+  return methodNames.has(name);
+}
+
+function isCallResult(node: estree.Expression | estree.Super): node is estree.CallExpression {
+  if (node.type === 'ChainExpression') {
+    return isCallResult(node.expression);
+  }
+  return node.type === 'CallExpression';
+}
```
guillemsarda: Rules S8932 and S8907 are also using this function. It is a good candidate to move into a shared helper.

packages/analysis/src/jsts/rules/S9114/rule.ts
```diff
@@ -0,0 +1,253 @@
...
+): MethodName | undefined {
+  const parts = fullyQualifiedName.replaceAll('/', '.').split('.');
+  if (parts.length !== 2) {
+    return undefined;
+  }
+  const [moduleName, qualifier] = parts;
+  if (
+    supportedModules.has(moduleName) &&
+    isMethodName(qualifier) &&
+    (syntacticMethod === null || syntacticMethod === qualifier)
```
guillemsarda: `(syntacticMethod === null || syntacticMethod === qualifier)` is always true here, given `supportedModules.has(moduleName) && isMethodName(qualifier)` already passed — it can be removed.

Reasoning:
- If `syntacticMethod === null`, the `||` is trivially true.
- If `syntacticMethod` is a `MethodName`, it was derived from `callee.property.name` on the call's `MemberExpression` (in `getSyntacticMethodName`). `getFullyQualifiedName`'s `qualifier` is built from the exact same node via `reduceTo` in `helpers/module.ts`, which reads `property.name`/`property.value` off that same `MemberExpression` — module resolution only ever supplies the prefix, never the trailing segment. So `qualifier === syntacticMethod` always holds in this branch too, and it's already confirmed to be a `MethodName` by the preceding `isMethodName(qualifier)` check.

Verified empirically too: stripped the check down to `supportedModules.has(moduleName) && isMethodName(qualifier) ? qualifier : undefined` and ran the rule's own `cb.test.ts` — all fixtures still pass identically. Also threw in two extra cases not covered by the current fixtures (a namespace import, `underscoreNs.debounce(...)`, and a plain aliased local, `const aliasedLib = _; aliasedLib.debounce(...)`) — no divergence there either.

Since both branches are unconditionally true, the whole clause — and the `syntacticMethod` parameter of `getSupportedMethodName` — can be dropped:

```ts
function getSupportedMethodName(fullyQualifiedName: string): MethodName | undefined {
  const parts = fullyQualifiedName.replaceAll('/', '.').split('.');
  if (parts.length !== 2) {
    return undefined;
  }
  const [moduleName, qualifier] = parts;
  return supportedModules.has(moduleName) && isMethodName(qualifier) ? qualifier : undefined;
}
```

(and the call site can stop passing `syntacticMethod` into it — it's still needed earlier, just not here).

packages/analysis/src/jsts/rules/S9114/cb.fixture.jsx
```diff
@@ -0,0 +1,65 @@
...
+    this.onChange = debounce(fetchResults, 300);
+  }
+
+  componentDidMount() {
+    const onChange = debounce(fetchResults, 300);
+    onChange();
+  }
+
+  render() {
+    const handleClick = () => debounce(fetchResults, 300);
```
guillemsarda: In `cb.fixture.jsx`, the `handleClick` compliant case doesn't reflect a realistic usage — `debounce(fetchResults, 300)` is created and its return value is discarded, so `fetchResults` never actually runs on click. Suggest making it call the debounced function so it represents the real anti-pattern (debounce recreated on every click, defeating it just as completely as the render-time case, just in a different way), and flagging in a comment that this is an accepted, known limitation of the rule rather than an oversight:

```jsx
render() {
  const handleClick = () => debounce(fetchResults, 300)(); // Compliant: known limitation — call is nested one function-level deep (inside handleClick), so it's not "directly in the render body". Still defeats debouncing in practice, just not detected.
  return <input onClick={handleClick} />;
}

File: packages/analysis/src/jsts/rules/S9114/rule.ts [hunk 6]
```diff
@@ -0,0 +1,253 @@
...
+  create(context: Rule.RuleContext) {
+    return {
+      CallExpression(node: estree.Node) {
+        const call = node as estree.CallExpression;
+        const syntacticMethod = getSyntacticMethodName(call.callee);
+        if (syntacticMethod === undefined) {
+          return;
+        }
+
+        const fullyQualifiedName = getFullyQualifiedName(context, call);
```
guillemsarda, line 67: **Potential false positive: ref-guard lazy-init pattern**

```jsx
function Component() {
  const ref = useRef();
  if (!ref.current) {
    ref.current = debounce(fetchResults, 300);
  }
  return <input onChange={ref.current} />;
}
```
This is a documented React pattern for creating a stable value without useMemo (used specifically because useMemo isn't a semantic guarantee — React can drop cached values). The if (!ref.current) guard means debounce(...) only ever runs once, so nothing is actually recreated on every render.

The rule currently flags this anyway: findFirstEnclosingFunction only stops at function boundaries, so an if/BlockStatement guard doesn't change the result — the call's enclosing function is still Component. isWrappedInMemoHook's memoHooks set only recognizes useMemo/useCallback, not a ref.current guard.

Worth a compliant fixture case either way — if excluding this pattern is out of scope for this PR, at least worth confirming it's an accepted false positive rather than an oversight.

packages/analysis/src/jsts/rules/S9114/cb.fixture.jsx
```diff
@@ -0,0 +1,65 @@
+import { debounce } from 'lodash';
+import React, { useMemo } from 'react';
+
+function Search() {
```
guillemsarda, line 4: **Potential false negative: bug hidden inside a custom hook**

```jsx
function useDebouncedSearch(onSearch) {
  return debounce(onSearch, 300); // same bug, not flagged
}

function SearchBox({ onSearch }) {
  const debouncedSearch = useDebouncedSearch(onSearch); // re-runs every SearchBox render
  return <input onChange={debouncedSearch} />;
}
```

`useDebouncedSearch` re-executes on every render of SearchBox, exactly like a component body does — same underlying bug (timer reset every render). But componentNamePattern = /^[A-Z]/ in isFunctionComponent only recognizes capitalized names, so a hook-convention use... function is never considered a "component" and the call inside it is never evaluated by getComponentType.

This is more of a scope question than a bug: does S9114 intend to cover only components, or also custom hooks with the same per-render lifecycle? If it's intentionally components-only, might be worth noting in the rule description; if custom hooks should be in scope, this'd need isFunctionComponent (or a sibling check) to also recognize /^use[A-Z]/-named functions.

packages/analysis/src/jsts/rules/S9114/rule.ts
guillemsarda: https://react.dev/reference/react/useRef#avoiding-recreating-the-ref-contents

guillemsarda (changes requested):
The ref pattern comment blocks the PR. If we decide to cover this case, an update of the rspec would be needed.

packages/analysis/src/jsts/rules/S9114/rule.ts
nathsou: Addressed in commit 8cdb1bfa6. S9114 now recognizes React useRef bindings and exempts debounce/throttle assignments to ref.current when they are directly guarded by !ref.current or a null comparison. The compliant fixture and RSPEC guidance are included; local uc check passes.

nathsou:
CI is now blocked only in `Prepare RSPEC rule data`, before SonarJS tests run. The refreshed RSPEC pin is correct at `432ea36127b62dc6beffc1124b9b5dc30c19427b`; the pinned RSPEC tree also exposes ready rule `S9128`, but this PR/base does not contain `resources/rule-data/javascript/S9128.json`. The deploy step therefore fails with `ENOENT`. This is an unrelated RSPEC baseline mismatch; I did not add unrelated S9128 data or bypass the guard. The S9114 local `uc check` remains green.

packages/analysis/src/jsts/rules/S9114/rule.ts
nathsou: Addressed in commit 8cdb1bfa6. `isCallResult` is now exported from `helpers/ast.ts` and reused by S8907, S8932, and S9114. Local `uc check` passes.

guillemsarda (approved):
Image

PR #7683, opened 2026-07-29
JS-2191 USER-2352 Support AWS CDK assertions in S2699

francois-mora-sonarsource:
## Summary

- recognize AWS CDK Template assertions from aws-cdk-lib/assertions in S2699
- support hasResourceProperties and resourceCountIs from Template.fromStack
- cover ordinary, type-aware, and setup-only boundary cases

## Root cause

S2699 did not recognize aws-cdk-lib/assertions as an assertion library, and its Template assertion calls were absent from both assertion-detection paths.

## Validation

- npx tsx --test packages/analysis/src/jsts/rules/S2699/unit.test.ts
- npm run bbf
- npx prettier --check packages/analysis/src/jsts/rules/helpers/assertion-detection.ts packages/analysis/src/jsts/rules/helpers/aws-cdk.ts packages/analysis/src/jsts/rules/S2699/unit.test.ts
- git diff --check a19f17468...HEAD

### DIFF
--- a/packages/analysis/src/jsts/rules/helpers/assertion-detection.ts
+++ b/packages/analysis/src/jsts/rules/helpers/assertion-detection.ts
@@ -38,6 +38,7 @@ import * as Vitest from './vitest.js';
 import * as Supertest from './supertest.js';
 import * as Cypress from './cypress.js';
 import * as Uvu from './uvu.js';
+import * as AwsCdk from './aws-cdk.js';
 import { getParent } from './ancestor.js';
 import { getFullyQualifiedName, importsOrDependsOnModule } from './module.js';
 import { getFullyQualifiedNameTS, importsModuleTS } from './module-ts.js';
@@ -53,6 +54,7 @@ const ASSERTION_LIBRARIES = [
   'bun:test',
   'node:assert',
   'node:assert/strict',
+  'aws-cdk-lib/assertions',
   'uvu/assert',
 ];
 // runners that expose assertion APIs as globals (no import required).
@@ -190,6 +192,7 @@ const SCRIPT_CAPABLE_DETECTORS: AssertionDetector[] = [
   Supertest.isAssertion,
   isFunctionCallFromNodeAssert,
   Uvu.isAssertion,
+  AwsCdk.isAssertion,
 ];
 
 const RUNNER_BOUND_DETECTORS: AssertionDetector[] = [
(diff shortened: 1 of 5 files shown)


packages/analysis/src/jsts/rules/helpers/aws-cdk.ts
```diff
@@ -0,0 +1,45 @@
...
+ * along with this program; if not, see https://sonarsource.com/license/ssal/
+ */
+import type { Rule } from 'eslint';
+import type estree from 'estree';
+import type { ParserServicesWithTypeInformation } from '@typescript-eslint/utils';
+import ts from 'typescript';
+import { getFullyQualifiedName } from './module.js';
+import { getFullyQualifiedNameTS } from './module-ts.js';
+
+const ASSERTION_METHOD_NAMES = new Set(['hasResourceProperties', 'resourceCountIs']);
```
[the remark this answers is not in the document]
francois-mora-sonarsource: Addressed in ed8b16afc. Detection now enumerates documented assertion methods by receiver: Template.fromStack/fromJSON/fromString, Annotations.fromStack, and Tags.fromStack. JavaScript and type-aware tests cover the added APIs, while query/accessor methods remain excluded.

packages/analysis/src/jsts/rules/helpers/assertion-detection.ts
```diff
@@ -157,6 +169,19 @@ export function hasSupportedAssertionLibrary(context: Rule.RuleContext): boolean
...
 
+/**
+ * Like {@link hasSupportedAssertionLibrary}, but also accepts the
+ * {@link UNCLASSIFIED_LIBRARIES}. Only for callers that do not reason about the
+ * script-capable / runner-bound classification.
+ */
+export function hasAssertionLibraryIncludingUnclassified(context: Rule.RuleContext): boolean {
+  return importsOrDependsOnModule(
+    context,
+    [...ASSERTION_LIBRARIES, ...UNCLASSIFIED_LIBRARIES],
```
guillemsarda, line 180: This function is fine for now. However, if a third rule wants to consume a different set of assertion libraries, we should perhaps think of modifying `hasAssertionLibrary` and adding an `allowList` parameter to make it more scalable. 

Right now this scales to one function per different subset of assertion libraries.

guillemsarda (approved):

packages/analysis/src/jsts/rules/helpers/assertion-detection.ts
francois-mora-sonarsource: I agree, and as discussed, whenever this happens we should start refactoring these shared helpers.

PR #7689, opened 2026-07-30
JS-2198 Block PR merges on failed ESLint plugin consumer checks

erwan-leforestier-sonarsource:
## Summary
This makes failed ESLint plugin consumer checks block PR promotion instead of slipping through as a green required gate. It also keeps all ESLint consumer matrix legs visible so the PR status reflects the real failure set.

## Changes
- run the `promote` job unconditionally after its dependencies complete and fail it explicitly when any needed job ends in `failure` or `cancelled`
- add a small tested helper that validates the GitHub Actions `needs` context and prints a targeted blocking message
- disable fail-fast on the ESLint plugin consumer matrix so all three ESLint versions report their status on the PR

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -379,6 +379,7 @@ jobs:
     needs: [setup, build_eslint_plugin]
     permissions: *read_permissions
     strategy:
+      fail-fast: false
       matrix:
         include:
           - eslint-version: 10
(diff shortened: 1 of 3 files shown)


erwan-leforestier-sonarsource:
Closing in favor of #7691. This work was repurposed from the original dependency fix into a CI merge-protection change, so I reopened the same implementation on a fresh correctly named branch and PR to keep the review history accurate.

PR #7701, opened 2026-08-03
JS-2205 USER-2389 Support Window onmessage in S2819

francois-mora-sonarsource:
Part of USER-2389

## What changed

S2819 now detects unverified `Window.onmessage` registrations alongside the existing `addEventListener("message", ...)` support.

Covered receivers are non-computed `onmessage` assignments on `window` and `globalThis`, including local aliases whose unique write resolves to one of them (`const target = window`). The receiver's `onmessage` property must be declared in `lib.dom.d.ts`, which leaves Worker and Node projects alone since they compile against `lib.webworker.d.ts` or no DOM lib at all.

Files that alias the global with a `window = self` / `window = global` shim are excluded: they run in a Worker, where messages come from the parent and there is no origin to verify. Types cannot tell this apart, since `self` and `window` are both `Window & typeof globalThis` once the DOM lib is loaded, so the shim is detected syntactically. `ace/lib/ace/worker/worker_v2.js` in the ruling sources is exactly this case.

Listener resolution is shared with the `addEventListener` path, which as a side effect also resolves handlers held in `const` variables (`const handler = event => {...}`) — previously only function declarations were followed, so `addEventListener` gains a few true positives too.

## Out of scope

- Computed and dynamic properties (`window["onmessage"]`).
- Compound assignments (`??=`, `||=`), which do register a handler and are a known false negative.
- Window-typed values that do not resolve to `window`/`globalThis`: parameters, `iframe.contentWindow`, `document.defaultView`.
- Non-Window transports such as WebSocket, `MessagePort` or `BroadcastChannel`.
- The same Worker shim on the `addEventListener` path, which reports it today. Harmonising the two paths would remove issues from the ruling baseline and belongs in its own change.

## Why

Assignments to `Window.onmessage` receive `postMessage` events and require the same origin verification as `addEventListener("message", ...)`.

## Validation

- `npx tsx --tsconfig packages/tsconfig.test.json --test packages/analysis/src/jsts/rules/S2819/unit.test.ts`
- `npx tsc --noEmit -p packages/tsconfig.test.json`
- `npm run bbf`
- `git diff --check origin/master...HEAD`
- Ruling: expected results unchanged.

### DIFF
--- a/packages/analysis/src/jsts/rules/S2819/rule.ts
+++ b/packages/analysis/src/jsts/rules/S2819/rule.ts
@@ -64,14 +64,9 @@ export const rule: Rule.RuleModule = {
   },
 };
 
-function isWindowObject(
-  node: estree.Node,
-  context: Rule.RuleContext,
-  allowWindowNameMatch = true,
-) {
+function isWindowObject(node: estree.Node, context: Rule.RuleContext) {
   const type = getTypeAsString(node, context.sourceCode.parserServices);
-  const hasWindowName =
-    allowWindowNameMatch && WindowNameVisitor.containsWindowName(node, context);
+  const hasWindowName = WindowNameVisitor.containsWindowName(node, context);
   return type.match(/window/i) || type.match(/globalThis/i) || hasWindowName;
 }
 
@@ -133,8 +128,15 @@ function checkOnMessageAssignment(
   checkMessageListener(context, assignment.right, left);
 }
 
+/**
+ * Unlike `isWindowObject`, which also accepts any identifier whose name contains 'window',
+ * the receiver of an `onmessage` assignment must resolve to `window` or `globalThis` itself.
+ * The name heuristic is too coarse here: `onmessage` is also a property of unrelated
+ * transports such as WebSocket, so a `wsWindowChannel.onmessage` assignment would be
+ * reported without this restriction.
+ */
 function isWindowMessageReceiver(node: estree.Node, context: Rule.RuleContext) {
-  if (isWorkerEnvironment(context) || isWindowAliasedToWorkerGlobal(node, context)) {
+  if (isWindowAliasedToWorkerGlobal(node, context)) {
     return false;
   }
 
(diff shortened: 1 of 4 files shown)


francois-mora-sonarsource:
Two known gaps, left deliberately out of this PR and recorded here rather than as tickets:

**1. Compound assignments are not detected.** `window.onmessage ??= handler` and `window.onmessage ||= handler` do register a handler, so they are false negatives. The guard is a single condition (`assignment.operator !== '='`), but widening it can add issues to the ruling baseline, so it is worth doing as its own change.

**2. The `addEventListener` path still reports Worker shims.** A file doing `window = self` and then `window.addEventListener("message", handler)` is reported, whereas the same file assigning `window.onmessage` is now excluded. `isWindowObject` accepts any identifier whose name matches `/window/i`, so it also reports non-Window receivers such as `const workerWindow = new WebSocket(...)`. Routing both paths through `isWindowMessageReceiver` would harmonise them and remove those false positives, which means updating the ruling expected results — again its own change.

For context on why the shim exclusion exists at all: dropping it made ruling fail with exactly one new issue, `ace/lib/ace/worker/worker_v2.js:71`, which aliases `window = self` at the top of a module factory and then registers `window.onmessage`. Messages there come from the parent page and `event.origin` is `""`, so there is nothing to verify. Types cannot separate the two cases, since `self` and `window` are both `Window & typeof globalThis` once the DOM lib is loaded.

packages/analysis/src/jsts/rules/S2819/rule.ts
```diff
@@ -100,7 +106,88 @@ function checkAddEventListenerCall(callExpr: estree.CallExpression, context: Rul
...
+ * The write is looked up on the `window` variable when the analysis configuration declares it as
+ * a global, and among the unresolved references of the enclosing scope otherwise. Only the second
+ * case is known to occur in practice; the first is covered so that the check does not depend on
+ * whether `window` happens to be declared.
+ */
+function isWindowAliasedToWorkerGlobal(node: estree.Node, context: Rule.RuleContext) {
+  const variable = getVariableFromName(context, 'window', node);
+  const writes = variable
+    ? variable.references.filter(reference => reference.isWrite())
+    : context.sourceCode.getScope(node).through.filter(reference => reference.isWrite());
```
erwan-leforestier-sonarsource, line 181: **False negative when a local `window` shadows the global.**

A local variable named `window` initialized from `self`/`global` looks identical to the Worker shim here, so the check bails out:

```js
const window = self;
globalThis.onmessage = e => console.log(e.data); // not reported
```

The local alias doesn't touch the real global, and the handler still needs origin verification.

Could we require the `window` variable to be the global one (e.g. `variable.scope.type === 'global'`) before treating it as a shim? A plain local binding would then fall through to the normal check.

packages/analysis/src/jsts/rules/S2819/rule.ts
```diff
@@ -100,7 +106,88 @@ function checkAddEventListenerCall(callExpr: estree.CallExpression, context: Rul
...
+  return writes.some(
+    reference =>
+      reference.identifier.name === 'window' &&
+      isIdentifier(reference.writeExpr ?? undefined, 'self', 'global'),
+  );
```
erwan-leforestier-sonarsource, line 186: **False positive when the shim and receiver are in different functions.**

The shim lookup reads `getScope(node).through`, but `node` is the receiver, so inside a function it only sees that function's escaped references. A shim in a sibling function is invisible:

```js
function init()   { window = self; }
function attach() { window.onmessage = e => console.log(e.data); } // reported
```

This contradicts the JSDoc above ("any shim in the file disqualifies every receiver, wherever the write sits"). The current tests only cover a module-level shim or a shim in the same factory, so this layout slips through.

Could we resolve `.through` from the program scope instead of the receiver's scope, so detection is genuinely file-wide?

francois-mora-sonarsource: Thanks. I created JS-2240 to address Worker-shim false positives coherently for both `onmessage` and the established `addEventListener("message", ...)` path: https://sonarsource.atlassian.net/browse/JS-2240. It needs a broader policy for genuine global shims and runtime/scope boundaries, so I am keeping this PR focused on the local-shadowing regression.

packages/analysis/src/jsts/rules/S2819/rule.ts
francois-mora-sonarsource: Thanks. Addressed in the latest commit: declared `window` bindings no longer count as global Worker shims, while configured globals and unresolved writes retain the existing behavior. Added the `const window = self; globalThis.onmessage = ...` regression, which now reports S2819.

erwan-leforestier-sonarsource (approved):

PR #7705, opened 2026-08-04
JS-2206: Implement S9115: Promises returned by Testing Library async events should be handled

nathsou:
## Summary

- Implement S9115 as a decorated `eslint-plugin-testing-library/await-async-events` rule.
- Gate findings on a known `@testing-library/user-event` minimum version of 14.0.0 or later.
- Restrict the first version to `userEvent`, preserve upstream quickfixes, and use the approved direct and wrapper messages.
- Add exhaustive unit tests, dependency-version fixtures, and a custom ruling baseline.

## Links

- Jira: https://sonarsource.atlassian.net/browse/JS-2206
- RSPEC PR: https://github.com/SonarSource/rspec/pull/7547

## Verification

- `npm run eslint-plugin:check`
- Focused S9115 unit tests
- Custom-jsts ruling test
- `uc check` (all checks pass locally except the expected no-open-PR Sonar issue check before this PR exists)


packages/analysis/src/jsts/rules/S9115/decorator.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  return {
+    ...interceptedRule,
+    create(context: Rule.RuleContext): Rule.RuleListener {
+      const dependencies = getDependenciesSanitizePaths(context);
+      const versionRange = dependencies.get(USER_EVENT_MODULE);
+      if (!supportsUserEvent(versionRange)) {
+        return {};
+      }
+      return interceptedRule.create(context);
```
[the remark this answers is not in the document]
nathsou: Addressed in 32a92504e: removed the redundant interceptReport wrapper, corrected the valid promise-handling fixture, and added S9115 to the decorated-rule catalog. `uc check` passes all seven checks.

packages/analysis/src/jsts/rules/S9115/unit.test.ts
```diff
@@ -0,0 +1,254 @@
...
+    ruleTester.run('dependency versions', rule, {
+      valid: [
+        ...[
+          'user-event-v13',
+          'user-event-mixed',
+          'user-event-unbounded',
+          'user-event-latest',
+          'user-event-invalid',
+        ].map(name => ({ code, filename: fixtures(name) })),
+        { code, filename: fixtures('user-event-missing') },
```
guillemsarda, line 216: The `'gates reports on a known v14 or newer dependency range'` test includes a `user-event-missing` case (line 216), but there's no `fixtures/user-event-missing/` directory in this PR — every other name in this test (`user-event-v13`, `user-event-v14`, `user-event-mixed`, etc.) has a matching fixture folder, this one doesn't.

The test still passes, but only because `getDependencies`'s directory walk (`find-minimatch.ts`) swallows `ENOENT` from a nonexistent path and falls through to "no manifest found", which happens to model "dependency absent" correctly. That's an implementation detail this test is leaning on implicitly rather than asserting directly.

Suggest either:
- adding an actual `fixtures/user-event-missing/` folder with a `package.json` that has no `@testing-library/user-event` dependency (making the "missing dependency" case explicit and fixture-consistent with the rest), or
- renaming the case/adding a comment noting it intentionally targets a nonexistent path to exercise the "no manifest resolves" branch.

packages/analysis/src/jsts/rules/S9115/decorator.ts
```diff
@@ -0,0 +1,101 @@
...
+    validRange(normalizedVersionRange) === null
+  ) {
+    return false;
+  }
+
+  const minimumVersion = getProjectMinVersion(normalizedVersionRange);
+  return minimumVersion !== null && gte(minimumVersion, USER_EVENT_MINIMUM_VERSION);
+}
+
+function isUnboundedVersionRange(versionRange: string): boolean {
```
guillemsarda, line 96: `isUnboundedVersionRange` in `decorator.ts` (line 95) is a byte-for-byte duplicate of the same-named helper already in `S8927/rule.ts` (line 144):

```ts
function isUnboundedVersionRange(versionRange: string): boolean {
  return (
    versionRange === '' ||
    versionRange === '*' ||
    ['latest', 'x'].includes(versionRange.toLowerCase())
  );
}
```

Both files already import getProjectMinVersion from helpers/validate-version.ts, so that seems like the natural home for this too. Suggest moving isUnboundedVersionRange there, exporting it, and importing it from both S9115/decorator.ts and S8927/rule.ts instead of keeping two private copies.

To be clear, only the detection function is identical — the two rules interpret the result differently on purpose (S8927 treats an unbounded range as "assume supported"; S9115 treats it as "assume unsupported", per the conservative-by-default requirement in JS-2206), and that branching should stay local to each rule. Only the wildcard/latest/x pattern-matching itself would move.

Not a blocker for this PR — S8927 is untouched otherwise, so this is more of a follow-up cleanup than something to fold into S9115's scope. Flagging so it doesn't get forgotten (e.g. if someone later tweaks the wildcard patterns in one copy and not the other).

guillemsarda (approved):
Image

packages/analysis/src/jsts/rules/S9115/decorator.ts
nathsou: Leaving this as follow-up cleanup rather than changing this focused S9115 PR. The duplicated pattern is real, but extracting it would also modify S8927, and the two rules intentionally interpret unbounded ranges differently. I am keeping the cross-rule refactor out of scope here.

packages/analysis/src/jsts/rules/S9115/unit.test.ts
nathsou: Fixed in 00794949f by adding packages/analysis/src/jsts/rules/S9115/fixtures/user-event-missing/package.json with an empty dependency set. The focused S9115 suite passes 7/7 tests, and the full uc check passes 7/7.

PR #7706, opened 2026-08-04
JS-2207: Implement S9135: Avoid mutating nested properties of shallow clones

nathsou:
## Summary

- Implement S9135 for direct nested mutations of uniquely bound Lodash, Lodash-es, and Underscore.js `clone` results.
- Report the mutation as primary and the shallow clone call as a secondary location.
- Provide `structuredClone()` and intentional-shared-state `NOSONAR` suggestions, with library-specific messages for Lodash and Underscore.js.
- Add focused unit and custom ruling coverage.

## Links

- Jira: https://sonarsource.atlassian.net/browse/JS-2207
- RSPEC: https://github.com/SonarSource/rspec/pull/7677


its/sources/custom/jsts/S9135.js
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE, either express or implied.
+ * See the Sonar Source-Available License for more details.
+ *
+ * You should have received a copy of the Sonar Source-Available License
+ * along with this program; if not, see https://sonarsource.com/license/ssal/
+ */
+import lodash from 'lodash';
```
guillemsarda, line 17: Out of curiosity, why have you added a custom its?

guillemsarda, line 17: Out of curiosity, why have you added a custom its?

packages/analysis/src/jsts/rules/S9135/rule.ts
```diff
@@ -0,0 +1,236 @@
...
+    return;
+  }
+
+  const memberChain = getStaticMemberChain(mutatedNode);
+  if (memberChain === undefined) {
+    return;
+  }
+
+  const cloneCall = getCloneCall(context, memberChain.root);
+  if (cloneCall === undefined || getScopeBoundary(cloneCall) !== getScopeBoundary(mutation)) {
```
guillemsarda: The `getScopeBoundary` check in `checkMutation` requires the clone call and the mutation to live in the exact same function/program boundary:

```js
if (cloneCall === undefined || getScopeBoundary(cloneCall) !== getScopeBoundary(mutation)) {
  return;
}
```
This misses the mutation when it happens inside a nested callback, which is a pretty common way this exact bug shows up in practice:
```js
function process(user) {
  const copy = _.clone(user);
  [1, 2, 3].forEach(() => {
    copy.address.city = 'Geneva'; // not flagged today
  });
}
```

Worth leaving a suite case for this to document it as a known limitation (known FN).

packages/analysis/src/jsts/rules/S9135/unit.test.ts
```diff
@@ -0,0 +1,374 @@
...
+          code: `
+const copy = clone(user);
+copy.address.city = 'Geneva';
+`,
+        },
+        {
+          code: `
+import _ from 'lodash';
+const copy = _.clone(user);
+copy['address'].city = 'Geneva';
```
guillemsarda: Worth adding as well this case where you modify an inner's array element property:
```js
copy.items[0].name = 'x';
```

packages/analysis/src/jsts/rules/S9135/unit.test.ts
```diff
@@ -0,0 +1,374 @@
...
+`,
```
guillemsarda, line 76: `getCloneCall`'s uniqueness check (via `getUniqueWriteReference`) looks at whether a variable has exactly one write reference across its whole scope, without considering the order of writes relative to the mutation. So a `let` that gets reassigned later still slips past detection for an earlier, genuine mutation:

```js
let copy = _.clone(user);
copy.address.city = 'Geneva'; // real bug — copy still aliases user's nested object here
copy = other;
```
I verified this against the actual rule — it reports 0 issues here, even though the mutation happens while copy still points at the shallow clone.

Worth deciding explicitly: should this be covered (e.g. only count write references that occur before the mutation), or called out as a known limitation (i.e., adding the FN in the test suite)?

guillemsarda (approved):
Image

PR #7707, opened 2026-08-04
JS-2208: Implement rule S6544: async predicates should not be passed to synchronous collection methods

nathsou:
## Summary

- Extend S6544 with async-predicate detection for synchronous Lodash, `lodash-es`, and Underscore methods.
- Cover `every`, `filter`, `find`, `findIndex`, `some`, and `reject` while preserving upstream native-array diagnostics and excluding intentional callback APIs such as `map` and `forEach`.
- Add JavaScript and TypeScript fixtures for imports, CommonJS bindings, callback references, exclusions, and deduplication.

## Links

- Jira: https://sonarsource.atlassian.net/browse/JS-2208
- RSPEC PR: https://github.com/SonarSource/rspec/pull/7679

## Verification

- Focused S6544 comment-based and unit tests pass.
- `uc check` reaches the implementation checks; repository-wide generated parser/gRPC artifacts are absent in this checkout and the Sonar issues check is expected to pass once this PR exists.


packages/analysis/src/jsts/rules/S6544/rule.ts
```diff
@@ -85,14 +118,164 @@ function isLazyInitPattern(
...
+
+function isSupportedLibraryPredicateCall(
+  context: Rule.RuleContext,
+  call: estree.CallExpression,
+): boolean {
+  const methodName = getSyntacticMethodName(call.callee);
+  if (methodName === undefined || !isDirectCallee(context, call.callee)) {
+    return false;
+  }
+  const fqn = getFullyQualifiedName(context, call)?.replaceAll('/', '.');
```
guillemsarda, line 174: ## Nit: alias-exclusion gate diverges from `getFullyQualifiedName` without a comment explaining why

`getFullyQualifiedName` already resolves arbitrary-depth variable aliasing (e.g. it would resolve `const lodashAlias = _; lodashAlias.filter(...)` to `lodash.filter` on its own). `isDirectCallee`/`isDirectReceiver`/`isDirectRequireReference` exist specifically to be *stricter* than that (single-hop only), which is what the "exclude aliases" fixtures need — but nothing in the code says so. Worth a one-line comment, otherwise a future reader could reasonably "simplify" this away and silently reintroduce the alias false positives V1 excludes.

packages/analysis/src/jsts/rules/S6544/rule.ts
```diff
@@ -85,14 +118,164 @@ function isLazyInitPattern(
...
+  if (isRequire(initializer)) {
+    return initializer;
+  }
+  if (initializer.type === 'MemberExpression' && isRequire(initializer.object)) {
+    return initializer.object;
+  }
+  return null;
+}
+
+function isUnshadowedRequire(
```
guillemsarda: ## Nit: `isUnshadowedRequire` duplicates a check buried inside `getUnshadowedRequireModuleName`

`isUnshadowedRequire` (rule.ts:257) re-derives the same scope-walk-and-check-defs logic that `getUnshadowedRequireModuleName` (`helpers/module.ts:170`) uses internally to decide shadowing. It can't simply be replaced by calling that helper directly, though — `getUnshadowedRequireModuleName` conflates three things behind one `undefined` return (not a require call / shadowed `require` / argument isn't a string literal), so treating its `undefined` as "shadowed" would change behavior on cases like `require(someVariable).filter(...)`.

Could we pull just the shadow check out into its own exported primitive, e.g.:

```ts
// module.ts
export function isRequireShadowed(sourceCode: SourceCode, requireCall: estree.CallExpression): boolean {
  const requireVariable = getVariableFromScope(sourceCode.getScope(requireCall), 'require');
  return !!requireVariable?.defs.length;
}
```
and have both getUnshadowedRequireModuleName and the rule's isUnshadowedRequire call that? That shares the actual check in one place instead of two copies that could drift.

guillemsarda (approved):
Image

packages/analysis/src/jsts/rules/S6544/rule.ts
nathsou: Added the one-line rationale above the direct-callee gate: `getFullyQualifiedName()` resolves aliases, while this check intentionally accepts only direct imports and `require` calls to avoid the excluded-alias false positives.

packages/analysis/src/jsts/rules/S6544/rule.ts
nathsou: Extracted `isRequireShadowed` into `helpers/module.ts` and use it from both `getUnshadowedRequireModuleName` and S6544. Added helper coverage for the global `require` and a parameter-shadowed `require`; the focused tests and TypeScript build pass.

PR #7708, opened 2026-08-04
JS-2209 Support AWS CDK assertions in S8784

francois-mora-sonarsource:
### S8784 — AWS CDK (original scope)
- classify AWS CDK assertions as script-capable for S8784
- retain S2699 TypeScript assignment fallback
- cover misplaced, standalone, and non-assertion AWS CDK calls

### Assertion helper refactor
Supporting the classification meant replacing the `isAssertion` / `isAssertionIncludingUnclassified`
pairs with a single catalog, so a library is defined once and rules opt in by profile.
- move the testing helpers under `helpers/testing/`
- add `helpers/testing/assertion-frameworks.ts`: one catalog entry per framework (module specifiers
  + AST and type-aware detectors); rules declare which keys they react to
- `allAssertionFrameworks()` derives the profile for S2699 and S8785 from the catalog, so a
  framework added later reaches them instead of being silently invisible
- `AssertionExecutionProfile` is total: a new framework must be classified in S8784 or the build
  fails, rather than being quietly dropped from the rule

### S2699 — beyond AWS CDK
Two changes were needed for the refactor to keep the vitest ruling baseline, and both affect rules
other than S8784. Flagging them explicitly since they are not CDK work:
- `vitest/node` counts as an assertion source, so S2699 / S8784 / S8785 now analyze files that only
  import it
- recognize `node:test` context assertions (`t.assert.*`, Node 22+). They were invisible to every
  detector — the file imports `node:test` rather than `node:assert`, and the receiver is a callback
  parameter, not a module binding — so S2699 reported "add at least one assertion" on tests that
  assert only that way. This also removes the same false positive for `node:test` + chai and
  `node:test` + vitest files, which predates this PR.

Baseline: drops `vitest:test/node-runner/test/cli.test.js:4`, which was that false positive.

## Validation
- focused S8784, S2699 and S8785 tests, plus new `assertion-frameworks` and `node-test` helper tests
- differential run of S2699 / S8784 / S8785 against a `master` worktree over a framework ×
  placement corpus: identical on the AST path; the only type-aware differences are the `t.assert.*`
  cases above
- npm run bbf
- Prettier check
- git diff --check


packages/analysis/src/jsts/rules/helpers/testing/node-test.ts
```diff
@@ -0,0 +1,187 @@
...
+/**
+ * Whether `callee` names a `node:test` entry point. The source file is already known to import
+ * `node:test`, so matching the callee name is enough to separate `test(...)` / `it(...)` from an
+ * unrelated call that happens to take a callback.
+ */
+function isNodeTestCalleeTS(callee: ts.Expression): boolean {
+  if (ts.isIdentifier(callee)) {
+    return TEST_FUNCTION_NAMES.has(callee.text);
+  }
+  return ts.isPropertyAccessExpression(callee) && TEST_FUNCTION_NAMES.has(callee.name.text);
```
nathsou: `test.skip` and `test.only` receive a TestContext, and S2699 already extracts them as test cases. In typed analysis their callbacks reach this path, but `skip`/`only` are rejected here, so `t.assert.*` does not count and S2699 raises a false positive whenever the file is otherwise enabled (for example by `vitest/node`). Please include these modifiers and add a typed S2699 regression.

packages/analysis/src/jsts/rules/helpers/testing/node-test.ts
```diff
@@ -0,0 +1,187 @@
...
+  const ancestors = context.sourceCode.getAncestors(
+    node as Parameters<typeof context.sourceCode.getAncestors>[0],
+  );
+  for (let i = ancestors.length - 1; i >= 0; i--) {
+    const enclosing = ancestors[i];
+    if (!isFunctionNode(enclosing)) {
+      continue;
+    }
+    const [first] = enclosing.params;
+    if (first?.type !== 'Identifier' || first.name !== name) {
```
nathsou: This only considers function parameters, not lexical bindings. With `test("x", t => { (() => { const t = { assert: { ok() {} } }; t.assert.ok(true); })(); })`, the inner arrow has no first parameter, so this scan falls through to the outer test context and treats the arbitrary local call as a node:test assertion. S2699 then misses the assertionless test. Resolve the actual scope binding (in both AST paths) and cover this shadowing case.

nathsou (approved):
Approved with two non-blocking follow-ups documented inline.

packages/analysis/src/jsts/rules/helpers/testing/node-test.ts
francois-mora-sonarsource: Addressed in d5cb417c1: the type-aware detector now follows `test.skip` / `test.only` to the node:test callback. Added typed S2699 coverage for `test.only`, plus AST helper coverage for both modifiers.

packages/analysis/src/jsts/rules/helpers/testing/node-test.ts
francois-mora-sonarsource: Addressed in d5cb417c1: both AST and TypeScript paths now compare the resolved receiver binding/symbol with the callback parameter, so a nested shadowed `t` cannot be treated as TestContext. Added helper and typed S2699 regressions.

PR #7712, opened 2026-08-05
JS-2217: Fix S8980 false positive for renderHook result actions

nathsou:
## Summary

- Treat calls through `renderHook(...).result.current` as tested-hook actions rather than Testing Library utilities.
- Keep reports for genuine `renderHook` utilities such as `rerender()`.
- Cover the upstream conditional regression, a mixed callback, and the `rerender()` boundary.

## Root cause

The upstream non-strict rule does not classify conditional hook-result actions as non-Testing-Library calls. It consequently reports a necessary `act()` before the hook state is read. SonarJS now recognizes the fully-qualified renderHook result path and suppresses that report, including mixed callbacks.

## Links

- Jira: https://sonarsource.atlassian.net/browse/JS-2217

## Verification

- `./run-node node_modules/tsx/dist/cli.mjs --test packages/analysis/src/jsts/rules/S8980/cb.test.ts`
- `npm run bbf`
- `uc check --base origin/master` (type check, unit tests, and lint passed; pre-PR issue check was pending)
- `./run-node node_modules/tsx/dist/cli.mjs --tsconfig packages/tsconfig.test.json --test packages/ruling/projects/searchkit.ruling.test.ts`

### DIFF
--- a/packages/analysis/src/jsts/rules/S8980/cb.fixture.tsx
+++ b/packages/analysis/src/jsts/rules/S8980/cb.fixture.tsx
@@ -1,4 +1,4 @@
-import { act, fireEvent, render, screen, waitFor } from '@testing-library/react';
+import { act, fireEvent, render, renderHook, screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 
 function Form() {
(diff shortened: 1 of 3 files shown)


guillemsarda (approved):
Image

PR #7717, opened 2026-08-07
JS-2174: Fix S9114: Handle direct useRef initializers

nathsou:
## Summary

- Suppress S9114 for direct React useRef initializers of Lodash or Underscore debounce/throttle wrappers.
- Preserve reporting for wrapper calls, aliases, nested expressions, and non-first arguments.
- Pin RSPEC PR #7711 until its specification update merges.

## Links

- Jira: https://sonarsource.atlassian.net/browse/JS-2174
- RSPEC: https://github.com/SonarSource/rspec/pull/7711

## Verification

- Focused S9114 comment-based tests
- TypeScript check
- custom-jsts ruling project


packages/analysis/src/jsts/rules/S9114/rule.ts
```diff
@@ -223,7 +224,64 @@ function isReactUseRef(context: Rule.RuleContext, refIdentifier: estree.Identifi
...
+  const { callee } = call;
+  if (callee.type === 'Identifier') {
+    return isReactUseRefImport(context, callee);
+  }
+  return (
+    callee.type === 'MemberExpression' &&
+    !callee.computed &&
+    isIdentifier(callee.property, 'useRef') &&
+    isReactModuleReference(context, callee.object)
+  );
```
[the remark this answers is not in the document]
nathsou: Fixed in 712c08b40. Direct initializers now reuse the FQN-based React useRef matcher, so renamed imports and stable aliases are handled consistently with lazy initialization; the fixtures cover both forms.

packages/analysis/src/jsts/rules/S9114/rule.ts
```diff
@@ -223,7 +224,20 @@ function isReactUseRef(context: Rule.RuleContext, refIdentifier: estree.Identifi
...
-  return init?.type === 'CallExpression' && getFullyQualifiedName(context, init) === 'react.useRef';
+  return init?.type === 'CallExpression' && isReactUseRefCall(context, init);
+}
+
+function isDirectRefInitializer(context: Rule.RuleContext, call: estree.CallExpression): boolean {
+  const parent = getNodeParent(call);
+  return (
+    parent.type === 'CallExpression' &&
+    parent.arguments[0] === call &&
+    isReactUseRefCall(context, parent)
```
francois-mora-sonarsource: Could we reuse `isReactUseRefCall(context, parent)` here instead of introducing a second React `useRef` resolver?

The existing lazy-initialization path uses `getFullyQualifiedName`, which already resolves React imports and aliases. The new resolver deliberately does not recognize `import { useRef as aliasedUseRef } from 'react'`, so `aliasedUseRef(debounce(...))` is still reported even though it is the same safe direct-ref initialization. Reusing the shared helper would keep both paths aligned, remove the duplicated resolution logic, and avoid that false positive.

francois-mora-sonarsource (changes requested):
Please unify the React `useRef` resolution so the existing lazy-initialization and new direct-initializer exemptions recognize the same valid React bindings.

packages/analysis/src/jsts/rules/S9114/rule.ts
```diff
@@ -223,7 +224,20 @@ function isReactUseRef(context: Rule.RuleContext, refIdentifier: estree.Identifi
...
   const init = variable.defs[0].node.init;
-  return init?.type === 'CallExpression' && getFullyQualifiedName(context, init) === 'react.useRef';
+  return init?.type === 'CallExpression' && isReactUseRefCall(context, init);
+}
+
+function isDirectRefInitializer(context: Rule.RuleContext, call: estree.CallExpression): boolean {
+  const parent = getNodeParent(call);
+  return (
+    parent.type === 'CallExpression' &&
+    parent.arguments[0] === call &&
```
francois-mora-sonarsource: Optional: could we unwrap TypeScript-only expressions before comparing the first argument? For example, `useRef(debounce(fn, 300) as any)` is equivalent for this rule, but the direct identity check leaves it reported because the argument is a `TSAsExpression`.

This would align the new `useRef` exemption with the rule's existing ancestor-based memo-hook handling, which already tolerates syntactic noise between the reported call and the hook. `unwrapTypeScriptExpression(parent.arguments[0]) === call` should preserve the intended distinction from runtime wrappers such as `useRef(wrap(debounce(...)))`.

francois-mora-sonarsource (commented):
Optional scope-consistency suggestion for TypeScript-only wrappers around direct `useRef` initializers.

packages/analysis/src/jsts/rules/S9114/rule.ts
nathsou: Already addressed in 712c08b40: direct initializers reuse isReactUseRefCall, so FQN resolution covers aliased imports and stable aliases consistently with lazy initialization.

packages/analysis/src/jsts/rules/S9114/rule.ts
nathsou: Fixed in 096fc1b57c. Direct initializers now tolerate transparent TypeScript wrappers and require the unwrapped first argument to be the debounce/throttle call; the new TypeScript fixture covers an `as any` assertion while runtime wrappers remain reported.

francois-mora-sonarsource (approved):
Approving — verified the fixes behave as intended, with no new false negatives.

nathsou:
> Approving — verified the fixes behave as intended, with no new false negatives.
> 
> One non-blocking follow-up, **pre-existing on master**, not introduced here:
> 
> `recreatedPerRender` (`packages/analysis/src/jsts/rules/S9114/rule.ts:53`) still ends with _"Move it outside the component or hook, or wrap it in `useMemo`."_ `useRef` has been an accepted remediation since the rule landed (null-guard form), and this PR adds the direct-initializer form — neither is offered in the message. `recreatedInClassRender` is fine as-is.

Thanks! While a debounce call wrapped in a `useRef` call is compliant, I think `useMemo` should be preferred so I'm not updating the issue message to mention useRef.

francois-mora-sonarsource:
> > Approving — verified the fixes behave as intended, with no new false negatives.
> > One non-blocking follow-up, **pre-existing on master**, not introduced here:
> > `recreatedPerRender` (`packages/analysis/src/jsts/rules/S9114/rule.ts:53`) still ends with _"Move it outside the component or hook, or wrap it in `useMemo`."_ `useRef` has been an accepted remediation since the rule landed (null-guard form), and this PR adds the direct-initializer form — neither is offered in the message. `recreatedInClassRender` is fine as-is.
> 
> Thanks! While a debounce call wrapped in a `useRef` call is compliant, I think `useMemo` should be preferred so I'm not updating the issue message to mention useRef.

I agree. Claude Opus 5 is just confused.

PR #7722, opened 2026-08-07
JS-2222: Implement S9144: Native APIs should be preferred over jQuery utility methods

nathsou:
## Summary
- Implements S9144 for replacing selected jQuery utility methods with native APIs.
- Covers deprecated methods with dedicated messages and inArray() with a dependency-reduction message.
- Adds unit fixtures and ruling baselines, including TypeScript import = require() support.

## Links
- Jira: https://sonarsource.atlassian.net/browse/JS-2222
- RSPEC: https://github.com/SonarSource/rspec/pull/7706


packages/analysis/src/jsts/rules/S9144/rule.ts
```diff
@@ -0,0 +1,179 @@
...
+}
+
+function isJQueryReceiver(
+  identifier: estree.Identifier,
+  context: Rule.RuleContext,
+  jqueryReceivers: Map<Scope.Variable, boolean>,
+): boolean {
+  const variable = getVariableFromName(context, identifier.name, identifier);
+  if (variable === undefined || variable.defs.length === 0) {
+    return JQUERY_IDENTIFIERS.has(identifier.name);
```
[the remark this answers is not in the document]
nathsou: Unresolved global `$` remains intentional here. SonarJS analyzes each file independently and cannot determine whether `$` came from a CDN/vendored jQuery script tag or another library. The matched method set is a strong signal: `parseJSON`, `inArray`, and `isArray` are jQuery-specific; `trim` and `now` have more collision potential. Local declarations, imports, and reassigned bindings are excluded, and S1848 follows the same global `$` convention. The RSPEC will document that an unresolved global `$` is assumed to be jQuery.\n\nI also verified the import forms in local npm projects. jQuery 4.0.0 exports default, `$`, and `jQuery` from both `jquery` and `jquery/slim`; with a DOM, `import $ from "jquery"`, `import { $ } from "jquery"`, and `import { jQuery } from "jquery"` resolve to the same function. jQuery 3.7.1 is CommonJS: default import interop works in Node ESM, but named `$` is not an available ESM export. S9144 now recognizes both default and named `$`/`jQuery` imports.

packages/analysis/src/jsts/rules/S9144/cb.fixture.ts
```diff
@@ -0,0 +1,90 @@
...
+
+function shadowedNames(jQuery: { trim: (value: string) => string }): void {
+  jQuery.trim(input);
+}
+
+function shadowedDollar($: { trim: (value: string) => string }): void {
+  $.trim(input);
+}
+
+jQuery['trim'](input);
```
guillemsarda, line 95: Worth flagging as a known FN the following:
`const { trim } = jQuery;` (it's true that detaches the `this` binding but still worth reporting it as a Compliant code

packages/analysis/src/jsts/rules/S9144/cb.fixture.ts
```diff
@@ -0,0 +1,90 @@
...
+jQuery?.trim(input);
+jQuery.trim.call(null, input);
+require('jquery').trim(input);
```
guillemsarda, line 98: `isDirectJQueryModuleBinding` has a couple of untested branches worth closing before merge:

1. **Named import via `default as`** — `isJQueryImportSpecifier`'s `isDefaultImportSpecifier` branch is only exercised through the type-only case (`import { type default as typeOnlySpecifier } from 'jquery'`), which is rejected by the FQN type-only filter, not by this function. There's no positive fixture proving:
   ```ts
   import { default as jq } from 'jquery';
   jq.trim(x); // should be Noncompliant

2. `import X = require(...)` rejection sibling — import-equals.fixture.ts only covers the TSExternalModuleReference case. The TSQualifiedName rejection path (line ~168-171) is never hit in either direction:
```js
import s3 = SomeNamespace.Member;
s3.trim(x); // should NOT be Noncompliant
```
Lower priority, but also missing:
- A redeclared-name case for the `variable.defs.length !== 1` guard (e.g. a parameter shadowed by var).
- A require-shape mismatch, e.g. `const jq = require('jquery')()`;

Could you add fixture cases for at least #1 and #2?

packages/analysis/src/jsts/rules/helpers/module.ts
```diff
@@ -421,6 +421,16 @@ function checkFqnFromImport(
   if (definition.type === 'ImportBinding') {
     const specifier = definition.node;
     const importDeclaration = definition.parent;
+    const importDeclarationNode = importDeclaration as TSESTree.Node;
```
guillemsarda: ## `packages/analysis/src/jsts/rules/helpers/module.ts` (checkFqnFromImport)

This change makes `getFullyQualifiedName` return `null` for type-only imports (`import type X`, `import { type X }`, `import type X = require(...)`). It's a shared helper used by ~80 other rules (S5527, S4830, S6321, the AWS-CDK helpers, etc.), but `module.test.ts` only tests `getCurrentFileModuleReferences` — there's no direct test for `checkFqnFromImport`/`getFullyQualifiedName` covering the new type-only branches. Right now the only coverage is indirect, via S9144's own fixtures (`typeOnlyJq.trim(input)` staying unflagged), which pins the *symptom* in one rule rather than the *behavior* of the shared helper itself.

Could you add a focused unit test in `module.test.ts` for `getFullyQualifiedName`/`checkFqnFromImport` covering the three type-only-import shapes directly? That way a future change to this helper can't silently regress it without every downstream rule's fixtures happening to catch it.

packages/analysis/src/jsts/rules/S9144/rule.ts
```diff
@@ -0,0 +1,218 @@
...
+
+function isJQueryModuleReference(
+  context: Rule.RuleContext,
+  identifier: estree.Identifier,
+): boolean {
+  const fullyQualifiedName = getFullyQualifiedName(context, identifier) ?? '';
+  return JQUERY_MODULES.has(fullyQualifiedName) || JQUERY_NAMED_IMPORTS.has(fullyQualifiedName);
+}
+
+function isDirectJQueryModuleBinding(variable: Scope.Variable, context: Rule.RuleContext): boolean {
```
guillemsarda, line 148: ## `packages/analysis/src/jsts/rules/S9144/rule.ts` (isDirectJQueryModuleBinding)

`isDirectJQueryModuleBinding` only accepts a variable whose *own* definition is directly an import specifier or a direct `require('jquery')` call. A plain alias of a genuine jQuery binding isn't detected:

```ts
const jq = require('jquery');
const alias = jq;
alias.trim(input); // NOT flagged, even though `alias` really is jQuery
``` 
If this an intentional scope limit, it would be worth to add a test case to express the intention of the rule.

guillemsarda (approved):
Image

packages/analysis/src/jsts/rules/S9144/cb.fixture.ts
nathsou: Addressed in commit 6e3e18f0fc: added `const { trim } = jQuery; trim(input);` to the compliant fixture to document this known false negative. No rule behavior changed because detached method calls are outside the current direct member-call scope; the focused S9144 suite passes.

packages/analysis/src/jsts/rules/S9144/cb.fixture.ts
nathsou: Addressed in commit 6e3e18f0fc: added positive coverage for `import { default as jqueryDefaultAlias } from 'jquery'` and compliant coverage for the `TSQualifiedName` import-equals rejection path. The focused S9144 suite passes.

packages/analysis/src/jsts/rules/helpers/module.ts
nathsou: Addressed in commit 6e3e18f0fc: added a direct `getFullyQualifiedName()` unit test covering `import type`, type-only named imports, and type-only import-equals. The focused module-helper suite passes.

packages/analysis/src/jsts/rules/S9144/rule.ts
nathsou: Addressed in commit 6e3e18f0fc: added a compliant fixture for an alias of a direct jQuery require, documenting that S9144 intentionally recognizes only direct import/require bindings. No behavior expansion was made; the focused S9144 suite passes.

PR #7734, opened 2026-08-10
JS-2232: Implement S9153: Testing Library disappearance waits should use non-throwing queries

nathsou:
## Summary

- Add original S9153 detection for unsafe Testing Library disappearance waits.
- Report throwing `getBy*` callback queries and promise-returning `findBy*` queries in the approved narrow import scope.
- Add unit and custom ruling coverage, including the existing S9026 overlap.

## Links

- Jira: https://sonarsource.atlassian.net/browse/JS-2232
- RSPEC: https://github.com/SonarSource/rspec/pull/7736


tools/generate-eslint-meta.ts
```diff
@@ -69,7 +69,10 @@ export async function generateMetaForRule(
       ___DESCRIPTION___: ruleRspecMeta.title.replace(/'/g, "\\'"),
       ___RECOMMENDED___: sonarWayProfile.ruleKeys.includes(sonarKey),
       ___TYPE_CHECKING___: `${tags.includes('type-dependent')}`,
-      ___FIXABLE___: ruleRspecMeta.quickfix === 'covered' ? "'code'" : undefined,
+      ___FIXABLE___:
```
guillemsarda: I think this is out of the scope of the PR. We should talk with Victor before bringing this change

packages/analysis/src/jsts/rules/S9153/rule.ts
```diff
@@ -0,0 +1,147 @@
...
+
+  const [statement] = node.body.body;
+  if (statement.type !== 'ReturnStatement' || !statement.argument) {
+    return null;
+  }
+
+  return getQuery(context, statement.argument);
+}
+
+function getQuery(context: Rule.RuleContext, node: estree.Node): Query | null {
```
guillemsarda, line 113: **Possible false negative:** `getQuery` checks `node.type !== 'CallExpression'` without unwrapping TypeScript assertion nodes first, so a query wrapped in `!` or `as X` (e.g. `screen.getByRole('alert')!`) won't be recognized and the rule stays silent. Sibling rule S9027's `getQueryCall` calls `unwrapTypeScriptExpression()` before the same check — could this do the same?

packages/analysis/src/jsts/rules/helpers/testing-library.ts
```diff
@@ -15,6 +15,84 @@
...
+    options.allowNamespaceImport &&
+    node.type === 'MemberExpression' &&
+    !node.computed &&
+    isIdentifier(node.object) &&
+    isIdentifier(node.property, importedName) &&
+    isDirectNamespaceImport(context, node.object, options.allowSubpathImport)
+  );
+}
+
+function isDirectNamedImport(
```
guillemsarda, line 55: `isDirectNamedImport`/`isDirectNamespaceImport` re-implement import-binding resolution that `getFullyQualifiedName` (`helpers/module.ts`) already provides, and this file already relies on it a few lines below for `waitFor` detection. Reusing it here would avoid two parallel import-resolution mechanisms drifting  
  apart over time.

packages/analysis/src/jsts/rules/S9153/rule.ts
```diff
@@ -0,0 +1,147 @@
...
+
+type ImportedName = 'screen' | 'waitForElementToBeRemoved';
+type QueryKind = 'get' | 'find';
+
+interface Query {
+  kind: QueryKind;
+  method: estree.Identifier;
+}
+
+export const rule: Rule.RuleModule = {
```
guillemsarda, line 44: This `CallExpression` listener has no `importsModule()`-style early exit, so it runs its checks (including scope resolution) on every call expression in every linted file, even files with no Testing Library import at all. Worth gating on `importsModule()` first, as other rules do?

packages/analysis/src/jsts/rules/S9153/rule.ts
```diff
@@ -0,0 +1,147 @@
...
+    return null;
+  }
+
+  return getQuery(context, statement.argument);
+}
+
+function getQuery(context: Rule.RuleContext, node: estree.Node): Query | null {
+  if (
+    node.type !== 'CallExpression' ||
+    node.callee.type !== 'MemberExpression' ||
```
guillemsarda: This re-checks `MemberExpression && !computed && Identifier property` manually, which `helpers/ast.ts`'s `isDotNotation()` already encapsulates. The same check is repeated again in S9027's `getQueryMethod` — using the shared guard would keep this in one place instead of three.

packages/analysis/src/jsts/rules/helpers/testing-library.ts
```diff
@@ -15,6 +15,84 @@
...
+  '@testing-library/angular',
+  '@testing-library/svelte',
+] as const;
+
+interface DirectTestingLibraryBindingOptions {
+  allowNamespaceImport: boolean;
+  allowSubpathImport: boolean;
+}
+
+export function isDirectTestingLibraryBinding(
```
guillemsarda, line 35: **Suggestion (perhaps as a follow-up task):**
`isDirectTestingLibraryBinding`'s two independent flags (`allowNamespaceImport`, `allowSubpathImport`) form a 4-state matrix, but callers only ever pass `{false, false}` (S9027) or `{true, true}` (S9153). Might read more clearly as two composable predicates than one function with two booleans that are always set together.

guillemsarda (changes requested):
The False negative case and the changes in the quickfix validation are blocking the PR .

tools/generate-eslint-meta.ts
nathsou: Yes definitely, I will revert

nathsou: Thanks, agreed. Removed the updates to tools/generate-eslint-meta.ts, tools/helpers.ts, and tools/validate-quickfix.ts in a73f66e3be. S9153 now pins the rebased RSPEC commit with quickfix=covered, so it uses the existing tooling unchanged. npm run validate-quickfix passes.

packages/analysis/src/jsts/rules/S9153/rule.ts
nathsou: Addressed in a73f66e3be. getQuery now unwraps TypeScript expression wrappers. Added fixture coverage for a callback non-null assertion and a direct as HTMLElement asynchronous query. Focused S9153 tests pass.

packages/analysis/src/jsts/rules/helpers/testing-library.ts
nathsou: Kept as-is. S9153 intentionally reports direct ESM imports only. getFullyQualifiedName also follows aliases and require bindings, which would expand the rule scope; the fixture deliberately covers both as no-issue cases.

packages/analysis/src/jsts/rules/S9153/rule.ts
nathsou: Addressed in a73f66e3be. importsTestingLibrary now gates the listener using cached file imports and retains subpath support, including @testing-library/react/pure. Focused S9153 tests pass.

packages/analysis/src/jsts/rules/S9153/rule.ts
nathsou: Addressed in a73f66e3be. The S9153 and S9027 query checks now use isDotNotation. Both focused test suites and the plugin type check pass.

packages/analysis/src/jsts/rules/helpers/testing-library.ts
nathsou: Deferred for now. The two flags express independent policy choices: S9027 accepts only direct root imports, while S9153 also accepts namespace and subpath imports. A new abstraction would add API shape without a third use case.

guillemsarda (approved):
Careful with the rspec.sha before merging.

Image

nathsou:
> Careful with the rspec.sha before merging.
> 
> We have a check now :)

PR #7736, opened 2026-08-10
JS-2233: Implement S9162: use retryable Cypress assertions

nathsou:
## Summary

Implement S9162 to recommend retryable Cypress `.should()` callbacks when `.then()` callbacks contain only assertions and retry-safe local calculations.

## Links

- Jira: https://sonarsource.atlassian.net/browse/JS-2233
- RSPEC: https://github.com/SonarSource/rspec/pull/7746

## Verification

- `uc check`: type-check, unit tests, ruling, and SonarJS lint pass; Sonar issues awaits this PR
- `npm run validate-quickfix`
- focused S9162 unit tests


packages/analysis/src/jsts/rules/S9162/rule.ts
```diff
@@ -0,0 +1,331 @@
...
+};
+
+type RetrySafeKind = 'value' | 'yielded' | 'string' | 'record';
+
+const STRING_METHODS = new Set(['trim', 'toLowerCase', 'toUpperCase']);
+
+export const rule: Rule.RuleModule = {
+  meta: generateMeta(meta, { messages, hasSuggestions: true }),
+  create(context: Rule.RuleContext): Rule.RuleListener {
+    if (!isTestRelatedFile(context.filename, context.settings?.testFileExtensions as string[])) {
```
guillemsarda, line 72: Isn't this check redundant? In the rule rspec we already set the scope to test so the rule should only be running in there.

packages/analysis/src/jsts/rules/S9162/rule.ts
```diff
@@ -0,0 +1,331 @@
...
+
+function retrySafeCallKind(
+  call: estree.CallExpression,
+  bindings: ReadonlyMap<string, RetrySafeKind>,
+): RetrySafeKind | null {
+  if (!isMethodCall(call) || call.arguments.length !== 0) {
+    return null;
+  }
+  const receiverKind = retrySafeKind(call.callee.object, bindings);
+  if (call.callee.property.name === 'text' && receiverKind === 'yielded') {
```
guillemsarda, line 472: **False positive: `.then()` → `.should()` suggestion is unsafe when the yielded value isn't a real jQuery element**

`retrySafeCallKind` (rule.ts:316-331) treats any 0-arg call named `text` on a `'yielded'`-kind receiver as pure, based on the method name alone — it never verifies that the receiver actually came from a Cypress DOM command (`cy.get`/`cy.contains`, etc.). Since `isGlobalCyChain` (rule.ts:99-106) accepts *any* `cy`-rooted chain, including `cy.wrap(...)`, a `.then()` callback that only "looks like" an assertion gets flagged and auto-fixed to `.should()`, even when the wrapped value has a side-effecting `text` method:

```js
let n = 0;
cy.wrap({
  text: () => {
    n++;
    sideEffectfulApiCall(n);
    return n > 2 ? 'Ready' : 'Loading';
  },
}).then(obj => {
  expect(obj.text()).to.equal('Ready'); // Noncompliant — but converting to .should() is NOT safe here
});
```

`.should()` re-invokes the callback on every retry, so the suggested fix turns a one-shot side effect into a repeated one. Since the rule can't statically prove `.text()` resolves to jQuery's side-effect-free getter, this specific heuristic (bare method-name matching on `'yielded'` kind) is unsound as a basis for an automatic code-fix.

Suggest either narrowing `isGlobalCyChain`/the chain-root check to only recognize chains that actually originate from a Cypress query command (`cy.get`, `cy.find`, `cy.contains`, etc.) rather than any `cy`-rooted chain, or dropping the auto-fix (keep it as a plain diagnostic) until that distinction can be made reliably.

packages/analysis/src/jsts/rules/S9162/rule.ts
guillemsarda: Also, since the rspec doesn't have Cypress as required dependency, it would be interesting to add the check: `importsOrDependsOnModule` to prevent the rule to run on non-Cypress projects.

packages/analysis/src/jsts/rules/S9162/rule.ts
```diff
@@ -0,0 +1,331 @@
...
+
+function hasTypeArguments(call: estree.CallExpression): boolean {
+  const typescriptCall = call as estree.CallExpression & {
+    typeArguments?: unknown;
+    typeParameters?: unknown;
+  };
+  return typescriptCall.typeArguments != null || typescriptCall.typeParameters != null;
+}
+
+function isGlobalCyChain(context: Rule.RuleContext, node: estree.Node): boolean {
```
guillemsarda: **Reuse: the TS-cast unwrap this rule needed should be fixed in the shared `chainStartsWithCy` helper, not duplicated here**

`findChainRoot` (rule.ts:108-120) calls `unwrapTypeScriptExpression` at every step so it can see through TS casts like `(cy.get('input') as Cypress.Chainable<...>).then(...)` — and the test suite proves this is load-bearing (unit.test.ts:350-359 only passes because of this unwrap).

The existing shared helper `chainStartsWithCy` (`helpers/testing/cypress.ts:57-71`), used directly by S2925, S8783, and S8959, does **not** do this unwrap — it only recognizes `Identifier`/`MemberExpression`/`CallExpression`/`ChainExpression`. That means those three already-shipped rules still miss a `cy`-chain hidden behind a TS cast today, e.g.:

```ts
(cy.get('input') as Cypress.Chainable<JQuery>).click({ force: true }); // evades S8783's forced-option check
```

Rather than adding a second, private copy of "walk to the cy root" in this new rule, could we fix `chainStartsWithCy` itself to unwrap TS expressions (mirroring what `findChainRoot` does here)? That would fix the latent gap in S2925/S8783/S8959 as a side effect, and let this rule call the shared helper instead of maintaining its own chain-walker that will now drift from it.

packages/analysis/src/jsts/rules/S9162/rule.ts
```diff
@@ -0,0 +1,331 @@
...
+    if (statement.type === 'VariableDeclaration') {
+      if (!acceptRetrySafeDeclaration(statement, bindings)) {
+        return false;
+      }
+      continue;
+    }
+    if (statement.type !== 'ExpressionStatement') {
+      return false;
+    }
+    const assertion = extractChaiAssertion(context, statement.expression, true);
```
guillemsarda, line 252: **False negative: any unsupported chai matcher rejects the entire callback, not just that statement**

`isAssertionOnlyCallback` calls `extractChaiAssertion(context, statement.expression, true)` (rule.ts:151) and bails out of the *whole* callback (`return false`) as soon as one statement doesn't parse as a recognized assertion. But `extractChaiAssertion`'s matcher vocabulary is limited — `getChaiComparison` (assertions-chai.ts:463-472) only recognizes `eql`/`eqls`/`equal`/`equals`/`eq`, and `getChaiPropertyPredicate` only covers a fixed predicate set. Any other common chai matcher falls through as `null`, taking down detection for the entire callback:

```js
cy.get('[data-cy=status]').then($status => {
  expect($status.text()).to.include('Read'); // Noncompliant, but never flagged
});
```

This is a pure, side-effect-free assertion-only `.then()` — exactly the flaky pattern S9162 is meant to catch — but `.include` (and likewise `.above`, `.match`, `.have.length`, `.contain`, etc.) isn't recognized, so `extractChaiAssertion` returns `null`, and the rule silently skips the whole callback instead of just not being able to reason about that one statement.

Worth either extending `getChaiComparison`/`getChaiPropertyPredicate` with the common missing matchers, or at minimum treating an unrecognized-but-still-a-chai-call statement as "unsafe" (reject the callback) only when it can't be shown to be side-effect-free, rather than conflating "we don't understand this matcher" with "this callback might have side effects" — the current behavior undercounts real flakiness whenever a test uses a matcher slightly outside the recognized set.

packages/analysis/src/jsts/rules/S9162/rule.ts
```diff
@@ -0,0 +1,331 @@
...
+function isAssertionOnlyCallback(
+  context: Rule.RuleContext,
+  callback: estree.CallExpression['arguments'][number] | undefined,
+): boolean {
+  if (
+    !callback ||
+    (callback.type !== 'ArrowFunctionExpression' && callback.type !== 'FunctionExpression') ||
+    callback.async ||
+    callback.generator ||
+    callback.body.type !== 'BlockStatement' ||
```
guillemsarda, line 223: **TLDR:** Why are callbacks without block statements directly discarded?


**False negative: expression-bodied arrow callbacks are skipped structurally, not for a real safety reason**

`isAssertionOnlyCallback` requires `callback.body.type === 'BlockStatement'` (rule.ts:131), so any arrow function with an implicit/expression body bails out before any assertion-safety analysis runs:

```js
cy.get('input').then($input => expect($input.text()).to.equal('Ready')); // never flagged
```

This is behaviorally identical to the block-bodied form the rule *does* flag and auto-fix today (same test file, `unit.test.ts:141-149`):

```js
cy.get('input').then($input => {
  expect($input.text()).to.equal('Ready'); // Noncompliant, flagged + auto-fixed
});
```

`.should()` accepts an expression-bodied callback exactly as well as a block-bodied one — there's no Cypress semantic difference here (Cypress's `.should(callback)` doesn't use the callback's return value the way `.then()` does, but that applies equally to both forms, since a bare `expect(...)` statement returns the same kind of unused value either way). The `BlockStatement` requirement looks like a pure AST-shape restriction rather than an intentional conservative safety choice, so this misses a common, equally-safe pattern.

Suggest relaxing the guard to also accept an expression body containing a single assertion expression (treat it the same as a one-statement block), rather than requiring braces.

packages/analysis/src/jsts/rules/S9162/rule.ts
```diff
@@ -0,0 +1,331 @@
...
+  }
+  const variable = getVariableFromName(context, root.name, root);
+  return !variable || variable.defs.length === 0;
+}
+
+function acceptRetrySafeDeclaration(
+  declaration: estree.VariableDeclaration,
+  bindings: Map<string, RetrySafeKind>,
+): boolean {
+  if (declaration.kind !== 'const' || declaration.declarations.length !== 1) {
```
guillemsarda: **False negative: a multi-declarator `const` is rejected outright, even when every declarator is independently retry-safe**

`acceptRetrySafeDeclaration` bails out as soon as a declaration has more than one declarator (rule.ts:202: `declaration.declarations.length !== 1`), before checking whether each declarator is individually safe:

```js
cy.get('input').then($input => {
  const text = $input.text(), expected = 'Ready'; // never flagged
  expect(text).to.equal(expected);
});
```

Both `text` (derived from the yielded `$input` via a retry-safe call chain) and `expected` (a literal) are individually retry-safe by the rule's own `retrySafeKind` logic — in fact, splitting this into two separate statements makes the rule accept and flag it immediately:

```js
cy.get('input').then($input => {
  const text = $input.text();      // each declarator on its own line...
  const expected = 'Ready';        // ...passes retrySafeKind independently
  expect(text).to.equal(expected); // Noncompliant, flagged + auto-fixed
});
```

There's no safety difference between the two forms — Cypress evaluates both declarators synchronously either way. The `declarations.length !== 1` check looks like it's guarding against a real multi-declarator hazard, but it's actually just rejecting a common comma-separated `const` style outright.

Suggest iterating over `declaration.declarations` and checking each declarator's initializer independently with `retrySafeKind`, rather than requiring exactly one declarator per statement.

packages/analysis/src/jsts/rules/S9162/rule.ts
guillemsarda: **Simplification: the same "unshadowed global identifier" check is implemented three times, two of them in this file**

`isGlobalCyChain` (rule.ts:99-106) and `isGlobalCypressAssertion` (rule.ts:189-196) both do the exact same thing — walk to a root identifier, call `getVariableFromName`, then return `!variable || variable.defs.length === 0` — differing only in which name(s) they check (`cy` vs. `expect`/`assert`). That's the same one-line predicate `isGlobalAssertReference` already implements in `helpers/assertions-chai.ts:176-179`, so this is now the third copy of the same idiom, with two of them living side by side in this new file.

```ts
// rule.ts:99-106
function isGlobalCyChain(context, node) {
  const root = findChainRoot(node);
  if (root.type !== 'Identifier' || root.name !== 'cy') return false;
  const variable = getVariableFromName(context, 'cy', root);
  return !variable || variable.defs.length === 0;
}

// rule.ts:189-196 — same shape, different name(s)
function isGlobalCypressAssertion(context, node) {
  const root = findChainRoot(node);
  if (root.type !== 'Identifier' || (root.name !== 'expect' && root.name !== 'assert')) return false;
  const variable = getVariableFromName(context, root.name, root);
  return !variable || variable.defs.length === 0;
}
```

Could these collapse into a single helper, e.g. `isUnshadowedGlobal(context, node, allowedNames: string[])` (using `findChainRoot` + the shared shadow check), ideally promoted to `helpers/ast.ts` alongside `getVariableFromName` so `isGlobalAssertReference` could eventually delegate to it too? No behavior change, just one idiom instead of three.

packages/analysis/src/jsts/rules/S9162/rule.ts
guillemsarda, line 252: **Architecture: this is the only rule that calls `extractChaiAssertion` directly instead of going through the shared `extractTestAssertion` dispatcher**

Every other assertion-detecting rule in the codebase (S5914, S9073, S1244, S5845, S5906, and S8784/S8785) calls `extractTestAssertion` (`helpers/assertions.ts:155-180`), which is the catalog-driven entry point that gates on imports/dependencies and dispatches across chai, cypress-chain, jest-like, jasmine, playwright, and node-assert styles. Its chai-like library entry already does exactly:

```ts
extractChaiAssertion(context, node, true) ?? extractCypressChainAssertion(node)
```

This rule instead imports and calls `extractChaiAssertion` directly (rule.ts:22, 151), making it the sole caller of that low-level primitive outside `assertions.ts` itself. It doesn't currently produce a wrong result — excluding cypress-chain style is fine here since those assertions are already retry-safe via `.should()`/`.and()` — but it does mean this rule has its own separate gating (`isGlobalCypressAssertion`) instead of `extractTestAssertion`'s import/dependency gating, and won't automatically inherit future improvements to the shared dispatcher (new matchers, refined negation handling, etc.).

Was going through `extractChaiAssertion` directly a deliberate choice (e.g. because `extractTestAssertion`'s broader style set would need extra filtering anyway), or would it be worth funneling through `extractTestAssertion` and just restricting the returned `style` the way this rule already does, so there's one canonical entry point for "give me the assertion structure" going forward?

packages/analysis/src/jsts/rules/S9162/rule.ts
```diff
@@ -0,0 +1,331 @@
...
+  expression: estree.Node,
+  bindings: ReadonlyMap<string, RetrySafeKind>,
+): RetrySafeKind | null {
+  const node = unwrapTypeScriptExpression(expression);
+  switch (node.type) {
+    case 'Literal':
+      return typeof node.value === 'string' ? 'string' : 'value';
+    case 'Identifier':
+      return bindings.get(node.name) ?? null;
+    case 'TemplateLiteral':
```
guillemsarda, line 375: **Test coverage gap: 4 of `retrySafeKind`'s expression branches are never exercised by any test**

`retrySafeKind` (rule.ts:217 onward) handles `TemplateLiteral`, `UnaryExpression` (via `retrySafeUnaryKind`), `BinaryExpression`/`LogicalExpression`, and `ConditionalExpression` (via `retrySafeConditionalKind`) — but none of these four branches has a single valid/invalid test case in `unit.test.ts` exercising it. No test constructs a ternary, a unary operator (`!x`, `-x`, `typeof x`), a binary/logical expression (`a + b`, `a && b`), or a template literal anywhere in a `.then()` callback's assertion arguments or local declarations.

That's shipped, unverified generality: real Cypress `.then()` callbacks overwhelmingly consist of member access on the yielded element, string/object literals, and simple derived strings — none of the test cases actually need this generality to pass. If any of these four branches has a latent bug (e.g. an operator that should be excluded but isn't, or an operand that should propagate `'yielded'`/`'record'` but gets coerced to `'value'` incorrectly), it will ship silently and only surface as a false positive/negative in real user code.

Could we either add a test per branch (e.g. `expect($input.text()).to.equal(cond ? 'Ready' : 'Done')`, `expect(!$input.is(':hidden')).to.equal(true)`, `` expect(`${prefix}Ready`).to.equal(...) ``) to lock in the intended behavior, or trim the branches that aren't needed to satisfy the RSPEC's "conservative" scope?

guillemsarda (changes requested):

packages/analysis/src/jsts/rules/S9162/rule.ts
nathsou: Kept the filename guard: scope Tests is metadata rather than an ESLint execution guard, and S2925/S8959 follow the same pattern. The dependency concern is already addressed: S9162 now declares Cypress as a required dependency, so the linter only enables it when Cypress is present.

packages/analysis/src/jsts/rules/S9162/rule.ts
nathsou: Fixed in ab21ccde16. S9162 now only considers direct DOM queries (cy.get with a literal non-alias selector or cy.contains), so cy.wrap and cy.get aliases remain unreported. Added regression coverage; the focused suite passes.

packages/analysis/src/jsts/rules/S9162/rule.ts
nathsou: Kept the local root walker: S9162 must identify the exact root to reject a shadowed cy, while chainStartsWithCy is a context-free boolean helper. Adding TypeScript-cast support there would be a separate improvement for the existing rules, not required for S9162 correctness.

packages/analysis/src/jsts/rules/S9162/rule.ts
nathsou: Keeping the conservative Chai subset. S9162 reports only assertion forms represented by extractChaiAssertion; supporting arbitrary Chai matcher semantics would expand shared assertion behavior beyond this rule scope.

packages/analysis/src/jsts/rules/S9162/rule.ts
nathsou: Keeping block-only callbacks. An expression-bodied then callback returns the assertion-chain value, and Cypress uses a non-null then return as the next subject. Replacing then only with should could therefore change a following chain.

packages/analysis/src/jsts/rules/S9162/rule.ts
nathsou: Fixed in ab21ccde16. Safe const declarators are now validated in source order, so a later declarator can use an earlier safe binding; unsafe, destructured, or uninitialized declarators still stop reporting. Added regression coverage.

packages/analysis/src/jsts/rules/S9162/rule.ts
nathsou: No code change here. Generalizing the unshadowed-global checks would be a refactor: the cy and Chai paths have different root and name requirements. The current implementation is behaviorally correct.

packages/analysis/src/jsts/rules/S9162/rule.ts
nathsou: No code change here. S9162 deliberately extracts only Chai styles and applies its own shadowing guard. Its generated Cypress dependency metadata provides linter gating; the broader dispatcher would add styles that this rule intentionally excludes.

packages/analysis/src/jsts/rules/S9162/rule.ts
nathsou: Fixed in ab21ccde16. Added focused cases for TemplateLiteral, UnaryExpression, BinaryExpression, LogicalExpression, and ConditionalExpression. The focused S9162 suite passes.

guillemsarda (approved):

PR #7737, opened 2026-08-10
JS-2228 Fix S125 with TypeScript project service

francois-mora-sonarsource:
## Summary
- Restore S125 detection of commented-out code when TypeScript-ESLint project service is enabled.
- Keep synthetic comment parsing independent of project-service file ownership.
- Add regression coverage using an in-project TypeScript source file.

## Root cause
S125 re-parsed comments through a synthetic placeholder filename after clearing project and programs, but retained projectService. TypeScript-ESLint rejected the synthetic non-project file and S125 treated the parse failure as non-code.

## Validation
- npx tsx --test packages/analysis/src/jsts/rules/S125/unit.test.ts
- npm run bbf
- git diff --check


victor-peixoto-sonarsource (approved):
LGTM

PR #7739, opened 2026-08-11
JS-2235: Implement S9169: vi.mock should be declared at module scope

nathsou:
## Summary

- Implement S9169 as an original SonarJS rule for nested Vitest `vi.mock()` calls.
- Report when `vi.mock` is not a module-scope expression statement; suggest module scope or `vi.doMock()`.
- Add focused JS/TS fixtures aligned with the RSPEC examples and binding/placement boundaries.
- Pin RSPEC to the open RSPEC PR head via `rspec.sha`.

## Links

- Jira: https://sonarsource.atlassian.net/browse/JS-2235
- RSPEC PR: https://github.com/SonarSource/rspec/pull/7752


packages/analysis/src/jsts/rules/S9169/rule.ts
```diff
@@ -0,0 +1,84 @@
...
+  ) {
+    return false;
+  }
+
+  const fqn = getFullyQualifiedName(context, call);
+  if (fqn !== null) {
+    return VITEST_MOCK_FQNS.has(fqn);
+  }
+
+  return isGlobalVitestNamespace(context, call.callee.object);
```
guillemsarda, line 62: `isGlobalVitestNamespace` flags any bare, unimported global named `vi`/`vitest` with a `.mock()` call as Vitest, with no check that the file/project actually uses Vitest. Reproduced: a file with zero `vitest` import/dependency, just `function helper() { vi.mock('x'); }` where `vi` is some unrelated global, still gets flagged. Consider gating on an actual vitest import/dependency signal rather than name alone. By requiring the dependency in the package.json, should be enough.

packages/analysis/src/jsts/rules/S9169/cb.fixture.ts
```diff
@@ -0,0 +1,70 @@
...
+  const vitest = require('vitest');
+  vitest.vi.mock('required-namespace-module'); // Noncompliant {{Move this call to module scope or replace it with vi.doMock(); vi.mock is hoisted and cannot provide runtime or per-test mocking.}}
+}
+
+function otherImport(): void {
+  otherVi.mock('other-module');
+}
+
+vi.doMock('do-mock-module');
+vi.unmock('unmock-module');
```
guillemsarda, line 67: Is unmock out of scope on purpose even it has the same behaviour as `mock`?

[packages/mocker/src/node/hoistMocks.ts in vitest-dev/vitest](https://github.com/vitest-dev/vitest/blob/main/packages/mocker/src/node/hoistMocks.ts)
```ts
/**
 * @default ["mock", "unmock"]
 */
hoistableMockMethodNames?: string[]
...
hoistableMockMethodNames = ['mock', 'unmock'],
```
`unmock` is in the same default array as `mock` and goes through the identical hoisting + top-level-scope validation (the same code path that throws when a hoisted node isn't found among ast.body's direct children).

Secondary, less authoritative but corroborating source:
DeepWiki's Vitest mocking/spying page - https://deepwiki.com/vitest-dev/vitest/3.3-mocking-and-spying
Describes the plugin as identifying "vi.mock, vi.unmock, and vi.hoisted calls" for hoisting.

guillemsarda: In case of covering `unmock` and/or `hoisted`, we should update the rspec too.

packages/analysis/src/jsts/rules/S9169/rule.ts
```diff
@@ -0,0 +1,84 @@
...
+    ('optional' in call.callee && call.callee.optional === true) ||
+    ('optional' in call && call.optional === true) ||
+    !isIdentifier(call.callee.property, 'mock')
+  ) {
+    return false;
+  }
+
+  const fqn = getFullyQualifiedName(context, call);
+  if (fqn !== null) {
+    return VITEST_MOCK_FQNS.has(fqn);
```
guillemsarda: `isVitestMockCall` (rule.ts:56-59) returns false whenever `getFullyQualifiedName` returns non-null, without checking whether that value is an actual import-resolved FQN. When `vi`/`vitest` are registered as globals (e.g. `sonar.javascript.environments=vitest`, which mirrors the existing jest/jasmine/mocha default-env support), `getFullyQualifiedName` returns the bare name (`"vi.mock"`) instead of null — this doesn't match `VITEST_MOCK_FQNS`, so the function returns false and never falls through to the `isGlobalVitestNamespace` fallback that exists to handle exactly this case. Net effect: nested `vi.mock()` calls go undetected under that config. Verified with a RuleTester run — same code, only the globals config differs, and it flips from caught to missed. Suggest falling through to `isGlobalVitestNamespace` whenever `fqn` isn't in `VITEST_MOCK_FQNS`, not only when `fqn === null`.

packages/analysis/src/jsts/rules/S9169/cb.fixture.ts
```diff
@@ -0,0 +1,70 @@
...
+vi.hoisted((): void => {});
+vi.mocked(service).mock();
+someObject.mock();
```
guillemsarda, line 70: Worth adding this as a conscious discarded option to not be considered for future maintainers of this rule since it is invalid vitest syntax but syntactically correct JS/TS.
```ts
const { mock } = vi;,
mock();
```

guillemsarda (changes requested):

packages/analysis/src/jsts/rules/S9169/rule.ts
nathsou: Good catch — fixed in b64cea0. `isGlobalVitestNamespace` now also requires `importsOrDependsOnModule(context, ['vitest'], ['vitest'])`, matching the convention used by `S8968`'s framework-detection fallback (import or package.json dependency signal, not just an unbound identifier name). Added a regression test in `unit.test.ts` (`does not flag an unbound "vi" name when the project has no Vitest signal`).

packages/analysis/src/jsts/rules/S9169/cb.fixture.ts
nathsou: Intentionally out of scope for this PR, not an oversight — `implementation-plan.md` explicitly lists `vi.unmock`, `vi.hoisted`, and `vi.doUnmock` under "Out of scope", and the Jira/RSPEC agreement is "Vitest `vi.mock` only" for v1. Since the rule doesn't cover `unmock`/`hoisted` here, the RSPEC (which only documents `vi.mock`) stays consistent, so no doc update is needed for this PR. Happy to file a follow-up ticket if we want to extend coverage to `unmock`/`hoisted` later — thanks for the pointers to `hoistMocks.ts` and the default `hoistableMockMethodNames`, that'll be useful context for that follow-up.

packages/analysis/src/jsts/rules/S9169/rule.ts
nathsou: Confirmed and fixed in b64cea0 — reproduced with a `RuleTester` run using `languageOptions.globals: { vi: 'readonly' }` (what `environments: ['vitest']` registers via the `globals` npm package, since `globals.vitest.vi === false`): detection flipped from caught to missed exactly as described. `isVitestMockCall` now only short-circuits to `true` when the FQN actually matches `VITEST_MOCK_FQNS`, and falls through to `isGlobalVitestNamespace` for any other FQN value (including a non-null bare name from a configured global), not only when the FQN is `null`. Added a regression test in `unit.test.ts` (`flags "vi.mock" when "vi" is registered as an environment global`).

packages/analysis/src/jsts/rules/S9169/cb.fixture.ts
nathsou: Good idea, added in b64cea0 to both `cb.fixture.ts` and `cb.fixture.js`:

    function destructuredMock() {
      const { mock } = vi;
      mock('destructured-module'); // Compliant: destructuring loses the vi/vitest receiver, out of scope
    }

guillemsarda (approved):
Image

PR #7741, opened 2026-08-11
JS-2221 Implement rule S9163: Reactive state should not be unconditionally mutated inside Vue's "updated" lifecycle hook

guillemsarda:
----
## Summary by Gitar

- **New rules:**
    - Implemented rule S9163 to detect unconditional mutation of reactive state inside Vue's `updated` lifecycle hook (`packages/analysis/src/jsts/rules/S9163/rule.ts`)


packages/analysis/src/jsts/rules/S9163/rule.ts
```diff
@@ -0,0 +1,309 @@
...
+      CallExpression(node: estree.CallExpression) {
+        if (getFullyQualifiedName(context, node) === VUE_ON_UPDATED_FQN) {
+          registerHook(node.arguments[0]);
+        }
+      },
+      Property(node: estree.Property) {
+        const isUpdatedKey =
+          isIdentifier(node.key, UPDATED_HOOK_NAME) ||
+          (isStringLiteral(node.key) && node.key.value === UPDATED_HOOK_NAME);
+        if (isUpdatedKey && getNodeParent(node)?.type === 'ObjectExpression') {
```
nathsou: [P1] Do not treat every object-literal `updated` method as a Vue lifecycle hook\n\nThis listener only checks the property name and its `ObjectExpression` parent, so it also reports unrelated objects such as:\n\n```js\nimport { ref } from 'vue';\nconst count = ref(0);\nconst cache = {\n  updated() {\n    count.value++;\n  },\n};\n```\n\n`cache.updated` is not registered as a Vue component lifecycle hook, so this is outside the rule scope and creates a false positive. Please restrict Options API matching to a known Vue component context (and add this regression case).

packages/analysis/src/jsts/rules/S9163/rule.ts
```diff
@@ -0,0 +1,309 @@
...
+}
+
+/**
+ * Signal 1: a mutation nested inside an `if`/ternary/logical guard, or reached only after an
+ * early-return guard clause earlier in the same block, is never flagged - proving that a guard
+ * actually converges is undecidable in general, so guard presence alone is a hard exemption.
+ */
+function isGuarded(mutation: Mutation): boolean {
+  const node = mutation as unknown as TSESTree.Node;
+  if (findFirstMatchingLocalAncestor(node, ancestor => GUARD_ANCESTOR_TYPES.has(ancestor.type))) {
```
nathsou: [P1] Distinguish guard branches from guard tests

findFirstMatchingLocalAncestor() returns true based only on the ancestor type, so this also suppresses mutations in an if/ternary test or in the always-evaluated left operand of a logical expression. For example, all of these currently produce no issue:

```js
onUpdated(() => {
  if (count.value++ > 100) { doSomething(); }
});
onUpdated(() => {
  count.value++ ? doA() : doB();
});
onUpdated(() => {
  count.value++ && doSomething();
});
```

Those mutations execute on every hook call and can retrigger the render, so they are in scope for this rule. Please retain the child position while walking ancestors: only the conditional branch of if/ternary, or the right operand of a logical expression, should be treated as guarded, and add regression tests for these cases.

packages/analysis/src/jsts/rules/S9163/rule.ts
```diff
@@ -0,0 +1,309 @@
...
+  node: estree.Node | null | undefined,
+): estree.Function | undefined {
+  if (!node) {
+    return undefined;
+  }
+  const resolved = resolveFunction(context, node);
+  if (resolved) {
+    return resolved;
+  }
+  if (node.type === 'Identifier') {
```
nathsou, line 150: [P2] Cover variable-initialized callback references

The fallback below handles a variable whose initializer is a FunctionExpression or ArrowFunctionExpression, which is not resolved by the shared resolveFunction() helper. The S9163 fixtures cover inline callbacks, named function declarations, and direct function expressions, but not the common pattern below:

```js
const handler = () => {
  count.value++;
};
onUpdated(handler); // should raise
```

The implementation reports this correctly, but the branch has no S9163 regression coverage. Please add this case to the JavaScript or Vue fixture.

packages/analysis/src/jsts/rules/S9163/rule.ts
```diff
@@ -0,0 +1,309 @@
...
+
+const VUE_ON_UPDATED_FQN = 'vue.onUpdated';
+const UPDATED_HOOK_NAME = 'updated';
+const REF_VALUE_PROPERTY = 'value';
+const FUNCTION_BOUNDARIES = new Set([
+  'FunctionExpression',
+  'ArrowFunctionExpression',
+  'FunctionDeclaration',
+]);
+const GUARD_ANCESTOR_TYPES = new Set(['IfStatement', 'ConditionalExpression', 'LogicalExpression']);
```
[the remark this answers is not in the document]
guillemsarda: Gitar fix

packages/analysis/src/jsts/rules/S9163/rule.ts
guillemsarda: Addressed in [this commit](https://github.com/SonarSource/SonarJS/pull/7741/changes/f2b1523dafcb07bd834eca1b7bb81ca612575299).

packages/analysis/src/jsts/rules/S9163/rule.ts
guillemsarda: Good catch — `resolveHookFunction`'s fallback branch (the `Variable`-initializer case not covered by `resolveFunction()`) was already resolving this correctly, it just had no regression test. Added it to `cb.fixture.vue`, right next to the existing `handleUpdate` (function declaration) case:

```js
const arrowHandlerCount = ref(0);
const arrowHandler = () => {
  arrowHandlerCount.value++; // Noncompliant
};
onUpdated(arrowHandler);
```

so both branches of `resolveHookFunction` are now exercised side by side. Confirmed the new case isn't vacuous: temporarily removed that fallback branch and reran — failed with `11 !== 12` (missing exactly this case) — then restored it.

packages/analysis/src/jsts/rules/S9163/rule.ts
guillemsarda: Thanks! Addressed in [5a79c58](https://github.com/SonarSource/SonarJS/pull/7741/commits/5a79c580d3a87d1e6536d53e00792d9bf0dd840f)

nathsou (approved):
Image

PR #7743, opened 2026-08-12
JS-2236: Implement S9332: disallow Playwright networkidle waits

nathsou:
## Summary

- Implement S9332 as an original AST rule to flag Playwright `networkidle` waits.
- Cover `waitForLoadState('networkidle')` and `waitUntil: 'networkidle'` on `goto`, `reload`, `goBack`, `goForward`, `setContent`, and `waitForURL`.
- Add comment-based fixtures and a custom-jsts ruling baseline.

## Links

- Jira: https://sonarsource.atlassian.net/browse/JS-2236
- RSPEC PR: https://github.com/SonarSource/rspec/pull/7857


packages/analysis/src/jsts/rules/S9332/rule.ts
```diff
@@ -0,0 +1,140 @@
...
+  if (!stateArgument || stateArgument.type === 'SpreadElement') {
+    return;
+  }
+  const stateLiteral = resolveLiteralValue(context, stateArgument);
+  if (isNetworkidleLiteral(stateLiteral)) {
+    context.report({ node: stateLiteral, message: MESSAGE });
+  }
+}
+
+function reportWaitUntilOption(
```
guillemsarda, line 88: **False negative: TS-wrapped options object bypasses detection**

`reportWaitUntilOption` resolves the options argument via `getValueOfExpression(context, optionsArgument, 'ObjectExpression')` (rule.ts:96), but neither `optionsArgument` nor the value it resolves through a variable is ever passed through `unwrapTypeScriptExpression`. `getValueOfExpression` only matches when the node (or the resolved variable's write-expression) is *directly* an `ObjectExpression`, so wrapping the whole options object in `as const`/`satisfies` makes the match fail silently.

Confirmed with the rule's own `NoTypeCheckingRuleTester` — both report 0 errors instead of 1:

```ts
// inline
await page.goto('/dashboard', { waitUntil: 'networkidle' } as const);

// via variable
const options = { waitUntil: 'networkidle' } as const;
await page.goto('/dashboard', options);

// satisfies
await page.goto('/dashboard', { waitUntil: 'networkidle' } satisfies object);
```
This affects every method routed through reportWaitUntilOption (goto, reload, goBack, goForward, setContent, waitForURL), not just goto. Note page.waitForLoadState('networkidle' as const) is not affected — resolveLiteralValue already unwraps at every step for the literal-argument path (the F02 fix), this gap is specifically in the options-object path that never got the same treatment.

Suggested fix: unwrap optionsArgument with unwrapTypeScriptExpression before calling getValueOfExpression, and also unwrap the resolved variable's write-expression before the ObjectExpression type check — mirroring what resolveLiteralValue already does one level down.

packages/analysis/src/jsts/rules/S9332/rule.ts
```diff
@@ -0,0 +1,140 @@
...
+  'Replace this "networkidle" wait with a web-first assertion or a specific readiness condition.';
+
+export const rule: Rule.RuleModule = {
+  meta: generateMeta(meta),
+  create(context: Rule.RuleContext): Rule.RuleListener {
+    if (!importsOrDependsOnModule(context, PLAYWRIGHT_MODULES, PLAYWRIGHT_MODULES)) {
+      return {};
+    }
+    return {
+      CallExpression(node: estree.Node): void {
```
guillemsarda, line 52: **FP risk: no receiver check — matches any object, not just Playwright `page`/`frame`**

The `CallExpression` visitor (rule.ts:53-73) only checks the method name (`call.callee.property.name`); it never inspects `call.callee.object`. Once a file/project passes the Playwright-dependency gate, *any* object with a matching method name is flagged — `goto`, `reload`, `goBack`, `goForward` aren't unique to Playwright (e.g. React Navigation's `navigation.goBack()`, generic router/history wrappers), so this is a realistic false-positive surface, not just a theoretical mock-object edge case.

There's already precedent in this codebase for narrowing this: S2925 restricts one of its method matches to receivers literally named `page`:

```ts
// S2925/rule.ts:54
} else if (methodName === 'waitForTimeout' && isIdentifier(call.callee.object, 'page')) {
```
Suggest applying the same heuristic here — restrict call.callee.object to page/frame identifiers (S9332's own fixtures use both, e.g. const frame = page.frames()[0]; await frame.waitForLoadState(...)). It won't fully close the gap (a mock could still be named page), but it's a low-cost improvement consistent with how the neighboring rule already mitigates the same risk.

Not a blocker, but worth considering before merge or as a fast-follow.

guillemsarda (approved):
Image

packages/analysis/src/jsts/rules/S9332/rule.ts
nathsou: Addressed in bc0f2a12e: `reportWaitUntilOption` now resolves options through the same TS-unwrapping identifier walk as the literal path (`resolveExpression`), so `{ waitUntil: 'networkidle' } as const`, the same via a variable, and `satisfies` are reported. Added those cases to `cb.fixture.ts`; `npx tsx --test packages/analysis/src/jsts/rules/S9332/cb.test.ts` passes.

packages/analysis/src/jsts/rules/S9332/rule.ts
nathsou: Leaving this unchanged for now.

The distinctive signal is already the `'networkidle'` literal (Puppeteer uses `networkidle0`/`networkidle2`), so a generic `navigation.goBack()` without that option does not match. Restricting receivers to identifiers named `page`/`frame` would miss common Playwright patterns that we intentionally want to flag, such as renamed pages (`dashboardPage.goto(...)`) and member access (`pom.page.goto(...)` / `this.page.goto(...)`).

The Playwright dependency gate already covers the no-Playwright case (see the CustomRouter example in `cb.test.ts`). Happy to revisit if Peach shows real FPs that need a smarter receiver model.

PR #7746, opened 2026-08-12
TASK-319 Guard S1481 against unsafe Babel Flow scopes

francois-mora-sonarsource:
## Summary

- identify Babel Flow type parameters through ESLint SourceCode scope data
- skip S1481 only when a Flow type parameter has no references, the upstream no-unused-vars crash shape
- retain S1481 on ordinary Babel JavaScript, TypeScript-parser sources, and Flow files whose type parameters are used

## Behavior

Babel exposes Flow generic definitions as the runtime-only TypeParameter scope-definition type. S1481 does not register the upstream listener for a Babel-fallback file containing an unused one, preventing the upstream crash. Because the upstream rule collects all unused variables at Program:exit, S1481 cannot report other unused variables in that same crashing file; all other rule activation remains unchanged.

## Validation

- ▶ S1481
  ✔ should skip Babel files with unused Flow type parameters (31.423141ms)
  ✔ S1481 (decorated: typescript-eslint/no-unused-vars) (49.565003ms)
✔ S1481 (81.593948ms)
ℹ tests 2
ℹ suites 1
ℹ pass 2
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 969.726293
- 
> sonarjs@1.0.0 bbf
> npm run bridge:build:fast

> sonarjs@1.0.0 bridge:build:fast
> npm run _:bridge:clear && npm run generate-meta && npm run bridge:compile

> sonarjs@1.0.0 _:bridge:clear
> rimraf --glob lib/*

> sonarjs@1.0.0 generate-meta
> tsx tools/generate-meta.ts

> sonarjs@1.0.0 bridge:compile
> npm run grpc:generate-proto && tsgo -p packages/tsconfig.app.json && tsgo -p packages/tsconfig.test.json && npm run grpc:copy-proto

> sonarjs@1.0.0 grpc:generate-proto
> tsx tools/generate-proto.ts

Generating packages/grpc/src/proto/analyze-project.js from packages/grpc/src/proto/analyze-project.proto
Generating packages/grpc/src/proto/analyze-project.d.ts from packages/grpc/src/proto/analyze-project.js
Generating packages/grpc/src/proto/language_analyzer.js from packages/grpc/src/proto/language_analyzer.proto
Generating packages/grpc/src/proto/language_analyzer.d.ts from packages/grpc/src/proto/language_analyzer.js
Generating packages/grpc/src/proto/health.js from packages/grpc/src/proto/health.proto
Generating packages/grpc/src/proto/health.d.ts from packages/grpc/src/proto/health.js
Generating packages/analysis/src/jsts/parsers/estree.js from packages/analysis/src/jsts/parsers/estree.proto
Generating packages/analysis/src/jsts/parsers/estree.d.ts from packages/analysis/src/jsts/parsers/estree.js
Proto generation complete.

> sonarjs@1.0.0 grpc:copy-proto
> tsx tools/generate-proto.ts --copy

Proto files copied to lib/.
-


packages/analysis/src/jsts/rules/S1481/meta.ts
```diff
@@ -17,3 +17,4 @@
 export const implementation = 'decorated';
 export const eslintId = 'no-unused-vars';
 export const externalRules = [{ externalPlugin: 'typescript-eslint', externalRule: eslintId }];
+export const requiresTypeScriptParser = true;
```
nathsou: Instead of introducing a boolean specifically for the TS parser, how about making it a bit more general, something like 

```typescript
export const supportedParsers = ['typescript-eslint', 'babel'];
```

Which if not exported, would default to all available parsers.

francois-mora-sonarsource: Thanks for the review, that's an option we should consider indeed. There's also an open point to clarify regarding the scope. As of now, only JavaScript files with Flow annotations are causing problems to the rule, therefore if the ESLint allows to determine this specific case, we could also deactivate the rule only in this narrower use-case. But that requires some investigations.

francois-mora-sonarsource:
Superseded by 30442c92c, which reverted the @typescript-eslint/no-unused-vars S1481 implementation that caused this crash. Closing rather than rebasing to avoid reintroducing the removed implementation.

PR #7770, opened 2026-08-17
JS-2261 Skip project analysis when all files are cached

vdiez:
Part of

Fixes the all-cache-hit project-analysis path reported in the [community thread](https://community.sonarsource.com/t/sonarjs-13-3-0-cache-hit-unknown-file-path-bug-report/186535).

When request construction consumes every input from the analysis cache, the resulting project-analysis request has no files. We now complete that analysis locally instead of starting an empty streaming gRPC call. Cache replay still happens while the request is built, so cached issues, CPD tokens, and ASTs continue to be processed.

The regression test verifies that an empty request completes without creating an `AnalyzeProjectServiceGrpc` stub. Existing transport-error tests now use non-empty requests so they continue to exercise the transport path.

Tested with:

```text
mvn -pl sonar-plugin/bridge -am -Dskip-nodejs -Dskip-java-generation "-Dtest=BridgeServerImplTest" "-Dsurefire.failIfNoSpecifiedTests=false" test
```


nathsou (approved):
Image

PR #7771, opened 2026-08-17
Narrow css:S4662 to plain CSS

jdkandersson:
## Summary

- run `css:S4662` only on plain CSS files and plain CSS blocks embedded in HTML/Vue
- stop reporting this rule on SCSS, Sass, and Less
- replace Sass-specific exclusions with supported Tailwind and css-loader directives
- add comprehensive coverage for standard CSS, every supported tool directive, preprocessors, and embedded blocks

## Why

False-positive analysis found that the rule primarily reports valid preprocessor syntax when applied outside plain CSS. Peachee classified 915/1000 sampled issues as false positives, dominated by Sass `@use` and `@forward`. SonarCloud also showed Less variables and SCSS control-flow directives among the main false-positive clusters.

A separate Sass-aware rule can follow under a new RSPEC key.

## Verification

- focused CSS and gRPC tests pass
- TypeScript bridge compilation passes
- Prettier and `git diff --check` pass


packages/analysis/tests/css/linter/at-rule-no-unknown.test.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  for (const directive of ['layer', 'container']) {
+    it(`accepts the standard @${directive} at-rule without an explicit exclusion`, () =>
+      ruleTester.valid({ code: `@${directive} example {}` }));
+  }
+
+  for (const directive of supportedToolDirectives) {
+    it(`accepts the supported @${directive} tool directive`, () =>
+      supportedToolRuleTester.valid({ code: `@${directive} value;` }));
+  }
```
erwan-leforestier-sonarsource: What is the purpose of this test ? Is it to test that the rule configuration work as expected or to test that these exact directives are allowed by the rule ? If the later, I suggest to try finding a way to reuse the default parameter defined in `metadata.ts` to not write twice the same list of directives.

sonar-plugin/css/src/test/java/org/sonar/css/CssRulesDefinitionTest.java
```diff
@@ -81,7 +81,9 @@ private void assertParameterProperties(Repository repository) {
     // AtRuleNoUnknown
     Param param = repository.rule("S4662").param("ignoreAtRules");
     assertThat(param).isNotNull();
-    assertThat(param.defaultValue()).startsWith("value,at-root,content");
+    assertThat(param.defaultValue()).isEqualTo(
+      "value,tailwind,screen,responsive,variants,apply,theme,source,utility,variant,custom-variant,reference,config,plugin"
+    );
```
erwan-leforestier-sonarsource: This test is not necessary imo and it's duplicating the default parameter value one more time. I suggest to remove it.

packages/analysis/src/css/rules/metadata.ts
```diff
@@ -225,12 +225,14 @@ export const cssRulesMeta: CssRuleMeta[] = [
...
   simpleRule('S4661', 'media-feature-name-no-unknown'),
+  // Stylelint already recognizes standard CSS at-rules. These defaults cover directives used
+  // in plain CSS by css-loader and current/legacy Tailwind versions.
   ignoreAtRulesRule(
     'S4662',
     'at-rule-no-unknown',
     'ignoreAtRules',
     'Comma-separated list of "at-rules" to consider as valid.',
-    'value,at-root,content,debug,each,else,error,for,function,if,include,mixin,return,warn,while,extend,use,forward,tailwind,apply,layer,container,theme,utility,custom-variant,source,plugin,config,reference,variant,/^@.*/',
+    'value,tailwind,screen,responsive,variants,apply,theme,source,utility,variant,custom-variant,reference,config,plugin',
```
erwan-leforestier-sonarsource: I ran the analyzer on this branch against a plain `.css` file using [postcss-mixins](https://github.com/postcss/postcss-mixins). With this default list, S4662 flags both mixin directives:

```
Unknown at-rule "@define-mixin"
Unknown at-rule "@mixin"
```

`@apply` (in the list) and standard `@media` produce zero issues, so the list works as intended.

postcss-mixins lets you reuse CSS blocks in plain `.css` without Sass:

```css
@define-mixin flex-center { display: flex; }
.card { @mixin flex-center; }
```

Since the rule now runs on plain CSS and `define-mixin`/`mixin` aren't in the default list, these valid directives get flagged as unknown at-rules.

packages/analysis/tests/css/linter/at-rule-no-unknown.test.ts
jdkandersson: Good point. I exported `supportedCssToolDirectives` from `metadata.ts` and the rule test now consumes that same list, so it verifies every configured default without duplicating it.

sonar-plugin/css/src/test/java/org/sonar/css/CssRulesDefinitionTest.java
jdkandersson: Agreed. I removed the assertion of the complete generated default value; the Java test retains the parameter presence, description, and type checks.

packages/analysis/src/css/rules/metadata.ts
jdkandersson: Thanks for validating this. I added `define-mixin` and `mixin` to the shared plain-CSS tool directive allowlist, with focused coverage for both.

erwan-leforestier-sonarsource (approved):

PR #7777, opened 2026-08-18
JS-2211 Load quality profiles from rule metadata

vdiez:
Part of [JS-2211](https://sonarsource.atlassian.net/browse/JS-2211). Follow-up to #7763.

## Why

PR #7763 generated aggregate quality-profile JSON files from the per-rule `Sxxxx.json` metadata, copied those files through Maven resources, and loaded them again when defining rules and profiles. That made the same profile membership exist in two forms and added generation, copying, and runtime parsing solely for derived data.

This PR keeps one persisted source of truth: each rule's RSPEC-generated `Sxxxx.json`. Profile membership is compiled into static Java metadata as part of the Java rule-class generation that already reads those files.

This gives us:

- **A single source of truth:** profile membership is authored and stored only in `Sxxxx.json`; the Java catalogs are derived build output, like the generated rule classes.
- **Less I/O:** no aggregate profile files are written, copied into module resources, discovered on the classpath, or parsed at runtime for profile membership.
- **A simpler build:** the existing `generate-java-rule-classes` step produces everything needed by Java consumers, without a separate package script or module-specific Maven executions.

## What changed

- Extend `generate-java-rule-classes` to emit immutable static profile catalogs:
  - JavaScript and TypeScript profile names and rule keys in generated `AllChecks`, exposed through `CheckList`.
  - CSS profile rule keys in generated `CssRules`.
- Make `JavaScriptProfilesDefinition` and `CssProfileDefinition` activate built-in rules directly from those generated catalogs.
- Make the JavaScript, TypeScript, and CSS rule definitions use the generated Sonar way membership to set `activatedByDefault`.
- Preserve both RSPEC forms:
  - an array applies to every compatible language;
  - a language map applies membership independently to JavaScript and TypeScript.
- Preserve all generated profile names, empty per-language memberships, compatible-language filtering, registrar-provided rules, and security rules.
- Remove aggregate-profile infrastructure:
  - `tools/generate-profiles.ts` and its package script;
  - generated-profile Maven executions and resource directories in the JS/TS and CSS modules;
  - obsolete profile JSON constants/loaders and the old Java profile-index test;
  - unused profile fields and return values left in `deploy-rule-data.ts`.
- Keep a reduced `quality-profiles.ts` helper as the single implementation of array-vs-map membership semantics used by both Java and ESLint metadata generation, with four focused contract tests.
- Avoid the intermediate `RuleProfileCatalog` runtime implementation entirely; neither it nor its test remains.
- Keep `deploy-rule-data.ts`, because it still deploys the per-rule RSPEC JSON/HTML that remains the source of truth.
- Update build documentation to describe the static metadata generation flow.

The normal `RuleMetadataLoader` still reads `Sxxxx.json` for rule descriptions and other rule metadata. What disappears is the extra aggregate-profile generation and runtime I/O used only to recover profile membership.

## Verification

- `npm run generate-java-rule-classes`
- `npm run bbf`
- Full TypeScript suite: 2,347 tests
- `npx knip --no-gitignore`
- Maven reactor compile for `javascript-checks`, `css`, and `sonar-javascript-plugin`
- 25 focused Java profile and rule-definition tests

[JS-2211]: https://sonarsource.atlassian.net/browse/JS-2211


tools/generate-java-rule-classes.ts
```diff
@@ -254,7 +258,85 @@ function generateBody(
...
+function generateJavaScriptProfileRuleKeys(profiles: QualityProfileRuleKeys<'js' | 'ts'>): string {
+  return [...profiles]
+    .map(
+      ([profileName, rulesByLanguage]) =>
+        `Map.entry("${escapeJavaString(profileName)}", Map.of(` +
+        `Language.JAVASCRIPT, ${javaStringCollection('Set', rulesByLanguage.get('js') ?? [])}, ` +
+        `Language.TYPESCRIPT, ${javaStringCollection('Set', rulesByLanguage.get('ts') ?? [])}))`,
+    )
+    .join(',');
+}
```
nathsou, line 278: Nice!

tools/generate-java-rule-classes.ts
```diff
@@ -254,7 +258,85 @@ function generateBody(
...
 
+type RuleQualityProfileMetadata<Language extends string> = {
+  ruleKey: string;
+  compatibleLanguages: Language[];
+  defaultQualityProfiles?: DefaultQualityProfiles;
+};
+
+type QualityProfileRuleKeys<Language extends string> = Map<string, Map<Language, string[]>>;
+
+function buildQualityProfileRuleKeys<Language extends string>(
```
nathsou: The array-vs-object membership tests in the deleted `quality-profiles.test.ts` are still the JS-2211 contract, they just lost their import when `quality-profiles.ts` went away.

`profileNameToFileName` / `aggregateProfileRuleKeys` can stay gone. The cases that should come back are:

- array form applied to every compatible language
- empty `[]` excluding a rule from every language
- object maps choosing JS vs TS independently
- `isInSonarWay` treating a map as recommended when any language lists Sonar way

Those don’t need Maven. `buildQualityProfileRuleKeys` is private in a script with top-level `await`, so it can’t be imported without regenerating every Java check. `CheckListTest.testDefaultQualityProfiles` only asserts catalog keys ⊆ implemented keys, which cannot catch a shared rule (S5906, S5976, S8754) assigned to the wrong language.

Could you extract `buildQualityProfileRuleKeys` (and `isInSonarWay` if it isn’t already easy to import) into a small module and restore those tests there?

nathsou (approved):
Looks good — compiling profile membership into generated Java is a cleaner follow-up to #7763 than aggregate JSON. One non-blocking ask on the deleted generator tests.

tools/generate-java-rule-classes.ts
vdiez: Addressed in 4fef733d0. I extracted the shared profile-membership logic (`buildQualityProfileRuleKeys` and `isInSonarWay`) into the reduced `quality-profiles.ts` module and restored the four contract cases you called out. The full TypeScript suite passes (2,347 tests).

PR #7784, opened 2026-08-19
JS-2280 Fix S7785 false positive for promise chains stored for later consumption

oleksandr-selehenenko-sonarsource:
Part of 

----
## Summary by Gitar

- **Rule Updates:**
    - Fixed false positives in `S7785` for promise chains stored or passed for later consumption by checking contextual types via `isContextualTypeThenable` in `decorator.ts` and `type.ts`

### DIFF
--- a/packages/analysis/src/jsts/rules/S7785/decorator.ts
+++ b/packages/analysis/src/jsts/rules/S7785/decorator.ts
@@ -31,9 +31,9 @@ import {
 import * as meta from './generated-meta.js';
 
 /**
- * A TypeScript assertion wrapper (`x as T`, `x satisfies T`, `x!`, `<T>x`) forces a contextual
- * type onto the wrapped expression without actually consuming or awaiting it. Such a chain is
- * still floating, so the assertion must not be mistaken for storage-for-later-consumption.
+ * A TypeScript assertion wrapper (`x as T`, `x satisfies T`, `x!`, `<T>x`) only pins the type of
+ * the wrapped expression; it neither consumes nor awaits it. Such wrappers must be looked through
+ * to find the node that occupies the real syntactic slot when deciding whether a chain is stored.
  */
 function isAssertionWrapper(node: Rule.Node) {
   const type = node.type as string;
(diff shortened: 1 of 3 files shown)


francois-mora-sonarsource:
**False negative:** `isStoredForLaterConsumption` (`decorator.ts:79-89`) suppresses based on contextual type alone, but type assertions also produce a contextual type without actually consuming the value:

```ts
Promise.resolve(42).then(x => x) as Promise<number>;
```

This chain is floating — nothing awaits it — but the `as` assertion makes `isContextualTypeThenable` return true, so we wrongly suppress it. Upstream still flags this correctly.

Suggested fix: reject assertion wrappers before trusting the contextual type.

```ts
const chainCall = memberExpr.parent;
const isAssertionWrapper = (n: estree.Node | undefined) =>
  n?.type === 'TSAsExpression' || n?.type === 'TSSatisfiesExpression' || n?.type === 'TSNonNullExpression';

const isStoredForLaterConsumption =
  chainCall?.type === 'CallExpression' &&
  chainCall.callee === memberExpr &&
  !isAssertionWrapper(chainCall.parent) &&
  isContextualTypeThenable(chainCall, services);
```

Also worth a look: `unit.test.ts:171` — the `getValue()` valid case doesn't actually exercise the new logic (upstream already ignores calls inside a function via `isTopLevelCallExpression`), and its comment is misleading.

francois-mora-sonarsource (changes requested):
**False negative:** `isStoredForLaterConsumption` (`decorator.ts:79-89`) suppresses based on contextual type alone, but type assertions also produce a contextual type without actually consuming the value:

```ts
Promise.resolve(42).then(x => x) as Promise<number>;
```

This chain is floating — nothing awaits it — but the `as` assertion makes `isContextualTypeThenable` return true, so we wrongly suppress it. Upstream still flags this correctly.

Suggested fix: reject assertion wrappers before trusting the contextual type.

```ts
const chainCall = memberExpr.parent;
const isAssertionWrapper = (n: estree.Node | undefined) =>
  n?.type === 'TSAsExpression' || n?.type === 'TSSatisfiesExpression' || n?.type === 'TSNonNullExpression';

const isStoredForLaterConsumption =
  chainCall?.type === 'CallExpression' &&
  chainCall.callee === memberExpr &&
  !isAssertionWrapper(chainCall.parent) &&
  isContextualTypeThenable(chainCall, services);
```

Also worth a look: `unit.test.ts:171` — the `getValue()` valid case doesn't actually exercise the new logic (upstream already ignores calls inside a function via `isTopLevelCallExpression`), and its comment is misleading.

francois-mora-sonarsource (approved):
Both review findings addressed — thanks! Assertion-wrapper handling looks correct and is even more thorough than my suggestion (also covers `!` and `<T>x`), backed by tests in both directions. Vacuous test case removed.

PR #7799, opened 2026-08-20
JS-2288: Suppress S6757 false positives in class members

francois-mora-sonarsource:
## Jira
JS-2288

## Scope
Suppress this reports lexically owned by any class member, including arrow closures. Retain reports across nested non-arrow function boundaries and in functional components.

## Validation
- npx tsx --test packages/analysis/src/jsts/rules/S6757/**/*.test.ts
- npm run bbf

## Peach baselines
- JS-2288-javascript-S6757-peach-before-20260820T092204Z.csv
- JS-2288-typescript-S6757-peach-before-20260820T092204Z.csv

No RSPEC change required.

### DIFF
--- a/packages/analysis/src/jsts/rules/S6757/decorator.ts
+++ b/packages/analysis/src/jsts/rules/S6757/decorator.ts
@@ -48,21 +48,33 @@ function isThisMemberExpression(
 }
 
 function isLexicallyBoundToClassMember(node: TSESTree.MemberExpression): boolean {
+  let previous: TSESTree.Node = node;
+
   for (const current of ancestorsChain(node, new Set<string>())) {
     if (isClassMember(current)) {
-      return true;
+      // Only the member value owns the instance receiver. A computed key or a member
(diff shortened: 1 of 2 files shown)


oleksandr-selehenenko-sonarsource (approved):
LGTM
Just a few moments:
1. Add a test for a function expression:
```
class Service {
  build() {
    const Row = function () {
      return <li>{this.props.value}</li>;
    };

    return <Row />;
  }
}
```

2. A computed-key test for accessor:
```
function MyComponent() {
  class Holder {
    accessor [this.props.key] = 1;
  }

  return <div>{Holder.name}</div>;
}
```

3. Maybe a getter case
```
class View {
  get rows() {
    const Row = item => <span>{this.format(item)}</span>;
    return <section>{[1].map(Row)}</section>;
  }
}
```

4. Static blocks
`static { this.x }` isn't handled, `this` there is the class itself. It might be nice to handle it, or at least add a unit test to pin the current behavior.

5. Decorators
The comment mentions member decorators as a reason for the `previous === current.value` check, but there's no test for it.

PR #7802, opened 2026-08-21
JS-2298 Don't raise S2486 when the unused catch clause contains only comments

oleksandr-selehenenko-sonarsource:
Part of JS-2298

Rspec PR: https://github.com/SonarSource/rspec/pull/7955

## What

`S2486` no longer raises when an unused `catch` clause is empty apart from comments. A `catch` block that contains no statements but holds at least one non-blank comment is treated as documented, and therefore compliant.

The issue message now offers the comment as a third way to resolve the issue: *"Handle this exception, don't catch it at all, or explain in a comment why it is ignored."*

## Notes

Comment **content** is not inspected beyond being non-empty. This is a deliberate call, matching ESLint core's `no-empty` and the other analyzers (`java`, `csharp`, `python`, `dart`, `rpg`): linter directives (e.g. `// eslint-disable-next-line no-empty`) and commented-out code (e.g. `// setError(err.message)`) will therefore also suppress the rule. Filtering directive prefixes was considered and rejected — it would be a maintained list that drifts and diverges from the other analyzers.

A comment with no textual content (a bare `//`) does **not** count as documentation and still raises.

----
## Summary by Gitar

- **Bug Fixes:**
    - Updated `S2486` rule to suppress warnings when an unused `catch` clause contains only comments
    - Added comprehensive test fixtures covering single-line, multi-line, and block comment variations

### DIFF
--- a/its/ruling/src/test/expected/angular.js/javascript-S2486.json
+++ b/its/ruling/src/test/expected/angular.js/javascript-S2486.json
@@ -2,9 +2,6 @@
 "angular.js:src/Angular.js": [
 1154
 ],
-"angular.js:src/ng/compile.js": [
-2778
-],
 "angular.js:src/ngSanitize/sanitize.js": [
 478
 ]
--- a/its/ruling/src/test/expected/console/typescript-S2486.json
+++ b/its/ruling/src/test/expected/console/typescript-S2486.json
@@ -6,9 +6,6 @@
 142
 ],
 "console:src/views/PermissionsView/PermissionPopup/ast.ts": [
-29,
-55,
-86,
 126
 ]
 }
--- a/its/ruling/src/test/expected/console/typescript-S2486.json
+++ b/its/ruling/src/test/expected/console/typescript-S2486.json
@@ -6,6 +6,9 @@
 142
 ],
 "console:src/views/PermissionsView/PermissionPopup/ast.ts": [
+29,
+55,
+86,
 126
 ]
 }
--- a/its/ruling/src/test/expected/courselit/typescript-S2486.json
+++ b/its/ruling/src/test/expected/courselit/typescript-S2486.json
@@ -1,7 +1,4 @@
 {
-"courselit:apps/web/components/admin/courses/lesson-editor.tsx": [
-168
-],
 "courselit:apps/web/components/admin/media/media-preview.tsx": [
 102
 ],
--- a/its/ruling/src/test/expected/desktop/typescript-S2486.json
+++ b/its/ruling/src/test/expected/desktop/typescript-S2486.json
@@ -3,8 +3,7 @@
 37
 ],
 "desktop:app/src/lib/api.ts": [
-1155,
-1204
+1155
 ],
 "desktop:app/src/lib/ci-checks/ci-checks.ts": [
 81
@@ -30,9 +29,6 @@
 ],
 "desktop:app/src/main-process/exception-reporting.ts": [
 82
-],
-"desktop:app/src/main-process/log.ts": [
-80
 ],
 "desktop:app/src/main-process/main.ts": [
 261
--- a/its/ruling/src/test/expected/eigen/typescript-S2486.json
+++ b/its/ruling/src/test/expected/eigen/typescript-S2486.json
@@ -5,14 +5,8 @@
 "eigen:src/app/Scenes/MyAccount/MyAccountDeleteAccount.tsx": [
 122
 ],
-"eigen:src/app/Scenes/MyAccount/updateMyUserProfile.ts": [
-68
-],
 "eigen:src/app/relay/RelayCache.ts": [
 59
-],
-"eigen:src/app/store/persistence.ts": [
-58
 ],
 "eigen:src/app/utils/LinkedAccounts/facebook.ts": [
 81
--- a/its/ruling/src/test/expected/rxjs/typescript-S2486.json
+++ b/its/ruling/src/test/expected/rxjs/typescript-S2486.json
@@ -1,6 +1,5 @@
 {
 "rxjs:src/observable/dom/AjaxObservable.ts": [
-49,
 54
 ]
 }
--- a/its/ruling/src/test/expected/sonar-web/javascript-S2486.json
+++ b/its/ruling/src/test/expected/sonar-web/javascript-S2486.json
@@ -1,8 +1,5 @@
 {
 "sonar-web:src/main/js/components/common/file-upload.js": [
 33
-],
-"sonar-web:src/main/js/components/workspace/models/items.js": [
-24
 ]
 }
--- a/its/ruling/src/test/expected/tailwindcss/javascript-S2486.json
+++ b/its/ruling/src/test/expected/tailwindcss/javascript-S2486.json
@@ -5,9 +5,6 @@
 "tailwindcss:src/index.postcss7.js": [
 43
 ],
-"tailwindcss:src/lib/getModuleDependencies.js": [
-32
-],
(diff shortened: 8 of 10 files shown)


francois-mora-sonarsource (changes requested):
Thanks for this — the direction is right and the evidence backs it up. `java`, `csharp`, `python`, `dart` and `rpg` all document *"When a block contains a comment, it is not considered to be empty"* for S2486, so this closes a gap where JavaScript was the outlier. The `node.body.body.length === 0` guard also matches how java/csharp/rpg scope it, and the predicate is byte-identical to ESLint core's `no-empty` (`no-empty.js:74`), which is good precedent to follow.

I checked the ruling diff independently and it is sound in shape: 12 removals (195 → 183), all 8 JSON files parse, no dangling commas, and the emptiness gate demonstrably works — `desktop:app/src/lib/api.ts:1155` correctly kept, and in `rxjs` line 49 removed while line 54 stayed.

Requesting changes on two blocking items plus a few smaller ones.

## Blocking

### 1. The companion RSPEC PR is missing

`rules/S2486/javascript/rule.adoc` documents exactly one exception — the single-statement `try` from JS-1995 — and says nothing about comments. I searched: there is no open or merged RSPEC PR for JS-2298, and no remote branch matching `S2486`/`2298`/`catch`.

Rule descriptions are pulled from RSPEC at build time (`docs/rspec-sync-requirements.md:30`), so as it stands SonarQube would render a description whose Exceptions section contradicts the analyzer.

The precedent is the adjacent exception in the same function: SonarJS #7445 and RSPEC #7306 were merged the same day. Could you open the RSPEC PR and link it here? It needs two edits — see item 4 for the second.

Suggested wording for `=== Exceptions`, phrased around intent so it stays true under item 3:

> A `catch` block containing no statements is not considered empty when it holds a comment explaining why the exception is ignored.

Worth adding a compliant example too, as `java` and `dart` both do.

### 2. Six pre-existing Compliant fixtures no longer test anything

`cb.fixture.js:4, 14, 39, 47, 55, 67` (`f`, `g`, `assignmentBody`, `variableDeclarationBody`, `throwBody`, `nestedTryBody`) all have the shape:

```js
} catch (err) { // Compliant

}
```

The `// Compliant` marker now sits inside the empty catch, so `isDocumentedEmptyCatch` short-circuits and `isSingleSimpleStatement` is never reached. These tests pass because of the annotation, not the code under test.

I confirmed this by mutation rather than inspection — deleting the entire `isSingleSimpleStatement` branch from `rule.ts` leaves the fixture green (`pass 1, fail 0`). So the PR silently removes regression coverage for the exception that shipped in #7445.

Same fix as the Noncompliant cases got — move the marker below the closing brace, which is safe because a comment after `}` is outside `node.body`:

```js
} catch (err) {

} // Compliant: try guards a single simple statement
```

Please re-run the mutation afterwards: those six should then **fail** with the branch removed. Any that still passes is shadowed by something else.

## Requested changes

### 3. Require non-blank comment text

```ts
context.sourceCode.getCommentsInside(node.body).some(c => c.value.trim().length > 0)
```

Three of the twelve ruling removals are `console:src/views/PermissionsView/PermissionPopup/ast.ts` lines 29, 55 and 86, where the entire comment is `    //` — zero characters of content. That documents nothing.

We deliberately went no further: linter directives and commented-out code will still suppress (see item 7). Filtering directive prefixes was considered and rejected as a maintained list that would drift and would diverge from the other analyzers.

Note this needs JS-2298 amended — the ticket description currently prescribes `getCommentsInside(...)` returning "at least one comment".

### 4. Update the issue message

Now that a comment is a third way to resolve the issue, the message should say so — `csharp` already does this. Agreed wording:

> Handle this exception, don't catch it at all, or explain in a comment why it is ignored.

11 occurrences, all in `S2486/rule.ts:54` and `S2486/cb.fixture.js`. The rspecator-view `=== Message` section of the RSPEC PR needs to match exactly.

### 5. Pin the residual behaviour in fixtures

```js
function pragmaOnlyCatch() {
    try {
        doSomething();
        doSomethingElse();
    } catch (err) {
        // eslint-disable-next-line no-empty
    } // Compliant: comment content is not inspected beyond being non-empty
}

function emptyCommentCatch() {
    try {
        doSomething();
        doSomethingElse();
    } catch (err) { // Noncompliant
        //
    }
}
```

The second is the executable form of item 3 — nothing would currently catch a regression that reverted it.

Three genuinely uncovered cases are also worth adding:

- a non-Identifier catch param (`catch ({ message })` raises nothing today — the selector is `CatchClause[param.type="Identifier"]` — and no fixture covers it)
- `EmptyStatement` + comment (`catch (e) { ; // c }` still raises, since `body.length === 1`)
- `catch (e) /* c */ {` (still raises — the exact `getCommentsInside` boundary)

### 6. Drop four redundant fixtures

`commentOnlyMultiLine`, `blockCommentOnlyMultiLine`, `commentAndCodeMultiLine` and `commentTrailingCodeSameLine` are the same input to the rule as their single-line counterparts. `blockCommentOnlyMultiLine` especially: a multi-line `/* */` is a *single* comment node, so it is literally identical to `blockCommentOnlySingleLine`. Suggest keeping `commentOnlySingleLine`, `blockCommentOnlySingleLine` and `commentAndCodeSingleLine`.

### 7. Fill in the PR body

It is currently a dangling `Part of ` plus the un-deleted template comment and the Gitar summary. Worth stating explicitly that comment *content* is not inspected beyond being non-empty, so linter directives and commented-out code suppress the rule — this matches ESLint `no-empty` and the other analyzers, and it is a deliberate call rather than an oversight. Two cases in the ruling diff show it in the wild:

- `tailwindcss:src/lib/getModuleDependencies.js:32` — the only comment is `// eslint-disable-next-line no-empty`
- `courselit:apps/web/components/admin/courses/lesson-editor.tsx:168` — the comment is `// setError(err.message)`, i.e. disabled error handling

Please also link the RSPEC PR from the body.

## Knock-on

Item 3 changes the ruling: removals drop from 12 to 9, and `console` 29/55/86 come back as expected issues.

---

Everything else checked out — no Java check class, rule option, `index.ts`/`meta.ts` or CHANGELOG change is needed, the `Noncompliant@-2` offsets are correct, and CI is green.

francois-mora-sonarsource (approved):
Thanks — all seven items addressed, and I verified each rather than taking the diff at face value.

The two coverage regressions are now genuinely pinned, confirmed by mutation:

- Removing the `isSingleSimpleStatement` branch makes the fixture **fail** (it passed before), with lines 4, 14, 39, 47, 55, 67 among the failures — the six formerly-shadowed cases.
- Reverting the predicate to `.length > 0` also **fails**, so the non-blank-comment behaviour is pinned by `emptyCommentCatch`.

Nice touches beyond what I asked for: relocating the `commentOnlySingleLine` / `blockCommentOnlySingleLine` markers too (the inside marker was the comment under test there as well), and putting the annotation outside the brace in `emptyCommentCatch`, which it has to be now that a `// Noncompliant` inside would count as non-blank content.

On the RSPEC side, #7955 has both edits and the `=== Message` text matches `rule.ts:54` byte-for-byte, so there is no spec/analyzer drift. Placing the new exception before the single-statement one reads better than the order I suggested.

CI green on `8d38e4357` (34 success, 0 failures), including `JS/TS Ruling` and `Ruling Test` against the regenerated console expectations.

One item left outside this PR: JS-2298's description still prescribes `getCommentsInside(...)` returning "at least one comment", which the non-blank predicate no longer matches — worth a one-line amendment so the ticket matches what shipped.

LGTM.

PR #7824, opened 2026-08-26
JS-2327 Avoid preloading unrelated generated-source config files

vdiez:
Part of

## Summary

- keep broad generated-source config candidates path-only during filesystem traversal
- load only configs declared by discovered package scripts during post-processing
- reuse content already read by another file store instead of reading a selected config twice

## Motivation

Generated-source initialization could retain every JavaScript, TypeScript, JSON, and YAML file under the project base directory even when the scanner supplied only a small authoritative input set. Large repositories with unrelated data files could therefore exhaust the Node.js heap before file analysis started.

## Testing

- `npx tsc -b packages`
- `npm run bbf`
- targeted file-store and generated-source tests
- `npm run bridge:test` (2,354 tests)

### DIFF
--- a/packages/analysis/src/file-stores/index.ts
+++ b/packages/analysis/src/file-stores/index.ts
@@ -19,7 +19,7 @@ import { DependencyManifestStore } from './dependency-manifests.js';
 import { GeneratedSourceStore } from './generated-sources/index.js';
 import { TsConfigStore } from './tsconfigs.js';
 import { findFiles } from '../common/find-files.js';
-import type { FileStore } from './store-type.js';
+import type { FileStore, FileStoreContext } from './store-type.js';
 import type { Configuration } from '../common/configuration.js';
 import {
   isRoot,
@@ -63,6 +63,8 @@ export async function initFileStores(configuration: Configuration, inputFiles?: 
     return;
   }
 
+  const context = createFileStoreContext(inputFiles);
+
   for (const store of pendingStores) {
     store.setup(configuration);
   }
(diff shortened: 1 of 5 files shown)


packages/analysis/src/file-stores/index.ts
```diff
@@ -130,19 +134,51 @@ async function processPendingFileStores(
...
+  const files = new Map<NormalizedAbsolutePath, Promise<File>>();
+
+  return {
+    retainFile(filePath) {
+      retainedPaths.add(filePath);
+    },
+    getFile(filePath, fileContent) {
+      let file = files.get(filePath);
+      if (!file) {
+        const inputFile = inputFiles?.[filePath];
```
erwan-leforestier-sonarsource, line 170: retainFile only activates caching when it is called before the first getFile for a path, but the type does not enforce that order. If a future store reads via getFile before another calls retainFile for the same path, caching silently will not kick in and the file gets read twice. Could retainFile return the handle, or the ordering be documented on the type?

packages/analysis/src/file-stores/store-type.ts
```diff
@@ -18,7 +18,17 @@ import type { File, NormalizedAbsolutePath } from '../../../shared/src/helpers/f
...
 
-export type FileProcessingMode = 'path' | 'content';
+/**
+ * `deferred-content` processes the path during the walk while preserving any shared content read
+ * for a later request from `postProcess`.
+ */
+export type FileProcessingMode = 'path' | 'content' | 'deferred-content';
+
+export type FileStoreContext = {
+  /** Preserve a file if another store reads it, without triggering a read itself. */
```
erwan-leforestier-sonarsource: This comment reads backwards: retention must be declared before any getFile for the path, so calling retainFile afterwards is a no-op for the cache. Something like "Mark a file path so its content is cached on first read" would be clearer.

packages/analysis/src/file-stores/index.ts
vdiez: Good point. The current dispatcher guarantees the ordering: it evaluates all `wantsFile` requests, calls `retainFile` for every `deferred-content` request, and only then performs any `getFile` call. Stores receive the context only during `postProcess`, after traversal has already declared retention. I agree that this invariant is not clear from the type; documenting the requirement to call `retainFile` before the first `getFile` is preferable to adding a handle here.

packages/analysis/src/file-stores/store-type.ts
vdiez: Agreed. The intended contract is: ‘Mark a file path so its content is cached on its first `getFile` call; this must happen before the first `getFile` for that path.’ That is clearer than the current wording.

erwan-leforestier-sonarsource (approved):

PR #7826, opened 2026-08-27
JS-2329 Fix test-file heuristic to preserve source metrics

vdiez:
Part of

## Summary

- Keep scanner/config source classification authoritative for metrics, highlighting, CPD, and analysis scope.
- Introduce a separate classification for rule selection when the fallback test-file heuristic matches.
- Cover JS/TS and CSS files that look like tests while remaining `MAIN` files.

## Testing

- `npm run generate-meta`
- `npm run bridge:compile`
- Targeted analysis, filtering, sanitization, and source-store tests (81 passed)
- Prettier check on all changed files

### DIFF
--- a/packages/analysis/src/common/filter/filter-path.ts
+++ b/packages/analysis/src/common/filter/filter-path.ts
@@ -79,11 +79,11 @@ export function filterPathAndGetFileType(
 }
 
 /**
- * Determines which rule set applies to a file. The filename heuristic deliberately affects only
+ * Returns the file type used for rule selection. The filename heuristic deliberately affects only
  * this classification: metrics and other file artifacts continue to use the scanner/path-derived
  * file type.
  */
-export function getRuleFileType(
+export function getFileTypeForRules(
   filePath: NormalizedAbsolutePath,
   fileType: FileType,
   params: FilterPathParams,
--- a/packages/analysis/src/common/filter/filter-path.ts
+++ b/packages/analysis/src/common/filter/filter-path.ts
@@ -67,7 +67,7 @@ export function filterPathAndGetFileType(
   filePath: NormalizedAbsolutePath,
   params: FilterPathParams,
 ): FileType | undefined {
-  if (matchesTestPath(filePath, params, true)) {
+  if (matchesTestPath(filePath, params)) {
     return 'TEST';
   }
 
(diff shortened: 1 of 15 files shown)


packages/analysis/src/common/filter/filter-path.ts
```diff
@@ -78,43 +78,32 @@ export function filterPathAndGetFileType(
   debug(`File ignored due to analysis scope filters: ${filePath}`);
 }
 
+/**
+ * Determines which rule set applies to a file. The filename heuristic deliberately affects only
+ * this classification: metrics and other file artifacts continue to use the scanner/path-derived
+ * file type.
+ */
+export function getRuleFileType(
```
erwan-leforestier-sonarsource: Name and descriptions are confusing. I suggest this:


```suggestion
/**
 * Returns the file type used for rule selection. The filename heuristic deliberately affects only
 * this classification: metrics and other file artifacts continue to use the scanner/path-derived
 * file type.
 */
export function getFileTypeForRules(
```

packages/analysis/src/common/filter/filter-path.ts
```diff
@@ -128,6 +117,27 @@ function matchesTestPath(
...
+  // `sonar.inclusions` narrows `sonar.sources`, so a matching file has been explicitly opted into
+  // MAIN rule selection and should not be second-guessed by the filename heuristic.
+  if (
+    inclusions.length > 0 &&
+    fileIsUnder(filePath, sourcesPaths) &&
+    inclusions.some(inclusion => inclusion.match(filePath))
+  ) {
+    return false;
+  }
+  return isTestRelatedFile(filePath, testFileExtensions);
```
erwan-leforestier-sonarsource, line 138: Not new in this PR, but documenting since the heuristic moved here. It only runs when `sonar.tests` is unset, and it is all-or-nothing. As soon as `sonar.tests` is set to anything, the heuristic is off for every file, so co-located tests silently become main code. A customer who adds `sonar.tests=test/` just to classify their integration tests unknowingly changes how their co-located unit tests are treated.

```
sonar.sources = src
sonar.tests   = test
src/Button.test.tsx     -> MAIN rules  (heuristic off, treated as main)
test/api.test.ts        -> TEST rules  (matched by path)
```

Legacy behavior, kept as-is. Flagging for a possible follow-up.

packages/analysis/src/common/filter/filter-path.ts
```diff
@@ -78,43 +78,32 @@ export function filterPathAndGetFileType(
...
+): FileType {
+  if (fileType === 'TEST' || !matchesTestFileHeuristic(filePath, params)) {
+    return fileType;
+  }
+
+  debug(
+    `Test file detected: ${filePath}. If this file should not use test rules, please configure sonar.tests or adjust your sonar.sources/sonar.inclusions to explicitly include it as MAIN.`,
+  );
+  return 'TEST';
+}
```
erwan-leforestier-sonarsource: A heuristic-matched file now stays `fileType=MAIN`, so it is counted as main code for metrics even though it gets TEST rules. Before this PR it was fully TEST and excluded from those metrics. The side effect is that a co-located test with no coverage entry now inflates main LOC and reads as 0%-covered main code, which drags down coverage.

```
sonar.sources = src   (sonar.tests unset)
src/Button.test.tsx

  before: fileType=TEST  -> excluded from main metrics
  after:  fileType=MAIN, ruleFileType=TEST
          -> TEST rules, but counted in main ncloc + coverage + duplication
```

Correct per the ticket, but likely to show up as a "why did my coverage drop?" report, so worth a changelog/doc note.

packages/analysis/src/common/filter/filter-path.ts
vdiez: Agreed. `getFileTypeForRules` makes the distinction much clearer. I applied the suggested name and JSDoc in `6bf20af7d9`.

packages/analysis/src/common/filter/filter-path.ts
vdiez: Good point. This is the existing fallback semantics, and the PR keeps them intentionally; there is already a regression test covering that `sonar.tests` disables the heuristic. I agree this should be handled separately because changing it requires a product decision about whether explicit test configuration remains authoritative or can be supplemented by the filename heuristic.

packages/analysis/src/common/filter/filter-path.ts
vdiez: Agreed. This is the intended consequence of JS-2329: the scanner/path-derived `fileType` remains authoritative for metrics, while the heuristic changes only rule selection. That means heuristic-matched MAIN files can now contribute to the coverage denominator. I agree the customer-visible impact should be called out in the Jira/release notes.

erwan-leforestier-sonarsource (approved):

PR #7835, opened 2026-08-28
JS-2333 Implement rule S9379: DOM elements should not use the "autofocus" attribute

guillemsarda:
----
## Summary by Gitar

    - Implemented rule `S9379` to flag DOM elements using the `autofocus` attribute
    - Added unit tests covering valid and invalid autofocus usage across modal dialogs and non-DOM components


packages/analysis/src/jsts/rules/S9379/decorator.ts
```diff
@@ -0,0 +1,130 @@
...
+import { getElementType, getRole } from '../helpers/accessibility.js';
+import * as meta from './generated-meta.js';
+
+// A capitalized keyword occurring anywhere in a PascalCase identifier is a real word boundary
+// by convention, so `CustomModal`, `DialogContent`, and `AppModalWrapper` all match; the
+// trailing negative lookahead rejects a lowercase continuation right after the keyword, so
+// `Dialogue`/`Modality` do not. `Sheet` is deliberately not in this list: it collides with too
+// many unrelated names (`StyleSheet`, `TimeSheet`, `SpreadSheet`, `BalanceSheet`) to be a
+// reliable modal signal, unlike the others.
+const MODAL_COMPONENT_NAME_PATTERN = /(?:Modal|Dialog|Popup|Drawer|Popover)(?![a-z])/;
```
martin-strecker-sonarsource: **[blocker]** Out of scope and overreaching: `MODAL_COMPONENT_NAME_PATTERN`/`MODAL_TRIGGER_NAME_PATTERN` (and their use in `isExemptFromReport`/`isModalOpeningElement`) aren't in JS-2333's design decision or in Biome's `noAutofocus` (which the ticket explicitly asks to mirror) -- both only exempt `<dialog>` descendants and `popover` elements. Guessing "is this a modal" from an identifier string is inherently unreliable and unlike anything else in this codebase's decorators (S6842 -- the decorator this PR mirrors for shape -- only ever checks real tags/roles/attributes, never names). It also risks diverging from the sonar-html companion check, which the ticket requires to report consistently on equivalent markup. gitar-bot's own review needed three rounds to patch bugs in this exact mechanism (whole-subtree suppression via `StyleSheetManager`, an inert trigger exclusion, an anchoring regression) -- recommend dropping it and relying on the dialog/popover/role checks only.

packages/analysis/src/jsts/rules/S9379/decorator.ts
```diff
@@ -0,0 +1,130 @@
...
+// Trigger/opener wrappers live in the page, not in the overlay, so autofocusing them, or an
+// element inside them, is a real issue even when they sit inside a matching modal ancestor.
+const MODAL_TRIGGER_NAME_PATTERN = /(?:Trigger|Button|Link|Toggle|Opener)$/;
+
+// ARIA roles that mark a hand-rolled (non-native-`<dialog>`) modal, per the WAI-ARIA spec.
+const MODAL_ROLES = new Set(['dialog', 'alertdialog']);
```
martin-strecker-sonarsource, line 29: **[suggestion]** Same scope question, lower risk: `role="dialog"`/`"alertdialog"` exemption isn't in JS-2333's spec either, though it's deterministic and consistent with how S6842 handles roles (and Angular ESLint added the same carve-out after user reports). Worth confirming this is an intentional scope increase, and that sonar-html gets the equivalent treatment.

packages/analysis/src/jsts/rules/S9379/decorator.ts
```diff
@@ -0,0 +1,130 @@
...
+ * the page rather than in the overlay it opens.
+ */
+export function decorate(rule: Rule.RuleModule): Rule.RuleModule {
+  return interceptReportForReact(
+    {
+      ...rule,
+      meta: generateMeta(meta, rule.meta),
+    },
+    (context, reportDescriptor) => {
+      const opening = openingElementOf(reportDescriptor);
```
martin-strecker-sonarsource, line 52: **[suggestion]** `rules['no-autofocus']` is used with no options, so jsx-a11y's `ignoreNonDOM` defaults to `false` -- this will also flag `autoFocus`-named props on custom components (e.g. a design-system `<Input autoFocus>`) that have nothing to do with real DOM focus. Well-documented FP class upstream (jsx-a11y added `ignoreNonDOM` for it; Angular ESLint added an equivalent option after angular-eslint#539). One-line guard here -- skip reporting when the opening element isn't a native DOM tag -- before the modal-exemption logic runs.

packages/analysis/src/jsts/rules/S9379/unit.test.ts
```diff
@@ -0,0 +1,84 @@
...
+ * along with this program; if not, see https://sonarsource.com/license/ssal/
+ */
+import { doesNotThrow } from 'node:assert';
+import { describe, it } from 'node:test';
+import { rule } from './index.js';
+import { NoTypeCheckingRuleTester } from '../../../../tests/jsts/tools/testers/rule-tester.js';
+
+const VALID_CASES = [
+  { code: `<div />;` },
+  // Autofocusing inside a `dialog` or a `popover` is exempt, direct child or nested descendant.
```
martin-strecker-sonarsource, line 24: **[nitpick]** This comment claims "direct child or nested descendant" for the whole block, but only the `<dialog>` pair below (lines 25-26) actually tests a nested descendant. The `popover` case (line 27) and both ARIA-role cases (lines 39-40) only cover the direct-child variant. `isExemptFromReport`'s ancestor walk is generic across all three exemption paths, so nesting almost certainly already works for these too -- just add the missing nested variants (e.g. `<div popover><div><input autoFocus /></div></div>`) so the coverage matches the comment's claim.

packages/analysis/src/jsts/rules/S9379/decorator.ts
```diff
@@ -0,0 +1,130 @@
...
+  opening: TSESTree.JSXOpeningElement,
+): boolean {
+  const tag = getElementType(context)(opening);
+  if (tag.toLowerCase() === 'dialog') {
+    return true;
+  }
+  if (getProp((opening as unknown as JSXOpeningElement).attributes, 'popover') !== undefined) {
+    return true;
+  }
+  const role = getRole(opening);
```
martin-strecker-sonarsource, line 117: **[FYI]** Small footgun for later maintainers: `getProp` (line 122) is jsx-ast-utils-x's own primitive and returns `undefined` when the attribute is absent -- there's no `getRole` in jsx-ast-utils-x at all, it's a SonarJS-local helper in `helpers/accessibility.ts` that independently chose `string | null` as its "no role" sentinel. The two checks happen to sit three lines apart using different sentinels (`!== undefined` vs `!== null`) purely because of that mismatch, not because of any semantic difference. Correct as written, but if someone later copy-pastes the `!== undefined` idiom onto a `getRole` call it will silently always be true.

its/ruling/src/test/expected/ant-design/javascript-S9379.json
```diff
@@ -0,0 +1,8 @@
+{
```
martin-strecker-sonarsource: **[FP]** Checked both entries in this file against jsts-test-sources@81a52e2c0: `Password.test.js:58` (`<Input.Password autoFocus>`) and `textarea.test.js:328` (`<TextArea autoFocus>`). Both are FP-nonDOM -- `Input.Password`/`TextArea` are this library's own custom wrapper components, not raw native elements, and `ignoreNonDOM` defaults to false here (see the earlier comment on decorator.ts line 64). Both hits are also inside unit tests asserting on rendered markup rather than production usage. Both would be fixed by the non-DOM guard suggested on decorator.ts:64 -- neither `Input.Password` nor `TextArea` resolves to a real DOM tag.

its/ruling/src/test/expected/ant-design/typescript-S9379.json
```diff
@@ -0,0 +1,11 @@
+{
+"ant-design:components/input/__tests__/index.test.tsx": [
+115
```
martin-strecker-sonarsource: **[FP]** `index.test.tsx:115` (`<Input ref={ref} autoFocus>`) is the same FP-nonDOM pattern as the other ant-design entries -- custom `Input` wrapper, inside a unit test -- and would be fixed by the non-DOM guard suggested on decorator.ts:64. (The third entry in this file, `IconDisplay/index.tsx:126`, is the one case in this whole sample that's arguably a genuine TP -- a docs-site search box statically autofocused on page load -- but it's still technically non-DOM, `Input.Search`, so the same guard would incidentally suppress it too -- worth flagging as a real tradeoff of that fix, not just an upside.)

its/ruling/src/test/expected/ant-design/typescript-S9379.json
```diff
@@ -0,0 +1,11 @@
...
+],
+"ant-design:components/modal/ConfirmDialog.tsx": [
+76
```
martin-strecker-sonarsource: **[FP]** This one is an unresolvable blind spot in the exemption mechanism itself, not just premise fallout: `ConfirmDialog.tsx`'s cancel button is built via `const cancelButton = <ActionButton autoFocus={autoFocusButton === 'cancel'} .../>` and only spliced into `<Dialog>` later via `{cancelButton}`. Because it's composed through an intermediate variable, the button's JSX node's `.parent` chain never actually includes `<Dialog>` in the parsed AST, so `isExemptFromReport`'s ancestor walk can't see it -- even though it unquestionably renders inside the modal at runtime. Proof: the sibling OK-button, written inline directly inside `<Dialog>` with no variable indirection, correctly does NOT appear in this snapshot. Extracting a JSX subtree into a local `const` before referencing it is an extremely common React idiom (conditional buttons/rows) -- this blind spot will misfire on real code far beyond this one file. Note: `ActionButton` also happens to be a non-DOM component, so the guard suggested on decorator.ts:64 would incidentally suppress this specific instance -- but only by coincidence, not because it fixes the underlying issue. Resolving it in general would require tracing the variable's usage back to its assignment -- i.e. data-flow/symbolic execution -- which is out of scope for this decorator. Recommend documenting this as a known limitation rather than trying to patch it here.

its/ruling/src/test/expected/console/typescript-S9379.json
```diff
@@ -0,0 +1,107 @@
+{
```
martin-strecker-sonarsource, line 1: **[FP]** Sampled all 41 entries in this file against jsts-test-sources@81a52e2c0. 39 are FP-reactive (the input/button only mounts when the user triggers an edit/add/search action -- e.g. all 7 `DatabrowserView/Cell/*.tsx` hits are `getEditCell` results that only render once a table cell enters edit mode; `DatabrowserView.tsx:267` and `FieldPopup/BaseSettings.tsx:290` are explicit `{condition && <input autoFocus>}`; most others are `state.isEntering*`/`editing*` ternaries). The remaining 2 (`AfterSignUpView.tsx:176`, `AlgoliaIndexPopupHeader.tsx:31`) are FP-nonDOM (`CustomInput`/`IndexInput` wrapper components) and would be fixed by the non-DOM guard suggested on decorator.ts:64. Zero confirmed TPs in this file -- every hit is a plausible false positive under the rule's current all-static-autofocus premise, but note the guard only accounts for 2 of the 41; the 39 FP-reactive ones are a distinct, unaddressed problem (timing of mount, not DOM-ness).

its/ruling/src/test/expected/desktop/typescript-S9379.json
```diff
@@ -0,0 +1,26 @@
+{
```
martin-strecker-sonarsource, line 1: **[FP]** Checked all 8 entries against jsts-test-sources@81a52e2c0. 6 are FP-nonDOM: `diff-search-input.tsx:37`, `authentication-form.tsx:107`, `enterprise-server-entry.tsx:58`, `fancy-text-box.tsx:44`, `filter-list.tsx:253`, `two-factor-authentication.tsx:81` all set `autoFocus` on this app's own `<TextBox>` wrapper, not a native element -- all 6 would be fixed by the non-DOM guard suggested on decorator.ts:64. The other 2 (`text-area.tsx:77`, `text-box.tsx:305`) are that wrapper's own internal `<textarea>`/`<input autoFocus={this.props.autoFocus}>` -- the real DOM mount point every FP-nonDOM hit above ultimately forwards into -- and would still be reported after the guard, since they're genuinely native elements. Net effect: the same underlying autofocus gets counted once at the true DOM boundary and again at every non-DOM pass-through call site; the guard collapses that down to just the one DOM-level report per wrapper.

its/ruling/src/test/expected/eigen/typescript-S9379.json
```diff
@@ -0,0 +1,51 @@
+{
```
martin-strecker-sonarsource: **[FP]** Checked all 16 files (17 hits, one file has 2) against jsts-test-sources@81a52e2c0. All 17 are FP-nonDOM: every single one sets `autoFocus` on a locally-defined wrapper component (`Input`, `PhoneInput`, `SearchInput`, `InquiryTextInput`), never a raw native/`TextInput`-from-'react-native' element -- all 17 would be fixed by the non-DOM guard suggested on decorator.ts:64. Several are also explicitly-conditional, accessibility-aware focus management, not careless static autofocus -- e.g. `OnboardingLogin.tsx:93`/`123` has a code comment explaining the conditional autofocus is driven by navigation params, and `RequestForPriceEstimateForm.tsx:74` uses `autoFocus={!values.requesterPhoneNumber}`. Zero confirmed TPs in this file.

its/ruling/src/test/expected/jira-clone/javascript-S9379.json
```diff
@@ -0,0 +1,11 @@
+{
```
martin-strecker-sonarsource: **[FP]** Checked all 3 entries against jsts-test-sources@81a52e2c0. All are FP: `BodyForm/index.jsx:34` and `IssueSearch/index.jsx:54` are FP-nonDOM (`Textarea`/`SearchInputDebounced` wrapper components); `Dropdown.jsx:183` is both FP-nonDOM and FP-reactive -- a textbook 'focus the search box when the dropdown opens' pattern on a non-native `DropdownInput`. All 3 would be fixed by the non-DOM guard suggested on decorator.ts:64, since none of the three tags resolve to a real DOM element -- for `Dropdown.jsx:183` that's coincidental (it suppresses the finding via the DOM check, not via anything that addresses the reactive-mount timing).

its/ruling/src/test/expected/redux/javascript-S9379.json
```diff
@@ -0,0 +1,5 @@
+{
+"redux:examples/todomvc/components/TodoTextInput.js": [
```
martin-strecker-sonarsource, line 2: **[FP]** FP-reactive: `TodoTextInput.js` (a real native `<input autoFocus="true">`) is shared between `Header.js`'s always-visible 'add new todo' box (static) and `TodoItem.js`'s edit box, which only renders `if (this.state.editing)` -- i.e. mounted reactively when the user double-clicks to edit. Same reported line covers both a plausible TP call site and a plausible FP one, since the rule can't distinguish which caller mounted it. Not affected by the non-DOM guard suggested on decorator.ts:64 -- `input` is a genuine native DOM tag here, so that guard wouldn't touch this one; it's purely a mount-timing problem.

martin-strecker-sonarsource (changes requested):
Correctly implements the dialog/popover exemption from JS-2333, but adds two suppression mechanisms (component-name heuristic, ARIA-role check) beyond the ticket's -- and Biome's -- stated scope, and is missing the standard non-DOM guard this same upstream rule needed elsewhere. Sanity-checked the new ruling snapshots against jsts-test-sources@81a52e2c0 across all 6 affected projects: of ~75 reported issues, ~72 are plausible false positives, split between non-DOM component reuse (fixed by the non-DOM guard proposed below) and reactive/conditional-mount focus management, which that guard can't touch (see per-file comments). The second bucket isn't noise -- it's the same focus-follows-user-action pattern ARIA's dialog-modal guidance recommends, plus explicit validation-guidance cases (e.g. `eigen/RequestForPriceEstimateForm.tsx:74`). We looked at exempting non-literal / prop-forwarded `autoFocus` expressions to shrink that bucket further and rejected both: neither can distinguish 'a parent computed this reactively' from 'a parent hardcoded `true` and it's just laundered through a wrapper' -- that requires tracing the value back through parent components, out of reach for a syntactic, single-file decorator, and both approaches risked hiding real TPs rather than just avoiding FPs. Separately, `ant-design/ConfirmDialog.tsx:76` exposes an unresolvable blind spot in the exemption mechanism: JSX built through an intermediate `const` before being spliced into `<Dialog>` defeats the `.parent`-based ancestor walk, and fixing it in general would need data-flow tracing rather than a syntactic tweak. Given the hit rate, I'd push back on merging as-is regardless of the heuristic-scope question.

packages/analysis/src/jsts/rules/S9379/decorator.ts
guillemsarda: Dropped the name-based heuristic (`MODAL_COMPONENT_NAME_PATTERN`/`MODAL_TRIGGER_NAME_PATTERN` and their use in `isExemptFromReport`/`isModalOpeningElement`) in 8b5ee2308affecd0d312ed4dda438a1726115637. The decorator now only exempts real `dialog` elements, `popover` elements, and ARIA `dialog`/`alertdialog` roles, matching Biome's `noAutofocus` and the sonar-html companion check.

packages/analysis/src/jsts/rules/S9379/decorator.ts
guillemsarda: Yes, it is intentional. I've left [a comment](https://sonarsource.atlassian.net/browse/JS-2333?focusedCommentId=997041) in the Jira ticket. Thanks for raising it up.

packages/analysis/src/jsts/rules/S9379/decorator.ts
guillemsarda: Applied in [c906d41](https://github.com/SonarSource/SonarJS/pull/7835/commits/c906d4164c9b8fbcea3ee161b707aa58a4ac0995), we are reusing now the eslint plugin option.

packages/analysis/src/jsts/rules/S9379/unit.test.ts
guillemsarda: Added them in [5ddfe2c](https://github.com/SonarSource/SonarJS/pull/7835/commits/5ddfe2c3c92f25e1219f848e1ac008c92880f78a)

packages/analysis/src/jsts/rules/S9379/decorator.ts
guillemsarda: Agreed, this deserves fixing properly rather than a quick patch here. Filed JS-2372 to refactor `getRole` to use `undefined` instead of `null`, with the call-site impact scoped out (only 3 files reference it today): https://sonarsource.atlassian.net/browse/JS-2372

its/ruling/src/test/expected/ant-design/typescript-S9379.json
guillemsarda: Agreed, patching this generically would need data-flow analysis which is out of scope for a decorator. Documented it as a known limitation on `isExemptFromReport` and added a test case reproducing the exact `const button = <input autoFocus />; return <dialog>{button}</dialog>;` pattern so it's pinned down and won't be mistaken for a regression later.

its/ruling/src/test/expected/ant-design/typescript-S9379.json
guillemsarda: Good flag on the tradeoff. Documented it as a known limitation on `withIgnoreNonDOM` — a custom component that forwards `autoFocus` straight to a real DOM element (like your `Input.Search` example) is now knowingly missed, since `ignoreNonDOM` only looks at the JSX tag name, not what the component renders to internally. Added a test case (`<Input.Search autoFocus />`) pinning down this false negative so it doesn't get mistaken for a regression later.

its/ruling/src/test/expected/console/typescript-S9379.json
guillemsarda: Good sampling, thanks. Agreed the non-DOM guard only accounts for 2 of the 41 — the reactive-mount pattern is a distinct, unaddressed gap and not something we should try to patch with a heuristic here (a conditionally-rendered element isn't necessarily gated by a user action; it can be true on first paint, so a blanket exemption would just trade this FP class for a new FN class — same failure mode as the modal-name heuristic dropped earlier in this PR). Documented it as a known, accepted limitation on `decorate` with test cases for both the `&&` and ternary variants, and logged this on JS-2333 to track it: watch S9379's real-world FP rate once it has usage data, and only invest in a proper fix (or reconsider the rule's default profile) if that data says it's warranted.

packages/analysis/src/jsts/rules/S9379/decorator.ts
```diff
@@ -0,0 +1,123 @@
...
+    },
+    (context, reportDescriptor) => {
+      const opening = openingElementOf(reportDescriptor);
+      if (opening !== undefined && isExemptFromReport(context, opening)) {
+        return;
+      }
+      context.report({ ...reportDescriptor, message: MESSAGE });
+    },
+  );
+}
```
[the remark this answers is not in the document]
guillemsarda: Good catch. Moved the non-DOM check into isExemptFromReport (report time, using aria-query's dom map directly) and dropped the forced-option override entirely, exactly as suggested — added aria-query and @types/aria-query as explicit dependencies since it was only transitive before. Also updated the rule-implementation notes we keep for this kind of decorator work with this lesson: forcing an upstream option isn't free if the option's own check runs before the rule's cheap filters, so it's worth reading where the check sits before deciding to force-vs-replicate.

its/ruling/src/test/expected/desktop/typescript-S9379.json
guillemsarda: Confirmed — 6 of the 8 are exactly the non-DOM guard's target and are now suppressed (see the fix applied on decorator.ts:64's thread). The other 2 (text-area.tsx:77, text-box.tsx:305) are the wrapper's own internal native elements, so they stay reported, as they should. Good catch on the double-counting pattern — will show up once /ruling is re-run for this project.

its/ruling/src/test/expected/eigen/typescript-S9379.json
guillemsarda: Confirmed — all 17 are the non-DOM pattern, now suppressed by the guard implemented on decorator.ts:64's thread. Will show up as zero once /ruling is re-run for this project.

its/ruling/src/test/expected/jira-clone/javascript-S9379.json
guillemsarda: Confirmed — all 3 are suppressed by the non-DOM guard, including Dropdown.jsx:183 (coincidentally, as you note — it's the DOM check catching it, not anything reactive-mount-aware). Will show up as zero once /ruling is re-run for this project.

its/ruling/src/test/expected/redux/javascript-S9379.json
guillemsarda: Good example — this is the same known, accepted limitation already documented on decorate() and tracked in JS-2333 (conditionally-mounted autofocus not recognized as equivalent to the dialog/popover "freshly revealed UI" exemption), just guarded by an if-statement early return rather than a JSX-level &&/ternary. Added a test case for this exact shape so it's covered alongside the other two variants. Not something we're patching here for the reasons laid out in that decorator docblock (a conditional branch isn't necessarily user-action-gated, so a blanket exemption would trade this FP class for a new FN class).

packages/analysis/src/jsts/rules/S9379/decorator.ts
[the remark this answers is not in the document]
guillemsarda: The fix is committed locally (as described) but hadn't been pushed yet — that's on me, not a different branch. No need for you to implement it separately; it'll be pushed to JS-2333-create-S9379-no-autofocus shortly.

packages/analysis/src/jsts/rules/S9379/decorator.ts
martin-strecker-sonarsource: Makes sense — confirmed it's documented in JS-2333 and matches sonar-html's role-based carve-out. No concerns.

packages/analysis/src/jsts/rules/S9379/decorator.ts
martin-strecker-sonarsource: Confirmed the guard is in — though worth noting for the record, the shipped version isn't the c906d41 approach (forcing ignoreNonDOM) you first mentioned. It ended up reworked into a direct dom.has() check in isExemptFromReport after gitar-bot's perf point, which is what's actually in the branch now. That version looks right to me.

packages/analysis/src/jsts/rules/S9379/decorator.ts
martin-strecker-sonarsource: Fair call to punt this to a follow-up rather than block here — thanks for filing JS-2372.

its/ruling/src/test/expected/ant-design/typescript-S9379.json
martin-strecker-sonarsource: Worth a correction for the record: this specific hit is gone from ruling too, but it's suppressed by the non-DOM guard (ActionButton isn't a native tag), not because the descendant-walk blind spot got fixed. That blind spot is still real for native elements and is now pinned by the `const button = <input autoFocus/>` test case — still the right call to document it as a known limitation rather than add data-flow tracing here.

its/ruling/src/test/expected/console/typescript-S9379.json
martin-strecker-sonarsource: Agreed there's no safe static heuristic for this without real control-flow analysis. Appreciate the write-up in JS-2333 for visibility.

its/ruling/src/test/expected/redux/javascript-S9379.json
martin-strecker-sonarsource: Confirmed, still flagged as expected — and the if-statement variant is now pinned by a test too.

martin-strecker-sonarsource (approved):
Re-verified all the fixes against the current code, tests, and ruling data — confirmed against the branch rather than taking the replies at face value. All blocker/suggestion items are resolved as requested. Two nuances noted inline (ignoreNonDOM implementation diverged from the first stated approach, and the ConfirmDialog ruling entry is suppressed for a different reason than the one documented) — left those threads open for visibility, the rest are resolved. LGTM.

martin-strecker-sonarsource:
`getRole()` returns `null` for any non-literal `role` (e.g. `role={isOpen ? 'dialog' : undefined}`), so a dynamically-role'd dialog gets flagged as a FP. sonar-html's `isDialogOrPopover` handles the same unresolvable case the opposite way (assumes dialog, doesn't report) for Vue/Angular bound roles - pre-existing, not new. Not a regression here, but the two engines are now inconsistent on the same rule. Suggest a follow-up ticket (linked to JS-2333) to align or document as accepted.

PR #7850, opened 2026-09-02
JS-2375 Add mise.toml declaring Java, Maven, and Node toolchain versions

francois-mora-sonarsource:
## Summary
- Adds `mise.toml` declaring the Java, Maven, and Node tool versions used by CI (`java = "21.0"`, `maven = "3.9"`, `node = "24.11.1"`).
- Removes the now-redundant inline `mise_toml` blocks from `.github/workflows/build.yml` wherever they exactly matched those defaults, or duplicated `java`/`maven` while only trying to exclude `node` — `jdx/mise-action` reads the checked-in `mise.toml` automatically when a job doesn't pass its own `mise_toml` input, and `install_args` restricts which tools get installed without re-declaring versions.
- For jobs that only need a different `node` version per matrix leg (`test_eslint_plugin`, `plugin_qa_with_node`, `plugin_qa_fast_with_node`), replaces the restated `java`/`maven`/`node` block with a job-level `MISE_NODE_VERSION` env override, plus an explicit `cache_key` template (`{{default}}-{{env.MISE_NODE_VERSION}}`) so mise-action's cache still keys per node version instead of collapsing to one shared entry.
- `sqaa-release.yml` checks out a caller-supplied `ref` that can predate `mise.toml`; instead of a hand-maintained inline `mise_toml` block (which drifted once already within this PR), it now checks whether the checked-out ref has its own `mise.toml` and only falls back to a hardcoded snapshot when it genuinely doesn't. `release_eslint_plugin.yml` needs none of this — it's `workflow_dispatch`-only with no `ref` input, so its file and checkout are always the same commit.
- `node` is pinned to an exact version (`24.11.1`) rather than left fuzzy like `java`/`maven`, because its resolved version feeds the `js-files-hash` skip-cache key and mise-action's own cache key, both of which only hash *declared text* — a fuzzy spec would let a patch release change the actual runtime under test with zero corresponding cache invalidation. A narrow `.github/renovate.json` override re-enables Renovate's patch updates for just this dependency (disabled by default for all mise-managed tools in the shared preset) so it still gets bumped automatically, just as a real commit each time.
- Documents all of this in `docs/CI.md`.

## Test plan
- [x] CI (`Build` workflow) passes on this PR.
- [x] Compared this PR's `Build` run job-by-job against a recent master run: identical 37 jobs, identical outcomes (only expected diff is `Releasability`, gated to the default branch).
- [x] Confirmed via job logs that `jdx/mise-action`'s resolved cache key differs per matrix node version (e.g. `...-22.12.0` vs `...-24.0.0`).
- [x] Verified GitHub Actions does **not** support YAML merge keys (`<<:`) after a prior commit silently broke workflow parsing with one — env vars are now spelled out explicitly instead.
- [x] Verified locally (exact pinned mise version + exact pinned Alpine image) that the `install_args: maven` Alpine QA change is inert: no shim/`JAVA_HOME` export for the declared-but-uninstalled tools.
- [x] Addressed @vdiez's review: `node` is now pinned exactly (was fuzzy `24.11`), and `sqaa-release.yml` no longer needs manual sync with `mise.toml`.


.github/workflows/build.yml
```diff
@@ -394,12 +384,12 @@ jobs:
...
       - uses: jdx/mise-action@c2a87611a18de5b3828c5652fe268e992400cb5c # v4.3.0
+        env:
+          MISE_NODE_VERSION: ${{ matrix.node-version }}
         with:
           version: 2026.8.14
           cache_save: ${{ needs.setup.outputs.is-default-branch == 'true' }}
-          mise_toml: |
-            [tools]
-            node = "${{ matrix.node-version }}"
+          install_args: node
```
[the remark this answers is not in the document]
francois-mora-sonarsource: Fixed — moved `MISE_NODE_VERSION` to job-level `env` (merged with the existing `shared_build_number` anchor via `<<:`) so it applies to every step in `test_eslint_plugin`, `plugin_qa_with_node`, and `plugin_qa_fast_with_node`, not just the mise-action step.

.github/workflows/build.yml
```diff
@@ -113,11 +113,6 @@ jobs:
...
           version: 2026.8.14
           cache_save: ${{ needs.setup.outputs.is-default-branch == 'true' }}
-          mise_toml: |
-            [tools]
-            java = "21.0"
-            maven = "3.9"
-            node = "24.11.0"
       - if: steps.cache.outputs.cache-hit != 'true'
         id: secrets
         name: Access vault secrets
```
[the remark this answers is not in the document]
francois-mora-sonarsource: Updated the PR description to reflect the actual workflow changes, and added a short note to docs/CI.md explaining the mise.toml + per-tool MISE_<TOOL>_VERSION override convention.

francois-mora-sonarsource:
Addressed the third finding from the Code Review summary too — `plugin_qa_without_node`/`plugin_qa_without_node_dev` (via `&mise_java_only`) and the Alpine QA job now use `install_args: java maven` / `install_args: maven` instead of an inline `mise_toml`, so `java`/`maven` versions live solely in the tracked `mise.toml` and no job overwrites that file on the runner anymore.

francois-mora-sonarsource:
Verified CI parity against master after the last two fixes:
- Compared job lists between this PR's latest `Build` run and a recent master run — identical 37 jobs, identical outcomes (only expected diff: `Releasability`, which is gated to the default branch).
- Caught and fixed two self-introduced regressions along the way, not from review:
  1. `<<: *shared_build_number` merge-key syntax in job-level `env:` — GitHub Actions supports YAML anchors/aliases but **not** merge keys, so this silently broke workflow parsing entirely for two commits (the `Build` workflow never ran on `pull_request` for those SHAs, only unrelated checks did, which made CI look green when it wasn't running at all).
  2. Switching matrix jobs to a `MISE_NODE_VERSION` env override collapsed mise-action's cache key (a hash of config files only) to one shared key across all node versions, since the tracked `mise.toml` no longer varies per leg. Fixed with an explicit `cache_key: '{{default}}-{{env.MISE_NODE_VERSION}}'` template, confirmed in logs to resolve to a distinct suffix per matrix leg (e.g. `...-22.12.0` vs `...-24.0.0`).

CI is now green and behaviorally equivalent to master.

.github/workflows/sqaa-release.yml
```diff
@@ -144,11 +144,6 @@ jobs:
       - uses: jdx/mise-action@c2a87611a18de5b3828c5652fe268e992400cb5c # v4.3.0
         with:
           version: 2026.8.14
```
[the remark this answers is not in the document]
francois-mora-sonarsource: Good catch — restored the explicit `mise_toml` in both `sqaa-release.yml` and `release_eslint_plugin.yml` since they can build a caller-selected ref that predates this PR's `mise.toml`. Documented the exception in docs/CI.md.

francois-mora-sonarsource:
Ran a parallel multi-agent review of this PR (GH Actions semantics / caching / blast radius across workflows / docs / adversarial). Three findings applied in 161345c37:

**1. `mise.toml` was missing from the `js-files-hash` skip-cache key — latent false green.** That hash gates `test_js` and `test_js_win`; on a hit, checkout/mise/npm/`bridge:test:cov` are all skipped while `Validate`/`Upload JS coverage reports` still run on the *restored* files. The tool versions previously lived in `build.yml`, which **is** in the hash, so a bump re-ran the tests. Moving them to `mise.toml` removed that: a future `node = "25.x"` bump would hit cache, run zero tests on the new runtime, and publish stale coverage to SonarQube NEXT — all green. Fixed by adding `mise.toml` to the find list.

**2. `release_eslint_plugin.yml` never needed its inline `mise_toml`, and my rationale comment was wrong.** It is `workflow_dispatch`-only with no `ref` input and a checkout that takes no `ref:`, so the workflow file and `mise.toml` always come from the same commit — dispatching an old tag runs *that tag's* file, inline block included. Removed. Only `sqaa-release.yml` is genuinely exposed (`workflow_call`-only, checks out `ref: inputs.ref`; `automated-release.yml` passes `refs/tags/<version>` and `docker-sqaa.yml` an arbitrary branch — decoupled from the file's own ref), so its block stays. This also removes a Renovate drift vector: Renovate's `mise` manager bumps `mise.toml` but not inline blocks.

**3. `docs/CI.md` taught half the pattern** — it described the `MISE_<TOOL>_VERSION` override without the `cache_key` companion it requires, so following it verbatim reintroduced the collapsed-cache bug. Now also documents why the env var must be job-level.

Verified clean by the reviewers, with evidence: matrix Node override genuinely takes effect (`mise ls` names `MISE_NODE_VERSION` as the source; 58 hits of `installs/node/22.12.0` in the QA log); cache keys are provably disjoint (`{{default}}` folds in an install-args hash, and restore is exact-match with no `restore-keys`); declared-but-uninstalled tools get no shim and no `JAVA_HOME`, so "Disable existing node" still works; and `java = "21.0"`/`maven = "3.9"` resolve to the identical 21.0.2/3.9.16 on master and here — no drift.

Two things a human should weigh, not applied here:
- **Nothing asserts the matrix Node override worked.** If it silently failed, the three `test_eslint_plugin` legs would all run 24.11.0 and still pass, quietly losing min-supported-Node coverage. A `node --version` guard would fix that, but it is scope beyond this PR.
- **The two Alpine QA jobs are `if: schedule` and were skipped in every run here**, so the `install_args: maven` change in a musl container is unexercised until the first post-merge nightly. Evidence from the Ubuntu equivalents says it is inert; worth a glance at that nightly.

francois-mora-sonarsource:
Closed the one open risk from my previous comment — the two Alpine QA jobs (`if: github.event_name == 'schedule'`) that no PR run can exercise.

Worth noting they are **nightly-cron-only**, not merely master-only: a `push` to master won't run them (`event_name` is `push`) and neither will a manual `workflow_dispatch`. Same for the DEV QA pair, IRIS, generated-files-freshness and shadow analysis — 9 gated jobs/steps whose only trigger is the midnight cron. I did *not* dispatch `build.yml` on a scratch branch to force them, because a dispatch satisfies `github.event_name != 'pull_request'` and would run `mvn deploy -Pdeploy-sonarsource,sign,release` plus `promote`, publishing and promoting artifacts in Repox from a throwaway branch.

Instead I reproduced the job's environment locally in the exact pinned image (`alpine/java:22-jdk@sha256:f1537f45…`), mounting this PR's `mise.toml` and running the changed `install_args: maven` path:

```
JAVA_HOME=/opt/java/openjdk ; openjdk version "22.0.2" ; musl libc (x86_64)

$ mise install maven          -> rc=0, maven@3.9.16 installed

$ mise ls
java   21.0 (missing)     /work/mise.toml  21.0
maven  3.9.16             /work/mise.toml  3.9
node   24.11.0 (missing)  /work/mise.toml  24.11.0

$ mise env --dotenv          -> rc=0, no JAVA_HOME emitted
$ ls $MISE_DATA_DIR/shims    -> mvn  mvnDebug  mvnyjp
```

So on musl: the install succeeds; `mise env --dotenv` returns 0 (the call that would otherwise fail the whole action); mise exports **no** `JAVA_HOME`, leaving the container's musl-native JDK 22 intact — which was the actual hazard of declaring a glibc `java = "21.0"` in a tracked file; and **no `java`/`node` shim is created**, so the declared-but-uninstalled tools are inert and "Disable existing node" is unaffected.

Caveat: probed with mise 2026.9.1 locally vs the 2026.8.14 pinned in CI. Still worth a glance at the first post-merge nightly, but this is no longer an unknown.

mise.toml
```diff
@@ -0,0 +1,4 @@
+[tools]
+java = "21.0"
+maven = "3.9"
+node = "24.11"
```
vdiez: The Node specification 24.11 is fuzzy: the current CI run resolves it to 24.11.1, but both the mise-action cache and js-files-hash encode only the unchanged config text. A future 24.11.x release can therefore reuse old JS coverage or the Windows marker and skip checkout, mise setup, and tests, so the new runtime is not validated. Please pin Node and enable Renovate patch updates here, or include a committed resolved version in both cache identities.

.github/workflows/sqaa-release.yml
```diff
@@ -144,11 +144,12 @@ jobs:
...
+          # This workflow builds a caller-supplied ref that may predate the tracked mise.toml.
           mise_toml: |
             [tools]
             java = "21.0"
             maven = "3.9"
-            node = "24.11.0"
+            node = "24.11"
```
vdiez: This inline mise_toml duplicates the canonical root configuration and unconditionally overwrites the configuration from the checked-out ref. Renovate updates only the root file, so a later tool update can silently leave SQAA on a different toolchain; the manual reminder in the documentation does not enforce synchronization. Please use the target configuration when present and generate or automatically validate the fallback against one canonical definition.

vdiez (changes requested):
I found two cache and toolchain consistency issues that should be addressed before merge. The matrix override itself is working correctly. Validation in a clean worktree passed npm ci, npm run bbf, Prettier, and git diff --check; the full GitHub Build is now green.

mise.toml
francois-mora-sonarsource: You're right — fixed. `node` is now pinned exactly (`24.11.1`, matching what's currently resolving in CI) instead of fuzzy, so the declared version and the resolved version are always the same value and the skip-cache/mise-action cache keys can never go stale relative to it. `java`/`maven` stay fuzzy since their exact patch doesn't feed any cache key that matters (no skip-cache job touches Maven or runs under a JVM).

My earlier fuzzy conversion was based on a wrong premise — I'd assumed the shared Renovate preset disables *all* updates for a patch-pinned mise tool, but it only disables *patch*-type updates; minor/major bumps still flow to an exact pin. So to keep getting Node's patch releases too (unlike Java/Maven, these matter for a runtime), I added a narrow override in `.github/renovate.json` re-enabling patch updates for just the `node` mise dependency — each bump now lands as a real commit that changes `mise.toml`'s text, which is exactly what keeps the cache keys honest going forward.

.github/workflows/sqaa-release.yml
francois-mora-sonarsource: Fixed by removing the manual-sync requirement entirely rather than trying to enforce it. The workflow now checks whether the checked-out ref already has its own `mise.toml`; if so it passes an empty `mise_toml` input (falsy, so mise-action skips overwriting the file and just reads what's checked out) — no drift possible for any ref recent enough to carry the file. The hardcoded block only remains as a fallback for refs old enough to predate `mise.toml` entirely, which makes it a frozen historical snapshot rather than something that needs to track future bumps.

docs/CI.md
```diff
@@ -818,6 +818,105 @@ These are the most important reusable components in the current pipeline.
...
+
+No other workflow needs any of this. In particular, [`release_eslint_plugin.yml`](../.github/workflows/release_eslint_plugin.yml) looks superficially similar (it's also a manually-triggered release workflow) but is _not_ exposed: it is `workflow_dispatch`-only with no `ref` input, and its checkout takes no `ref:` either, so the workflow file and the checked-out tree always come from the same commit — dispatching an old ref just runs _that ref's own_ copy of the file. Don't add a `mise_toml` block back there. The one thing this doesn't cover: dispatching a ref old enough to predate `mise.toml` provisions _no_ toolchain at all rather than failing loudly — mise installs nothing and the job silently falls back to whatever Node/Java happens to be ambient on the runner. That's an availability risk for very old refs, not a correctness one, and isn't worth the same fallback machinery for a workflow that always runs against a ref you just chose from the dispatch UI.
```
vdiez: This final caveat contradicts the preceding explanation and repository history. An old ref runs its own copy of release_eslint_plugin.yml, and refs predating mise.toml still contain the former inline mise_toml, so they do provision a toolchain. Please remove or rephrase this scenario; this is non-blocking.

vdiez (approved):
The two requested changes are correctly addressed. Node is now exactly pinned with a targeted Renovate patch-update override, and SQAA preserves the checked-out configuration while using a stable fallback only when needed. Clean bootstrap, Prettier, Renovate validation, actionlint, diff checks, and the full GitHub Build pass.

docs/CI.md
francois-mora-sonarsource: Fixed — removed the trailing sentence entirely. You're right that it contradicted the preceding explanation: dispatching an old ref just runs that ref's own copy of release_eslint_plugin.yml, and any ref old enough to predate the tracked mise.toml still carries the former inline mise_toml block in its own copy of the file, so it does provision a toolchain from that. There's no gap to caveat.

PR #7852, opened 2026-09-02
JS-2377 Remove redundant logging dependency handling

vdiez:
Part of

## Summary

- Remove redundant SLF4J and Logback version overrides, transitive constraints, and update exceptions from plugin-bound code.
- Rely on Sonar Plugin API v14 and its test fixtures for SLF4J 2.0.18 and Logback 1.6.3 where they are part of the platform contract.
- Keep logging dependencies explicit only in standalone code or tests that use the implementation API directly.


aurelien-coet-sonarsource (approved):
LGTM

PR #7854, opened 2026-09-02
JS-2368 S5332: False positive on http:// URLs used as schema/namespace identifiers (CycloneDX schema, SNOMED/SOAP/XMPP/ESCO namespaces)

oleksandr-selehenenko-sonarsource:
﻿## Summary
- S5332 was flagging http:// URLs used purely as spec-mandated identifiers (JSON-Schema $schema, coding-system/namespace URIs) that are never dereferenced.
- Extends the existing NAMESPACE_URI_AUTHORITIES allowlist (from JS-1877) with cyclonedx.org, snomed.info, adlnet.gov, jabber.org, etherx.jabber.org.
- Refactored the allowlist check from a single alternation regex to a plain string array + exact-authority helper, avoiding a regex-complexity issue on the extended list.

## Test plan
- [x] Added CycloneDX $schema and SNOMED coding-system reproducers, confirmed red before the fix, green after
- [x] Added subdomain-spoofing negative tests per new host (e.g. cyclonedx.org.evil.com) to prove exact-authority anchoring
- [x] Full S5332 suite green

Jira: JS-2368

### DIFF
--- a/packages/analysis/src/jsts/rules/S5332/rule.lib.ts
+++ b/packages/analysis/src/jsts/rules/S5332/rule.lib.ts
@@ -54,9 +54,41 @@ const INSECURE_PROTOCOLS = Object.keys(CLEARTEXT_PROTOCOL_ALTERNATIVES).map(s =>
 const SAFE_HOSTS =
   /(?:^localhost|^127(?:\.\d+){1,3}|^\[(?:0*:){7}:?0*1]|^\[::1]|^169\.254\.\d+\.\d+|^\[fd00:ec2::254]|^168\.63\.129\.16|^100\.100\.100\.200|^metadata\.google\.internal|^metadata\.internal|^host\.docker\.internal|^gateway\.docker\.internal|\.svc\.cluster\.local)(?=:|$)/i;
 
-// XML namespace URI authorities — port of CleartextProtocolFilter.NAMESPACE_URI_AUTHORITIES, extended with pre-existing sonar-js exceptions
-const NAMESPACE_URI_AUTHORITIES =
-  /(?:^www\.w3\.org|^schemas\.android\.com|^schemas\.microsoft\.com|^schemas\.xmlsoap\.org|^www\.sap\.com|^www\.opengis\.net|^hl7\.org|^unitsofmeasure\.org|^purl\.org|^docs\.oasis-open\.org|^xmlns\.com|^json-ld\.org|^schema\.org|^www\.springframework\.org|^maven\.apache\.org|^dublincore\.org|^ogp\.me|^xml\.apache\.org|^schemas\.openxmlformats\.org|^rdfs\.org|^schemas\.google\.com|^a9\.com|^ns\.adobe\.com|^ltsc\.ieee\.org|^docbook\.org|^graphml\.graphdrawing\.org|^json-schema\.org)(?=:|$)/i;
+// Namespace / spec-mandated identifier URI authorities — port of CleartextProtocolFilter.NAMESPACE_URI_AUTHORITIES, extended with pre-existing sonar-js exceptions
+const NAMESPACE_URI_AUTHORITIES = [
+  'www.w3.org',
+  'schemas.android.com',
+  'schemas.microsoft.com',
+  'schemas.xmlsoap.org',
+  'www.sap.com',
+  'www.opengis.net',
+  'hl7.org',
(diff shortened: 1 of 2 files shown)


francois-mora-sonarsource (changes requested):
## Review: S5332 false-positive fix (JS-2368)

This is a false-positive fix, so the bar for "does the narrowed exemption stay tight enough" is high — one finding below needs to be addressed before merge; the rest are optional follow-ups.

### Mandatory: `jabber.org` entry is too broad (new false negative)

`NAMESPACE_URI_AUTHORITIES` exempts on **host alone**, with no path constraint. That's fine for the other four new entries (`cyclonedx.org`, `snomed.info`, `adlnet.gov`, `etherx.jabber.org`), which are verified to be namespace-only strings never used as live endpoints. But `jabber.org` itself is a real, long-running public XMPP server (jabberd → ejabberd → Prosody since 1999), and this exemption also silently suppresses genuinely insecure connections to it, e.g.:

```js
url = 'http://jabber.org:5280/http-bind'; // a real BOSH (XMPP-over-HTTP) endpoint
```

This is exactly the kind of cleartext transmission S5332 exists to catch — once host-allowlisted, it becomes invisible rather than occasionally missed.

**Suggested fix** — scope the `jabber.org` exemption to the well-known XEP namespace path prefix (`/protocol/`), since every real XMPP namespace on that host is registered there, while a real endpoint (e.g. `/http-bind`) is not:

```ts
type NamespaceAuthority = string | { host: string; pathPrefix: string };

const NAMESPACE_URI_AUTHORITIES: NamespaceAuthority[] = [
  // ...unchanged entries...
  'cyclonedx.org',
  'snomed.info',
  'adlnet.gov',
  // jabber.org is also a live public XMPP server, not just a namespace string —
  // only exempt its well-known XEP namespace paths (all registered under /protocol/);
  // anything else on this host is a real network target.
  { host: 'jabber.org', pathPrefix: '/protocol/' },
  'etherx.jabber.org',
];

function isNamespaceAuthority(host: string, pathname: string | undefined, entry: NamespaceAuthority) {
  const authority = typeof entry === 'string' ? entry : entry.host;
  if (host !== authority && !host.startsWith(`${authority}:`)) {
    return false;
  }
  return typeof entry === 'string' || (pathname !== undefined && pathname.startsWith(entry.pathPrefix));
}
```

`hasExceptionHost` needs to thread `pathname` through (from `url.pathname` on the strict-parse branch; `undefined` on the lenient fallback, which then fails closed for the path-scoped entry):

```ts
function hasExceptionHost(value: string) {
  let host: string;
  let pathname: string | undefined;

  try {
    const url = new URL(value);
    host = url.hostname;
    pathname = url.pathname;
    if (host.length === 0) return false;
  } catch {
    const match = LENIENT_AUTHORITY.exec(value);
    if (!match) return false;
    host = match[1];
  }

  return isSafeHost(host, pathname);
}
```

Please add a regression test locking in the narrowed behavior, e.g.:

```js
url = 'http://jabber.org/http-bind'; // real BOSH endpoint — must still be flagged
```

alongside the existing `http://jabber.org/protocol/muc` valid case (which should stay unflagged).

### Optional (test-only, non-blocking)

- The `isNamespaceAuthority` port-handling branch (`host.startsWith(`${authority}:`)`) is untested for every authority, old and new. It's only reachable via the `LENIENT_AUTHORITY` fallback (e.g. `http://cyclonedx.org:${port}/...` with a template placeholder) — worth a test case.
- That same untested branch hides a latent case-sensitivity gap: matching is now `===`/`startsWith` against lowercase literals, so a mixed-case host that only reaches the lenient fallback (e.g. `HTTP://CYCLONEDX.ORG:${port}/x`) would silently fail to match. Masked on the normal `new URL()` path since WHATWG lowercases `hostname` automatically, but real on the fallback path.

Everything else — the regex→array refactor (anchoring, case sensitivity, port handling), and the other four new allowlist entries — checked out clean; verified independently against spec/usage docs and via a full code-review pass.

francois-mora-sonarsource (approved):
Both mandatory findings from my previous review are addressed: the `jabber.org` exemption is now scoped to `/protocol/` and `/features/` path prefixes (not the whole host), verified with regression tests for the BOSH endpoint (`/http-bind`) still being flagged and the XMPP stream-feature namespaces now being exempt. Unit tests pass and `tsc` is clean. LGTM.

PR #7857, opened 2026-09-02
JS-2373 S107: False positive on the AMD/UI5 define() factory callback (dependency list counted as parameters)

oleksandr-selehenenko-sonarsource:
﻿## Summary
- S107 (max function parameters) was counting the injected AMD dependency list of a define()/require()/sap.ui.define() factory callback as hand-written parameters.
- Added an isAmdFactoryCallback exception: skips the parameter-count check when the flagged function is the last argument to a define/require/sap.ui.define call, immediately preceded by an array-literal dependency list.
- Scoped narrowly to that call/argument-position shape; ordinary functions and callbacks to unrelated array-preceded calls are unaffected.

## Test plan
- [x] Added valid cases: the reported sap.ui.define repro, bare define, bare require
- [x] Added an invalid negative-control case (unrelated array-preceded call) proving the carve-out doesn't leak
- [x] Full S107 suite green

Jira: JS-2373

### DIFF
--- a/packages/analysis/src/jsts/rules/S107/rule.ts
+++ b/packages/analysis/src/jsts/rules/S107/rule.ts
@@ -62,16 +62,13 @@ function getMax(options: FromSchema<typeof meta.schema>[0]) {
 }
 
 /**
- * The two trailing arguments inspected for an AMD/UI5 factory-callback shape: the
- * dependency array literal followed by the factory function itself.
- */
-const AMD_FACTORY_TRAILING_ARGUMENTS = 2;
-
-/**
  * An AMD/UI5 factory callback, e.g., `define([...], function (a, b) {})`,
- * `require([...], function (a, b) {})` or `sap.ui.define([...], function (a, b) {})`.
- * Its parameters are module dependencies injected by the loader, one per entry of the
- * preceding dependency array, not a hand-written signature.
+ * `require([...], function (a, b) {})`, `sap.ui.define([...], function (a, b) {})` or
+ * `sap.ui.require([...], function (a, b) {})`. Its parameters are module dependencies
+ * injected by the loader, one per entry of the immediately preceding dependency array,
+ * not a hand-written signature. The factory is not necessarily the last argument: the
+ * RequireJS `require([...], factory, errback)` form appends an error callback after it.
+ * Parameters beyond the dependency count are hand-written and remain reported.
  */
 function isAmdFactoryCallback(context: Rule.RuleContext, functionLike: TSESTree.FunctionLike) {
   const call = functionLike.parent;
@@ -79,9 +76,11 @@ function isAmdFactoryCallback(context: Rule.RuleContext, functionLike: TSESTree.
     return false;
   }
 
-  const [precedingArg, lastArgument] = call.arguments.slice(-AMD_FACTORY_TRAILING_ARGUMENTS);
+  const factoryIndex = call.arguments.indexOf(functionLike as TSESTree.CallExpressionArgument);
+  const dependencies = (factoryIndex > 0 ? call.arguments[factoryIndex - 1] : undefined) as
+    estree.Node | undefined;
   return (
-    lastArgument === functionLike && isArrayExpression(precedingArg as estree.Node | undefined)
+    isArrayExpression(dependencies) && functionLike.params.length <= dependencies.elements.length
   );
 }
 
@@ -96,7 +95,7 @@ function isAmdDefineOrRequireCall(
   return (
     (isIdentifier(callee, 'define', 'require') && !isShadowed(context, callee)) ||
     (callee.type === 'MemberExpression' &&
-      isIdentifier(callee.property, 'define') &&
+      isIdentifier(callee.property, 'define', 'require') &&
       isMemberExpression(callee.object as estree.Node, 'sap', 'ui'))
   );
 }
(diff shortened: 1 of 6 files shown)


francois-mora-sonarsource:
## Review: 1 mandatory finding

**Too broad — shadow guard doesn't cover `sap.ui.define`/`sap.ui.require`**

`isAmdDefineOrRequireCall` checks `isShadowed(context, callee)` for the bare `define`/`require` identifier branch, but the `sap.ui.*` member-expression branch has no equivalent check on `sap`:

```js
var sap = { ui: { define: function(factory) { return factory(); } } };
sap.ui.define(['a', 'b', 'c', 'd', 'e'], function (a, b, c, d, e) { ... });
```

`sap` is an ambient global here (injected by the UI5 bootstrap, same mechanism as RequireJS's global `define`/`require` — no import involved), so this is the same shadowing case already handled for the bare identifier form, just needing the check applied to the root identifier of the member chain instead. Left as-is, a locally-mocked `sap` still gets the AMD carve-out, i.e. S107 can miss a real long-parameter-list violation.

**Request:** extend the existing shadow check to the `sap.ui.define`/`sap.ui.require` branch (walk to the root identifier `sap` and reuse `isShadowed`).

**Nit:** `isShadowed` (rule.ts) duplicates `isRequireShadowed` (`packages/analysis/src/jsts/rules/helpers/module.ts`) — same `getVariableFromScope(...)?.defs.length` pattern, one hardcoded to `'require'`, the other generalized. Worth consolidating into one helper.

packages/analysis/src/jsts/rules/S107/rule.ts
```diff
@@ -54,6 +60,55 @@ function getMax(options: FromSchema<typeof meta.schema>[0]) {
...
+): node is TSESTree.CallExpression {
+  if (node?.type !== 'CallExpression') {
+    return false;
+  }
+  const callee = node.callee as estree.Node;
+  return (
+    (isIdentifier(callee, 'define', 'require') && !isShadowed(context, callee)) ||
+    (callee.type === 'MemberExpression' &&
+      isIdentifier(callee.property, 'define', 'require') &&
+      isMemberExpression(callee.object as estree.Node, 'sap', 'ui'))
```
francois-mora-sonarsource: Too broad: this `sap.ui.*` branch has no shadow check on `sap`, unlike the bare `define`/`require` branch above (`!isShadowed(context, callee)`). `sap` is an ambient global (injected by the UI5 bootstrap, no import involved) just like `define`/`require`, so a locally-mocked `sap` still gets the AMD carve-out here:

```js
var sap = { ui: { define: function(factory) { return factory(); } } };
sap.ui.define(['a', 'b', 'c', 'd', 'e'], function (a, b, c, d, e) { ... });
```

This is a new false negative — S107 can miss a real long-parameter-list violation. Please extend the shadow check to the root identifier of the member chain (`sap`) and reuse `isShadowed` on it, same as the fix already applied to the bare identifier form.

packages/analysis/src/jsts/rules/S107/rule.ts
```diff
@@ -54,6 +60,55 @@ function getMax(options: FromSchema<typeof meta.schema>[0]) {
...
+}
+
+/**
+ * True when `identifier` resolves to a local binding, e.g., a local `function define(...) {}`
+ * or a test double named `require`, rather than the global AMD loader function of that name.
+ */
+function isShadowed(context: Rule.RuleContext, identifier: estree.Identifier) {
+  return !!getVariableFromScope(context.sourceCode.getScope(identifier), identifier.name)?.defs
+    .length;
+}
```
francois-mora-sonarsource, line 109: Nit: this duplicates `isRequireShadowed` in `packages/analysis/src/jsts/rules/helpers/module.ts` (same `getVariableFromScope(...)?.defs.length` pattern, hardcoded to `'require'`). Worth consolidating into one helper.

francois-mora-sonarsource (changes requested):
Mandatory finding on the shadow guard, plus a small reuse nit — see inline comments.

packages/analysis/src/jsts/rules/S107/rule.ts
oleksandr-selehenenko-sonarsource: Fixed in a19fc0b74. The `sap.ui.define`/`sap.ui.require` branch now checks the root of the member chain with the same shadow guard as the bare identifier form, so a locally-mocked `sap` no longer gets the carve-out. Added the invalid test case with your exact repro.

packages/analysis/src/jsts/rules/S107/rule.ts
oleksandr-selehenenko-sonarsource: Fixed in a19fc0b74. Consolidated into a single `isGlobalShadowed(sourceCode, node, name)` in `helpers/module.ts`; `isRequireShadowed` is gone and its three call sites (`getUnshadowedRequireModuleName`, S6544, S9144) now pass `'require'` explicitly. S107 reuses the same helper for `define`/`require`/`sap`.

packages/analysis/src/jsts/rules/S107/rule.ts
francois-mora-sonarsource: Verified — a19fc0b74 adds the sap shadow check and a new test case. Looks good.

packages/analysis/src/jsts/rules/S107/rule.ts
francois-mora-sonarsource: Verified — consolidated into isGlobalShadowed and reused across S107/S6544/S9144. Thanks.

francois-mora-sonarsource (approved):
Both review findings from the earlier CHANGES_REQUESTED review are fixed in a19fc0b74 and verified against the diff + tests (all 22 relevant tests pass locally):

- The `sap.ui.define`/`sap.ui.require` carve-out now checks whether `sap` itself is shadowed, closing the false-positive-suppression-on-a-mock gap.
- `isShadowed`/`isRequireShadowed` are consolidated into a single `isGlobalShadowed` helper, reused by S107, S6544, and S9144.

LGTM.

PR #7858, opened 2026-09-02
JS-2379 Reset analysis caches before scanner requests

vdiez:
## Summary

- centralize project-analysis cache initialization in `initFileStoresForAnalysis`
- reset shared file stores and the source-file content cache before independent scanner analyses from SQS, SQC, and SQAA
- preserve caches for SonarQube for IDE requests to retain incremental-analysis performance
- remove the endpoint-specific SQAA reset and document the shared cache lifecycle

## Why

The Node.js process is reused across analyses, and its caches are shared between requests. Independent scanner analyses must not reuse cached data from an earlier project analysis, especially when requests contain overlapping file paths. Applying the reset policy at the common file-store initialization boundary gives both gRPC endpoints the same lifecycle while keeping SonarQube for IDE caches incremental.

Sparse or omitted-file scanner requests remain supported: their caches are reset before files are rediscovered from the filesystem.

## Validation

- `npm run bbf`
- `npx knip --no-gitignore`
- focused analyze-project normalization tests (7 passing)
- `npm run bridge:test` (2,375 passing)
- `git diff --check`
- Prettier check for all changed files

### DIFF
--- a/docs/node-analysis-caches.md
+++ b/docs/node-analysis-caches.md
@@ -168,9 +168,10 @@ The main entrypoints are:
 - `sanitizeProjectAnalysisInput()` in `packages/analysis/src/common/input-sanitize.ts`
 - `normalizeAnalyzeProjectRequest()` in `packages/grpc/src/analyze-project-normalize.ts`
 
-One additional nuance in `normalizeAnalyzeProjectRequest()`:
+All entrypoints initialize their file stores through `initFileStoresForAnalysis()`:
 
-- when the request has no explicit files and filesystem access is enabled, it explicitly resets the four file stores before rediscovering the project from disk
+- scanner analyses reset the shared caches before initializing the request
+- SonarQube for IDE analyses retain the caches for incremental reuse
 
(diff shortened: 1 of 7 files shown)


francois-mora-sonarsource (approved):
Reviewed the cache-reset logic (initFileStoresForAnalysis and its three call sites), and it's correct: the sonarlint/scanner split is safe everywhere, ordering is right, and no dangling references to the removed exports.

Optional follow-ups (non-blocking):
- docs/node-analysis-caches.md: step 3 of "Initialization Flow" still says "Call initFileStores(...)" — should reference initFileStoresForAnalysis() to match the section right after it.
- docs/node-analysis-caches.md: the "Cache Invalidation" and "SonarQube Full-Project Analysis" sections describe the source-file-content cache clear as happening only "at the end" via analyzeWithProgram() / only in the standalone gRPC service — now it's also cleared up front by initFileStoresForAnalysis() for every scanner call site.
- packages/grpc/src/transformers/request.ts:131 has a stale JSDoc comment referencing initFileStores instead of initFileStoresForAnalysis.

Approving.

PR #7869, opened 2026-09-03
JS-2386 Fix ruling bot baseline for pull requests

vdiez:
## Summary

- use the synthetic merge commit's first parent as the ruling report baseline for pull requests
- retain both merge parents in the checkout and fail explicitly when they are unavailable
- remove the stale pull-request event base SHA fallback

## Why

Pull request workflows analyze GitHub's synthetic merge commit. The ruling report must therefore compare the analyzed merge tree with the exact base tree included in that merge.

### Observed failure

The problem surfaced in [PR #7858](https://github.com/SonarSource/SonarJS/pull/7858):

- [the offending ruling comment](https://github.com/SonarSource/SonarJS/pull/7858#issuecomment-5512486607) attributed 42 S9379 issues to the PR
- the workflow had actually checked out synthetic merge `c24fd37055d184376978c4fd63717553baeeb78d`, whose first parent was `e731ed5c1132532abf806065e75ea848c01701d8`
- the bot instead compared against stale event base SHA `9a99bcbfc06876393e0563380432369fe4aa27bc`
- the S9379 expected-result files had already been added to `master` by [PR #7835](https://github.com/SonarSource/SonarJS/pull/7835); PR #7858 did not change them

The ruling run consequently passed against the correct merge tree while the comment incorrectly reported unrelated changes from `master`.

### Why the previous fix was insufficient

[PR #7520](https://github.com/SonarSource/SonarJS/pull/7520) attempted to derive the tested base with `git merge-base HEAD origin/$BASE_REF`. However, the default depth-1 checkout marked the synthetic merge commit as shallow, so Git could not traverse its parents. Fetching the base branch did not change that shallow boundary. The failed merge-base lookup was ignored and the bot silently fell back to the stale event base SHA.

## Fix

The checkout now retains both parents of the synthetic merge commit. On pull requests, the bot uses `HEAD^1` as the report base, verifies that `HEAD^2` is also available, and fails explicitly instead of falling back to unrelated history.

The first parent is the exact base tree included in the analysis, so it remains correct even if `master` advances again while the workflow is running.

### Why depth 2 is sufficient

GitHub's synthetic pull-request merge has two immediate parents: `HEAD^1` is the tested base and `HEAD^2` is the pull-request head. A depth-2 checkout retains `HEAD` and both of those parents. The bot only reads the direct parents; it does not need to traverse either branch's earlier history. Base advancement, force-pushes after checkout, and merge commits within either branch therefore do not require additional depth.

The bot verifies that both parents are available and fails explicitly if the checkout is not the expected synthetic merge. Depth 2 would not necessarily be sufficient for a general `git merge-base` calculation, but it is sufficient for the direct `HEAD^1` lookup used here.

## Validation

- parsed the modified workflow and action YAML successfully
- Prettier check for the modified workflow, action, script, and CI documentation
- `git diff --check`
- verified that PR #7858's tested base and synthetic merge tree have no ruling expected-result differences


francois-mora-sonarsource (approved):
Image

PR #7892, opened 2026-09-07
JS-1728 Resolve root Bun catalogs in package.json

erwan-leforestier-sonarsource:
Copy of https://github.com/SonarSource/SonarJS/pull/7009

----
## Summary by Gitar

- **Dependency Resolution:**
    - Added support to resolve root Bun catalogs in `package.json` manifest resolver via `getCatalogSource` helper
    - Added tests verifying default and named Bun root catalog resolution


packages/analysis/src/jsts/rules/helpers/dependency-manifests/resolvers/package-json.ts
```diff
@@ -40,32 +40,31 @@ export const packageJsonManifestResolver: ManifestResolver = {
...
-        : undefined;
-      if (parsedPnpmWorkspace) {
-        parsedPackageJson = injectWorkspacePackages(parsedPackageJson, parsedPnpmWorkspace);
-        catalogSource = parsedPnpmWorkspace;
-      }
+    // Bun only reads catalogs from the workspace root: a package.json declaring `workspaces` is
+    // itself a root, otherwise the closest parent package.json with catalogs is the root.
+    const closestParent = isWorkspaceRoot
+      ? undefined
+      : findClosestParentPackageJsonWithCatalogs(dir, topDir, fileSystem);
```
vdiez, line 66: [P2] Check ancestor workspace membership before treating this as a root

A package declaring `workspaces` can still be a member of an ancestor workspace. With outer `workspaces: ["sub"]`, and `sub` declaring `workspaces: ["child"]` plus `dependencies: { leaf: "catalog:" }`, Bun 1.4.2 resolves `leaf` from the outer catalog when installing from the outer root. This branch skips that catalog, so the PR leaves `leaf` unresolved if `sub` has no catalog, or uses its ignored child catalog if one exists. The base resolver resolves both cases correctly.

Please preserve ancestor catalog resolution for included workspace members and add a regression fixture where the nested project is included in the outer `workspaces`. The new nested-workspace-root fixture only covers an independent nested project, excluded from the outer workspace.

vdiez (changes requested):
The root catalog self-reference fix works, but the new workspace-root check regresses catalog resolution for a nested package included in an outer workspace (see inline finding). Please address that case before merging.

Validated commit `5d9d4573bb`: `npm ci`, `npm run bbf`, and all 44 dependency-manifest tests passed. Additional checks passed for default and named root catalogs, both at the top level and inside `workspaces`. The regression was reproduced with Bun 1.4.2 and compared against the base resolver.

packages/analysis/src/jsts/rules/helpers/dependency-manifests/resolvers/package-json.ts
```diff
@@ -209,6 +212,60 @@ function findClosestParentPackageJsonWithCatalogs(
...
+  let currentDir = getParentDirPath(dir);
+  const cache = closestPatternCache.get(PACKAGE_JSON, fileSystem).get(topDir);
+
+  while (currentDir !== null) {
+    const file = cache.get(currentDir);
+    if (!file) {
+      return false;
+    }
+
+    const ancestorDir = dirnamePath(file.path);
```
vdiez: [P1] Use the renamed File property after updating from master

This new access does not compile in the PR's merge with current master: `File.path` was renamed to `File.filePath` in [#7310](https://github.com/SonarSource/SonarJS/pull/7310), merged on June 18, 2026. The CI merge commit `24d8de2` fails here (line 243 in the merged file) with `TS2339: Property 'path' does not exist on type 'File'`, blocking compilation and downstream checks. Please merge or rebase master and use `file.filePath` here. The PR head alone still has the old `File` definition, which explains why its standalone build passes.

packages/analysis/src/jsts/rules/helpers/dependency-manifests/resolvers/package-json.ts
```diff
@@ -209,6 +212,60 @@ function findClosestParentPackageJsonWithCatalogs(
...
+  ancestorDir: NormalizedAbsolutePath,
+  dir: NormalizedAbsolutePath,
+): boolean {
+  const { workspaces } = packageJson;
+  const patterns = Array.isArray(workspaces) ? workspaces : workspaces?.packages;
+  if (!patterns?.length || !dir.startsWith(`${ancestorDir}/`)) {
+    return false;
+  }
+  const relativeDir = dir.slice(ancestorDir.length + 1);
+  return patterns.some(pattern => new Minimatch(pattern).match(relativeDir));
```
vdiez: [P2] Match Bun's workspace exclusion and path semantics

`patterns.some(...)` does not reproduce Bun's workspace matching. With `workspaces: ["sub/*", "!sub/excluded"]`, Bun excludes `sub/excluded`, but the positive pattern makes this return `true`. A self-contained project there consequently inherits the outer catalog instead of using its own. It also misses valid declarations `workspaces: ["./sub"]` and `workspaces: ["sub/"]`: neither pattern matches the relative directory `sub`, so an included member that declares its own `workspaces` is still treated as independent and its inherited `catalog:` references remain unresolved.

All three cases were reproduced with Bun 1.4.2 and this resolver. Please normalize workspace paths and honor exclusions when combining patterns, with regression tests for these layouts.

vdiez (changes requested):
The two originally reported included-member cases are fixed and now have regression tests. Two issues still need addressing before approval: compilation against current master and Bun workspace-pattern matching.

Validation on `77a2c29a4c`: `npm ci`, `npm run bbf`, and all 46 dependency-manifest tests passed on the PR head. CI builds the merge with current master and fails on the renamed `File` property. Additional reproductions against Bun 1.4.2 confirmed the workspace exclusion and path-normalization issues described inline.

packages/analysis/src/jsts/rules/helpers/dependency-manifests/resolvers/package-json.ts
```diff
@@ -212,6 +212,84 @@ function findClosestParentPackageJsonWithCatalogs(
...
+  const matches = (workspacePattern: string) =>
+    new Minimatch(workspacePattern, { nonegate: true }).match(relativeDir);
+
+  return (
+    normalizedPatterns.some(
+      ({ isExclusion, workspacePattern }) => !isExclusion && matches(workspacePattern),
+    ) &&
+    !normalizedPatterns.some(
+      ({ isExclusion, workspacePattern }) => isExclusion && matches(workspacePattern),
+    )
```
vdiez: [P2] Preserve explicitly included workspace members

For `workspaces: ["sub/*", "!sub/excluded", "sub/excluded"]`, Bun includes `sub/excluded` and resolves its catalog dependency from the outer root. This final exclusion check overrides the explicit entry. If `sub/excluded` declares its own `workspaces` and has `dependencies: { leaf: "catalog:" }` but no own catalog, the resolver incorrectly treats it as independent and leaves `leaf` unresolved.

Confirmed with Bun 1.4.2 and commit `610d5e102a`: Bun's lockfile includes the workspace and resolves `leaf` from the outer catalog, while this resolver returns the literal `catalog:` value. Please match Bun's explicit workspace inclusion behavior and add a regression test for this layout.

vdiez (changes requested):
The previously reported build integration issue and workspace exclusion/path-normalization cases are fixed. One P2 remains: an explicitly included workspace member is incorrectly excluded, leaving its outer catalog references unresolved (see inline finding).

Validation on `610d5e102a`: `npm ci`, `npm run bbf`, and all 54 dependency-manifest tests passed; CI is green. The remaining mismatch was reproduced with Bun 1.4.2 and the latest resolver.

packages/analysis/src/jsts/rules/helpers/dependency-manifests/resolvers/package-json.ts
erwan-leforestier-sonarsource: Fixed in d61edabfb.

I checked Bun 1.4.2’s `WorkspaceMap.process_names_array`: it registers literal workspace paths before expanding globs, and later negated globs only filter glob expansion. The resolver now follows that model, so an explicit `sub/excluded` remains an outer workspace member despite `!sub/excluded`.

Added regressions for the reported `["sub/*", "!sub/excluded", "sub/excluded"]` layout and `["sub/*", "!sub/excluded", "sub/excluded", "!sub/excluded"]`; both resolve the nested member’s `catalog:` dependency from the outer catalog, matching Bun 1.4.2.

Validated with `npm run bbf` and all 56 dependency-manifest tests.

vdiez (approved):
Approved: all prior review findings are addressed on d61edabfb, including the workspace-pattern cases. Focused validation passed (npm run bbf; 56 dependency-manifest tests).

packages/analysis/src/jsts/rules/helpers/dependency-manifests/resolvers/package-json.ts
```diff
@@ -212,6 +212,125 @@ function findClosestParentPackageJsonWithCatalogs(
...
+    return true;
+  }
+
+  const workspaceGlobs = normalizedPatterns.filter(hasWorkspaceGlobSyntax).map(pattern => ({
+    isExclusion: pattern.startsWith('!'),
+    workspacePattern: pattern.startsWith('!') ? pattern.slice(1) : pattern,
+  }));
+  const matches = (workspacePattern: string) =>
+    new Minimatch(workspacePattern, { nonegate: true }).match(relativeDir);
+
```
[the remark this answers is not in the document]
erwan-leforestier-sonarsource: Not applying this change: Bun 1.4.2 includes `sub/excluded` for both `["!sub/excluded", "sub/*"]` and `["sub/*", "!sub/excluded", "sub/ex*"]` (verified from freshly generated Bun lockfiles).

This is intentional in Bun’s `WorkspaceMap.process_names_array`: while expanding a glob, it only checks `workspace_globs[i + 1..]` for a negated match. The resolver follows that same later-negated-glob rule. Making exclusions position-independent would diverge from Bun.

packages/analysis/src/jsts/rules/helpers/dependency-manifests/resolvers/package-json.ts
```diff
@@ -212,6 +212,125 @@ function findClosestParentPackageJsonWithCatalogs(
...
+  if (
+    normalizedPatterns.some(pattern => !hasWorkspaceGlobSyntax(pattern) && pattern === relativeDir)
+  ) {
+    return true;
+  }
+
+  const workspaceGlobs = normalizedPatterns.filter(hasWorkspaceGlobSyntax).map(pattern => ({
+    isExclusion: pattern.startsWith('!'),
+    workspacePattern: pattern.startsWith('!') ? pattern.slice(1) : pattern,
+  }));
```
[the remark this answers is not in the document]
erwan-leforestier-sonarsource: Fixed in cd06099d0.

Bun 1.4.2 includes an escaped literal workspace path (`"sub/\\*"`) when the directory is literally `sub/*`. The resolver now normalizes literal workspace paths with Bun-compatible separators before comparing them, and a regression test verifies that its `catalog:` dependency resolves from the ancestor catalog.

I did not add extglob matching: Bun 1.4.2 rejects `"sub/+(a|b)"` as `Workspace not found`, so accepting it would diverge from Bun.

packages/analysis/src/jsts/rules/helpers/dependency-manifests/resolvers/package-json.ts
```diff
@@ -212,6 +213,132 @@ function findClosestParentPackageJsonWithCatalogs(
...
+    : normalizedPattern;
+  while (normalizedPattern.endsWith('/')) {
+    normalizedPattern = normalizedPattern.slice(0, -1);
+  }
+  return isExclusion ? `!${normalizedPattern}` : normalizedPattern;
+}
+
+function normalizeLiteralWorkspacePath(pattern: string): string {
+  return path.posix.normalize(pattern.replaceAll('\\', '/'));
+}
```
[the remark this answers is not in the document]
erwan-leforestier-sonarsource: Fixed in b3ac614e6. `normalizeLiteralWorkspacePath` now removes escape characters before POSIX normalization, so escaped literal characters remain literal rather than becoming separators. The dependency-manifest suite passes (57 tests), along with `npm run bbf`.

packages/analysis/tests/fixtures-package-jsons/bun-workspace-patterns-escaped-literal/sub/[star]/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "star",
+  "workspaces": ["child"],
+  "dependencies": {
+    "react": "catalog:"
+  }
+}
```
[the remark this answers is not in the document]
Comment (erwan-leforestier-sonarsource) [reply packages/analysis/tests/fixtures-package-jsons/bun-workspace-patterns-escaped-literal/sub/[star]/package.json:7]: Fixed in b3ac614e6. The fixture now uses the Windows-valid literal directory `sub/[star]` and an escaped workspace pattern (`sub/\\[star\\]`), so it exercises the same escape path without an uncheckoutable filename. The dependency-manifest suite passes (57 tests), along with `npm run bbf`.

PR #7895, opened 2026-09-07
JS-2406 Implement rule S9381: Promises should not be nested

martin-strecker-sonarsource:
Fixes JS-2406

## Summary

Wraps eslint-plugin-promise's `no-nesting` rule as a decorated rule (`implementation = 'decorated'`) rather than reimplementing it, with a small decorator that suppresses one class of upstream false positives.

- Adds `eslint-plugin-promise` as a dependency and registers it as a new `promise` external plugin in `registry.ts`, alongside the existing react/unicorn/vue/etc. wraps
- Adds a `typings/eslint-plugin-promise` shim since the plugin ships no TypeScript types (same pattern as the existing `eslint-plugin-jsx-a11y` shim)
- `rspec.sha` is pinned to [SonarSource/rspec#8123](https://github.com/SonarSource/rspec/pull/8123) (still unmerged) per `docs/BUILD.md`'s documented process for a PR that depends on an in-flight RSPEC change, so local builds/CI pick up the real RSPEC content instead of placeholder defaults
- **Added to the `Sonar way` default profile** — reversed from an earlier "opt-in only" decision after discussion on the RSPEC PR: this is one of the "Gap Analysis" rules where the default-profile signal is wanted to see how users actually interact with it. Still noted in that discussion as closer to a styling/personal-preference rule than a typical correctness rule.
- `decorator.ts` suppresses reports where an intervening non-promise function (a `.map()`/`.reduce()`/`.each()` callback, a deferred event-handler closure, a `Promise` executor, a transaction-driver callback, ...) separates the reported call from its enclosing promise callback — the upstream rule's own nesting-tracker has no concept of this boundary, so it flags cases that aren't avoidable sequential nesting. Deliberately does not special-case IIFEs: not observed anywhere in the ITS corpus, and there's no real motivating reason to write one specifically inside a `.then()` callback.
- Added unit test coverage (`unit.test.ts`) per review request, covering: flat chain, the closure-variable exception, the suppressed intervening-function patterns, a regression case proving genuine nesting *inside* an intervening function still gets reported, and a documented known gap (nesting inside a plain named function invoked synchronously nearby — not an IIFE, but not caught either).

## Test plan

- [x] `npm run bbf` and `npm run eslint-plugin:check` pass
- [x] `unit.test.ts`: 7 valid / 3 invalid cases covering the rule's own behavior plus the decorator's suppression logic
- [x] Full ruling suite (59 projects): 59/59 pass after syncing the `S9381` baseline
- [x] Triaged all 171 real-world ITS findings by hand before the decorator (150 TP / 21 FP, ~88% precision) — see [PR comment](https://github.com/SonarSource/SonarJS/pull/7895#issuecomment-5581785882)
- [x] Re-ran the full corpus after the decorator: 21 reports suppressed across 14 files, verified against real source that the suppressions generalize correctly (not over-suppressing) — see [follow-up comment](https://github.com/SonarSource/SonarJS/pull/7895#discussion_r3967969069)


martin-strecker-sonarsource:
Went through all 171 new `S9381` findings (grouped by file, read surrounding context for each) — **150 TP / 21 FP** (~88% precision).

**FP breakdown:**

| Category | Count | Example |
|---|---|---|
| Misidentification (parallel fan-out via `.map()`/`Promise.all()`, deferred event-handler closures, DB transaction callback boundaries, deliberately-scoped distinct error paths, etc.) | 17 | `Ghost/core/client/app/controllers/setup/three.js:175` — `.then()` inside a `.map()` building independent promises for `RSVP.Promise.all()`, not sequential nesting. Checked upstream for a tracked issue on this — none exists; appears undocumented. |
| Error-recovery / retry nesting | 3 | `Ghost/core/server/api/users.js:340-341` — re-fetch-after-failed-send recovery flow; matches a known upstream gap, [eslint-plugin-promise#58](https://github.com/eslint-community/eslint-plugin-promise/issues/58) |
| Skip-level data dependency | 1 | `Ghost/core/server/models/user.js:557` — `.catch` needs `user` from two levels up. Not a separately tracked issue, but it's the unaddressed edge of the exact fix in [eslint-plugin-promise#123](https://github.com/eslint-community/eslint-plugin-promise/issues/123)/[PR#361](https://github.com/eslint-community/eslint-plugin-promise/pull/361) — that fix only checks the *closest* enclosing scope, so a dependency one level further out still slips through. |

No FPs fell outside these three buckets.

**Takeaway:** the dominant FP mode (17/21) is the rule having no concept of "this `.then()` is building a promise for parallel aggregation" vs. genuine sequential-habit nesting — a structural limitation inherited from the upstream rule we're wrapping, not something introduced by the wrap, and apparently not reported upstream either. Given S9381 is already opt-in (not in `Sonar way`), this seems like an acceptable trade-off rather than something to patch with a decorator.

its/ruling/src/test/expected/angular.js/javascript-S9381.json
```diff
@@ -0,0 +1,9 @@
+{
```
guillemsarda, line 1: I think this false-positive pattern is feasible to address with a small decorator. In `Ghost/core/client/app/controllers/setup/three.js:175,180`, each `users.map(...)` iteration builds an independent promise and converts its success or failure into an invitation result before `Promise.all()` collects those results. Moving those handlers onto the outer chain would change their meaning: they would handle the aggregate operation instead of each invitation.

The upstream rule keeps tracking the outer promise callback across the intervening `.map()` callback. We could use `interceptReport` to suppress a report when an intervening non-promise function separates the reported call from the enclosing promise callback, while looking through direct IIFEs. This also covers deferred event handlers and transaction callbacks, and still reports genuine nested promise chains inside those functions.

A local prototype using our existing interceptor reproduced all 171 ITS findings and suppressed 21 reports across 14 files. Eleven focused checks preserved ordinary nesting, nesting inside `.map()` callbacks, and IIFEs. These are not the same 21 reports classified in the earlier comment; retry and scoped error-handling cases remain outside this proposed fix.

Could we add this bounded decorator, together with basic rule tests for nested promises, flat chains, the closure-variable exception, and these callback-boundary cases? Even though the implementation comes from an external rule, that coverage would document the expected behavior for future maintainers and protect dependency upgrades.

packages/analysis/src/jsts/rules/S9381/index.ts
```diff
@@ -0,0 +1,18 @@
+/*
```
guillemsarda, line 1: Even though the `new-rule` skill doesn't consider it (it will be changed soon), we should add basic test coverage for all kinds of rules (external or decorators included) rule. This helps future maintainers to understand quicker the rule behavior. Please, add basic tests coverage.

guillemsarda (changes requested):
If we add the rule in the "Sonar Way" profile (TBD in the rspec PR), as you mentioned, we should reduce the potential FPs  by decorating it.

its/ruling/src/test/expected/angular.js/javascript-S9381.json
martin-strecker-sonarsource: Agreed — one simplification first: drop the IIFE-transparency handling from the walk. IIFEs run synchronously in place, so nesting inside one is just as flattenable as without it, but it doesn't show up anywhere in the 171 ITS findings and there's no real reason to write one specifically inside a `.then()` callback (it already has its own scope). So the walk becomes binary: nearest enclosing function is a promise callback (report) or it isn't (suppress) — smaller decorator, one less thing to test.

Separately: this is a bug in the upstream algorithm itself, not our wrap — raw `eslint-plugin-promise` would misfire the same way on these shapes, and nothing matching it is filed on their tracker. Worth opening an issue/PR upstream with the same fix? Would help everyone using the plugin directly and let us retire the decorator if it lands, but it's extra process for a rule we may not stay deeply invested in. What's your read?

martin-strecker-sonarsource: Implemented and validated against the full ITS ruling corpus (59 projects, 171 S9381 findings): **21 reports suppressed across 14 files** — exactly matching your prototype's numbers.

Two things worth calling out:

**IIFEs made zero difference on this corpus.** Since none of the 171 findings involve one, a version with IIFE-transparency and the simplified version without it produce identical results here — confirms dropping that check was safe, at least against real data.

**The suppressions generalize correctly beyond the 3 examples I originally showed you** — `_.map()`/`Promise.map()`/`.each()` fan-out in several more files, and a `new Promise((resolve, reject) => {...})` executor case (the executor's parent is a `NewExpression`, not a `.then()`/`.catch()` call, so it's structurally the same "intervening function" shape). I verified each against the real source rather than trusting the count — several of these were actually called TP in the original manual triage, using reasoning like "extractable into a named helper" that doesn't actually distinguish TP from FP (you can always extract *anything* into a helper). So the decorator's structural rule is more consistent than that manual pass was.

**One gap remains, and it's not caused by dropping IIFE-handling** — a literal IIFE check wouldn't catch it either: `Ghost/core/server/controllers/frontend/index.js:190` nests inside a plain named function (`function render() {...}`) that's declared and then invoked synchronously nearby (`return render();`), not deferred. It's not an IIFE, so IIFE-transparency wouldn't have helped — it's a general blind spot of "nearest enclosing function is/isn't a promise callback" that only a real synchronous-invocation analysis could close, which felt like disproportionate complexity for one instance.

sonar-plugin/javascript-checks/src/main/resources/org/sonar/l10n/javascript/rules/javascript/S9381.json
```diff
@@ -0,0 +1,33 @@
...
+  "quickfix": "infeasible",
+  "code": {
+    "impacts": {
+      "MAINTAINABILITY": "MEDIUM"
+    },
+    "attribute": "CLEAR"
+  },
+  "defaultQualityProfiles": [
+    "Sonar way"
+  ]
```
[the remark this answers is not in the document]
martin-strecker-sonarsource: Confirmed, but fixed via the PR description rather than the code — the code is actually correct here. `defaultQualityProfiles: ["Sonar way"]` is intentional: we reversed the earlier opt-in decision after discussion with @guillemsarda on the RSPEC PR (it's one of the "Gap Analysis" rules where the default-profile signal is wanted). The PR description was just stale from before that reversal — updated it to reflect the current state instead of reverting the code.

packages/analysis/src/jsts/rules/S9381/decorator.ts
```diff
@@ -0,0 +1,54 @@
...
+  const parent = fn.parent;
+  return (
+    !!parent &&
+    parent.type === 'CallExpression' &&
+    parent.callee.type === 'MemberExpression' &&
+    !parent.callee.computed &&
+    parent.callee.property.type === 'Identifier' &&
+    (parent.callee.property.name === 'then' || parent.callee.property.name === 'catch')
+  );
+}
```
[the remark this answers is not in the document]
martin-strecker-sonarsource: Tested this directly — the suppression is real, confirmed with the exact example: bare upstream `no-nesting` reports 1 issue on `a().then(() => b().finally(() => c().then(() => d())))`, our decorated version reports 0. So the behavior you describe is accurate.

On the citation though: `is-promise.js` is a real file and does treat `finally` as a promise expression — but `no-nesting.js` doesn't import it. It only imports `has-promise-callback.js` and `is-inside-promise.js`, and both of those check *only* `then`/`catch`, explicitly excluding `finally`. `is-promise.js` is used by other rules (`always-return`, `catch-or-return`, `prefer-await-to-then`, etc.), not this one.

More importantly, I don't think the suggested fix is correct even setting the citation aside. Ran a real runtime check:

```js
Promise.resolve('X').finally(() => Promise.resolve('Y').then(y => console.log(y))).then(v => console.log(v));
// logs: Y, then X
```

`.finally()` always passes through the *original* settled value to whatever comes after it, regardless of what its own callback resolves to. So `.finally(() => c().then(() => d()))` and `.finally(() => c()).then(() => d())` aren't equivalent — `d` gets fed a completely different value in each (the finally callback's own resolution vs. the untouched original value). That's the same shape as our error-recovery FP category: flattening isn't behavior-preserving, so treating a `.finally()` callback as "intervening" (not avoidable nesting) looks intentional-correct rather than a bug. Not applying this one — happy to revisit if I'm missing a case where it would actually be equivalent.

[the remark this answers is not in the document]
martin-strecker-sonarsource: Added a regression test (`unit.test.ts`) codifying this exact case as `valid` with the runtime-verified reasoning inline, so it reads as a deliberate decision rather than an oversight for whoever looks at this next. Not applying the suggested fix, per the reasoning above.

packages/analysis/src/jsts/rules/S9381/decorator.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export function decorate(rule: Rule.RuleModule): Rule.RuleModule {
+  return interceptReport(
+    { ...rule, meta: generateMeta(meta, rule.meta!) },
+    (context, descriptor) => {
+      const node = (descriptor as unknown as { node: TSESTree.Node }).node;
+      const enclosingFunction = findFirstMatchingAncestor(node, n =>
+        FUNCTION_NODES.includes(n.type),
+      );
```
guillemsarda, line 45: (nit) you can use `isFunctionNode` from `packages/analysis/src/jsts/rules/helpers/ast.ts` which is the same exact logic that you are expressing here.


```suggestion
      const enclosingFunction = findFirstMatchingAncestor(node, isFunctionNode)
```

its/ruling/src/test/expected/angular.js/javascript-S9381.json
guillemsarda: > Separately: this is a bug in the upstream algorithm itself, not our wrap — raw `eslint-plugin-promise` would misfire the same way on these shapes, and nothing matching it is filed on their tracker. Worth opening an issue/PR upstream with the same fix? Would help everyone using the plugin directly and let us retire the decorator if it lands, but it's extra process for a rule we may not stay deeply invested in. What's your read?

For what I could experience, we didn't have this habit in SonarJS. For now, I'd keep it simple and lean toward not opening an Issue. Nevertheless, I'll bring the topic in the next discussion with the ex web-squad.

guillemsarda (approved):
Image

guillemsarda (approved):
LGTM. Small optional nit: https://github.com/SonarSource/SonarJS/pull/7895#discussion_r3970139721

PR #7896, opened 2026-09-07
JS-2407 Implement rule S9383: Promises should not be left unhandled

martin-strecker-sonarsource:
Part of JS-2407

## Summary
- Implements S9383, wrapping typescript-eslint's `no-floating-promises` the same way S6544 wraps `no-misused-promises`.
- Default options are inherited from upstream unchanged (`ignoreVoid: true`, `checkThenables: false`, `ignoreIIFE: false`, `allowForKnownSafeCalls`/`allowForKnownSafePromises` empty).
- Upstream's "add await"/"add void" suggestions are passed through unmodified (not stripped), following the established pattern for decorated wraps of suggestion-producing typescript-eslint rules (e.g. S6572, S6582). RSPEC declares `quickfix: covered` to match this codebase's `validate-quickfix.ts`, which only accepts `covered` as the pass state for `hasSuggestions=true` (not `partial`, even though that's also a documented RSPEC value).
- Fixes a false positive found during the manual ruling review directly in the decorator: upstream's `isValidRejectionHandler()` misclassifies an `any`-typed rejection handler as "not a function", since `any` structurally has zero call signatures despite being callable at runtime. Filed [typescript-eslint/typescript-eslint#12848](https://github.com/typescript-eslint/typescript-eslint/issues/12848) with a proposed upstream fix — **closed by the maintainer as working-as-intended** (deliberate FP/FN tradeoff on their side, not a bug). `S9383/rule.ts` now intercepts the two `floatingUselessRejectionHandler(Void)` messageIds and suppresses the report locally when the handler's resolved type is exactly `any`, covering all 4 real-world occurrences from the ruling review (see follow-up comment below for details).
- RSPEC is merged: [SonarSource/rspec#8128](https://github.com/SonarSource/rspec/pull/8128) landed the S9383 rule metadata on the RSPEC default branch. The local RSPEC mirror (`S9383.json`) and the temporary root `rspec.sha` pin (used per `docs/BUILD.md#baseline-ci-mismatches` / `docs/DEV.md` while the rule wasn't yet upstream) have both been removed now that CI's `prepare_rspec_rule_data` job picks up the real, merged metadata.

## Ruling
- Ran the full ruling suite; every project producing an S9383 hit had a new baseline added (25 files across 24 projects, 791 total findings).
- Spot-checked the two highest-volume projects before syncing:
  - `desktop` (360 issues/76 files): unawaited `async` `conditionalVersion(...)` calls inside DB constructors (Dexie migration setup) — real fire-and-forget async work.
  - `eigen` (257 issues/186 files): spread ~1/file, e.g. unawaited async calls inside `useEffect` and unawaited `navigate(...)` in ternaries — genuine floating promises, not a detection bug.
- `npm run ruling` is green (59/59) after `ruling-sync`; re-verified green after the suggestion pass-through change (locations/messages unaffected).
- Full manual TP/FP review of all 791 ruling findings is in progress; will post results as a follow-up comment.

## Test plan
- [x] `npx tsx --test packages/analysis/src/jsts/rules/S9383/cb.test.ts` passes (12 comment-based scenarios: floating statement, awaited, returned, `.catch()`, `.then()` with/without rejection handler, `void`-ignored, floating promise array vs. `Promise.all()`, floating async IIFE, `any`-typed rejection handler (not reported), non-function rejection handler (still reported)), including exact quick-fix output assertions (`[[qfN]]`/`fix@`/`edit@`) for the scenarios that carry suggestions.
- [x] `npm run bbf` builds clean.
- [x] `mvn install -DskipTests` builds clean (validated `validate-quickfix` and RSPEC refresh against the repinned commit).
- [x] `npm run ruling` green after baseline sync.
- [x] Merge/sync SonarSource/rspec#8128 before merging this PR.
- [x] Remove root `rspec.sha` pin before merging to master.


martin-strecker-sonarsource:
## S9383 ruling validation — 791 findings

| | Count | % |
|---|---|---|
| **True Positive** | 775 | 98.0% |
| **Confirmed False Positive** | 6 | 0.8% |
| **Borderline / low-value TP** (not a defect — see below) | 10 | 1.3% |

### Confirmed false positives — 2 distinct root causes, both pre-existing upstream limitations (not introduced by our wrapping)

**Root cause A — `any`-typed rejection handler misread as "not a function" (4 instances)**
Upstream's `isValidRejectionHandler()` checks `getCallSignatures().length > 0` on the handler's resolved TS type. `any` structurally has zero call signatures even though it's callable at runtime by definition, so a genuinely-callable handler gets misreported as "*A rejection handler that is not a function will be ignored*."
Not previously tracked upstream (checked) — filed [typescript-eslint/typescript-eslint#12848](https://github.com/typescript-eslint/typescript-eslint/issues/12848) with a verified minimal repro and a proposed one-line fix (`tsutils.isIntrinsicAnyType` short-circuit, using a helper already imported in that file), confirmed locally to resolve all four occurrences without weakening detection of genuinely non-callable handlers.
- `ace/src/ext/code_lens_test.js:114` and `:139` — `.catch(next)`, untyped Mocha callback parameter.
- `p5.js/src/core/preload.js:93` — `.catch(errorCallback)`, untyped rest-args reassignment.
- `eigen/scripts/algolia-example-galleries-near-me.ts:82` — `.catch(console.error)`; project's `tsconfig.json` has no `"node"`/DOM lib, so `console` is untyped there.

**Root cause B — array-mutation method misread as a floating promise array (2 instances)**
- `vitest/packages/expect/src/utils.ts:36` and `vitest/packages/runner/src/artifact.ts:148` — `.splice()` on an already-`.finally()`-handled `Promise[]` tracking array. `.splice()`'s return type is coincidentally `Promise[]` too, which trips the array-of-promises heuristic even though nothing new is being launched.

### Borderline cases (10) — technically true positives per the rule's literal contract, but low real-world signal
- **8×** `vuetify/packages/vuetify/src/components/VOverlay/useActivator.tsx` — timer-backed promises whose executor structurally never calls `reject` (verified). Matches the exception our own rule.adoc already documents ("can be ignored for promises you know will always resolve like timers").
- **1×** `oak/http_server_bun.ts:210` — flagged from an interface's declared `void | Promise<void>` union; this concrete override never returns a promise at runtime.
- **1×** `vitest/packages/vitest/src/runtime/moduleRunner/nativeModuleMocker.ts:199` — the derived `.finally()` chain is technically unhandled even though the original promise is awaited by the caller elsewhere.

### Non-issue worth a one-line mention
`ant-design/components/form/__tests__/list-noStyle.test.tsx:48-49` — correct TP, but S9383 is `scope: Main`; a real SonarQube scan excludes test files at the Java-plugin level. The ruling harness doesn't apply that filter, so it surfaces here but wouldn't in production.

**Bottom line: zero false positives traceable to our wrapper** — both FP categories are pre-existing upstream `typescript-eslint` limitations (one now filed and fix-proposed), and at 6/791 (0.8%) they're narrow enough that I wouldn't block on them.

martin-strecker-sonarsource:
Re: https://github.com/SonarSource/SonarJS/pull/7896#issuecomment-5585396188

Checked both points against current state rather than the diff snapshot the bot analyzed:

**Point 2 (description contradicts implementation) — already stale.** The PR description was updated two commits ago to reflect that suggestions are passed through (not stripped) and RSPEC declares `quickfix: covered` (not `partial`/`unknown`). No further action needed there.

**Point 1 (no automated guard against `rspec.sha` reaching `master`) — real, but the framing is slightly off.** A guard already exists: the "Reject root rspec.sha" step inside the `promote` job (`if [ -e rspec.sha ]; then exit 1; fi`) — it's currently failing on this PR, exactly as intended. So it's not that no CI step exists.

The actual gap: I checked this repo's branch-protection rulesets via the API (`gh api repos/SonarSource/SonarJS/rulesets/<id>`), and `promote` is **not** a required status check on `master` — the only required check is `StepSecurity Required Checks` (a security scan, unrelated). So nothing currently stops a merge while that step is red; the existing guard just isn't wired up as *required*.

That's a repo-level branch-protection/ruleset setting, not something a code change in this PR's diff can fix — happy to raise it separately with whoever administers branch protection for this repo, since it'd protect every future `rspec.sha` pin, not just this one. For this PR specifically, the pin removal is already tracked as an explicit checklist item in the description.

packages/analysis/src/jsts/rules/S9383/rule.ts
```diff
@@ -0,0 +1,31 @@
...
+import * as meta from './generated-meta.js';
+
+const noFloatingPromisesRule = tsEslintRules['no-floating-promises'];
+
+export const rule: Rule.RuleModule = {
+  meta: generateMeta(meta, { ...noFloatingPromisesRule.meta }),
+  create(context: Rule.RuleContext) {
+    return noFloatingPromisesRule.create(context);
+  },
+};
```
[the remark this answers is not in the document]
martin-strecker-sonarsource: description updated

martin-strecker-sonarsource:
**Update (2026-09-10):** typescript-eslint/typescript-eslint#12848 (filed above for root cause A) was closed by the maintainer (`bradzacher`) as `NOT_PLANNED`/working-as-intended — treating an `any`-typed rejection handler as reportable is a deliberate FP/FN tradeoff on their side, not a bug they're planning to fix. Since this is now a permanent upstream characteristic rather than a pending fix, root cause A's 4 false positives are suppressed locally in this PR's decorator instead: `S9383/rule.ts` intercepts the `floatingUselessRejectionHandler(Void)` messageIds and drops the report when the rejection handler's resolved type is exactly `any`. Root cause B (the `.splice()`-on-tracking-array FPs) is unaffected by this and remains open.

packages/analysis/src/jsts/rules/S9383/rule.ts
```diff
@@ -0,0 +1,99 @@
...
+    return args[0] ?? null;
+  }
+  if (callee.property.name === 'then') {
+    return args[1] ?? null;
+  }
+  return null;
+}
+
+// `any` has no call signatures though it's callable at runtime; upstream closed this
+// as working-as-intended (typescript-eslint/typescript-eslint#12848), so fix it here.
```
guillemsarda, line 62: Thanks for the Issue!

packages/analysis/src/jsts/rules/S9383/cb.fixture.ts
```diff
@@ -0,0 +1,68 @@
...
+
+function floatingAsyncIife() {
+  (async () => { // Noncompliant [[qf5,qf6=0]] {{Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator.}}
+  // fix@qf5 {{Add void operator to ignore.}}
+  // edit@qf5 {{  void (async () =>}}
+  // fix@qf6 {{Add await operator.}}
+  // edit@qf6 {{  await (async () =>}}
+    await fetchData();
+  })();
+}
```
guillemsarda, line 68: Worth adding two regression tests for the **Optional chaining** and **Computed member access** cases.

Optional chaining:
```ts
declare function fetchData(): Promise<void>;

function handle(next: any) {
  fetchData()?.catch(next);
}

declare const maybePromise: Promise<void> | undefined;

function handle(next: any) {
  maybePromise?.then(undefined, next);
}
```

Computed member access:
```ts
declare function fetchData(): Promise<void>;

function handle(next: any) {
  fetchData()['catch'](next);
}

function handle(next: any) {
  fetchData()[`then`](undefined, next);
}

**Review by guillemsarda — APPROVED** [review]
LGTM. Worth adding the two tests.

File: packages/analysis/src/jsts/rules/S9383/rule.ts [hunk 4]
```diff
@@ -0,0 +1,99 @@
+/*
```
guillemsarda, line 1: (Optional) This rule puts the decoration logic in `rule.ts`, but the established convention across the codebase (~95 decorated rules) is to split it into a dedicated `decorator.ts`, with `index.ts` doing `export const rule = decorate(getBaseRule(...))`. Only a few rules (S6544, S6572, S6582) currently deviate from that and inline everything in `rule.ts`.

Could we rename `rule.ts` → `decorator.ts` here (and adjust `index.ts` accordingly) to follow the majority pattern, rather than growing the outlier group to 4? Makes it easier for reviewers/future maintainers to find the FP-suppression/quick-fix logic in a consistent place across rules.

guillemsarda (approved):

martin-strecker-sonarsource:
Reviewed the Peach validation results (929 new / 0 lost issues): all 929 are `S9383` (673 TypeScript, 256 JavaScript), no other rule regressed. Spot-checked source snippets across several projects (ag-grid, rollup, jest, koel, ngx-admin) - findings range from clear true positives (e.g. unhandled `HTMLMediaElement.play()`, `fetch().then().then()` chains with no `.catch`) to plausible fire-and-forget calls consistent with the rule's intent. Nothing suspicious found - no FP patterns, no decoder/location mismatches.

PR #7900, opened 2026-09-08
JS-2409 Implement rule S9382: Promises should not be awaited sequentially in a loop

martin-strecker-sonarsource:
Fixes [JS-2409](https://sonarsource.atlassian.net/browse/JS-2409).

## Summary

- New rule `S9382` flags `await` used inside a loop body, which forces each iteration's async operation to run sequentially instead of in parallel (matches Biome's `noAwaitInLoops`, which is itself just a re-exposure of ESLint core's `no-await-in-loop`).
- Implemented as a plain wrap of ESLint core's `no-await-in-loop` (`external/core.ts`), following the existing `S1751`/`S1763`/`S878` pattern — no custom decorator/behavior change for now. That's a deliberate scope decision, not an oversight: see "Open design question" in JS-2409 about whether to add a Sonar-side decorator suppressing the early-exit false-positive case ESLint's own maintainers declined to fix ([eslint#14700](https://github.com/eslint/eslint/issues/14700)).
- Not enabled in the default "Sonar way" profile (`defaultQualityProfiles: []`), matching ESLint's own `recommended: false` and Biome's off-by-default posture — this rule has a known high-noise profile on intentionally-sequential code (dependent iterations, retries, rate limiting, ordered writes), documented in the rule's `Exceptions` section.
- `quickfix: infeasible` — rewriting to `Promise.all()` changes execution/error semantics, not a safe mechanical fix.

## Dependency

- RSPEC key `S9382` reserved via [SonarSource/rspec#8127](https://github.com/SonarSource/rspec/pull/8127) (still in review as of this PR). This PR includes a local mirror of that rule's `metadata.json` at `sonar-plugin/javascript-checks/.../rules/javascript/S9382.json`, since `npm run generate-meta` reads from this vendored copy rather than the live rspec repo. If the rspec content changes during review, this file needs to stay in sync.

## Test plan

- [x] `unit.test.ts` covers the `for-await-of` exclusion plus `for...of`/classic `for`/`while`/`do...while`/`for...in` invalid cases.
- [x] Full `npm run bridge:test` suite: 2371/2388 passing; the 17 failures are all pre-existing, unrelated gRPC server port-binding flakiness in `analyze-project-server.test.ts`, not touched by this change.
- [x] `npm run ruling` run against the full third-party corpus; new findings spot-checked for correctness (e.g. `vuetify/.../form.ts:89` sequential validation loop, `sizzle/tasks/karma-tests.js:30` sequential subprocess spawning). Synced via `npm run ruling-sync` — 11 new expected-result files added, no existing baselines changed.

[JS-2409]: https://sonarsource.atlassian.net/browse/JS-2409?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


martin-strecker-sonarsource:
**Ruling results: 78 TP / 169 FP across 247 findings (32% TP rate)**

Ran the full ruling corpus and manually classified every finding. Full breakdown in [JS-2409](https://sonarsource.atlassian.net/browse/JS-2409).

The FP volume is expected — `S9382` wraps ESLint core's `no-await-in-loop` verbatim, which is purely syntactic (any `await` lexically inside a loop, no data/control-flow analysis of whether iterations are actually independent). Most FPs are legitimate and not fixable by us: shared mutable state / lock safety (git index ops, XML stream writers), test/runtime isolation by design, retry/polling loops, and already-implemented rate-limited concurrency.

One category is different: ~14 of the 169 FPs (~8%) are "first-match-wins" early-exit searches — the loop's only `await` sits on a branch that unconditionally `return`s/`break`s right after, so the loop runs at most once in practice and there's nothing to parallelize. This is the exact FP class ESLint's own maintainers declined to fix upstream ([eslint#14700](https://github.com/eslint/eslint/issues/14700), closed as "working as intended"). It's the one class we could plausibly suppress ourselves via a decorator.

`S9382` is currently **not in the default Sonar Way profile** (`defaultQualityProfiles: []`), given this FP profile.

**Draft spec for a suppression decorator (unsettled in places — see below)**

Both the `await` node and its enclosing loop are accessible at report time — ESLint's own `context.report({ node: awaitNode, ... })` passes the `AwaitExpression` itself, `.parent` links are intact (confirmed via source, not assumed), and `interceptReport` passes the node through unmodified.

Sketch: walk forward from the await's containing statement; if every reachable path hits `break`/`return`/`throw` before the loop could iterate again, suppress.

```
exitsLoopOnEveryPath(position, loopNode):
  next = nextStatement(position)   // pop to parent's next statement when a block ends
  if next is "fell off the end of loopNode's body": return false
  match next.type:
    BreakStatement                  → targetsLoop(next, loopNode)
    ContinueStatement                → false
    ReturnStatement, ThrowStatement  → true
    BlockStatement                   → exitsLoopOnEveryPath(enter(next), loopNode)
    everything else (if/switch/nested-loop/plain statements) → [see open questions]
```

Two correctness knobs still open, both narrow edge cases:
1. **A later `if`/`else` where both arms `break`/`return`.** Handling this needs recursing into both branches; skipping it (treating the `if` as opaque) just means we miss a rare additional suppression opportunity — safe, just less thorough.
2. **A later `if`/`switch`/nested-loop that itself contains a `continue`** (e.g. `await x(); if (cond) { continue; } break;`). Fully skipping over these (not inspecting their contents at all) risks the *opposite* mistake — wrongly suppressing a real finding, since one branch could re-enter the loop. Bailing out conservatively on these avoids that risk but adds a bit more logic.

Nested loops (a `break` that only exits an inner loop while an outer loop can still re-trigger the same await) are out of scope entirely for v1 — deferred as a documented follow-up regardless of how (1) and (2) are resolved.

**Given this is a non-Sonar-Way, opt-in rule — do we want to dig deeper into (1)/(2) here, or ship the passthrough as-is (matching S878/S1751/S1763 — no decorator, just the vanilla wrap) and revisit only if it comes up in practice?**

@guillemsarda


[JS-2409]: https://sonarsource.atlassian.net/browse/JS-2409?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

packages/analysis/src/jsts/rules/S9382/unit.test.ts
```diff
@@ -0,0 +1,105 @@
...
+    }`),
+
+        valid(`
+    async function foo(arr) {
+      for (const x of arr) {
+        bar(x);
+      }
+    }`),
+
+        valid(`
```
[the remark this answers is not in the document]
martin-strecker-sonarsource: Applied — added the function-boundary case verbatim as suggested, plus a second valid case for the for-await-of header expression (`for await (const x of await getAsyncIterable())`), confirmed empirically against ESLint's actual source: `isBoundary` treats a for-await-of as a boundary purely from `node.await === true`, before the body/header distinction is even checked, so an await in either position is exempt. Pushed in 0169e8978.

packages/analysis/src/jsts/rules/S9382/unit.test.ts
```diff
@@ -0,0 +1,119 @@
+/*
```
guillemsarda, line 1: There's a known false positive that isn't mitigated: `await` immediately followed by a loop-terminating `return`/`break` still gets flagged, even though there's no real sequential-execution cost since the loop won't continue.

```js
async function foo(arr) {
  for (const x of arr) {
    const r = await fetch(x);
    if (r.ok) return r; // loop ends right after — flagging this await is noisy
  }
}
```

This matches upstream eslint/eslint#14700, which was closed as "working as intended" since it's out of scope for the base `no-await-in-loop` rule. Since we're wrapping it for a Sonar-specific rule, could we add a decorator to suppress the report when the `await` is the last statement before a `return`/`break` that exits the loop? Otherwise this will be a live source of noise on early-exit-after-await code, which is a pretty common pattern.

packages/analysis/src/jsts/rules/S9382/unit.test.ts
```diff
@@ -0,0 +1,119 @@
...
+ * SonarQube JavaScript Plugin
```
guillemsarda, line 2: Test coverage gap: `unit.test.ts` doesn't exercise a few cases that the underlying ESLint `no-await-in-loop` logic does handle:

- `await` in the test/update clause of a classic `for(;;)` loop, e.g. `for (let i = 0; await cond(i); i++) { ... }`
- `await using` declarations inside a loop, e.g. `for (const x of arr) { await using r = acquire(); }`
- `await` nested inside intermediate control flow (e.g. `if`) within the loop body, to confirm detection still works through non-function nesting:
  ```js
  async function foo(arr) {
    for (let i = 0; i < arr.length; i++) {
      if (arr[i]) {
        await bar(arr[i]);
      }
    }
  }
  ```

Could we add valid/invalid cases for these? Without them, a future ESLint upgrade or refactor of the decorator could silently regress any of these without any test catching it.

guillemsarda (changes requested):
Requesting changes until we decide if we cover the FP.

packages/analysis/src/jsts/rules/S9382/unit.test.ts
martin-strecker-sonarsource: Thinking through how to implement the suppression, I see three options:

1. **Only suppress when the await is unconditionally followed by return/break** (no branching). This means the loop always exits after its first iteration — nonsense code, since it wouldn't be written as a loop.
2. **Suppress whenever a return/break appears anywhere later in the loop body**, regardless of whether it's on the same path as the await, e.g.:
   ```js
   for (const id of ids) {
     const item = await fetch(id);
     if (item.matches(query)) return item;
   }
   ```
   Easy to implement, but carries the potential for a high false-negative rate, since the presence of a later return/break doesn't tell us whether that path is actually taken on a given iteration.
3. A heuristic somewhere in **between 1 and 2**
4. **Real reachability/CFG analysis** to determine whether all paths after the await exit the loop. SonarJS doesn't currently have a control-flow-graph engine for JS/TS rules, so this would require building that piece first.

See also https://github.com/eslint/eslint/issues/14700 where this FP was discussed, but no heuristic was ever really speced.

What did you have in mind? If it's 4), that's a separate SE-engine question we'd want to scope on its own rather than bundle into this rule's first cut.

guillemsarda: I'd lean toward option 2.  Even a `return/break` later in the loop body that isn't strictly on the await's execution path is a strong enough signal. That shape of code (search-and-early-exit) is exactly where sequential awaiting is idiomatic, not accidental. It shows the author already intended each promise to resolve one at a time before deciding whether to keep going, so I'd rather accept the occasional false negative than keep flagging that pattern as a bug. What do you think?

martin-strecker-sonarsource: Ran the actual numbers instead of guessing. Classified all 247 ruling findings by control-flow shape relative to the flagged await:

- **8/247 (3.2%)** are option-1 territory: every path after the await unconditionally exits the loop.
- **30/247 (12.1%)** are option-2-only: a return/break exists later in the loop body, but on a different path than the await's own.
- **206/247 (83.4%)** have no later return/break at all.

Option 2 would additionally suppress those 30. I classified each of the 30 for TP/FP against the independent-iteration test (shared state / dependency between iterations / retry-backoff / rate-limited / resource-bounded / inherently-sequential = FP; independent = TP): **14 TP / 16 FP**.

So option 2 would introduce ~14 new false negatives (real, independent-iteration cases it would wrongly suppress) — things like `for (const path of paths) { if (await pathExists(path)) return path; }` in `editors/darwin.ts`/`linux.ts`/`win32.ts`, `find-account.ts`, `dev_build_cache.ts` — the "search a small list, return on first hit" idiom, occurring independently across several codebases in the corpus, not a one-off. It would also incidentally suppress 16 findings that were already FP for unrelated reasons (pagination, listener chains, retry loops, etc.) — those would go away regardless.

Does a ~14-in-247 (5.7% of all findings) false-negative rate seem like it's worth taking, against the noise reduction it buys?

packages/analysis/src/jsts/rules/S9382/decorator.ts
```diff
@@ -0,0 +1,142 @@
...
+import type { Rule, SourceCode } from 'eslint';
+import type estree from 'estree';
+import { childrenOf, getNodeParent } from '../helpers/ancestor.js';
+import { isFunctionNode, isLoopLike, type LoopLike } from '../helpers/ast.js';
+import { interceptReport } from '../helpers/decorators/interceptor.js';
+import { generateMeta } from '../helpers/generate-meta.js';
+import * as meta from './generated-meta.js';
+
+export function decorate(rule: Rule.RuleModule): Rule.RuleModule {
+  return interceptReport(
```
guillemsarda, line 26: In PVF the code runs in test files in the `ag-grid` project because the settings are not properly defined. This pitfall causes FPs like [this one](https://github.com/ag-grid/ag-grid/blob/latest/testing/behavioural/src/overlays/overlays-react-unmount.test.tsx#L166), it simulates real-time pointer events in the second, so `Promise.all` isn't a valid alternative.

I suggest adding a defensive check where if it's a test file, the rule bails out. We already have precedent for a rule-side safety net against this: `S2187`, `S2925`, `S8959`, and `S9162` all bail out early in `create()` using `isTestFile`/`isTestRelatedFile` from `helpers/test-file-pattern.ts`.

guillemsarda (approved):
I left [a comment](https://github.com/SonarSource/SonarJS/pull/7900#discussion_r4045439309) to add a defensive gate against test files. LGTM.

packages/analysis/src/jsts/rules/S9382/decorator.ts
martin-strecker-sonarsource: Applied — added the `isTestRelatedFile` bail-out in the decorator, matching the pattern used in S2187/S2925/S8959/S9162 (see `unit.test.ts` for a case using your `overlays-react-unmount.test.tsx` filename).

That said, I want to flag the underlying inconsistency this papers over rather than pretend it's clean: whether a file "is a test file" shouldn't be a per-rule decision. Right now it effectively is — some rules (S2187, S2925, S8959, S9162, and now S9382) each carry their own local `isTestRelatedFile`/`isTestFile` check, while others (e.g. S2068) have none at all, relying solely on the scanner-layer fallback in `getFileTypeForRules`. That fallback only kicks in when a project has *no* `sonar.tests` configured — once any test path is declared (even one that misses this file, as here with ag-grid's PVF settings), the fallback is skipped entirely and each rule is on its own. That's exactly why this fix had to be rule-local instead of automatic.

This traces back to [JS-1501](https://sonarsource.atlassian.net/browse/JS-1501), which originally set out to *remove* rule-level MAIN/TEST filename gating entirely and centralize it in the linter layer — but that goal was never fully reached. [JS-1906](https://sonarsource.atlassian.net/browse/JS-1906) is a still-open, rule-specific instance of the same root cause for S2068. [JS-1948](https://sonarsource.atlassian.net/browse/JS-1948) proposed a fix but was scoped to S2068/S6418 specifically and closed `Won't Do` on the claim that `filterPathAndGetFileType` already calls `isTestRelatedFile` — it doesn't; only `getFileTypeForRules` does, and only when `sonar.tests` is entirely unconfigured. Neither ticket covers the general case, and I couldn't find one that does.

@guillemsarda would you mind filing a new ticket for the general problem — centralizing test-file classification for rule selection so it's consistent regardless of whether `sonar.tests` is fully, partially, or not configured, rather than leaving it to each rule author to decide (inconsistently) on their own?

guillemsarda: @martin-strecker-sonarsource I'll create a Jira ticket to tackle this issue.

PR #7907, opened 2026-09-08
JS-2421 Use SonarLint Core's Sonar API in SQ-IDE QA

vdiez:
The shared root POM managed `sonar-plugin-api` at the analyzer's compilation version, 14.0.0.4498. That overrode SonarLint Core 11.9.0.86162's API 13.4.2.4284 in SQ-IDE integration tests, masking compatibility problems.

Move API and API test-fixture dependency management into the analyzer modules under `sonar-plugin`. Preserve API 14 and its existing scopes explicitly for the three integration-test consumers that need it. SonarLint tests then resolve Core's API transitively, without duplicating its API version in our configuration.

The SonarLint test POM retains its parent, inherited project version, license checks, and local/CI artifact selection. During test-resource generation it copies Core's published POM; a pre-analysis guard compares Core's declared API version with the runtime API and logs the loaded JAR. The guard also fails if test discovery finds no tests. Workflow commands are unchanged. Other shared dependency management remains inherited.

Validation:

- `npm ci` and `npm run bbf` passed.
- `mvn -pl sonar-plugin/sonar-javascript-plugin -am -DskipTests package` passed.
- The runtime guard accepted Core 11.9.0.86162's API 13.4.2.4284 and failed before analysis when a local dependency-management override injected API 14.0.0.4498.
- `mvn -f its/plugin/sonarlint-tests/pom.xml -DskipTests=true verify` passed.
- A focused functional test passed the guard and loaded the selected Darwin artifact, but cannot complete on this macOS host: SonarLint Core selects `/usr/local/bin/node` (22.11.0), below the analyzer's 22.12.0 minimum. CI covers the supported QA environments.

### DIFF
--- a/its/plugin/sonarlint-tests/src/test/java/com/sonar/javascript/it/plugin/sonarlint/tests/SonarLintRuntimeApiTest.java
+++ b/its/plugin/sonarlint-tests/src/test/java/com/sonar/javascript/it/plugin/sonarlint/tests/SonarLintRuntimeApiTest.java
@@ -1,0 +1,73 @@
+/*
+ * SonarQube JavaScript Plugin
+ * Copyright (C) SonarSource Sàrl
+ * mailto:info AT sonarsource DOT com
+ *
+ * You can redistribute and/or modify this program under the terms of
+ * the Sonar Source-Available License Version 1, as published by SonarSource Sàrl.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the Sonar Source-Available License for more details.
+ *
+ * You should have received a copy of the Sonar Source-Available License
+ * along with this program; if not, see https://sonarsource.com/license/ssal/
+ */
+package com.sonar.javascript.it.plugin.sonarlint.tests;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.net.JarURLConnection;
+import javax.xml.XMLConstants;
+import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.xpath.XPathFactory;
+import org.junit.jupiter.api.Test;
+import org.sonar.api.batch.sensor.SensorContext;
+import org.sonarsource.sonarlint.core.plugin.commons.ApiVersions;
+
+class SonarLintRuntimeApiTest {
+
+  @Test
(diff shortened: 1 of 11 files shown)


its/plugin/sonarlint-tests/src/test/java/com/sonar/javascript/it/plugin/sonarlint/tests/SonarLintRuntimeApiTest.java
```diff
@@ -0,0 +1,73 @@
...
+import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.xpath.XPathFactory;
+import org.junit.jupiter.api.Test;
+import org.sonar.api.batch.sensor.SensorContext;
+import org.sonarsource.sonarlint.core.plugin.commons.ApiVersions;
+
+class SonarLintRuntimeApiTest {
+
+  @Test
+  void runtime_api_matches_sonarlint_core() throws Exception {
```
erwan-leforestier-sonarsource, line 32: The check runs after the analysis tests, not before. In the Windows QA job, `SonarLintIntegrationTest` runs its 8 tests first, then `SonarLintRuntimeApiTest` starts 55 seconds later. Nothing sets the order.

Would it be better as a blocking condition before the tests run? A mismatch still fails the build, but only after 8 confusing analysis failures, so we would waste time digging into those instead of seeing the real cause.

Simplest fix: move the comparison into `TestUtils` and call it from a `@BeforeAll` in `SonarLintIntegrationTest`, so no analysis runs on a mismatch. Avoid `assumeTrue`, which skips tests and keeps the build green.

erwan-leforestier-sonarsource (approved):

its/plugin/sonarlint-tests/src/test/java/com/sonar/javascript/it/plugin/sonarlint/tests/SonarLintRuntimeApiTest.java
vdiez: Thanks. This separation is intentional: the check POM is a test-only resource generated in generate-test-resources and is not needed to run the SonarLint functional tests. Putting the check in SonarLintIntegrationTest @BeforeAll would make all of those tests fail in an IDE or direct Surefire run that has not generated that resource. We do want SonarLintRuntimeApiTest itself to fail in that situation, but not to make an instrumentation check a precondition for the functional suite. In CI the resource is generated and an API mismatch still fails the job, so we are keeping the guard isolated and accepting the later diagnostic.

PR #7949, opened 2026-09-14
JS-2452 Implement rule S1101

guillemsarda:
----
## Summary by Gitar

- **New lint rule:**
    - Implemented rule `S1101` ensuring links with identical texts have identical targets
    - Added fixtures and unit tests under `packages/analysis/src/jsts/rules/S1101/`


packages/analysis/src/jsts/rules/S1101/rule.ts
```diff
@@ -0,0 +1,572 @@
...
+
+// Only two links sharing the same root are exclusive, and not when both merely follow it as a guard.
+function shareConditionalRoot(a: LinkInfo, b: LinkInfo): boolean {
+  if (a.conditionalRoot === undefined || a.conditionalRoot !== b.conditionalRoot) {
+    return false;
+  }
+  return !(a.followsGuard && b.followsGuard);
+}
+
+function registerBaseline(siblingBaselines: Map<string, LinkInfo>, link: LinkInfo) {
```
martin-strecker-sonarsource: **[blocker]** `registerBaseline` never replaces a *conditional* baseline occupant with a later conditional link — once a conditional anchor seeds an empty `(scope, accessibleName)` slot, it permanently blocks comparison of subsequent conditional siblings against each other.

Example: `<div>{cond ? <a href="/a">Same</a> : <a href="/b">Same</a>}{isAdmin && <a href="/a">Same</a>}</div>`. The ternary consequent (`/a`) seeds the baseline; the alternate (`/b`) is correctly suppressed via `shareConditionalRoot` but never replaces it (line 152: `!link.conditional || !existing`). The `&&` branch (`/a`) then only ever gets compared against the ternary consequent, matches, and nothing is reported — even though at runtime `cond=false && isAdmin=true` renders `/b` and `/a` together with the same accessible name and different targets, a genuine violation.

packages/analysis/src/jsts/rules/S1101/rule.ts
```diff
@@ -0,0 +1,572 @@
...
+  attributes: JsxAttributes,
+  context: Rule.RuleContext,
+  elementType: (node: TSESTree.JSXOpeningElement) => string,
+): string | null {
+  const ariaLabelAttribute = getProp(attributes, 'aria-label') as JSXAttribute | undefined;
+  if (ariaLabelAttribute) {
+    const staticValue = getStaticText(ariaLabelAttribute.value);
+    if (staticValue === undefined) {
+      return null;
+    }
```
martin-strecker-sonarsource: **[blocker]** When `aria-label` is present but dynamically unresolvable, `computeAccessibleName` returns `null` and skips the anchor entirely, instead of falling back to text content per the stated `aria-label > text > title` precedence (comment on line 391).

Example: `<a href="/a" aria-label={dynamicLabel}>Same</a>` next to `<a href="/b">Same</a>` — the first anchor's static text ("Same") matches the second's, but the anchor is dropped outright because of the unresolvable `aria-label`, so the pair is never compared.

packages/analysis/src/jsts/rules/S1101/rule.ts
```diff
@@ -0,0 +1,572 @@
...
+  return {
+    scope,
+    conditional,
+    conditionalRoot: match?.root,
+    followsGuard: match?.followsGuard ?? false,
+  };
+}
+
+// `child` is conditional if it's a direct branch of `parent`, or follows a guard inside it.
+function matchConditional(
```
martin-strecker-sonarsource: **[suggestion]** The ticket asked to export/reuse `renderedChildPositions` from `helpers/jsx.ts` rather than re-deriving the "which JSX positions render conditionally" predicate. This PR only exports `isArgumentOfRenderingCall` and reimplements the rest (`getConditionalRoot`/`matchConditional`/`isScopeBoundary`) independently. Not a bug today, but the two classifications can silently drift apart if `renderedChildPositions` is later extended or fixed.

packages/analysis/src/jsts/rules/S1101/rule.ts
```diff
@@ -0,0 +1,572 @@
...
+  if (expression.type === 'Literal') {
+    const { value } = expression;
+    if (typeof value === 'string') {
+      return value;
+    }
+    // Numbers render as visible text; null/booleans render nothing.
+    return typeof value === 'number' || typeof value === 'bigint' ? String(value) : '';
+  }
+  if (expression.type === 'TemplateLiteral' && expression.expressions.length === 0) {
+    return expression.quasis.map(quasi => quasi.value.cooked ?? '').join('');
```
martin-strecker-sonarsource: **[nitpick]** Both `getStaticHref` (line 385) and this function map an unparsable template-literal escape (`cooked === undefined`) to an empty string rather than treating the whole value as unresolvable. Two hrefs/texts with different invalid escapes could collapse to the same empty contribution. Narrow edge case, low real-world likelihood.

martin-strecker-sonarsource (changes requested):
Well-designed implementation that already documents most subtle edge cases (ternary branches, guard-style if/else, unrelated `&&` conditions, switch/case, dynamic-value conservatism) in its own fixture. Two real functional gaps need fixing before merge; one design deviation from the ticket and one minor edge case are worth a look.

packages/analysis/src/jsts/rules/S1101/rule.ts
guillemsarda: Moved the logic into the shared helper: [82e900f](https://github.com/SonarSource/SonarJS/pull/7949/commits/82e900f4f49575fcdb924c4b67dcd054bd877079)

packages/analysis/src/jsts/rules/S1101/rule.ts
guillemsarda: Reverted this in 55ccc514c after a follow-up bot comment (r4027886908) caught a real regression it introduced: falling back to text content when `aria-label` is unresolvable made the rule flag its own recommended fix — e.g. `<a href="/posts/1" aria-label={`Read more about ${a.title}`}>Read more</a>` next to the same shape for `/posts/2` was reported, even though the runtime accessible names (the dynamic labels) are almost certainly distinct.

Restored the original conservative behavior: an unresolvable `aria-label` now excludes the anchor entirely, same as every other unresolvable value in this rule (dynamic `href`, unresolvable spreads, etc.). Two reasons this is the right call rather than falling back:

1. An unresolvable `aria-label` is far more likely to be a legitimate, non-empty per-item label (the idiomatic disambiguation pattern) than to actually be empty at runtime — falling back to text trades a rare false negative for a much more common false positive.
2. Per the browser's accessible-name computation, `aria-label` (once present and non-empty) fully overrides text content and title — they aren't consulted at all. So using text as a substitute isn't just a risky guess, it doesn't match how the name is actually computed.

Added a unit test covering exactly this idiomatic pattern (`unresolvable aria-label excludes the anchor`) so it can't regress the same way again. Thanks for catching this originally — the fallback direction made sense from the precedence-chain angle alone, it just didn't hold up against the "unresolvable != empty" case.

packages/analysis/src/jsts/rules/S1101/rule.ts
```diff
@@ -0,0 +1,535 @@
...
+// Joins an expression-free template literal's quasis, or undefined if any has an unparsable escape sequence.
+function cookTemplateLiteral(expression: estree.TemplateLiteral): string | undefined {
+  const cooked = expression.quasis.map(quasi => quasi.value.cooked);
+  return cooked.every((value): value is string => value !== undefined)
+    ? cooked.join('')
+    : undefined;
+}
+
+// Accessible name precedence per S6827: aria-label > text content > title.
+function computeAccessibleName(
```
martin-strecker-sonarsource, line 383: **[blocker]** `computeAccessibleName` never reads `aria-labelledby`, even though it's the highest-precedence source in the real accessible-name computation (`aria-labelledby > aria-label > text content > title`), and using it is literally the W3C-recommended technique for this exact scenario — [ARIA7: Using aria-labelledby for link purpose](https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA7). Concretely:
- **FP**: two links with identical visible text but distinct `aria-labelledby` references (the textbook ARIA7 fix) still get flagged, since the rule falls back to the identical text.
- **FN**: two links with different visible text but the same `aria-labelledby` target (same computed accessible name) pointing to different destinations are never compared.

The fixture already tests the `aria-label` version of this exact pattern (`distinctAriaLabelsSameParent`); there's no equivalent for `aria-labelledby`. The repo has precedent for treating `aria-labelledby` as name-establishing (`S6842/decorator.ts:191-197`), so this reads as an oversight rather than an intentional v1 exclusion (unlike `role="link"`, which is called out explicitly).

packages/analysis/src/jsts/rules/S1101/rule.ts
```diff
@@ -0,0 +1,535 @@
...
+        if (elementType(opening).toLowerCase() !== 'a') {
+          return;
+        }
+
+        const attributes = (opening as unknown as JSXOpeningElement).attributes;
+        if (!isSpreadSafe(attributes, context)) {
+          return;
+        }
+
+        if (isLinkHidden(attributes, context)) {
```
martin-strecker-sonarsource: **[suggestion]** `isLinkHidden` only inspects the anchor's own attributes, never its ancestors. A link hidden via a wrapping `<div aria-hidden="true">` (or ancestor `hidden`/`display:none`) is fully removed from the accessibility tree, but S1101 will still compare it against a visible sibling with the same text and report a false conflict. `resolveScope` already walks `.parent` for scope/conditional-exclusivity, and `isLinkHidden`/`isAriaHidden`/`isHiddenAttributeSet`/`isDisplayNone` already take a plain `attributes` array, so this is a small addition reusing existing machinery — not a new mechanism.

packages/analysis/src/jsts/rules/S1101/rule.ts
```diff
@@ -0,0 +1,535 @@
...
+import { report, toSecondaryLocation } from '../helpers/location.js';
+import * as meta from './generated-meta.js';
+
+const messages = {
+  identicalTextDifferentTarget:
+    'Use a distinct text or label, or point to the same target for this link and the one on line {{line}}.',
+};
+
+// Props whose presence in a spread makes the anchor unresolvable.
+const RELEVANT_PROPS = ['href', 'aria-label', 'title', 'hidden', 'aria-hidden', 'style'];
```
martin-strecker-sonarsource: **[nitpick]** `RELEVANT_PROPS` doesn't include `aria-labelledby`. Once the comment above is addressed, a spread like `<a {...props} href="/a">Text</a>` where `props` could set `aria-labelledby` should also be treated as conservatively unresolvable, same as `aria-label` today.

packages/analysis/src/jsts/rules/S1101/rule.ts
```diff
@@ -0,0 +1,535 @@
...
+
+      const hiddenState = ariaHiddenState(attributes);
+      if (hiddenState === 'unknown') {
+        return null;
+      }
+      if (hiddenState === 'hidden') {
+        return '';
+      }
+
+      if (elementType(opening).toLowerCase() === 'img') {
```
martin-strecker-sonarsource: **[suggestion]** This special-cases `` for a nested element's own accessible name, but nothing else — e.g. a nested `<svg aria-label="Download">` icon falls through to recursing into its (probably empty) children instead of using its own name, per accname's "name from content" step. Minor compared to the `aria-labelledby`/ancestor-`aria-hidden` items above, worth a look while touching this area.

martin-strecker-sonarsource (changes requested):
Solid design overall (parent-scoping, conditional-exclusivity, host-preserving destination normalization all look right after the last round). But the accessible-name computation is missing `aria-labelledby` entirely — which happens to be the W3C-recommended fix for exactly the problem this rule detects — so it will false-positive on the correct remediation. One more real gap (ancestor `aria-hidden` not checked) plus two smaller follow-on items from the same root cause.

PR #7974, opened 2026-09-16
JS-2479 JS-2480 Restrict Composition API to Vue 3

guillemsarda:
----
## Summary by Gitar

- **Vue rules update:**
    - Restricted Composition API checks in `isVue2OrEarlier` to target Vue 3 instead of Vue 2.7+
    - Updated `S8961`, `S9145`, and `S9150` rules and corresponding test fixtures to handle the Vue version requirement change


packages/analysis/src/jsts/rules/helpers/vue.ts
```diff
@@ -26,7 +26,7 @@ type VChildElement = AST.VElement | AST.VText | AST.VExpressionContainer | AST.V
 
 export type VueReactiveBindingKind = 'ref' | 'reactive';
 
-const VUE_COMPOSITION_API_MIN_VERSION = '2.7.0';
+const VUE_COMPOSITION_API_MIN_VERSION = '3.0.0';
```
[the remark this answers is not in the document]
guillemsarda: This change is too verbose and doesn't help the readability. I'll keep the name as it is with the proper description on top to reason about the cutoff at Vue 3.

packages/analysis/src/jsts/rules/S9145/fixtures/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "type": "module",
+  "dependencies": {
```
martin-strecker-sonarsource, line 3: CI is red: the "Knip" check fails with `Unused files (2)` for `S9145/cb-fixtures/cb.fixture.ts` and `S9150/cb-fixtures/cb.fixture.ts` — the `.ts` fixtures left over from moving `cb.fixture.*` into this folder. They're not referenced by anything anymore. Please delete the two stray `.ts` fixtures (or wire them into a test if they're meant to be used) so the build passes.

packages/analysis/src/jsts/rules/helpers/vue.ts
```diff
@@ -50,27 +51,16 @@ export function isInsideVueSetupScript(node: estree.Node, ctx: Rule.RuleContext)
...
- * rule. Ranges whose floor is already 2.7+ (e.g. "^2.7.0", "^2.7.0 || ^3.0.0", "^3.0.0") keep
- * reporting. Unknown/unparseable ranges (catalog:, workspace:, git:, missing dependency, ...) also
- * keep reporting.
- */
+/** Returns true when the project's Vue dependency is missing or cannot resolve to Vue 3+: Vue 2 users aren't moving off the Options API just because 2.7 backported the Composition API. */
 export function lacksCompositionApi(context: Rule.RuleContext): boolean {
   const vueVersionRange = getVueVersion(context);
-  if (!vueVersionRange || !validRange(vueVersionRange)) {
+  if (!vueVersionRange) {
+    return true;
```
martin-strecker-sonarsource, line 58: This flips the "no Vue dependency declared at all" case from reporting (old `return false`) to silencing (`return true`). JS-2480's description says the fix is to gate on the range floor and that unparseable/missing-dependency cases should "keep being flagged" — this contradicts that. Was the missing-dependency change intentional, or is the ticket description just stale relative to what shipped? If intentional, worth updating JS-2480 to describe the actual behavior; if not, this should probably go back to `return false`.

packages/analysis/src/jsts/rules/S8961/decorator.ts
```diff
@@ -22,7 +22,8 @@ import { generateMeta } from '../helpers/generate-meta.js';
 import { getVueVersion } from '../helpers/dependency-manifests/dependencies.js';
 import * as meta from './generated-meta.js';
 
-const VUE_3_OR_LATER_RANGE = '>=3.0.0';
+// "-0" makes the lower bound prerelease-inclusive, so an exact pin like "3.0.0-rc.13" still counts as Vue 3+
+const VUE_3_OR_LATER_RANGE = '>=3.0.0-0';
```
martin-strecker-sonarsource: `isVue2OrEarlier` right below this line still documents "When the Vue version cannot be determined the rule still reports" — i.e., missing dependency → still flagged, the opposite of the new `lacksCompositionApi` behavior in `helpers/vue.ts` (see comment there). These two version-gating helpers are now inconsistent with each other and nothing here explains why. Should S8961 be updated to match, or is there a reason S8961 and S9145/S9150 should diverge on this?

martin-strecker-sonarsource (changes requested):
CI is currently failing (Knip flags 2 dead fixture files from the rename) — that's a hard blocker. Separately, the `lacksCompositionApi` rewrite and the parallel `isVue2OrEarlier` gate in S8961 now disagree on how to handle a missing Vue dependency, and this diverges from what JS-2480 describes — flagging both for reconciliation before merge.

packages/analysis/src/jsts/rules/helpers/vue.ts
guillemsarda: Yes, it was intentional. I've updated the Jira ticket accordingly.

packages/analysis/src/jsts/rules/S8961/decorator.ts
guillemsarda: I actually ended up merging the two helpers because the rules were gating on Vue 3 features.

packages/analysis/src/jsts/rules/S8961/vue3-prerelease-project/unit.test.ts
```diff
@@ -0,0 +1,42 @@
...
+ */
+import { rule } from '../index.js';
+import { join } from 'node:path/posix';
+import { NoTypeCheckingRuleTester } from '../../../../../tests/jsts/tools/testers/rule-tester.js';
+import { describe } from 'node:test';
+import parser from 'vue-eslint-parser';
+
+describe('S8961 on an exact Vue 3 prerelease pin', () => {
+  const dirname = join(import.meta.dirname, 'fixtures');
+  process.chdir(dirname);
```
martin-strecker-sonarsource: I think this isn't needed — `getClosestDependencyManifestDir` checks `context.filename` before `context.cwd`.

packages/analysis/src/jsts/rules/S8961/vue3-prerelease-project/unit.test.ts
```diff
@@ -0,0 +1,42 @@
...
+ * along with this program; if not, see https://sonarsource.com/license/ssal/
+ */
+import { rule } from '../index.js';
+import { join } from 'node:path/posix';
+import { NoTypeCheckingRuleTester } from '../../../../../tests/jsts/tools/testers/rule-tester.js';
+import { describe } from 'node:test';
+import parser from 'vue-eslint-parser';
+
+describe('S8961 on an exact Vue 3 prerelease pin', () => {
+  const dirname = join(import.meta.dirname, 'fixtures');
```
martin-strecker-sonarsource: If chdir can be removed, `const dirname = join(import.meta.dirname, 'fixtures');` and `filename: join(dirname, 'component.js'),` should be merged into a single variable or inlined.

martin-strecker-sonarsource (changes requested):

packages/analysis/src/jsts/rules/S8961/vue3-prerelease-project/unit.test.ts
guillemsarda: True, I've just updated it!

martin-strecker-sonarsource (approved):
LGTM

PR #7995, opened 2026-09-18
JS-2503 S4123: False positive on await of an async JS function documented with a JSDoc @return tag

oleksandr-selehenenko-sonarsource:
Part of JS-2503

----
## Summary by Gitar

- **Bug fixes:**
    - Fixed false positive in `S4123` (`await-thenable`) when awaiting an `async` function with a non-Promise declared return type from JSDoc or TypeScript annotations


francois-mora-sonarsource (approved):
LGTM

PR #850, opened 2017-11-30
Improve DeadStoreCheck: cover '++', '--', '+=' etc

vilchik-elena:
Fixes #653

PR #8003, opened 2026-09-21
JS-2515 Extend S1077: cover `svg` a11y

guillemsarda:
----
## Summary by Gitar

- **Rule enhancements:**
    - Extended rule `S1077` to check for accessible names on inline `<svg>` elements.
    - Added support for verifying `title` children, `aria-label`, `aria-labelledby`, and decorative roles.


packages/analysis/src/jsts/rules/helpers/accessibility.ts
```diff
@@ -101,6 +101,45 @@ function isPotentiallyNonEmptyTemplateLiteralAttribute(attribute: JSXAttribute):
...
+            c.expression.type !== 'JSXEmptyExpression' &&
+            !(c.expression.type === 'Literal' && !c.expression.value) &&
+            !(c.expression.type === 'Identifier' && c.expression.name === 'undefined')),
+      ),
+  );
+}
+
+/**
+ * Checks whether an <svg> has an accessible name via aria-labelledby, aria-label, or a
+ * direct-child <title> element (in that precedence order). Id references in aria-labelledby
```
martin-strecker-sonarsource, line 131: **[question]** `hasSvgAccessibleName` treats `aria-labelledby` as a pure presence check, explicitly not resolving the id to its target's text (well-documented here, and mirrored by the sibling HTML PR #850). The ticket asks to "Check referenced text, not just ID existence... avoid reporting unresolved references." Since both the JS and HTML implementations independently landed on the same simplification, is this an agreed cross-platform scope decision worth capturing in a shared follow-up ticket, rather than two silent local limitations?

packages/analysis/src/jsts/rules/S1077/svg.ts
```diff
@@ -0,0 +1,137 @@
...
+  const roleValue = getLiteralPropValue(roleProp);
+  if (typeof roleValue !== 'string') {
+    return false;
+  }
+  const firstToken = roleValue.trim().split(/\s+/)[0]?.toLowerCase();
+  return DECORATIVE_ROLES.has(firstToken);
+}
+
+// An ancestor's aria-hidden="true" can't be overridden by a descendant's "false", so only "true" short-circuits.
+function isHiddenFromAssistiveTech(
```
martin-strecker-sonarsource, line 100: **[question]** `getAriaHiddenState`/`isHiddenFromAssistiveTech` is now the fourth independent `aria-hidden`-checking implementation in this rule set (S5264, S5256, S6819's `isDecorativeSvg` each have their own, none shared via `helpers/`) — and the only one that handles ancestor inheritance and string/boolean normalization correctly. Worth a follow-up to extract this into a shared helper and point S5264/S5256/S6819 at it, so they stop missing the same ancestor-inheritance case this PR just fixed for S1077?

packages/analysis/src/jsts/rules/helpers/accessibility.ts
```diff
@@ -101,6 +101,45 @@ function isPotentiallyNonEmptyTemplateLiteralAttribute(attribute: JSXAttribute):
   );
 }
 
+/**
+ * Checks whether an element has a direct-child <title> element with non-empty content.
+ * Independent of whitespace and of the title's position among its siblings.
+ */
+function hasSvgTitleChild(node: TSESTree.JSXOpeningElement): boolean {
```
martin-strecker-sonarsource, line 108: **[suggestion]** Neither `hasSvgTitleChild` nor `hasSvgAccessibleName` actually verify the node is an `<svg>` — both rely entirely on callers guarding `elementName === 'svg'` first. The "Svg" naming overpromises: a `<title>` JSX child is equally valid under `<head>`/`Head` (Next.js, Remix, react-helmet), unrelated to SVG accessible-name semantics. Suggest dropping "Svg" from both names. Doc comments should be updated to match.

packages/analysis/src/jsts/rules/S1077/svg.ts
```diff
@@ -0,0 +1,137 @@
...
+  // name (e.g. aria-label="." or <title>-</title>) is treated as accessible, matching
+  // upstream jsx-a11y/alt-text and Biome's no-svg-without-title.
+  if (hasSvgAccessibleName(node)) {
+    return;
+  }
+
+  context.report({ node: node.name as unknown as Node, message: MESSAGE });
+}
+
+function isRoleDynamic(attributes: JSXOpeningElement['attributes']): boolean {
```
martin-strecker-sonarsource, line 74: **[suggestion]** `isRoleDynamic` and `isDecorativeRole`'s guard clauses both re-derive the same `getProp`/`getLiteralPropValue`/non-string check. Suggest merging into one method, with the "unresolvable" and "decorative" branches collapsed into a single short-circuiting expression:
```ts
function isRoleDecorativeOrUnresolvable(attributes: JSXOpeningElement['attributes']): boolean {
  const roleProp = getProp(attributes, 'role');
  if (!roleProp) return false;
  const roleValue = getLiteralPropValue(roleProp);
  return (
    typeof roleValue !== 'string' ||
    DECORATIVE_ROLES.has(roleValue.trim().split(/\s+/)[0].toLowerCase())
  );
}
```
`isRoleDynamic` goes away entirely. Also note: the original `[0]?.toLowerCase()` optional chaining is dead code — `String.split(/\s+/)` always returns at least one element, so `[0]` can never be `undefined` once `roleValue` is a string; dropped above.

packages/analysis/src/jsts/rules/S1077/svg.ts
martin-strecker-sonarsource, line 100: **[question]** `isHiddenFromAssistiveTech` only checks `aria-hidden`, not the native `hidden` attribute or `style={{ display: 'none' }}`/`visibility: 'hidden'`. Per WAI-ARIA's "Excluding Elements from the Accessibility Tree," these are listed as parallel, equally-valid exclusion mechanisms alongside `aria-hidden="true"`: *"Elements... that have host language semantics specifying that the element is hidden, such as CSS `display:none` or `visibility:hidden` or HTML5 `hidden` attribute."* The `hidden` attribute case is trivially checkable statically (a boolean attribute, no CSS resolution needed) — is this a known/accepted gap, or worth adding given it's a common visibility-toggle pattern (initially-hidden tab/accordion icons)?

packages/analysis/src/jsts/rules/S1077/svg.ts
```diff
@@ -0,0 +1,137 @@
...
+  }
+  const roleValue = getLiteralPropValue(roleProp);
+  if (typeof roleValue !== 'string') {
+    return false;
+  }
+  const firstToken = roleValue.trim().split(/\s+/)[0]?.toLowerCase();
+  return DECORATIVE_ROLES.has(firstToken);
+}
+
+// An ancestor's aria-hidden="true" can't be overridden by a descendant's "false", so only "true" short-circuits.
```
martin-strecker-sonarsource, line 99: **[suggestion]** `getAriaHiddenState`'s `boolean | undefined` return has a state (`false` vs `undefined`) neither caller distinguishes — collapse to a plain boolean. Separately, `isHiddenFromAssistiveTech` checks the node's own attributes, then walks ancestors checking the same thing again — starting the walk at `node.parent` instead of `node.parent?.parent` folds the "self" check into the loop's first iteration for free, so it becomes one loop calling the check once per level instead of a pre-check plus a separate ancestor loop:
```ts
function isAriaHidden(attributes: JSXOpeningElement['attributes']): boolean {
  const prop = getProp(attributes, 'aria-hidden');
  if (!prop) return false;
  const value = getLiteralPropValue(prop);
  return value !== false && value !== 'false';
}

function isHiddenFromAssistiveTech(node: TSESTree.JSXOpeningElement): boolean {
  let ancestor: TSESTree.Node | undefined = node.parent;
  while (ancestor) {
    if (
      ancestor.type === 'JSXElement' &&
      isAriaHidden((ancestor.openingElement.attributes as unknown) as JSXOpeningElement['attributes'])
    ) {
      return true;
    }
    ancestor = ancestor.parent;
  }
  return false;
}
```
The `attributes` parameter drops out (derivable from `node`), and the cast narrows to just `.attributes` instead of the whole opening element — more honest about what's actually used, and closer to the shape overlap TS needs to accept the cast.

martin-strecker-sonarsource (changes requested):
Requesting changes. Two comments (role resolution and aria-hidden resolution) include concrete simplifications worth applying before merge; the rest are non-blocking questions/suggestions. Needs a follow-up review round once addressed.

packages/analysis/src/jsts/rules/helpers/accessibility.ts
guillemsarda: It was left out on purpose. I'd say that by adding test cases in both PRs, documenting this known limitation it's enough (in SonarJS we have `S1077/unit.test.ts:44-46`). What do you think?

PR #8012, opened 2026-09-22
JS-2522 Fix catalog resolution outside ESLint working directory

vdiez:
## Summary

- resolve pnpm and Bun catalog definitions from their nearest ancestor workspace even when ESLint is started inside a workspace package
- keep dependency manifest collection bounded by ESLint's working directory
- remove dependency-manifest diagnostics from stdout so ESLint formatter output remains machine-readable

This addresses the behavior reported in [the Sonar Community report](https://community.sonarsource.com/t/eslint-plugin-sonarjs-pnpm-catalogs-unresolved-when-linting-from-a-workspace-package/188188).

## Lookup boundaries

Above ESLint's working directory, lookup is limited to:

- `pnpm-workspace.yaml`, to find the nearest pnpm catalog definition
- `package.json`, to resolve Bun catalogs and determine ancestor workspace ownership

These catalog-specific searches stop at the filesystem root (`/` or the Windows drive root). Regular dependency-manifest collection remains bounded by the working directory, including ordinary `package.json`, `deno.json`, and `deno.jsonc` manifests. Ancestor `package.json` files are consulted only for Bun workspace and catalog semantics; they are not added as normal dependency manifests.

## Testing

- `npm run bbf`
- focused dependency-manifest test suites (70 tests)
- `npm run eslint-plugin:build`
- packaged-plugin reproduction with ESLint started from the pnpm workspace package; JSON formatter output parses successfully and stderr is empty


packages/analysis/src/jsts/rules/helpers/dependency-manifests/resolvers/package-json.ts
```diff
@@ -184,16 +186,10 @@ function findClosestParentPackageJsonWithCatalogs(
...
-    // No point in searching for parent package.json if we're already at the top directory.
-    return undefined;
-  }
-
   let currentDir = getParentDirPath(dir);
-  const cache = closestPatternCache.get(PACKAGE_JSON, fileSystem).get(topDir);
 
   while (currentDir !== null) {
-    const file = cache.get(currentDir);
+    const file = closestPackageJson(currentDir, topDir, fileSystem);
```
francois-mora-sonarsource, line 198: `findClosestParentPackageJsonWithCatalogs` accepts the first ancestor `package.json` that has `catalog`/`catalogs`, without checking that it actually declares `dir` as a workspace member. Since the search can now climb to the filesystem root, any unrelated ancestor with a catalog (e.g. a sibling project, or a `package.json` in a shared parent directory) will silently supply a version to a project it has no relationship to.

Repro: project not a member of an ancestor's workspace, ancestor declares `catalog: { react: 1.0.0-STRAY }`:
- master: `react` stays `"catalog:"` (unresolved, safe)
- this branch: `react` resolves to `"1.0.0-STRAY"` (wrong, silent)

This isn't a case where full resolution is too expensive to justify a heuristic — the membership check already exists in this file: `declaresWorkspaceDir` (used by `isIncludedInAncestorWorkspace` below) tests exactly this on data that's already been parsed. Suggest gating the `hasCatalogs` match here the same way:

```ts
if (parsed && hasCatalogs(parsed) && declaresWorkspaceDir(parsed, dirnamePath(file.filePath), dir)) {
  return parsed;
}
```

Without this, the search should probably keep climbing past a catalog-bearing-but-non-owning ancestor rather than stopping at it, so a genuine owner further up is still found.

vdiez: Good catch. Fixed in b9a3fb6e9d: ancestor Bun catalogs are now accepted only when `declaresWorkspaceDir` confirms that the ancestor workspace includes the package. Non-owning ancestors are skipped so a genuine owner further up can still be found.

I added the stray-ancestor regression and corrected the existing intermediate-manifest fixture so its workspace glob actually includes the package. All 59 focused dependency-manifest tests pass.

packages/analysis/src/jsts/rules/helpers/dependency-manifests/resolvers/package-json.ts
```diff
@@ -244,15 +232,34 @@ function isIncludedInAncestorWorkspace(
...
+): File | undefined {
+  const cache = closestPatternCache.get(PACKAGE_JSON, fileSystem);
+  const rootDir = getPathRoot(from);
+  if (from === topDir || from.startsWith(`${topDir}/`)) {
+    const file = cache.get(topDir).get(from);
+    if (file || topDir === rootDir) {
+      return file;
+    }
+    const parentOfTopDir = getParentDirPath(topDir);
+    return parentOfTopDir === null ? undefined : cache.get(rootDir).get(parentOfTopDir);
```
francois-mora-sonarsource, line 265: This fallback (and the pnpm one at line 56) reaches the ancestor via `closestPatternCache.get(pattern, fileSystem).get(rootDir)`, which is never preloaded and always falls back to real `fs.readdirSync`/`readFileSync`/`statSync` (`find-up/closest.ts:30`). There's no in-memory `Filesystem` alternative in the codebase — this always hits real disk.

That breaks `canAccessFileSystem: false` (SonarLint / SonarQube for IDE), where analysis must run only off supplied file content — e.g. the ancestor `pnpm-workspace.yaml`/`package.json` defining a catalog typically isn't part of what SonarLint sends, and reading it from disk anyway can even contradict unsaved editor content.

Repro (project with **no catalogs at all**, `canAccessFileSystem: false`, `fs` instrumented after `initFileStores`):
```
master        fsCallCount: 0
this branch   fsCallCount: 6   readdirSync(<baseDir's parent>), readdirSync(<its parent>), readdirSync(/)
```
The PR's own new no-fs test doesn't catch this — it only covers the preloaded `topDir`-cache hit, not this fallthrough.

Suggested fix: skip this fallback (return "not found") whenever `canAccessFileSystem` is false / `fileSystem` isn't the real `fs`.

vdiez: Fixed in 5e2f18ac5e. Preloaded manifest caches now retain the analysis filesystem-access policy. Both pnpm and Bun ancestor fallbacks stop at `topDir` when `canAccessFileSystem` is false, while direct ESLint lookups (which have no preloaded cache policy) continue to search ancestors on disk.

I added a cache-miss regression that instruments `readdirSync`, `readFileSync`, and `statSync` after initialization and confirms zero calls. All 60 focused dependency-manifest tests and the ESLint plugin type-check pass.

francois-mora-sonarsource (approved):
Reviewed the catalog-resolution and diagnostics-removal changes.

Two High-severity issues were raised on the ancestor-lookup logic added for above-`topDir` catalog resolution:
- unrelated ancestor workspaces could hijack a resolved dependency version (no membership check)
- the `canAccessFileSystem: false` (SonarLint) contract was bypassed by an unconditional disk-backed fallback

Both are fixed in `b9a3fb6e9` and `5e2f18ac5`, verified locally (isolated + full suite, 60/60 pass) and via targeted repros against master for each fix.

The `console.debug` removal is correct and complete for the ESLint-plugin-bundled surface (no stdout writers remain under `tsconfig-plugin.json`'s `rootDir`).

Approving.

PR #8015, opened 2026-09-22
JS-1821 Clean up S7739 thenable-contract exception per post-merge review

francois-mora-sonarsource:
## Jira

JS-1821 — follow-up cleanup for the S7739 thenable-contract exception merged in #7863.

## Context

#7863 was merged before its final human review round (by @oleksandr-selehenenko-sonarsource) could be applied. This PR applies that round's code-quality feedback. It is pure cleanup — no change to what the rule reports.

## Changes

- **Drop `thenableContractCache`.** The WeakMap only guarded a cheap JSDoc/heritage-clause scan on `interceptReport`'s already-cold path (it only runs on nodes the upstream rule already decided to report), so the caching gave no measurable benefit while adding module-level mutable state — and forced a test case (`AccessorThenable`) that existed solely to hit its cache-hit branch. That test's comment is reworded to describe the real assertion it still makes (both members of a get/set `then` pair are suppressed).
- **Reuse `getNodeParent`/`ancestorsChain`** from `helpers/ancestor.ts` instead of the hand-rolled parent-chain walks, deleting the duplicate logic from `S7739/helpers.ts`.
- **Replace ad-hoc structural casts with TSESTree narrowing** for the class-member (`MethodDefinition`/`PropertyDefinition`) and heritage-clause (`ClassDeclaration`/`ClassExpression.implements`) checks. Kept the `implements ?? []` fallback — a first pass without it broke the plain-JavaScript test suite, since a JS-parsed class has no `implements` field at runtime even though TSESTree's type declares it non-optional.
- **Pin the Closure-style `@implements {!IThenable<?>}` nullability prefix** with a new valid test case. The regex already matched it; nothing previously asserted so, and the PR description for #7863 flagged this shape as a risk for a future exact-match refactor.

## Tests

`packages/analysis/src/jsts/rules/S7739/no-validation-lib/unit.test.ts` — added the `!IThenable<?>` valid case; reworded the `AccessorThenable` case's comment (no behavior change). Full `S7739` suite passes (4/4 suites).

## Boundary

No behavior change — same suppression boundary as #7863.


oleksandr-selehenenko-sonarsource (approved):
LGTM

PR #8017, opened 2026-09-22
CLP-960 Make concurrency queueing explicit so pending runs are not silently cancelled in SonarJS

frederic-tingaud-sonarsource:
Part of CLP-959

`concurrency.queue` defaults to `single`, which allows only **one** run to sit in `pending`.
With `cancel-in-progress: false` the *running* job is protected, but as soon as a third run
joins the group the **pending** run is silently cancelled: it never executes, nothing retries
it, and no failure is reported anywhere.

This PR makes the queueing policy explicit in every concurrency block of this repository.

| Workflow | `queue` | Rationale |
|---|---|---|
| `.github/workflows/pvf-comment.yml` | `max` | Each run dispatches the specific `/pvf` comment body to peachee; a dropped run loses that request entirely. |

Blocks already using `cancel-in-progress: true` are untouched: they cannot drop a pending run.


alex-meseldzija-sonarsource (approved):
LGTM

PR #8018, opened 2026-09-22
JS-2522 Search ancestor manifests for standalone ESLint

vdiez:
## Summary

- search ancestor dependency manifests up to the filesystem root when rules run in standalone ESLint
- keep dependency lookup bounded by the configured project directory in the Sonar runtime
- add regression coverage for Jest rules when ESLint starts in a nested application directory

This is a follow-up to [JS-2522](https://sonarsource.atlassian.net/browse/JS-2522) and fixes the behavior reported in [the Sonar Community report](https://community.sonarsource.com/t/eslint-plugin-sonarjs-skips-test-rules-because-of-the-project-working-directory/188359).

## Testing

- focused `importsOrDependsOnModule`, S5976, and S8754 test suites
- packaged-plugin reproduction with ESLint's `cwd` below the root Jest dependency
- `npm run bbf`
- `npm run eslint-plugin:build`

[JS-2522]: https://sonarsource.atlassian.net/browse/JS-2522?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/analysis/src/jsts/rules/helpers/dependency-manifests/dependencies.ts
```diff
@@ -137,12 +143,13 @@ export function withCurrentFileInlineDependencies(manifest: DependenciesList): D
...
-      normalizeToAbsolutePath(context.cwd),
-    ),
-  );
+  const filePath = normalizeToAbsolutePath(context.filename);
+  // ESLint can lint from a nested working directory while Node still resolves packages from
+  // ancestor directories. Sonar analysis has an explicit project boundary and must remain inside it.
+  const topDir = isSonarRuntime(context)
+    ? normalizeToAbsolutePath(context.cwd)
+    : getPathRoot(filePath);
+  return withCurrentFileInlineDependencies(getDependencies(dirnamePath(filePath), topDir));
```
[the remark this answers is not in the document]
vdiez: Good catch — confirmed. The dependency cache was keyed only by the closest manifest directory, so lookups with different search boundaries could reuse the wrong result and become rule-order dependent.

Fixed in `066871b588` by:

- including `topDir` in the dependency-cache identity;
- centralizing standalone-vs-Sonar boundary selection for dependency, React, and Vue lookups;
- adding regressions for both lookup orders and for nested packages inheriting ancestor framework dependencies.

Validated with the focused rule/helper suites, `npm run bbf`, and `npm run eslint-plugin:build`.

packages/analysis/src/jsts/rules/helpers/dependency-manifests/dependencies.ts
```diff
@@ -137,12 +174,18 @@ export function withCurrentFileInlineDependencies(manifest: DependenciesList): D
...
+}
+
+function getDependencyTopDir(
+  context: Rule.RuleContext,
+  filePath: NormalizedAbsolutePath,
+): NormalizedAbsolutePath {
+  // ESLint can lint from a nested working directory while Node still resolves packages from
+  // ancestor directories. Sonar analysis has an explicit project boundary and must remain inside it.
+  return isSonarRuntime(context) ? normalizeToAbsolutePath(context.cwd) : getPathRoot(filePath);
```
francois-mora-sonarsource: The search has to stop somewhere as it walks up looking for `package.json` files.

Today it stops at `context.cwd`. That's too early — when ESLint runs inside `apps/myapp`, it never sees the root `package.json`, so jest is invisible. That's the bug being fixed.

This PR makes it stop at the filesystem root instead. That's too late — it now also reads `package.json` files from directories that have nothing to do with the project:

```
linting a project from its own root; an unrelated ancestor declares jest
master        hasJest: false   (correct)
this branch   hasJest: true    (taken from the unrelated ancestor)
```

The trouble is that both options pick a *distance*, and no distance is right: the file we want might be three levels up, the file we must ignore might be one level up.

Suggestion: stop at the git repository root. Everything inside the repo can reasonably be called this project; anything outside it clearly isn't. (It needs a plain existence check — `closestPatternCache` only matches files, so it would silently skip a `.git` directory.)

The line 186 comment is right that Node can resolve modules from any ancestor, but that doesn't make those packages part of this project, which is what the rules are asking about.

I prototyped it:

| | `cwd` | `getPathRoot` | repo root |
|---|---|---|---|
| root `package.json` has jest, ESLint runs in `apps/myapp` | missed | found | found |
| unrelated ancestor has jest, ESLint runs at project root | clean | leaked | clean |
| same as row 1 but no lockfile committed | missed | found | found |
| git submodule inside a monorepo | missed | found | **missed** |
| no `.git` at all (vendored copy, extracted tarball) | missed | found | **missed** |

I couldn't produce a false positive, since the search never leaves the repo. But the last two rows are real misses, so it's not strictly better — it trades false positives that hit ordinary projects for false negatives in two unusual layouts.

francois-mora-sonarsource:
`getPackageJsonManifestsSanitizePaths` (`all-in-parent-dirs.ts:57`) still passes `context.cwd`, so it didn't get the new boundary.

It's the final case in S1607's switch (`S1607/rule.ts:64`), right after the jest / vitest / mocha / cypress / playwright checks. Those five go through `getDependenciesSanitizePaths` and now find ancestor manifests; the sixth still stops at `cwd`. So within one switch, a project using jest gets the fix and a project using the plain Node.js test runner doesn't — it keeps the exact bug this PR exists to fix, in the same nested-cwd layout from the linked report.

Worth applying whatever boundary you settle on to this helper too, so the switch behaves consistently.

packages/analysis/src/jsts/rules/helpers/dependency-manifests/dependencies.ts
francois-mora-sonarsource: `getPackageJsonManifestsSanitizePaths` (`all-in-parent-dirs.ts:57`) still passes `context.cwd`, so it didn't get the new boundary.

It's the final case in S1607's switch (`S1607/rule.ts:64`), right after the jest / vitest / mocha / cypress / playwright checks. Those five go through `getDependenciesSanitizePaths` and now find ancestor manifests; the sixth still stops at `cwd`. So within one switch, a project using jest gets the fix and a project using the plain Node.js test runner doesn't — it keeps the exact bug this PR exists to fix, in the same nested-cwd layout from the linked report.

Worth applying whatever boundary you settle on to this helper too, so the switch behaves consistently.

vdiez: Addressed both points in `1ce028024f`. Standalone lookups now stop at the nearest Git repository root, including worktrees with a `.git` file, so unrelated ancestor manifests are excluded. `getPackageJsonManifestsSanitizePaths` uses the same boundary, covering S1607's Node test-runner branch; there is a direct nested-`cwd` regression test.

I also checked the CI failures. Both unit jobs failed because the previous commit added Vue to a shared fixture whose existing test expects only `foo`. React/Vue now have a dedicated fixture, so those tests no longer read SonarJS's own root package. The full JavaScript unit suite passes locally (2,333 passed, 0 failed), as do `npm run bbf` and the plugin build.

The PR description calls out the remaining Git-submodule and no-Git fallback behavior.

packages/analysis/src/jsts/rules/helpers/dependency-manifests/top-dir.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  const fileDir = dirnamePath(filePath);
+  const repositoryRoot = repositoryRootCache.get(fileDir);
+  if (repositoryRoot) {
+    return repositoryRoot;
+  }
+
+  // Without a repository marker, keep the existing working-directory boundary.
+  // A file outside cwd cannot be bounded by cwd, so use its own directory.
+  return relativeToAncestorPath(fileDir, cwd) === undefined ? fileDir : cwd;
```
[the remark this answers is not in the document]
vdiez: Good catch. Commit `b3c999408e` keeps `cwd` as the boundary when it contains the nested repository root, preserving parent-workspace manifests in that layout. Otherwise standalone lookup still stops at the nearest Git root. I added a regression covering dependency, React/Vue, and `package.json` lookup; the focused suites pass.

francois-mora-sonarsource (approved):
LGTM

